PR #17, opened 2023-01-10
fix: ensure coder.coder-remote dir exists

jsjoeio:
- chore(deps): add fs-extra
- fix: ensure coder.coder-remote dir exists

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -269,6 +269,7 @@ export class Storage {
     const token = await this.getSessionToken()
     if (token) {
       axios.defaults.headers.common["Coder-Session-Token"] = token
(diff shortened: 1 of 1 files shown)


src/storage.ts
```diff
@@ -247,6 +248,7 @@ export class Storage {
     const url = this.getURL()
     axios.defaults.baseURL = url
     if (url) {
+      await ensureDir(this.globalStorageUri.fsPath)
```
kylecarbs, line 251: You should add this for the session token too!

jsjoeio: Ah good call. Let me do that now!

PR #33, opened 2023-01-17
fix: keep binary in same place, update if needed

### LINKED ISSUE #23: Windows Firewall prompt on every new binary
Every time there's a new version on the server (which is multiple times per day in dev.coder.com) the extension will download a new binary and Windows firewall will prompt to allow network access.

We should figure out a way to prevent the windows firewall thing from popping up (we only need to access over localhost anyways).

jsjoeio:
## Description

@deansheather filed a bug related to the Windows Firewall prompt coming up with every new binary (i.e. every time we download `coder` to an updated version).

After some initial testing, it appears keeping the binary in the same place will make the prompt only appear on the first time.

Previous implementation kept the version in the path.

We decided to refactor things to add support for [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) on the server and then use those changes on the client (aka in the extension).

This means now the client will tell the server "Here's the version I have. What should I do?" and send the appropriate headers.

### Checklist
- [x] if needed, change path/name of location to not include version
- [x] if the binary exists, SHA1 it and use it as the `If-None-Match` header in your request
- [x] Server Response changes
  - [x] on 200 OK: move the binary on top of the old one
  - [x] on 304 Not Modified: delete the temp file and use the old binary

### Testing Plan

TBD

Fixes #23

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "build": "webpack",
     "watch": "webpack --watch",
     "package": "webpack --mode production --devtool hidden-source-map",
+    "package:prerelease": "npx vsce package --pre-release",
     "lint": "eslint . --ext ts,md",
     "tsc:compile": "tsc",
     "tsc:watch": "tsc -w",
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -114,103 +114,117 @@ export class Storage {
       // Ignore all errors so we can catch a 404!
       validateStatus: () => true,
     })
-    if (resp.status === 404) {
-      vscode.window
-        .showErrorMessage(
-          "Coder isn't supported for your platform. Please open an issue, we'd love to support it!",
-          "Open an Issue",
+
+    switch (resp.status) {
+      case 200: {
+        const contentLength = Number.parseInt(resp.headers["content-length"])
+
+        // Ensure the binary directory exists!
+        await fs.mkdir(path.dirname(binPath), { recursive: true })
+        const tempFile = binPath + ".temp-" + Math.random().toString(36).substring(8)
+
+        const completed = await vscode.window.withProgress<boolean>(
+          {
+            location: vscode.ProgressLocation.Notification,
+            title: `Downloading the latest binary (${buildInfo.version} from ${baseURI.authority})`,
+            cancellable: true,
+          },
+          async (progress, token) => {
+            const readStream = resp.data as IncomingMessage
+            let cancelled = false
+            token.onCancellationRequested(() => {
+              controller.abort()
+              readStream.destroy()
+              cancelled = true
+            })
+
+            let contentLengthPretty = ""
+            // Reverse proxies might not always send a content length!
+            if (!Number.isNaN(contentLength)) {
+              contentLengthPretty = " / " + prettyBytes(contentLength)
+            }
+
+            const writeStream = createWriteStream(tempFile, {
+              autoClose: true,
+              mode: 0o755,
+            })
+            let written = 0
+            readStream.on("data", (buffer: Buffer) => {
+              writeStream.write(buffer, () => {
+                written += buffer.byteLength
+                progress.report({
+                  message: `${prettyBytes(written)}${contentLengthPretty}`,
+                  increment: (buffer.byteLength / contentLength) * 100,
+                })
+              })
+            })
+            try {
+              await new Promise<void>((resolve, reject) => {
+                readStream.on("error", (err) => {
+                  reject(err)
+                })
+                readStream.on("close", () => {
+                  if (cancelled) {
+                    return reject()
+                  }
+                  writeStream.close()
+                  resolve()
+                })
+              })
+              return true
+            } catch (ex) {
+              return false
+            }
+          },
         )
-        .then((value) => {
-          if (!value) {
-            return
-          }
-          const params = new URLSearchParams({
-            title: `Support the \`${os}-${arch}\` platform`,
-            body: `I'd like to use the \`${os}-${arch}\` architecture with the VS Code extension.`,
-          })
-          const uri = vscode.Uri.parse(`https://github.com/coder/vscode-coder/issues/new?` + params.toString())
-          vscode.env.openExternal(uri)
-        })
-      return
-    }
-
-    const contentLength = Number.parseInt(resp.headers["content-length"])
-
-    // Ensure the binary directory exists!
-    await fs.mkdir(path.dirname(binPath), { recursive: true })
-    const tempFile = binPath + ".temp-" + Math.random().toString(36).substring(8)
-
-    const completed = await vscode.window.withProgress<boolean>(
-      {
-        location: vscode.ProgressLocation.Notification,
-        title: `Downloading the latest binary (${buildInfo.version} from ${baseURI.authority})`,
-        cancellable: true,
-      },
-      async (progress, token) => {
-        const readStream = resp.data as IncomingMessage
-        let cancelled = false
-        token.onCancellationRequested(() => {
-          controller.abort()
-          readStream.destroy()
-          cancelled = true
-        })
-
-        let contentLengthPretty = ""
-        // Reverse proxies might not always send a content length!
-        if (!Number.isNaN(contentLength)) {
-          contentLengthPretty = " / " + prettyBytes(contentLength)
+        if (!completed) {
+          return
         }
-
-        const writeStream = createWriteStream(tempFile, {
-          autoClose: true,
-          mode: 0o755,
-        })
-        let written = 0
-        readStream.on("data", (buffer: Buffer) => {
-          writeStream.write(buffer, () => {
-            written += buffer.byteLength
-            progress.report({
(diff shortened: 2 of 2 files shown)


src/storage.ts
```diff
@@ -80,32 +81,8 @@ export class Storage {
...
-          this.output.appendLine("The cached binary cannot be executed: " + ex)
-          resolve(false)
-        }
-      })
-      if (valid) {
-        return binPath
-      }
-    }
+    const binPath = this.binaryPath()
+    const exists = await this.checkBinaryExists(binPath)
```
jsjoeio, line 86: Most of this is simply refactoring to make the code easier to follow

src/storage.ts
```diff
@@ -114,6 +91,23 @@ export class Storage {
...
+      } else {
+        // Even if the file exists, it could be corrupted.
+        // We run `coder version` to ensure the binary can be executed.
+        this.output.appendLine(`Using existing binary: ${binPath}`)
+        const valid = await this.checkBinaryValid(binPath)
+        if (valid) {
+          return binPath
+        }
+      }
+    }
```
jsjoeio: This is the meat of the changes

src/storage.ts
```diff
@@ -236,6 +230,10 @@ export class Storage {
     return path.join(this.globalStorageUri.fsPath, "url")
   }
 
+  public getBinaryETag(): string {
+    return crypto.createHash("sha1").update(this.binaryPath()).digest("hex")
+  }
```
jsjoeio: I believe this is the same `sha1` hashing the backend is using. Although I'd love input on easiest way to check (less familiar on the Go side).

src/storage.ts
```diff
@@ -260,16 +258,57 @@ export class Storage {
     }
   }
 
-  private binaryPath(version: string): string {
+  private binaryPath(): string {
     const os = goos()
     const arch = goarch()
-    let binPath = path.join(this.getBinaryCachePath(), `coder-${os}-${arch}-${version}`)
+    let binPath = path.join(this.getBinaryCachePath(), `coder-${os}-${arch}`)
```
jsjoeio, line 328: 🔨  We no longer use the version in the binaryPath. This should fix the Windows prompt issue.

src/storage.ts
```diff
@@ -260,16 +258,57 @@ export class Storage {
...
+
+    switch (resp.status) {
+      case 200:
+        return true
+      case 304:
+      default:
+        return false
+    }
+  }
+
```
jsjoeio: New logic added based off: https://github.com/coder/coder/pull/5750

jsjoeio:
What's the easiest way to test these changes locally @kylecarbs ?

src/storage.ts
deansheather: SHA1 is universal, so as long as it's a SHA1 hash with hex encoding (lowercase) then it's good.

src/storage.ts
deansheather: I think this logic should be more like:
```
if (exists) {
  const valid = checkBinValid();
  if (valid) {
    return binPath;
  }
  deleteBin();
}

downloadBinary()
```

and the request to the bin endpoint should only be made once. The axios request should be piped to a temp file in the same dir, then once the request is finished:
- if 200: move the temp file on top of the binPath
- if 304: delete the temp file and use the old binPath

deansheather: Using two requests for this is not necessary

deansheather (approved):

deansheather (dismissed):
oops meant to request changes

src/storage.ts
jsjoeio: But doesn't that mean we're always downloading a new binary even if we don't have to?

jsjoeio: Also if the logic is like that, then we'd never get to the `downloadBinary()` right?

jsjoeio: I'll see if I can clean up to make it only one request.

deansheather: oh you're right, I meant this:
```
if (exists) {
  const valid = checkBinValid();
  if (!valid) {
    deleteBin();
  }
}

let tempFile = binPath + ".temp-" + randString(8);
let res = await request(serverBinPath, streamToFile: tempFile);

if (res.status == 200) {
  move(tempFile, binPath);
  return binPath;
}
remove(tempFile);
if (res.status != 304) {
  throw new Error("failed to make request to server");
}
return binPath;
```

jsjoeio: OH, okay that makes more sense. Thank you!

src/storage.ts
```diff
@@ -140,15 +129,12 @@ export class Storage {
         })
       return
     }
-    if (resp.status !== 200) {
```
deansheather, line 147: Could avoid piping to the temp file here by checking resp.status == 304 and returning the existing binary path.

deansheather: And also error if the status isn't 200 or 304 here so you don't have to pipe to the file first. I didn't think you could do that but you can judging by the similar check for 404 above.

deansheather (commented):
the logic seems OK but I haven't tested it

src/storage.ts
jsjoeio: Good idea. I'll refactor.

jsjoeio:
@deansheather re: testing

Let me make these changes, build to a `vsix` then send to you via Slack to test

jsjoeio:
I tested locally on macOS and something is off 

image

Seems like it's trying to use the binary before it's done renaming it?

jsjoeio:
I'm going to try rebasing on `main` and then rebuilding and testing again.

jsjoeio:
I figured it out - I was trying to `rm` the file after `mv`ing and so it was failing because it couldn't remove a file that doesn't exist 🤦🏼‍♂️

jsjoeio:
Tested locally and seems to work 👍🏼 

image

jsjoeio:
@deansheather ready for a new review!

src/storage.ts
```diff
@@ -114,106 +92,152 @@ export class Storage {
...
+    this.output.appendLine(`Using binPath: ${binPath}`)
+    this.output.appendLine(`Using ETag: ${etag}`)
+
     const resp = await axios.get("/bin/" + binName, {
       signal: controller.signal,
       baseURL: baseURL,
       responseType: "stream",
       headers: {
         "Accept-Encoding": "gzip",
+        "If-None-Match": `"${await this.getBinaryETag()}"`,
```
deansheather:
```suggestion
        "If-None-Match": `"${etag}"`,
```

src/storage.ts
```diff
@@ -114,106 +92,152 @@ export class Storage {
...
+      if (!valid) {
+        const removed = await this.rmBinary(binPath)
+        if (!removed) {
+          vscode.window.showErrorMessage("Failed to remove existing binary!")
+          return undefined
+        }
+      }
+    }
+    const etag = await this.getBinaryETag()
+    this.output.appendLine(`Using binPath: ${binPath}`)
```
deansheather, line 109:
```suggestion
    this.output.appendLine(`Using binName: ${binName}`)
    this.output.appendLine(`Using binPath: ${binPath}`)
```

deansheather (approved):
Looks good and works on Windows for me. The logic seems OK but I'm not a big typescript guy so someone else should approve too

package.json
```diff
@@ -85,6 +85,7 @@
     "build": "webpack",
     "watch": "webpack --watch",
     "package": "webpack --mode production --devtool hidden-source-map",
+    "package:prerelease": "npx vsce package --pre-release",
```
Kira-Pilot, line 88: This script looks like it publishes a pre-release version of a package. Could you help me understand the general workflow here and how it relates to this issue?

src/storage.ts
```diff
@@ -73,39 +74,16 @@ export class Storage {
...
-          this.output.appendLine("The cached binary cannot be executed: " + ex)
-          resolve(false)
-        }
-      })
-      if (valid) {
-        return binPath
-      }
-    }
+    const binPath = this.binaryPath()
+    const exists = await this.checkBinaryExists(binPath)
```
Kira-Pilot, line 86: Do we need to worry about error handling here?

Kira-Pilot (approved):
Really appreciate the comments you left!

package.json
jsjoeio: This will build the project and output a `.vsix` file which makes it easier to test (or send to someone else to test).

Eventually, I'd like to a GitHub workflow to build on PRs and upload the `.vsix` as an artifact to make it easier to test. Not directly related to this PR but should help for future troubleshooting!

src/storage.ts
jsjoeio: Good question! So this method calls `fs.stat` and has a `catch` method which returns `false` basically saying the path/file doesn't exist. I think it's okay as is for now!

Kira-Pilot: Oh ya! Looks great.

Issue #42, opened 2023-01-19
Removes first matching Host block from ~/.ssh/config


aaronlehmann:
If there is a `Host` block in ~/.ssh/config that matches the name used for the remote connection (for example `Host *`), this gets removed from the file when attempting to connect to a workspace.

I suspect this is caused by:

```ts
parsedConfig.remove({ Host: computedHost.Host })
```

`computedHost.Host` would evaluate to the `Host` value from the first block that matched (`*` in my example). It would be safer to pass an explicit value to remove.

mafredri:
Linking issue #30 as it's kind of related.

kylecarbs:
Ahh, certainly. Will fix ASAP.

kylecarbs:
This should be fixed now! We have starting and ending comments. v0.1.6 has the fix.

PR #48, opened 2023-01-25
fix: Inject coder ssh configuration between comments

BrunoQuaresma:


BrunoQuaresma:
Closes https://github.com/coder/vscode-coder/issues/42

src/remote.ts
```diff
@@ -386,7 +388,27 @@ export class Remote {
...
     }
-    const parsedConfig = SSHConfig.parse(sshConfigRaw)
+
+    // We are going to extract only the coder config
+    let coderConfigRaw = ""
+    const startOfCoderConfig = sshConfigRaw.indexOf(startBlockComment)
+    const endOfCoderConfig = sshConfigRaw.lastIndexOf(endBlockComment)
+
+    // There is an existent coder config
+    if (startOfCoderConfig > -1 && endOfCoderConfig > -1) {
```
mafredri: We should also check `startOfCoderConfig < endOfCoderConfig`. See https://github.com/coder/coder/pull/5859.

src/remote.ts
```diff
@@ -386,7 +388,27 @@ export class Remote {
       // Probably just doesn't exist!
       sshConfigRaw = ""
     }
-    const parsedConfig = SSHConfig.parse(sshConfigRaw)
+
+    // We are going to extract only the coder config
+    let coderConfigRaw = ""
+    const startOfCoderConfig = sshConfigRaw.indexOf(startBlockComment)
+    const endOfCoderConfig = sshConfigRaw.lastIndexOf(endBlockComment)
```
mafredri: This (lastIndexOf) seems a bit unsafe considering there could be malformed comments. We should ensure there's only one of each exists or ask the user to correct it.

src/remote.ts
BrunoQuaresma: Good catch!

src/remote.ts
BrunoQuaresma: A good one too!

mafredri:
> Does it look right?

Yup, that sounds perfect @BrunoQuaresma. 👍🏻

BrunoQuaresma:
I wanted to add some tests to the `SSHConfig` but I figure out the testing runner for this project needs some extra work so I decided to not do it on this PR.

src/SSHConfig.ts
```diff
@@ -0,0 +1,105 @@
...
+    return `  ${text}`
+  }
+
+  private async save() {
+    await ensureDir(path.dirname(this.filePath))
+    return fs.writeFile(this.filePath, this.getRaw(), "utf-8")
+  }
+
+  private getRaw() {
+    if (!this.raw) {
```
mafredri: Should this check explicitly for undefined? Will match empty string as well.

src/SSHConfig.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (startBlockIndex === -1) {
+      throw new SSHConfigBadFormat("Start block not found")
+    }
+
+    if (startBlockIndex === -1) {
+      throw new SSHConfigBadFormat("End block not found")
+    }
+
+    if (endBlockIndex < startBlockIndex) {
+      throw new SSHConfigBadFormat("End block is starting before the start block")
```
mafredri:
```suggestion
      throw new SSHConfigBadFormat("Malformed config, end block is before start block")
```

src/SSHConfig.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private appendBlock({ Host, ...otherValues }: SSHValues) {
+    const lines = [this.startBlockComment, `Host ${Host}`]
+    const keys = Object.keys(otherValues) as Array<keyof typeof otherValues>
+    keys.forEach((key) => {
+      lines.push(this.withIndentation(`${key} ${otherValues[key]}`))
+    })
+    lines.push(this.endBlockComment)
+    this.raw = `${this.raw}\n${lines.join("\n")}`
```
mafredri: Have you tested how this affects a config when a block is repeatedly removed and added back, or repeatedly updated/replaced?

For instance, when doing the above the newline count in the config should not constantly be growing. I.e. one risk is there are 100s of newlines at the end after a bunch of operations.

src/SSHConfig.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private withIndentation(text: string) {
+    return `  ${text}`
```
mafredri: Suggestion (future improvement), parse config to figure out indentation used by other hosts (tabs, 2 spaces, 4 spaces etc), and use the same if unified across the file.

src/SSHConfig.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private async save() {
+    await ensureDir(path.dirname(this.filePath))
```
mafredri: If ssh dir doesn't exist and we create it here, we should ensure it has good permissions. Usually `chmod 0700`, i.e. only owner has rwx permission, not group or everyone.

Likewise for `~/.ssh/config`, if we create it then a good permission would be `chmod 0600` (owner rw).

mafredri (commented):
Looking good. Still had some pretty minor but probably important (for good user experience) nits.

src/SSHConfig.ts
BrunoQuaresma: Good catch! And I didn't know JS would consider an empty string as false. 🤦

src/SSHConfig.ts
BrunoQuaresma: Good catch!

mafredri (approved):
Thanks for fixing the nits. Code looks good to me, but I haven't tested the extension so might be good to have another set of eyes on it.

src/SSHConfig.ts
```diff
@@ -0,0 +1,142 @@
+import { ensureDir } from "fs-extra"
```
kylecarbs: What do you think about camel casing this name instead? So there's uniformity with other files.

kylecarbs (approved):

src/SSHConfig.ts
BrunoQuaresma: I just started it with capital case because it is a class, but I can do it.

PR #52, opened 2023-02-02
fix: open workspaces in new windows

kylecarbs:
Fixes coder/coder#5703


mafredri (approved):
Looks good 👍🏻. If we can check if this is an "empty" window (like no folder opened), we could consider enabling the reuse in that case.

Reason I'm suggesting this is that the common workflow, I assume, will be to open VS Code in and empty window, find your workspace and launch it. Now you have two VS Code windows.

mafredri (approved):
Nice! 😎

PR #53, opened 2023-02-02
fix: default to https in login url

### LINKED ISSUE #44: No web browser window opens for copying API key when logging in to server without "https://" prefix
On version `v0.1.5` of the extension, I tried logging into my Coder deployment by providing the URL but I didn't include the "https://" prefix. I was then prompted to paste the API key from the Coder website that the extension said should have been launched in a web browser window, but no web browser window was ever launched.

I tried once more but this time provided the "https://" prefix and then VS Code prompted me to allow it to open the Coder website in a web browser window, and from there I was able to login successfully.

kylecarbs:
Fixes #44.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -21,6 +21,11 @@ export class Commands {
     if (!url) {
       return
     }
+    if (!url.startsWith("http://") && !url.startsWith("https://")) {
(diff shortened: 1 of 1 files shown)


mafredri (approved):
LGTM!

PR #55, opened 2023-02-06
feat: show workspace updates in the ui

kylecarbs:
Listens for workspace updates and displays them in the UI.

Eventually, we can show the template version upate message!

Fixed coder/coder#6050

### DIFF
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -72,9 +72,8 @@ export class Remote {
     }
 
     // Find the workspace from the URI scheme provided!
-    let workspace: Workspace
     try {
-      workspace = await getWorkspaceByOwnerAndName(parts[0], parts[1])
+      this.storage.workspace = await getWorkspaceByOwnerAndName(parts[0], parts[1])
     } catch (error) {
       if (!axios.isAxiosError(error)) {
         throw error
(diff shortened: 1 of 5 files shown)


src/remote.ts
```diff
@@ -262,11 +264,44 @@ export class Remote {
...
       // TODO: Add debug output that we got an error here!
     })
+
+    const workspaceUpdatedStatus = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, 999)
+    disposables.push(workspaceUpdatedStatus)
+
+    const refreshWorkspaceUpdatedStatus = (newWorkspace: Workspace) => {
+      // If the newly gotten workspace was updated, then we show a notification
+      // to the user that they should update.
+      if (!this.storage.workspace?.outdated && newWorkspace.outdated) {
```
deansheather: If the workspace is already pending an upgrade it seems to me from the code that the notification won't show because of this condition. I would expect the notification to always show when first connecting in this case.

deansheather (approved):
Code LGTM

src/remote.ts
kylecarbs: I see. I'll make this display on first connection too!

kylecarbs: Done!

PR #59, opened 2023-02-13
feat: multiple agent support

ytcheng:
add multiple agent support
**Screenshot**


src/commands.ts
```diff
@@ -158,24 +159,67 @@ export class Commands {
...
       workspaceName = args[1]
-      // workspaceAgent is reserved for args[2], but multiple agents aren't supported yet.
+      workspaceAgent = args[2]
       folderPath = args[3]
     }
 
     // A workspace can have multiple agents, but that's handled
     // when opening a workspace unless explicitly specified.
-    const remoteAuthority = `ssh-remote+${Remote.Prefix}${workspaceOwner}--${workspaceName}`
+    const remoteAuthority = `ssh-remote+${Remote.Prefix}${workspaceOwner}--${workspaceName}--${workspaceAgent}`
```
kylecarbs: This should be optionally appended, because when opening a stopped workspace there may be no agents.

src/commands.ts
```diff
@@ -158,24 +159,67 @@ export class Commands {
...
       }, [] as WorkspaceAgent[])
 
       if (agents.length === 1) {
         folderPath = agents[0].expanded_directory
-      }
+        workspaceAgent = agents[0].name
+      }else{
+          const agentQuickPick = vscode.window.createQuickPick()
+          let lastAgents: WorkspaceAgent[]
+          agentQuickPick.title = `Connect to a agent`;
```
kylecarbs:
```suggestion
          agentQuickPick.title = `Select an agent`;
```

kylecarbs (approved):
Thank you for the contribution! 🥳

Some lines have `;` and some don't. If you could run the prettier formatter, that'd be great! Have you manually tested this? I'll give it a shot in a bit too.

ytcheng:
> Thank you for the contribution! 🥳
> 
> Some lines have `;` and some don't. If you could run the prettier formatter, that'd be great! Have you manually tested this? I'll give it a shot in a bit too.

Yes, I've manually tested. It works

src/commands.ts
ytcheng: I've Changed this code

kylecarbs (approved):

kylecarbs:
@ytcheng I'll test, merge, and do a release in the morning (it's 1am here) ;p.

Thanks again!

smolinari:
Just a small improvement suggestion and a question. 

The text should be "Connect to an Agent" (a to an). :grin: 

Also, what would be multiple agents? In a workspace?

Scott

kylecarbs:
@smolinari some users will have a workspace that contains a database and a container that they code in, or maybe multiple containers in more abstract scenarios.

In these situations, you might have multiple agents you can connect to inside a single workspace!

smolinari:
Hmm... interesting. Didn't think about having more than one agent in a workspace. Cool. More possibilities.... :grin: 

Scott

rodrimaia:
✅ Updated branch with most recent changes
✅ Added support to also selecting agents on Sidebar:

When the workspace only has multiple agents, now we list them and allow users to open an agent directly:
image

For workspaces with a single agent, use the default view:
image

PR #69, opened 2023-03-22
feat: Add feature to allow config-ssh value overrides

Emyrk:
Also supports unsetting ssh config options.


src/remote.ts
```diff
@@ -440,6 +441,37 @@ export class Remote {
...
+            "Your session expired...",
+            {
+              useCustom: true,
+              modal: true,
+              detail: "You must login again to access your workspace.",
+            },
+            "Login",
+          )
+          throw error
+        }
```
Emyrk: What is the best way to handle this?

Emyrk: This just throws some error if there is an unauthorized response code. It shouldn't happen because we are logged in, so it's an edge case that we don't need to handle all too well?

Emyrk:
Fixes https://github.com/coder/coder/issues/6826

Issue #64, opened 2023-03-07
CRUD workspaces


bpmct:
Support creating, updating, and deleting workspaces from the VS Code extension, with full parameter support.

VS Code supports WebViews, should it be done via this?

PR #72, opened 2023-03-28
WIP: feat: crud workspaces

rodrimaia:
#64 

https://www.loom.com/share/269af08124544e81a5e58471ffe2a67c 

- [x] View Workspaces:
  - we are displaying two views: My Workspaces and All Workspaces (the user can still use the quickPick command to filter and open all workspaces)
  - We have also a context menu showing a link to the workspace dashboard page 
- [x] Create Workspaces:
  - A button pointing to the dashboard/create extension page

image

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -81,6 +81,9 @@ export class Commands {
         throw new Error("Failed to get authenticated user")
       }
       await vscode.commands.executeCommand("setContext", "coder.authenticated", true)
+      if (user.roles.find((role) => role.name === "owner")) {
+        await vscode.commands.executeCommand("setContext", "coder.isOwner", true)
+      }
       vscode.window
         .showInformationMessage(
           `Welcome to Coder, ${user.username}!`,
--- a/package.json
+++ b/package.json
@@ -50,7 +50,8 @@
           "id": "allWorkspaces",
           "name": "All Workspaces",
           "visibility": "visible",
-          "icon": "media/logo.svg"
+          "icon": "media/logo.svg",
+          "when": "coder.authenticated && coder.isOwner"
         }
       ]
     },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -20,8 +20,13 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   vscode.window.registerTreeDataProvider("allWorkspaces", allWorkspacesProvider)
 
   getAuthenticatedUser()
-    .then(() => {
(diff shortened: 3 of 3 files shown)


BrunoQuaresma:
I see some plugins asking to confirm the "name" of the thing you want to delete. Probably we want to do the same. What do you think?

bpmct (commented):
Any plans to add support for an "edit' button which goes to the workspace settings?

src/api-helper.ts
```diff
@@ -0,0 +1,16 @@
...
+): [agents: WorkspaceAgent[], folderPath: string | undefined] {
+  // TODO: multiple agent support
+  const agents = workspace.latest_build.resources.reduce((acc, resource) => {
+    return acc.concat(resource.agents || [])
+  }, [] as WorkspaceAgent[])
+
+  let folderPath = undefined
+  if (agents.length === 1) {
+    folderPath = agents[0].expanded_directory
+  }
```
Emyrk, line 14: If there is more than 1 agent, does everything fail? Should we randomly choose one?

src/commands.ts
```diff
@@ -79,6 +81,9 @@ export class Commands {
         throw new Error("Failed to get authenticated user")
       }
       await vscode.commands.executeCommand("setContext", "coder.authenticated", true)
+      if (user.roles.find((role) => role.name === "owner")) {
+        await vscode.commands.executeCommand("setContext", "coder.isOwner", true)
+      }
```
Emyrk, line 86: Why do we need to check for the owner role? We prefer to do capability checks. You can send a payload to `https://dev.coder.com/api/v2/authcheck` to see if you can do something. Eg:
```json
{
   "checks":{
      "readAllUsers":{
         "object":{
            "resource_type":"user"
         },
         "action":"read"
      },
   },
}
```

Is this just to see if we have arbitrary execution of all workspaces?

Emyrk: I think this is ok for now, we but we might want to change this soon, especially with https://github.com/coder/coder/pull/6875

src/api-helper.ts
rodrimaia: That is a good question. I am not changing the actual behavior, just extracted it to a new function. the current version also does this `agents[0]` choice =/ . I am not sure what is the best way to handle multiple agents...

src/commands.ts
rodrimaia: We decided to only show the "all Workspaces" view for users who have the owner role. sure we can revisit this after the owner connections PR

BrunoQuaresma (approved):
FE code LGTM

PR #75, opened 2023-04-05
feat: Align remote ssh naming with coder cli

Liorba:
coder's cli tool allow users to configure ssh conection to their remote hosts.
as part of it's configuration, he generate ssh config  host to the dedicated user's machine.
The patten used by the cli tool, is not align with the one user by the vscode extension. 
This mismatch between  coder cli and the vscode extension, cause errors when trying to connect to remote host.

Key differences:
1. the host pattern [generated](https://github.com/coder/coder/blob/929589ddfa90ee55928fe0830d225f030cba7344/cli/configssh.go#L352) by the coder cli is `coder.<workspace-name>.<agent>`. currently the default agent is main
   the vscode extension is expecting the pattern `coder-vscode--<workspace-owner>--<workspace-name>--<agent?>`
2. owner / user is not encoded in the ssh host name genereated by coder cli
3. the cli [allows](https://github.com/coder/coder/blob/929589ddfa90ee55928fe0830d225f030cba7344/cli/configssh.go#L507) the user to set different prefix other then `coder`

This PR is set to close the gap between the 2 and make sure hosts created by the cli can be connected from the vscode extension

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -217,7 +217,8 @@ export class Commands {
 
     // A workspace can have multiple agents, but that's handled
     // when opening a workspace unless explicitly specified.
-    const remoteAuthority = `ssh-remote+${Remote.Prefix}${workspaceOwner}--${workspaceName}`
+    // search for workspace name in the ssh config
+    const remoteAuthority = `ssh-remote+${Remote.Prefix}.${workspaceName}.main` //main is the default agent
 
     let newWindow = true
     // Open in the existing window if no workspaces are open.
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -17,6 +17,7 @@ import * as os from "os"
 import * as path from "path"
 import prettyBytes from "pretty-bytes"
 import * as semver from "semver"
+
 import * as vscode from "vscode"
 import * as ws from "ws"
 import { SSHConfig, defaultSSHConfigResponse, mergeSSHConfigValues } from "./sshConfig"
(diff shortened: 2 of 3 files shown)


src/commands.ts
```diff
@@ -217,7 +217,8 @@ export class Commands {
 
     // A workspace can have multiple agents, but that's handled
     // when opening a workspace unless explicitly specified.
-    const remoteAuthority = `ssh-remote+${Remote.Prefix}${workspaceOwner}--${workspaceName}`
+    // search for workspace name in the ssh config
+    const remoteAuthority = `ssh-remote+${Remote.Prefix}.${workspaceName}.main` //main is the default agent
```
kylecarbs, line 221: `main` is not the default name for an agent. It can be anything that a user defines!

src/remote.ts
```diff
@@ -38,18 +40,13 @@ export class Remote {
     // If the URI passed doesn't have the proper prefix ignore it. We don't need
     // to do anything special, because this isn't trying to open a Coder
     // workspace.
-    if (!authorityParts[1].startsWith(Remote.Prefix)) {
-      return
-    }
+    // if (!authorityParts[1].startsWith(Remote.Prefix)) {
+    //   return
+    // }
```
kylecarbs, line 45: This needs to be checked, otherwise Coder will activate on any remote SSH connection, which will cause errors to appear on non-Coder connections.

kylecarbs (changes requested):
Hey @Liorba, I'm a bit confused by this contribution. We don't want `coder config-ssh` to interact with the VS Code extension at all, because they are truly separate experiences.

Could you explain more of the problem this solves?

Liorba:
The reason for that is the ability to use the vscode extension to connect to workspaces that was configured using the config ssh option.
For example. I have a workspace that I'm connected to using ssh that I already configured via the cli.
I want to seamlessly be able to connect to it and work.with it using the extension.

code-asher:
Closing as stale, but happy to take another look if this becomes active again.

Issue #76, opened 2023-04-05
pass headers to coder api 


Liorba:
Hello coder team!
Our coder deployment is put behind a vpn so in order to send api calls to our control plan every request needs to have additional header holding  our vpn session token. 

in the vscode-coder code base, I noticed the user of axios, which can accept additional headers.
The problem resides in the fact that some of the api called are in the __coder__  codebase and I'm not sure what's the preferred way to propagate those headers to the axios client in the coder coder base.

please advice on the preferred path and I'll be happy contribute!

Liorba:
@kylecarbs I've added support for VPN headers in #81

code-asher:
Accidentally closed this, still need to merge in the plugin PR for this to be complete.

PR #81, opened 2023-04-17
pass headers to coder api

Liorba:
Hey coder team this pr tries to tackle issue #76
This pr added the following functionality to the vscode extension:
1. new set of configuration to set a VPN header 
2. option to set the token directly in `settings.json` or in file 
3. if header name is set and non of the option above was not provided a user will be presented by prompt widow to add his token (similar to session token)
4. if a valid header it will be added to `Axios` global headers.
5. To deal with a case that the provided token expired of invalid after some time, I've added new Axios interceptor. which will check a set of conditions and prompt the user to put new case it discover a bad response due to invalid token
6. pass the vpn header name and token to the SSH config `proxyCommand` 

There is an issue with the coder `vpnssh` command that unlike the `ssh` command do not pass the headers correctly to the underline ssh command. when I replace it with the "regular" ssh command it worked. any Idea what might causing this?


kylecarbs:
@Liorba I fixed the header problem in the attached PR: https://github.com/coder/coder/pull/7181

I'll review the rest shortly!

src/commands.ts
```diff
@@ -1,4 +1,4 @@
-import axios from "axios"
+import axios  from "axios"
```
deansheather:
```suggestion
import axios from "axios"
```

src/commands.ts
```diff
@@ -8,8 +8,8 @@ import { Storage } from "./storage"
 import { WorkspaceTreeItem } from "./workspacesProvider"
 
 export class Commands {
-  public constructor(private readonly vscodeProposed: typeof vscode, private readonly storage: Storage) {}
 
+  public constructor(private readonly vscodeProposed: typeof vscode, private readonly storage: Storage) {}
```
deansheather:
```suggestion
  public constructor(private readonly vscodeProposed: typeof vscode, private readonly storage: Storage) {}

```

deansheather (commented):
This should probably just be a map of header name to string or file path so people can set multiple headers

Liorba:
@deansheather,I thought about this approach at first,  but this is no regular header. it sort of like the session token which is a header but has a special treatments.

kylecarbs:
@rodrimaia let's make this more generic with `coder.headers` instead to allow for a series of arbitrary headers!

bpmct:
We did this in https://github.com/coder/vscode-coder/pull/119!

Issue #87, opened 2023-04-24
Coder extension installing but login command not found


mjasten32:
When I either manually install Coder Remote or from the web UI Dekstop button, I do not see the CLI download.

If I'm in the extension, the left pane is blank and if I click the login into Coder button, it says it is not found.

I have tried uninstalling, rebooting, and reinstalling the extension and have witnessed the same experience.

Any suggestions? I'm in an M1 Macbook, with plenty of memory 16GB. I used the Arm64 version of VS Code.

<img width="1024" alt="image" src="https://user-images.githubusercontent.com/106293630/234023324-6853bd80-500c-4e48-a85a-5d449b257c23.png">

<img width="288" alt="image" src="https://user-images.githubusercontent.com/106293630/234023446-3f0aac8e-afff-46fc-a77f-5a25e087b4d5.png">

leonhausdorf:
I got the same error but with the MacBook Pro 2020 M1.

I also tried to install the extension into the Insiders version from VSCode and it happened to work normally at first but after a workspace restart/timeout it got "broken" again. Deleting the most known cache and reinstalling VSCode does not fix it. 

It works well on other systems like Windows. The Apple Silicon platform seems to be the only one with this problem.

I will do further investigation.

leonhausdorf:
After further investigation it seems to have a problem with the CSRF Token while activating.

<img width="1374" alt="image" src="https://user-images.githubusercontent.com/59312981/234715557-8dbd6fee-772a-42fc-8827-d3d38d16c731.png">

PR #88, opened 2023-04-26
fixed sessionToken execution error and stop with try-catch

leonhausdorf:
Hello coder team 

this pull request solves the error in #87 

It seemed like the MacOS (Apple Silicon) version from VSCode have some issues with processing the stored sessionToken in the SecretStorage. When the extension tried to get the token the code execution got interrupted by the error mentioned in #87.

I fixed that interruption by using a try-catch and return undefined if the process could not be completed. If it's not possible to retrieve the session token, the user needs to reauthenticate.

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -52,7 +52,11 @@ export class Storage {
   }
 
   public async getSessionToken(): Promise<string | undefined> {
-    return this.secrets.get("sessionToken")
+    try {
+      return this.secrets.get("sessionToken");
+    } catch {
+      return undefined;
+    }
   }
 
   // getRemoteSSHLogPath returns the log path for the "Remote - SSH" output panel.
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -53,9 +53,9 @@ export class Storage {
 
   public async getSessionToken(): Promise<string | undefined> {
     try {
-      return this.secrets.get("sessionToken");
+      return this.secrets.get("sessionToken")
     } catch {
-      return undefined;
+      return undefined
     }
   }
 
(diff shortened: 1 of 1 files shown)


kylecarbs (commented):
Actually... this would never throw because we aren't awaiting the `this.secrets.get` function. Are you sure this resolves it?

leonhausdorf:
I tracked down the error to this line of code where the error (I posted it in the issue) got thrown. Building this try-catch resolved it for me on my MacBook and MacMini (both M1). 

I will add the comment in a moment

leonhausdorf:
Okay I did more testing today to really understand what's happening there. First I added the await since this is an async function (also it helps the try-catch). 

If it fails to get the token, following message pops up.
image

I logged the error:
`Error: Unable to parse decrypted password at S.$getPassword (vscode-file://vscode-app/private/var/folders/rw/jrm9d3p96hvf869w2wbxc6tc0000gn/T/AppTranslocation/DB56FBC8-41D5-4D4A-9D24-2883FE87323D/d/Visual%20Studio%20Code%20-%20Insiders.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:1527:47885) {name: 'Error', message: 'Unable to parse decrypted password', stack: 'Error: Unable to parse decrypted password ' ....`

This might need to get some more attention to find out what's really happening there since the Apple Silicon Devices are the only one being affected by this (at least this is what I figured).

kylecarbs:
@leonhausdorf I was able to reproduce this and fixed it today! A new release should be out shortly :)

PR #92, opened 2023-05-02
feat: show agent metadata

### LINKED ISSUE #71: display agent metadata
we would like to display agent metadata on vscode extension, possibly on the Status Bar ( bottom bar)

rodrimaia:
~~the UI alignment is not working as it is supposed to, I guess something related to the vscode tooltip internals (because the string itself is correct). I can investigate this later, but I don't think that is a blocker.~~

- We added the metadata to the sidebar instead of having a statusbaricon: 
image

Closes #71


matifali:
Wow loved it. Awesome addition.

kylecarbs:
@rodrimaia the agent name is already shown in the workspace title, so I'm not sure we should add another item to the bottom bar all the time.

e.g. if I have no metadata right now, what shows up?

BrunoQuaresma (approved):
The code looks good to me. My only considerations are:
- What happens if it is empty or the agent is not connected
- If there is no way to align the value text to the right, I'm ok with this version

kylecarbs:
@rodrimaia could you update the screenshot with the new display?

rodrimaia:
> @rodrimaia could you update the screenshot with the new display?

@kylecarbs 
image

matifali (approved):

kylecarbs:
@rodrimaia thoughts on putting metadata in the Coder sidebar instead?

kylecarbs:
The reason is that it's kinda odd for dynamic data to go in the tooltip, and in the UI we have a "Copy" button, which might be useful in the extension too.

rodrimaia:
@kylecarbs I am experimenting with showing the metadata on the sidebar. I will post it here when I reach a working version.

rodrimaia:
image

PR #96, opened 2023-05-04
feat: validate ssh properties before launching workspaces

kylecarbs:
This was reported by a member in Discord, and displayed a poor error by VS Code when it failed before.

Users cannot override our config options otherwise connections may fail or have unexpected behavior.

### DIFF
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -20,7 +20,7 @@ import * as semver from "semver"
 import * as vscode from "vscode"
 import * as ws from "ws"
 import { SSHConfig, SSHValues, defaultSSHConfigResponse, mergeSSHConfigValues } from "./sshConfig"
-import { sshSupportsSetEnv } from "./sshSupport"
+import { computeSSHProperties, sshSupportsSetEnv } from "./sshSupport"
 import { Storage } from "./storage"
 
 export class Remote {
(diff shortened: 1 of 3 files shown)


src/remote.ts
```diff
@@ -528,6 +528,34 @@ export class Remote {
...
+      if (computedProperties[key] === sshValues[key]) {
+        continue
+      }
+
+      const result = await this.vscodeProposed.window.showErrorMessage(
+        "Unsupported SSH Config Option",
+        {
+          useCustom: true,
+          modal: true,
+          detail: `Your SSH config is overriding the "${key}=${computedProperties[key]}" option for the "${hostName}" host. Please remove it and try again!`,
```
mafredri: We could add "overriding the key=correct with key=wrong ..." to help trace the issue?

src/sshSupport.ts
```diff
@@ -36,3 +36,55 @@ export function sshVersionSupportsSetEnv(sshVersionString: string): boolean {
...
+        properties: Record<string, string>
+      }
+    | undefined
+  const configs: Array<typeof currentConfig> = []
+  config.split("\n").forEach((line) => {
+    line = line.trim()
+    if (line === "") {
+      return
+    }
+    const [key, ...valueParts] = line.split(/\s+/)
```
mafredri: Separator could be `=` or space.

src/remote.ts
kylecarbs: Good point!

src/sshSupport.ts
kylecarbs: Ahh, great point! I'll add a test case.

kylecarbs:
@mafredri 
image

PR #99, opened 2023-05-10
fix: open from sidebar with agent

rodrimaia:
Fix open workspace from sidebar

### DIFF
--- a/src/workspacesProvider.ts
+++ b/src/workspacesProvider.ts
@@ -57,7 +57,9 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
     if (element) {
       if (element instanceof WorkspaceTreeItem) {
         const agents = extractAgents(element.workspace)
-        const agentTreeItems = agents.map((agent) => new AgentTreeItem(agent, element.watchMetadata))
+        const agentTreeItems = agents.map(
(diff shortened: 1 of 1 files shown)


src/workspacesProvider.ts
```diff
@@ -57,7 +57,9 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
     if (element) {
       if (element instanceof WorkspaceTreeItem) {
         const agents = extractAgents(element.workspace)
-        const agentTreeItems = agents.map((agent) => new AgentTreeItem(agent, element.watchMetadata))
+        const agentTreeItems = agents.map(
+          (agent) => new AgentTreeItem(agent, element.workspaceOwner, element.workspaceName, element.watchMetadata),
```
BrunoQuaresma, line 61: Why not just pass `element` directly into the `AgentTreeItem`?

BrunoQuaresma (approved):
Just one minor consideration

src/workspacesProvider.ts
rodrimaia: Ideally, you could instantiate an agentTreeItem without a WorkspaceTreeItem, but it doesn't matter in practice. Thank you for the suggestion I will squeeze this in later :)

PR #110, opened 2023-06-21
feat: add VS code notifications for autostop

Kira-Pilot:
- part of the resolution for [#7360](https://github.com/coder/coder/issues/7360)
 - the rest will follow after https://github.com/coder/coder/pull/7989 is merged up
Screenshot 2023-06-15 at 3 43 14 PM

PR #111, opened 2023-06-21
feat: add VS code notifications for workspace actions

Kira-Pilot:
Opening this PR in place of #110 as this solution is a little cleaner.
This resolves https://github.com/coder/coder/issues/7360
Screenshot 2023-06-22 at 12 33 03 PM

### DIFF
--- a/src/WorkspaceAction.ts
+++ b/src/WorkspaceAction.ts
@@ -1,6 +1,8 @@
 import { getWorkspaces } from "coder/site/src/api/api"
 import { Workspace, WorkspacesResponse } from "coder/site/src/api/typesGenerated"
 import { formatDistanceToNowStrict } from "date-fns"
+import { Storage } from "./storage"
+import axios from "axios"
 import * as vscode from "vscode"
 
 interface NotifiedWorkspace {
@@ -16,7 +18,11 @@ export class WorkspaceAction {
   #workspacesApproachingAutostop: NotifiedWorkspace[] = []
   #workspacesApproachingDeletion: NotifiedWorkspace[] = []
 
-  private constructor(private readonly vscodeProposed: typeof vscode, ownedWorkspaces: Workspace[]) {
+  private constructor(
+    private readonly vscodeProposed: typeof vscode,
+    private readonly storage: Storage,
+    ownedWorkspaces: Workspace[],
+  ) {
     this.#ownedWorkspaces = ownedWorkspaces
 
     // seed initial lists
@@ -28,15 +34,21 @@ export class WorkspaceAction {
     this.pollGetWorkspaces()
   }
 
-  static async init(vscodeProposed: typeof vscode) {
+  static async init(vscodeProposed: typeof vscode, storage: Storage) {
     // fetch all workspaces owned by the user and set initial public class fields
     let ownedWorkspacesResponse: WorkspacesResponse
     try {
       ownedWorkspacesResponse = await getWorkspaces({ q: "owner:me" })
     } catch (error) {
+      if (!axios.isAxiosError(error) || error.response?.status !== 401) {
+        storage.writeToCoderOutputChannel(
+          `Failed to fetch owned workspaces. Some workspace notifications may be missing: ${error}`,
+        )
+      }
+
       ownedWorkspacesResponse = { workspaces: [], count: 0 }
     }
-    return new WorkspaceAction(vscodeProposed, ownedWorkspacesResponse.workspaces)
+    return new WorkspaceAction(vscodeProposed, storage, ownedWorkspacesResponse.workspaces)
   }
 
   seedNotificationLists() {
@@ -98,10 +110,15 @@ export class WorkspaceAction {
         this.seedNotificationLists()
         this.notifyAll()
       } catch (error) {
+        errorCount++
+        if (!axios.isAxiosError(error) || error.response?.status !== 401) {
+          this.storage.writeToCoderOutputChannel(
+            `Failed to poll owned workspaces. Some workspace notifications may be missing: ${error}`,
+          )
+        }
         if (errorCount === 3) {
           clearInterval(this.#fetchWorkspacesInterval)
         }
-        errorCount++
       }
     }, 1000 * 5)
   }
--- a/src/WorkspaceAction.ts
+++ b/src/WorkspaceAction.ts
@@ -51,7 +51,14 @@ export class WorkspaceAction {
   }
 
   filterImpendingAutostopWorkspaces(workspace: Workspace) {
-    if (workspace.latest_build.transition !== "start" || !workspace.latest_build.deadline) {
+    // a workspace is eligible for autostop if the last build was successful,
+    // and the workspace is started,
+    // and it has a deadline
+    if (
+      workspace.latest_build.job.status !== "succeeded" ||
+      workspace.latest_build.transition !== "start" ||
+      !workspace.latest_build.deadline
+    ) {
       return false
     }
 
--- a/src/WorkspaceAction.ts
+++ b/src/WorkspaceAction.ts
@@ -1,5 +1,5 @@
-import { Workspace, WorkspacesResponse } from "coder/site/src/api/typesGenerated"
 import { getWorkspaces } from "coder/site/src/api/api"
+import { Workspace, WorkspacesResponse } from "coder/site/src/api/typesGenerated"
 import * as vscode from "vscode"
 
 interface NotifiedWorkspace {
--- a/src/WorkspaceAction.ts
+++ b/src/WorkspaceAction.ts
@@ -26,7 +26,7 @@ export class WorkspaceAction {
     this.#ownedWorkspaces = ownedWorkspaces
 
     // seed initial lists
-    this.seedNotificationLists()
+    this.updateNotificationLists()
 
     this.notifyAll()
 
@@ -51,14 +51,18 @@ export class WorkspaceAction {
     return new WorkspaceAction(vscodeProposed, storage, ownedWorkspacesResponse.workspaces)
   }
 
-  seedNotificationLists() {
+  updateNotificationLists() {
     this.#workspacesApproachingAutostop = this.#ownedWorkspaces
       .filter(this.filterWorkspacesImpendingAutostop)
-      .map((workspace: Workspace) => this.transformWorkspaceObjects(workspace, workspace.latest_build.deadline))
+      .map((workspace: Workspace) =>
+        this.transformWorkspaceObjects(workspace, this.#workspacesApproachingAutostop, workspace.latest_build.deadline),
+      )
 
     this.#workspacesApproachingDeletion = this.#ownedWorkspaces
       .filter(this.filterWorkspacesImpendingDeletion)
-      .map((workspace: Workspace) => this.transformWorkspaceObjects(workspace, workspace.deleting_at))
+      .map((workspace: Workspace) =>
+        this.transformWorkspaceObjects(workspace, this.#workspacesApproachingDeletion, workspace.deleting_at),
+      )
   }
 
   filterWorkspacesImpendingAutostop(workspace: Workspace) {
@@ -89,14 +93,13 @@ export class WorkspaceAction {
     return Math.abs(new Date().getTime() - new Date(workspace.deleting_at).getTime()) <= dayMilli
   }
 
-  transformWorkspaceObjects(workspace: Workspace, deadlineField?: string) {
+  transformWorkspaceObjects(workspace: Workspace, workspaceList: NotifiedWorkspace[], deadlineField?: string) {
     // the below line is to satisfy TS; we should always pass a deadlineField, e.g
     // workspace,deleting_at or workspace.latest_build.deadline
     if (!deadlineField) {
       return { workspace, wasNotified: true, impendingActionDeadline: "" }
     }
-    const wasNotified =
-      this.#workspacesApproachingAutostop.find((wn) => wn.workspace.id === workspace.id)?.wasNotified ?? false
+    const wasNotified = workspaceList.find((nw) => nw.workspace.id === workspace.id)?.wasNotified ?? false
     const impendingActionDeadline = formatDistanceToNowStrict(new Date(deadlineField))
     return { workspace, wasNotified, impendingActionDeadline }
   }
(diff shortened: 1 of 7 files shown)


src/WorkspaceAction.ts
```diff
@@ -0,0 +1,151 @@
...
+        this.#ownedWorkspaces = workspacesResult.workspaces
+        this.seedNotificationLists()
+        this.notifyAll()
+      } catch (error) {
+        if (errorCount === 3) {
+          clearInterval(this.#fetchWorkspacesInterval)
+        }
+        errorCount++
+      }
+    }, 1000 * 5)
```
Kira-Pilot: Not ideal, but this is the same interval we use to poll in the Dashboard.

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,151 @@
...
+    // set up polling so we get current workspaces data
+    this.pollGetWorkspaces()
+  }
+
+  static async init(vscodeProposed: typeof vscode) {
+    // fetch all workspaces owned by the user and set initial public class fields
+    let ownedWorkspacesResponse: WorkspacesResponse
+    try {
+      ownedWorkspacesResponse = await getWorkspaces({ q: "owner:me" })
+    } catch (error) {
```
code-asher, line 49: Should we add some kind of log or notification for this request failure?  At least, if it fails for any reason other than not being logged in.

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,151 @@
...
+    // return workspaces with a deleting_at  that is 24 hrs or less
+    return Math.abs(new Date().getTime() - new Date(workspace.deleting_at).getTime()) <= dayMilli
+  }
+
+  transformWorkspaceObjects(workspace: Workspace, deadlineField?: string) {
+    // the below line is to satisfy TS; we should always pass a deadlineField, e.g
+    // workspace,deleting_at or workspace.latest_build.deadline
+    if (!deadlineField) {
+      return { workspace, wasNotified: true, impendingActionDeadline: "" }
+    }
```
code-asher: If we wanted to fix this I think we would need to modify the filter return types to assert that the fields exist.  Something like:

```typescript
type WithRequired<T, K extends keyof T> = T & Required<Pick<T, K>>
type WorkspaceWithDeadline = Workspace & {latest_build: WithRequired<WorkspaceBuild, "deadline">}

// ...

    this.#workspacesApproachingAutostop = this.#ownedWorkspaces
      .filter(this.filterWorkspacesImpendingAutostop)
      .map((workspace) => this.transformWorkspaceObjects(workspace, workspace.latest_build.deadline))

// ...

  filterWorkspacesImpendingAutostop(workspace: Workspace): workspace is WorkspaceWithDeadline {

// ...
```

But, just a thought, not something we have to do.

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,151 @@
...
+    const wasNotified =
+      this.#workspacesApproachingAutostop.find((wn) => wn.workspace.id === workspace.id)?.wasNotified ?? false
```
code-asher: Since `transformWorkspaceObject` is also called for workspaces approaching deletion I think we would fail to preserve the `wasNotified` status for all of those as we only check autostop notifications here.

We probably do not want to simply check both here either since a workspace could be notified for autostop and then later notified for deletion (or vice versa) and checking both could prevent one from firing?  Maybe we pass in the list to check into `transformWorkspaceObject`.

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,151 @@
...
+    let ownedWorkspacesResponse: WorkspacesResponse
+    try {
+      ownedWorkspacesResponse = await getWorkspaces({ q: "owner:me" })
+    } catch (error) {
+      ownedWorkspacesResponse = { workspaces: [], count: 0 }
+    }
+    return new WorkspaceAction(vscodeProposed, ownedWorkspacesResponse.workspaces)
+  }
+
+  seedNotificationLists() {
```
code-asher: Very minor nit: `seed` implies to me we call it once but we call it on every poll; maybe `updateNotificationList` instead?

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  async pollGetWorkspaces() {
+    let errorCount = 0
+    this.#fetchWorkspacesInterval = setInterval(async () => {
+      try {
+        const workspacesResult = await getWorkspaces({ q: "owner:me" })
+        this.#ownedWorkspaces = workspacesResult.workspaces
+        this.seedNotificationLists()
+        this.notifyAll()
+      } catch (error) {
```
code-asher, line 115: Thoughts on whether we should raise these errors in some way?  Even if just logging it somewhere out of sight.  I think the plugin has an output channel somewhere.

src/WorkspaceAction.ts
code-asher: Makes sense to me!  I am tempted to say we throw it in a named variable and mention it matches the dashboard interval.

code-asher (commented):
Looks great!  Love to see the `#` syntax at work.

I have not tried it out yet since when I build it tells me `deleting_at` does not exist.  Do we need to update the `coder` dependency maybe?

src/WorkspaceAction.ts
code-asher: One other thought on the notification list: if you get notified for a workspace being autostopped then you start it again leaving VS Code open, would you not get notified for the next autostop?  Maybe if the deadline changes we want to reset `wasNotified`.

code-asher: Ah I guess it would only be a problem if it autostops within an hour of you starting it again, otherwise the workspace would have been removed from the list thus essentially resetting `wasNotified`.

src/remote.ts
```diff
@@ -126,6 +127,9 @@ export class Remote {
       this.registerLabelFormatter(remoteAuthority, this.storage.workspace.owner_name, this.storage.workspace.name),
     )
 
+    // Initialize any WorkspaceAction notifications (auto-off, upcoming deletion)
+    const Action = await WorkspaceAction.init(this.vscodeProposed)
```
code-asher: Is there a special meaning behind having this variable capitalized?

src/WorkspaceAction.ts
Kira-Pilot: I had to think about this for a minute! I think you're saying that as we poll, we are redefining `workspacesApproachingAutostop`, effectively resetting `wasNotified`. If a workspace is temporarily removed from that list and then added back, `wasNotified` will be reinitialized to `true`. I agree - should only be an issue if the workspace is eligible to autostop within the hour, in which case I think the extra notification is forgivable, but let me know if you feel that will be annoying to the user.

code-asher: I was thinking about the opposite case where we might miss a notification because `wasNotified` was set for an earlier/different deadline and never cleared.

Specifically I was thinking about it happening when a workspace automatically turns off and then the user turns it on again but I just realized that would not happen even if the auto-stop was set to one hour because the job status and transition would no longer be `succeeded`/`start` and it would get kicked out of the list, resetting `wasNotified`, and all is well, unless they somehow manage to start it so fast the poll does not catch it in the off state.  :facepalm:  Sorry for the noise.

I suppose we could technically hit the issue if the deadline is extended (by one hour or less)? 

1. User gets notified that a workspace is about to auto-stop (or be deleted)
2. `wasNotified` set to true
3. User extends deadline (or delete deadline is moved) by one hour or less
4. No new notification for the next deadline since already notified for the previous one

But since VS Code itself is supposed to keep the workspace active we would never see this case anyway?  And shifting a delete by one hour seems like a rather unlikely case as well.  So I think you can ignore everything I am saying. :sweat_smile: 

If we do need to account for this I think we could do it easily by only preserving `wasNotified` if `deadline`/`deleting_at` is the same, that way we show a new notification for each separate deadline even if they happen back-to-back.  For example:

```typescript
    const wasNotified =
      this.#workspacesApproachingAutostop.find((wn) => wn.workspace.id === workspace.id && wn.latest_build.deadline === workspace.latest_build.deadline)?.wasNotified ?? false
```

src/WorkspaceAction.ts
Kira-Pilot: Yes! I didn't know about output channel and was worried about spamming the user - thanks for pointing me in the right direction below.
We now have logging for both `catch` blocks:
Screenshot 2023-06-23 at 11 19 37 AM

src/WorkspaceAction.ts
Kira-Pilot: Done!

Kira-Pilot:
> I have not tried it out yet since when I build it tells me `deleting_at` does not exist. Do we need to update the `coder` dependency maybe?
Done! Let me know if you still see that error.

src/remote.ts
Kira-Pilot: The meaning is that I was really tired - my bad 😵‍💫

src/WorkspaceAction.ts
Kira-Pilot: Oh clever! Added. I will be using this elsewhere going forward.

src/WorkspaceAction.ts
Kira-Pilot: Oof, thanks for catching this.

src/WorkspaceAction.ts
Kira-Pilot: Gotcha! I am going to be working on adding CTAs as a fast-follow so I will investigate extending auto-stop then!

src/WorkspaceAction.ts
```diff
@@ -11,6 +11,11 @@ interface NotifiedWorkspace {
   impendingActionDeadline: string
 }
 
+type WithRequired<T, K extends keyof T> = T & Required<Pick<T, K>>
+
+type WorkspaceWithDeadline = Workspace & { latest_build: WithRequired<WorkspaceBuild, "deadline"> }
+type WorkspaceWithDeletingAt = Workspace & WithRequired<Workspace, "deleting_at">
```
code-asher: Very minor nit, this could be simplified to:
```suggestion
type WorkspaceWithDeletingAt = WithRequired<Workspace, "deleting_at">
```

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.pollGetWorkspaces()
+  }
+
+  static async init(vscodeProposed: typeof vscode, storage: Storage) {
+    // fetch all workspaces owned by the user and set initial public class fields
+    let ownedWorkspacesResponse: WorkspacesResponse
+    try {
+      ownedWorkspacesResponse = await getWorkspaces({ q: "owner:me" })
+    } catch (error) {
+      if (!axios.isAxiosError(error) || error.response?.status !== 401) {
```
code-asher: Is it safe to ignore Axios errors? I think that covers things like timeouts or connection refused which might be useful information to relay, although I could see an argument against it because if the server is not responding then most likely the workspace would not be connectable anyway so it might already be obvious something is wrong with the network.

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,173 @@
...
+    let errorCount = 0
+    this.#fetchWorkspacesInterval = setInterval(async () => {
+      try {
+        const workspacesResult = await getWorkspaces({ q: "owner:me" })
+        this.#ownedWorkspaces = workspacesResult.workspaces
+        this.updateNotificationLists()
+        this.notifyAll()
+      } catch (error) {
+        errorCount++
+        if (!axios.isAxiosError(error) || error.response?.status !== 401) {
```
code-asher: Similar thoughts on Axios errors here.

src/WorkspaceAction.ts
```diff
@@ -0,0 +1,173 @@
...
+      )
+  }
+
+  filterWorkspacesImpendingAutostop(workspace: Workspace): workspace is WorkspaceWithDeadline {
+    // a workspace is eligible for autostop if the last build was successful,
+    // and the workspace is started,
+    // and it has a deadline
+    if (
+      workspace.latest_build.job.status !== "succeeded" ||
+      workspace.latest_build.transition !== "start" ||
```
code-asher: Completely FYI and optional but there is another `status` property that is the composite of these two:

https://github.com/coder/coder/blob/140683813d794081a0c0dbab70ec7eee16c5f5c4/coderd/workspacebuilds.go#L898-L930

It is found on `latest_build.status` and the equivalent value in this case is `running`.

src/remote.ts
```diff
@@ -126,6 +127,9 @@ export class Remote {
       this.registerLabelFormatter(remoteAuthority, this.storage.workspace.owner_name, this.storage.workspace.name),
     )
 
+    // Initialize any WorkspaceAction notifications (auto-off, upcoming deletion)
+    const action = await WorkspaceAction.init(this.vscodeProposed, this.storage)
```
code-asher, line 131: It makes sense to me that we have this in `Remote` since `Remote` activates once you actually connect to a workspace and there seems little value in notifying a workspace is going offline if it is not being actively used but that makes me wonder if we want to filter the notifications by the currently connected workspace?  Currently if I connect to some other random workspace I get notified about other workspaces shutting down.

But, I am not sure how easy it is to tell what the currently connected workspace is.

If we do want to scope to the current workspace I think doing it as a followup at some point would make sense.

code-asher (approved):
Works like a charm!

src/remote.ts
Kira-Pilot: I think that makes a lot of sense with, say, an auto-stop notification and we might want to tweak it in the future, especially as I don't want to bombard users with messaging.
The reason we didn't scope notifications more narrowly for this first pass is because we want workspace owners to be informed about inactive workspaces that are about to be locked or deleted (locked status coming in follow-up PR). Naturally, they won't be using these workspaces; they will be inactive.
Down the road, it would be nice to have users be able to toggle notification settings for specific workspaces or turn them off entirely. Not sure how accommodating the VS Code API will be here.

src/WorkspaceAction.ts
Kira-Pilot: Hmm, good question. Honestly, this is a pattern I copied from elsewhere in the repo without much consideration and I'm not getting a lot of inspiration from the limited [Axios documentation](https://github.com/axios/axios/blob/main/README.md).
I'm thinking it makes the most sense to simplify and take it out - the error message is broadly applicable and if the workspace isn't connectable at all, we should bail out before we ever get to this gate.

Kira-Pilot: Ha, I think were following this pattern in the past to silence TS. I'll use a different pattern.

src/remote.ts
code-asher: Makes sense!

Issue #116, opened 2023-07-21
Link to docs when the server has a partial chain instead of "allow insecure" prompt loop


bpmct:
Description in here: https://github.com/coder/vscode-coder/issues/105#issuecomment-1645999830

PR #118, opened 2023-08-01
Improve UX around certificate errors

code-asher:
Unfortunately the insecure bypass does not seem to work so I removed the button for now.  Interestingly, according to the tests I added it should actually work for all cases including the partial chain case, but for some reason in VS Code it appears to simply do nothing?

In addition to the tests, I manually ran the plugin in VS Code against domains provided by badssl.com.  The only one I have not tested manually yet is the missing key usage one since badssl.com has no sample for that but there is a test for it.

Closes https://github.com/coder/vscode-coder/issues/116


src/error.ts
```diff
@@ -1,51 +1,157 @@
...
+}
+
+// X509_ERR contains human-friendly versions of TLS errors.
+export enum X509_ERR {
+  PARTIAL_CHAIN = "Your Coder deployment's certificate cannot be verified because a certificate is missing from its chain. To fix this your deployment's administrator should bundle the missing certificates or you can add the missing certificates directly to this system's trust store.",
+  // NON_SIGNING can be removed if BoringSSL is patched and the patch makes it
+  // into the version of Electron used by VS Code.
+  NON_SIGNING = "Your Coder deployment's certificate is not marked as being capable of signing. VS Code uses a version of Electron that does not support certificates like this even if they are self-issued. The certificate should be regenerated with the certificate signing capability.",
+  UNTRUSTED_LEAF = "Your Coder deployment's certificate does not appear to be trusted by this system. The certificate should be added to this system's trust store.",
+  UNTRUSTED_CHAIN = "Your Coder deployment's certificate chain does not appear to be trusted by this system. The root of the certificate chain should be added to this system's trust store. ",
```
code-asher, line 20: @bpmct the issue https://github.com/coder/vscode-coder/issues/116 says to link to docs but do you think these instructions are sufficient?  Or do you think we will also need a docs page that goes into more detail, like what exactly it means to bundle, how to serve a bundled certificate, how to add to the trust store to each operating system, how to add the key signing capability, that sort of thing?

code-asher: Alternatively we could offload all the work to docs, basically if we get any TLS error we link to the docs and then in the docs we describe how to figure out what the problem is with your certificate (rather than trying to detect it automatically here) and how to fix it.  That might be a bit more flexible.

kylecarbs (approved):
These tests are nuts and I love it.

src/error.ts
code-asher: Merging for now, but happy to make changes later!

PR #119, opened 2023-08-04
Add customizable header command

code-asher:
The command will be executed before requests and before configuring SSH to set custom headers.

Some notes:

1. When I use `--header` I seem to get the value set twice.  For example if I use `coder --header X-MY-CUSTOM-HEADER=hello-there` then on the server end I see `x-my-custom-header: hello-there, hello-there` so we should look into that.
2. I went with environment variables to pass data to the credentials process for now since I figured that would be easier than parsing command-line arguments but opinions are welcome here.  Only `CODER_URL` right now.  Are there others needed?
3. ~More importantly, I am not sure about the config SSH portion.  Since configuration might only happen on the first connect, the headers we hardcode into the SSH config could expire.~

~I have two main ideas for fixing the SSH config issue:~

1. ~Modify the CLI to call out to the same process to get the headers.  `--credentials-process` or something like that.  The CLI executes the process just as the extension does and sets the headers.  We set `--credentials-process` in the SSH config on the plugin side and are off to the races.~
2. ~Add a second VS Code setting for a proxy command wrapper which gets passed the path to the Coder binary and is responsible for adding the appropriate `--header` flags and passing through any other flags.  This way this credential process is called every time SSH is invoked.  Or instead of a separate setting it could be an environment variable/flag and we reuse the same process; it can switch behavior depending on what task it needs to do (output JSON or execute the Coder binary).~

~An example of the second option: `ProxyCommand my-wrapper --wrapper-flags -- --coder-flags` where `my-wrapper` is essentially `exec $CODER_BINARY --header my-custom-header "$@"` except everything after `--` should get passed through to the Coder binary and the rest can be flags to the wrapper itself in case we need them.~

~Thoughts?  I am inclined to go the first route since the second requires more complexity on the user side but the second does at least have the advantage (or danger) of allowing more control over how the Coder binary executes.~

~I also considered substitution in the proxy command (something similar to `coder $(/path/to/credentials-process) ssh dev.dev --stdio`) but I am not sure that is portable.~

Sorted the above, `--header-command` has been added to the CLI although we will need the next release before this can be used in the plugin.

Closes https://github.com/coder/vscode-coder/issues/76 and supersedes https://github.com/coder/vscode-coder/pull/81.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -46,7 +46,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   axios.interceptors.response.use(
     (r) => r,
     async (err) => {
-      throw await CertificateError.maybeWrap(err, err?.config?.baseURL, storage)
+      throw await CertificateError.maybeWrap(err, axios.getUri(err.config), storage)
     },
   )
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -61,7 +61,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 
   // Add headers from the header command.
   axios.interceptors.request.use(async (config) => {
-    Object.entries(await storage.getHeaders()).forEach(([key, value]) => {
+    Object.entries(await storage.getHeaders(config.baseURL || axios.getUri(config))).forEach(([key, value]) => {
       config.headers[key] = value
     })
     return config
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -61,13 +61,10 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 
   // Add headers from the header command.
   axios.interceptors.request.use(async (config) => {
-    return {
-      ...config,
-      headers: {
-        ...(await storage.getHeaders()),
-        ...creds.headers,
-      },
-    }
+    Object.entries(await storage.getHeaders()).forEach(([key, value]) => {
+      config.headers[key] = value
+    })
+    return config
   })
 
   const myWorkspacesProvider = new WorkspaceProvider(WorkspaceQuery.Mine, storage)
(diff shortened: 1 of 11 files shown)


kylecarbs:
Option 1 seems correct to me as well.

For the naming, should we call this `proxy-credential-process` or something else instead? `credential-process` sounds idiomatic, but could easily be confused for Coder credentials instead of middleware credentials.

code-asher:
Or `--config-process` if we think we might need to be more generic.

liorb-canva:
@code-asher option 1 seems fine as well for us

code-asher:
I am not sure if we need to check the Coder version before setting the new flag but since it only sets if you provide the value maybe it is OK to just rely on the user only adding it if their version supports the new flag?

kylecarbs (approved):

src/extension.ts
```diff
@@ -59,6 +59,17 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   const storage = new Storage(output, ctx.globalState, ctx.secrets, ctx.globalStorageUri, ctx.logUri)
   await storage.init()
 
+  // Add headers from the header command.
+  axios.interceptors.request.use(async (config) => {
+    return {
+      ...config,
+      headers: {
+        ...(await storage.getHeaders()),
+        ...creds.headers,
```
liorb-canva: hey @code-asher  this does not compile..
I've tried to run it locally and it failed

code-asher: Yeah just realized I messed this up, fixing now.

code-asher: Pushed a fix!

code-asher: Turns out I had another issue where the URL in storage was not set yet (for the login request) resulting in no headers.  I think everything is working now, will test a bit more then merge this in.  Let me know if you find anything else wrong!

PR #128, opened 2023-08-21
feat: add options to support client tls

### LINKED ISSUE #126: add configuration options to support mtls
This is related to https://github.com/coder/coder/issues/8889

We run the coder api behind a TLS proxy and would like to be able to specify the client cert/key on the vscode plugin requests.

It appears we just need to add `cert` and `key` options to the `axios` default client:
https://github.com/coder/vscode-coder/blob/main/src/extension.ts#L39-L43

I hacked on the plugin a bit it seems these were the only changes necessary:
```
  const applyInsecure = () => {
    const cfg = vscode.workspace.getConfiguration()
    const insecure = Boolean(cfg.get("coder.insecure"))
    const certFile = String(cfg.get("coder.certFile"))
    const keyFile = String(cfg.get("coder.keyFile"))
    const caFile = String(cfg.get("coder.caFile"))

    axios.defaults.httpsAgent = new https.Agent({
      cert: certFile === "" ? undefined : fs.readFileSync(certFile),
      key: keyFile === "" ? undefined : fs.readFileSync(keyFile),
      ca: caFile === "" ? undefined : fs.readFileSync(caFile),
      // rejectUnauthorized defaults to true, so we need to explicitly set it to false
      // if we want to allow self-signed certificates.
      rejectUnauthorized: !insecure,
    })
  }
```

I know little typescript, so there is likely a better way to do it, and we would likely want to change the `applyInsecure` name for the expanded scope.

coryb:
adding configuration properties:
  * coder.tlsCertFile
  * coder.tlsKeyFile
  * coder.tlsCaFile

fixes #126

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,21 @@
           "markdownDescription": "An external command that outputs additional HTTP headers added to all requests. The command must output each header as `key=value` on its own line. The following environment variables will be available to the process: `CODER_URL`.",
           "type": "string",
           "default": ""
+        },
+        "coder.tlsCertFile": {
+          "markdownDescription": "Path to file for TLS client cert",
+          "type": "string",
+          "default": ""
+        },
+        "coder.tlsKeyFile": {
+          "markdownDescription": "Path to file for TLS client key",
+          "type": "string",
+          "default": ""
+        },
+        "coder.tlsCaFile": {
+          "markdownDescription": "Path to file for TLS certificate authority",
+          "type": "string",
+          "default": ""
         }
       }
     },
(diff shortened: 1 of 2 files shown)


kylecarbs (approved):

kylecarbs:
@code-asher why might this fail?

code-asher:
Ah I think the certificate fixtures I generated expired, for some reason I just assumed they had no expiration if I never specified one but apparently there is a default.  I will change it to either generate them as part of the test or make the expiration really far in the future.

code-asher (approved):

Issue #173, opened 2023-11-28
Logging out or connected to different deployment causes url load to fail with no such file or directory


spikecurtis:
When logged out of the Coder extension or connecting to a different deployment of the Coder than the one previously logged in. I get the below error with VSCode not even trying to connect.

I think the relevant error is:

```console
[14:48:57.703] Running script with connection command: ssh -T -D 51504 -o ConnectTimeout=20 'coder-vscode--matifali--coder' bash
[14:48:58.001] > Encountered an error running "coder vscodessh"
> read url: open /Users/matifali/Library/Application Support/Code/User/globalStorage/coder.coder-remote/url: no such file or directory
> kex_exchange_identification: Connection closed by remote host
```

Contents of .ssh/config:

```
Host github.com
  AddKeysToAgent yes
  UseKeychain yes
  IdentityFile ~/.ssh/id_ed25519

# ------------START-CODER-----------
# This section is managed by coder. DO NOT EDIT.
#
# You should not hand-edit this section unless you are removing it, all
# changes will be lost when running "coder config-ssh".
#
Host coder.coder
	HostName coder.coder
	ConnectTimeout=0
	StrictHostKeyChecking=no
	UserKnownHostsFile=/dev/null
	LogLevel ERROR
	ProxyCommand /opt/homebrew/bin/coder --global-config "/Users/matifali/Library/Application Support/coderv2" ssh --stdio coder
Host coder.coder.dev
	HostName coder.coder.dev
	ConnectTimeout=0
	StrictHostKeyChecking=no
	UserKnownHostsFile=/dev/null
	LogLevel ERROR
	ProxyCommand /opt/homebrew/bin/coder --global-config "/Users/matifali/Library/Application Support/coderv2" ssh --stdio coder.dev
# ------------END-CODER------------


# --- START CODER JETBRAINS dev.coder.com
Host coder-jetbrains--coder.dev--dev.coder.com
  HostName coder.coder.dev
  ProxyCommand "/Users/matifali/Library/Application Support/coder-gateway/dev.coder.com/coder-darwin-arm64" --global-config "/Users/matifali/Library/Application Support/coder-gateway/dev.coder.com/config" ssh --stdio coder.dev
  ConnectTimeout 0
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
  LogLevel ERROR
  SetEnv CODER_SSH_SESSION_TYPE=JetBrains
# --- END CODER JETBRAINS dev.coder.com

[...]

PR #174, opened 2023-11-28
Ensure we have a URL

code-asher:
It looks like when we switched to using the VS Code desktop module the URL was no longer included, which caused the extension to query localhost if you had never logged in before, which errors and causes the configuration to never happen.

I will fix the module as well, but I figure we can check the URL in the plugin just in case.  **Edit**: module PR: https://github.com/coder/modules/pull/107

Fixes https://github.com/coder/coder/issues/10896

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -8,6 +8,34 @@ import { Remote } from "./remote"
 import { Storage } from "./storage"
 import { OpenableTreeItem } from "./workspacesProvider"
 
+// maybeAskUrl asks the user for the URL if it was not provided and normalizes
+// the returned URL.
+export async function maybeAskUrl(
+  providedUrl: string | undefined | null,
+  lastUsedUrl?: string,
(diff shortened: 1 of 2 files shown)


matifali:
@code-asher, Does this also close #173? And do we now support the seamless transition between multiple coder deployments?

code-asher:
It will partially support multiple deployments.  If you use the "VS Code desktop" button in one deployment, then you use it in another deployment, that should work since the plugin will re-configure SSH for the new deployment.

But, if you then try to re-open the previous deployment (for example from the recent menu) it will not work, because the plugin is now configured for the new deployment.

To truly support multiple deployments each deployment needs their own host in the SSH config, but right now they all use the same host `coder-vscode--*` so they clobber each other.

Either that or we need some way of associating a URL with a recent entry so we can re-configure SSH every time it is connected.

PR #176, opened 2023-12-06
feat: expand `${userHome}` in tls settings paths

coryb:
For parity with the jetbrains-coder plugin allow for HOME directory expansion in the path locations for tls configuration settings.  Using the `${userHome}` variable as this is used in other VSCode configurations like launch.json.


coryb:
@code-asher would appreciate a review of this small PR.  Thanks!

PR #194, opened 2024-01-26
Override connect timeout

code-asher:
See https://github.com/coder/coder/pull/10473

After this change you can see in the logs:

```
[13:50:25.629] > local-server-1> Running ssh connection command: "-v -T -D 43233 -o ConnectTimeout=1800 coder-vscode--code-asher--dev27--dev bash"
```

### DIFF
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -276,6 +276,7 @@ export class Remote {
     // VS Code ignores the connect timeout in the SSH config and uses a default
     // of 15 seconds, which can be too short in the case where we wait for
     // startup scripts.  For now we hardcode a longer value.
+    // If microsoft/vscode-remote-release#8519 is resolved we can remove this.
     const connectTimeout = 1800
 
     let settingsContent = "{}"
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -265,20 +265,17 @@ export class Remote {
       agent = matchingAgents[0]
     }
 
-    let remotePlatforms = this.vscodeProposed.workspace
+    const hostname = authorityParts[1]
+    const remotePlatforms = this.vscodeProposed.workspace
       .getConfiguration()
-      .get<Record<string, string>>("remote.SSH.remotePlatform")
-    remotePlatforms = {
-      ...remotePlatforms,
-      [`${authorityParts[1]}`]: agent.operating_system,
-    }
+      .get<Record<string, string>>("remote.SSH.remotePlatform", {})
+    const connTimeout = this.vscodeProposed.workspace.getConfiguration().get<number | undefined>("remote.SSH.connectTimeout")
 
-    // VS Code ignores the connect timeout in the SSH config and uses a default
-    // of 15 seconds, which can be too short in the case where we wait for
-    // startup scripts.  For now we hardcode a longer value.
-    // If microsoft/vscode-remote-release#8519 is resolved we can remove this.
-    const connectTimeout = 1800
-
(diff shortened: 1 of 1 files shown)


src/remote.ts
```diff
@@ -273,17 +273,30 @@ export class Remote {
       [`${authorityParts[1]}`]: agent.operating_system,
     }
 
+    // VS Code ignores the connect timeout in the SSH config and uses a default
+    // of 15 seconds, which can be too short in the case where we wait for
+    // startup scripts.  For now we hardcode a longer value.
+    const connectTimeout = 1800
```
mafredri: Btw, `1800` was my best guess at a time that should be enough for most people (30 minutes), whilst avoiding being infinite. This is because changing this setting will affect non-Coder use of the Remote SSH extension too.

If `1800` is too low, we can always bump it, too. And if users don't like us changing this, we can implement some improved conditional logic, I think.

One idea I had was to read in the setting, establish connection, and then reset the setting to the old value. But constantly touching users settings also feels wrong.

PS. Should we mention https://github.com/microsoft/vscode-remote-release/issues/8519 here?

src/remote.ts
```diff
@@ -273,17 +273,30 @@ export class Remote {
...
-    const parsed = jsonc.parse(settingsContent)
-    parsed["remote.SSH.remotePlatform"] = remotePlatforms
-    const edits = jsonc.modify(settingsContent, ["remote.SSH.remotePlatform"], remotePlatforms, {})
+
+    settingsContent = jsonc.applyEdits(
+      settingsContent,
+      jsonc.modify(settingsContent, ["remote.SSH.remotePlatform"], remotePlatforms, {}),
+    )
+
+    settingsContent = jsonc.applyEdits(
```
mafredri: Should we add a note that we're overwriting user settings here? It's not pretty but I don't think there's a lot of better ways we can do this, unless we implement logic to prompt the user to update it, and keep track off if we've changed it or not `¯\_(ツ)_/¯`.

src/remote.ts
code-asher: That makes sense!  Yeah, touching the settings like this does feel pretty janky, but I think you are right about trying to reset it after connecting.  Will add.

Also, I am not completely sure why we are directly munging it with jsonc, I will try going through the extension API since that might feel a bit less sketchy.

Good call on adding the issue link!

src/remote.ts
code-asher: I think maybe if we go through the API it will at least feel a bit less sketchy, so I will try that (and will also add a comment).

code-asher: Nevermind, trying to update settings properly through the extension API inexplicably hangs indefinitely.

mafredri: That’s too bad, it was a nice idea!

code-asher:
Made the changes!  One open question is whether there is a better way to detect when we are connected.  I am using `findSSHProcessID()` but I suppose we could reset the value early if we are unable to find it or it takes too long to spawn.  It might be good enough though.

mafredri (approved):
Nice changes! Good point about SSH process, would it be possible that it exits very fast and won't be found? Maybe we could use a race between find process and a timeout promise. But if we do, we should make sure the timeout starts after we've activated the Remote SSH extension.

code-asher:
Yup it would be possible that it exits fast, but `findSSHProcessID()` does also have a timeout (15 seconds) so it will eventually resolve and then we will revert the setting.  The remote extension activates right after we set this up so I think it is as good as we will get it!

Issue #182, opened 2023-12-17
URL should not be  required on login when it is predefined  on global storage


liorb-canva:
**Context:**
For self hosteed coder server, a private url needed to be provided in login screen
This url is not alweys intuative and easy to remember

**Accepted behaviour**
If the URL is defined in global config or in the `CODER_URL` environment variable, do not show the user a prompt to fill the url, and redirect to the pre-defined url to login.

**Current Behaviour**
The login function, ignored the url stored in `Storage` class
![Screenshot 2023-12-18 at 10 15 38 am](https://github.com/coder/vscode-coder/assets/125318364/b04176a1-3488-4d98-b29d-b691322bda86)

matifali:
While doing this, make sure it can also handle multiple coder deployments gracefully.

code-asher:
I think we might still want to show the login prompt but just prefill it, otherwise what if you want to connect to a different deployment but it keeps auto-logging you into a different one?

matifali:
The best would be to add the ability to list recently used URLs and prefilled the last connected one.

PR #200, opened 2024-01-30
Add URL history and CODER_URL

code-asher:
We default to the last used URL or CODER_URL, if set, and you can choose from a list of past URLs.

Closes https://github.com/coder/vscode-coder/issues/182.

In this screenshot the "prefilled" value is from CODER_URL and dev.coder.com is from a past connection.

screenshot

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -18,7 +18,7 @@ export class Commands {
    * Ask the user for the URL, letting them choose from a list of recent URLs or
    * CODER_URL or enter a new one.  Undefined means the user aborted.
    */
-  private askURL(selection?: string): Promise<string | undefined> {
+  private async askURL(selection?: string): Promise<string | undefined> {
     const quickPick = vscode.window.createQuickPick()
     quickPick.value = selection || process.env.CODER_URL || ""
     quickPick.placeholder = "https://example.coder.com"
@@ -42,10 +42,12 @@ export class Commands {
 
     quickPick.show()
 
-    return new Promise<string | undefined>((resolve) => {
+    const selected = await new Promise<string | undefined>((resolve) => {
       quickPick.onDidHide(() => resolve(undefined))
       quickPick.onDidChangeSelection((selected) => resolve(selected[0]?.label))
     })
+    quickPick.dispose()
+    return selected
   }
 
   /**
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -8,42 +8,79 @@ import { Remote } from "./remote"
 import { Storage } from "./storage"
 import { OpenableTreeItem } from "./workspacesProvider"
 
-// maybeAskUrl asks the user for the URL if it was not provided and normalizes
-// the returned URL.
-export async function maybeAskUrl(
-  providedUrl: string | undefined | null,
-  lastUsedUrl?: string,
-): Promise<string | undefined> {
-  let url =
-    providedUrl ||
-    (await vscode.window.showInputBox({
-      title: "Coder URL",
-      prompt: "Enter the URL of your Coder deployment.",
-      placeHolder: "https://example.coder.com",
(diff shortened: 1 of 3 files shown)


src/commands.ts
```diff
@@ -8,42 +8,79 @@ import { Remote } from "./remote"
...
     private readonly storage: Storage,
   ) {}
 
+  /**
+   * Ask the user for the URL, letting them choose from a list of recent URLs or
+   * CODER_URL or enter a new one.  Undefined means the user aborted.
+   */
+  private askURL(selection?: string): Promise<string | undefined> {
+    const quickPick = vscode.window.createQuickPick()
+    quickPick.value = selection || process.env.CODER_URL || ""
```
Kira-Pilot, line 23: Do we need to validate the user-entered URL in any way?

src/commands.ts
```diff
@@ -8,42 +8,79 @@ import { Remote } from "./remote"
...
+    // Initial items.
+    quickPick.items = this.storage.withUrlHistory(process.env.CODER_URL).map((url) => ({
+      alwaysShow: true,
+      label: url,
+    }))
+
+    // Quick picks do not allow arbitrary values, so we add the value itself as
+    // an option in case the user wants to connect to something that is not in
+    // the list.
+    quickPick.onDidChangeValue((value) => {
```
Kira-Pilot, line 36: What happens if the user enters a duplicate or partial URL?

src/commands.ts
```diff
@@ -8,42 +8,79 @@ import { Remote } from "./remote"
...
+      quickPick.items = this.storage.withUrlHistory(process.env.CODER_URL, value).map((url) => ({
+        alwaysShow: true,
+        label: url,
+      }))
+    })
+
+    quickPick.show()
+
+    return new Promise<string | undefined>((resolve) => {
+      quickPick.onDidHide(() => resolve(undefined))
```
Kira-Pilot, line 46: Do we need to use the `dispose` handler here? I see them using that in [their example](https://github.com/microsoft/vscode-extension-samples/blob/main/quickinput-sample/src/extension.ts#L27), but not sure exactly what it cleans up.

Kira-Pilot (approved):
Couple of questions but approving so I'm not blocking you! Thanks for looking at this.

src/commands.ts
code-asher: `withUrlHistory` removes duplicates (using a fancy `Set`).

src/commands.ts
code-asher: Good call, I had no idea!  I am not sure either but it seems smart to call it.  Will fix.

src/commands.ts
code-asher: Good question, thinking about it we could validate that the characters are allowable in a URL and we could also maybe check if the protocol is not something unexpected like `file:///` although I think it might already be enough that these "bad" URLs will fail in the login step before we store the URL.

Kira-Pilot: Ahh that seems fine then!

src/commands.ts
code-asher: Oh and a partial URL will show at the top of the list, like this:

screenshot

Here is the duplicate removal at work, if you type something that matches, it only shows it once:

screenshot

code-asher: One improvement we could make is that if they omit the `https://` we could consider it still the same, right now it would show twice, even though if they accept the `dev.coder.com` option we will add `https://` to it (when we save the URL history we will only save one though).

screenshot

Kira-Pilot: That seems reasonable; what you have right now isn't offensive, either. I assume everyone is using `https` but I suppose there's a chance someone could add `dev.coder.com` and want to use a less secure protocol.

PR #202, opened 2024-02-01
Prevent parallel refreshes

code-asher:
While adding an automatic refresh I discovered it was easy to
accidentally cause two refreshes to go off at the same time which causes
some races.

Also added a minor tweak to pass in the URL instead of pulling from the
global state since that felt race-y too.

### DIFF
--- a/src/workspacesProvider.ts
+++ b/src/workspacesProvider.ts
@@ -67,7 +67,7 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
       const showMetadata = this.getWorkspacesQuery === WorkspaceQuery.Mine
       if (showMetadata) {
         const agents = extractAgents(workspace)
-        agents.forEach((agent) => this.monitorMetadata(agent.id, token)) // monitor metadata for all agents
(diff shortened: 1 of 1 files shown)


src/workspacesProvider.ts
```diff
@@ -22,26 +23,54 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
...
+    const url = this.storage.getURL()
+    const token = await this.storage.getSessionToken()
+    if (!url || !token) {
+      throw new Error("not logged in")
+    }
+
+    const resp = await getWorkspaces({ q: this.getWorkspacesQuery })
+
+    // We could have logged out while waiting for the query.
+    if (!url || !token) {
```
Kira-Pilot: Can `url` or `token` change while `getWorkspaces` is fetched, i.e. do we have to fetch them from storage again to assert they're really there?

Kira-Pilot (approved):

src/workspacesProvider.ts
code-asher: Holy smokes I completely missed that.  Good catch!  Will fix.

Issue #184, opened 2024-01-04
periodically refresh workspace listing


stirby:
The "My Workspaces" tab falls out of sync with the workspaces in a deployment without manual reload. This can present misleading or inaccurate listings unless the user remembers to keep it up-to-date.

A customer requested that this tab automatically refreshes periodically. 

 
<img width="306" alt="Screenshot 2024-01-04 at 2 54 08 PM" src="https://github.com/coder/vscode-coder/assets/58410745/adfe7c5a-1f90-4617-95f1-43c958e1195a">

PR #203, opened 2024-02-01
Automatically refresh workspaces

code-asher:
Closes https://github.com/coder/vscode-coder/issues/184.

One existing issue with refreshing that is made more apparent with the auto refresh is that it reestablishes all the agent metadata watchers which both feels wasteful and causes the metadata to flash as it disappears for a moment then comes back when the watcher comes online.  In a followup PR I think I will refactor this to reuse existing watchers when possible.

Stacked on https://github.com/coder/vscode-coder/pull/202.

### DIFF
--- a/src/workspacesProvider.ts
+++ b/src/workspacesProvider.ts
@@ -45,16 +45,22 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
     // needs to be cleared.
     this.cancelPendingRefresh()
 
+    let hadError = false
     try {
       this.workspaces = await this.fetch()
-      this.fetching = false
-      this.maybeScheduleRefresh()
     } catch (error) {
+      hadError = true
       this.workspaces = []
-      this.fetching = false
     }
 
+    this.fetching = false
+
     this.refresh()
+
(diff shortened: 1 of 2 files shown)


src/workspacesProvider.ts
```diff
@@ -14,34 +14,47 @@ export enum WorkspaceQuery {
...
+    // needs to be cleared.
+    this.cancelPendingRefresh()
+
     try {
       this.workspaces = await this.fetch()
+      this.fetching = false
+      this.maybeScheduleRefresh()
     } catch (error) {
       this.workspaces = []
+      this.fetching = false
```
Kira-Pilot: Should we move this to a `finally` block?

Kira-Pilot (approved):

src/workspacesProvider.ts
code-asher: I really wanted to move it, but it has to be set before calling `this.maybeScheduleRefresh()` (since it checks whether a fetch is currently underway) so a `finally` would be too late.  So either way it would have to be set twice.

But it feels brittle.  I think maybe I refactor a bit to only set it once, add an error boolean or something.

PR #204, opened 2024-02-02
Reuse agent metadata watchers

code-asher:
Followup PR to reuse watchers to make the automatic refresh less janky.

This has two effects:

1. Lower overhead from establishing the watchers over and over.
2. No more flashing in the tree as the watchers get recreated and we wait for the first update.

I renamed `CoderTreeItemType` to make clear it was for openable items only, not for the metadata items and not for the new error item.

Because I needed to refactor the watchers a little I added a line for when something with the metadata fails, rather than quietly disposing (tested by manually adding `throw new Error("hey")`).

screenshot

### DIFF
--- a/src/api-helper.ts
+++ b/src/api-helper.ts
@@ -1,12 +1,26 @@
 import { Workspace, WorkspaceAgent } from "coder/site/src/api/typesGenerated"
 import { z } from "zod"
 
+export function errToStr(error: unknown, def: string) {
+  if (error instanceof Error && error.message) {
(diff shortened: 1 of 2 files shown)


code-asher:
@Kira-Pilot Just FYI I was going to push this directly to main to get a test release out, but I decided to open a PR first and merge it in that way so it could be reviewed later just in case I missed something!

Kira-Pilot (commented):
This looks solid to me!!

Issue #210, opened 2024-02-09
`Host *` blocks overwritten in SSH config files, leading to disconnects


ericpaulsen:
a large customer identified an issue where our VS Code extension overwrites `Host *` entries in SSH configs, instead of respecting them (`coder config-ssh` [does this here](https://github.com/coder/coder/blob/main/cli/configssh.go#L720S)).

> 
> There's a Host * block in our default SSH config and the VSCode plugin was putting it's blocks after those. It seemed as those even after moving the blocks around, the plugin would remove the blocks and place them at the end of the file.
> 
> We have a sample config file that includes empty comment markers for coder's config-ssh and VS Code blocks above a Host *. config-ssh obeys the placement but VS Code does not.

this line of code is where the `Host *` block is appended to the end of the config file: https://github.com/coder/vscode-coder/blob/main/src/sshConfig.ts#L188

the customer confirmed that commenting out the `Host *` block did resolve the issue.

> First we tried just moving the Host * to the end of the file, but that got moved right back on the next connection. Eventually I just told them to comment out that block, which isn't ideal

ericpaulsen:
to compare, OpenSSH searches the file from top to bottom and uses the parameter from the first block that matches, so if there's a block earlier in the file that has a parameter that's also in the VSCode block, it'll use the earlier block's parameter:

```
Host *
  foo star

Host vscode
  foo vscode
  bar vscode
```

this shakes out to be `{foo: star, bar: vscode}`

PR #211, opened 2024-02-09
fix: replace ssh config block in-place

mafredri:
This PR introduces two changes:
1. Replace vscode block in-place in the SSH config
2. Adds an additional guard against replacing user-added blocks that _look_ like an old-form SSH config block (without block markings)

TBH, we could probably remove the old block detection entirely instead of this additional check, but didn't want to make that decision here.

### DIFF
--- a/src/sshConfig.test.ts
+++ b/src/sshConfig.test.ts
@@ -174,6 +174,40 @@ Host coder-vscode--*
   })
 })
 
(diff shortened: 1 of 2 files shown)


ericpaulsen:
related issue: #210

kylecarbs (approved):

PR #221, opened 2024-03-01
Enhancing SSH Connection Compatibility with Cursor

RyousukeUchino:
## Overview

I have been a dedicated user of both Coder and Cursor. However, I've encountered persistent issues when attempting to connect to the Coder workspace via Cursor ([Cursor](https://cursor.sh/)). These issues range from receiving 404 errors to SSH connection failures, preventing any access to the workspace. The root cause of these problems lies in the necessity to use Cursor's built-in Open Remote - SSH extension, `anysphere.open-remote-ssh`, for successful SSH connections.

## Proposed Changes

To address this issue, I propose two modifications:

1. **Remove Remote SSH from Extension Dependencies**: This change aims to eliminate the direct dependency on the Remote SSH extension, which conflicts with Cursor's operation.
2. **Support for `anysphere.open-remote-ssh` in Extension Acquisition**: By officially supporting `anysphere.open-remote-ssh`, we ensure a smoother integration and functionality for users relying on Cursor for their SSH connections.

## Considerations

I acknowledge that these changes might introduce confusion for users who do not use Cursor, as it could affect the automatic resolution of dependencies. Finding a more elegant solution that accommodates all users would be ideal. Nonetheless, my goal is to enable seamless operation with Cursor by incorporating support for `anysphere.open-remote-ssh`.

**Note**: To use `anysphere.open-remote-ssh`, it is necessary to uninstall `ms-vscode-remote.remote-ssh`.

I look forward to feedback and suggestions on this proposal to enhance our workspace connectivity and user experience.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## [v0.1.34](https://github.com/coder/vscode-coder/releases/tag/v0.1.34) (2024-03-03)
+
+### Changes
+
+- Improve fetching the Coder binary. This is mostly just better logging but it
+  also will avoid fetching if the existing binary version already matches, to
+  support scenarios where the ETag is ignored.
+
 ## [v0.1.33](https://github.com/coder/vscode-coder/releases/tag/v0.1.33) (2024-02-20)
 
 ### Bug fixes
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "displayName": "Coder",
   "description": "Open any workspace with a single click.",
   "repository": "https://github.com/coder/vscode-coder",
-  "version": "0.1.33",
+  "version": "0.1.34",
   "engines": {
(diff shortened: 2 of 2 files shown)


code-asher:
I think we are not able to merge this because we want a one-click experience from the Coder dashboard, and that means we need the extension dependencies to automatically install, which I think only happens if we have that key in the `package.json`.

Pretty cool they made their own version of the Remote SSH plugin though.  I would love to see that become open source and completely take over the official plugin, then we could just use it by default!

We can merge the code portion, but I am not sure how to handle the `package.json`.  Maybe we will need to repackage the extension, remove that part from the package.json, then distribute it separately as `coder-cursor` or something.  Or maybe this is even something that should be part of `Cursor` itself?  They could automatically replace the `"ms-vscode-remote.remote-ssh"` dependency with their own when they load extensions.

PR #225, opened 2024-03-06
Enhancing SSH Connection Compatibility with Cursor

RyousukeUchino:
Previous MR https://github.com/coder/vscode-coder/pull/221

## Overview

I have been a dedicated user of both Coder and Cursor. However, I've encountered persistent issues when attempting to connect to the Coder workspace via Cursor ([Cursor](https://cursor.sh/)). These issues range from receiving 404 errors to SSH connection failures, preventing any access to the workspace. The root cause of these problems lies in the necessity to use Cursor's built-in Open Remote - SSH extension, `anysphere.open-remote-ssh`, for successful SSH connections.

### Changes Made

I've confirmed that users utilizing Cursor can simply specify `anysphere.open-remote-ssh` explicitly within the Coder plugin without any issues coexisting with `ms-vscode-remote.remote-ssh`. Therefore, I've made a straightforward update to check for the existence of `anysphere.open-remote-ssh` and use it preferentially if available. This eliminates the need to remove any dependencies.

With this MR, I'm hopeful for a seamless integration between Cursor and Coder, which I personally believe is the best combination.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -22,6 +22,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   const openRemoteSSHExtension = vscode.extensions.getExtension("anysphere.open-remote-ssh")
   // If the "anysphere.open-remote-ssh" extension is available, it is used with priority
   // If it is not found, then the extension "ms-vscode-remote.remote-ssh" is sought as a fallback.
+  // This is specifically for non-official VS Code distributions, such as Cursor.
   const useRemoteSSHExtension = openRemoteSSHExtension
     ? openRemoteSSHExtension
     : vscode.extensions.getExtension("ms-vscode-remote.remote-ssh")
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -19,13 +19,15 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   //
   // This is janky, but that's alright since it provides such minimal
   // functionality to the extension.
-  const msRemoteSSHExtension = vscode.extensions.getExtension("ms-vscode-remote.remote-ssh")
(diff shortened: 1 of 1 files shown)


kylecarbs (changes requested):
Awesome description, thanks for the contribution!

I'd rather we only get the other extension if the official remote extension isn't found.

src/extension.ts
```diff
@@ -19,15 +19,20 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   //
   // This is janky, but that's alright since it provides such minimal
   // functionality to the extension.
-  const remoteSSHExtension = vscode.extensions.getExtension("ms-vscode-remote.remote-ssh")
-  if (!remoteSSHExtension) {
+  const openRemoteSSHExtension = vscode.extensions.getExtension("anysphere.open-remote-ssh")
+  // If the "anysphere.open-remote-ssh" extension is available, it is used with priority
+  // If it is not found, then the extension "ms-vscode-remote.remote-ssh" is sought as a fallback.
```
kylecarbs, line 24: Nit: Let's add a comment to explicitly mention this is for non-official VS Code distributions. Just so in the future we know!

kylecarbs (approved):

code-asher:
Released in 0.1.35!

stobias123:
is there configuration required to make this work? I'm still getting 404 errors when trying to open a workspace with cursor.

kylecarbs:
Seems like something might be wrong with Cursor's Remote SSH here.

I'd be surprised if it's supposed to download Microsoft's official server.

code-asher:
Do you have the open remote SSH extension installed?  Maybe it is using the official one?  You should only have the open remote one installed.

PR #260, opened 2024-05-09
Tweak certificate errors

code-asher:
Closes https://github.com/coder/customers/issues/442

If testing, can use URLs from https://badssl.com (like https://incomplete-chain.badssl.com/).

Tweak error wording and show the detailed error under the input, so it will display in two places now.

screenshot

Also remove the "view more details" button from the notification, for reasons detailed in the commit message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


Kira-Pilot (approved):
🏆 #1

Issue #193, opened 2024-01-26
Add option to disable downloading Coder binary even if out of date or missing


jfhovinne:
- [x] Option to disable downloading
- [ ] If the binary is read-only, automatically skip downloading, maybe output a warning < Decided not to do this at least for now in favor of being explicitly and predictably controlled by settings
- [x] Check the version beforehand in case etags are not supported

In some environments, it is not allowed to download then execute arbitrary binaries; such binaries have to be approved, specifically packaged and then installed from a central application repository through e.g. UI tooling.

#122 adds settings for the Coder binary source and destination; would it be possible to also allow pointing to an existing coder executable (e.g. "C:\Program Files\Coder\bin\coder-windows-amd64.exe"), and disable the remote download?

code-asher:
Yeah we should add an option to disable downloading like our JetBrains plugin has.

That said, the VS Code plugin should only try to download if the binary is the wrong version, so setting `binaryDestination` to `C:\Program Files\Coder\bin\coder-windows-amd64.exe` should work as long as it is the right version.  Is that not the behavior you are seeing?

jfhovinne:
@code-asher No, as I understand it from my testing and checking `storage.fetchBinary()`, the extension downloads the binary and tries to save it in `binaryDestination` (which is read-only), then tries to replace the existing one. This leads to the below error, which might be confusing for the end-user:

`Error: ENOENT: no such file or directory, rename 'C:\Program Files\Coder\bin\coder-windows-amd64.exe.temp-*****' -> 'C:\Program Files\Coder\bin\coder-windows-amd64.exe'`

The extension seems usable afterwards though.

code-asher:
Huh, that is strange, assuming it is the same version (the hash matches, sent via the `If-None-Match` header) the request should result in a 304.

https://github.com/coder/vscode-coder/blob/ae170658581cf1b99eefb3e1d23cf33a091633d3/src/storage.ts#L226-L229

I suppose this implies that the endpoint is returning a 200.

https://github.com/coder/vscode-coder/blob/ae170658581cf1b99eefb3e1d23cf33a091633d3/src/storage.ts#L214-L216

Is the binary being downloaded from somewhere that is not supporting the `If-None-Match` header, maybe?

In the JetBrains plugin one thing we do is check the version before making the request with `coder version --json` in case the source does not support etags, so we might want to do the same here as well.

code-asher:
[...]

PR #263, opened 2024-05-10
Add option to disable downloads

code-asher:
Closes https://github.com/coder/vscode-coder/issues/193

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,11 @@
           "type": "string",
           "default": ""
         },
+        "coder.enableDownloads": {
+          "markdownDescription": "Allow the plugin to download the CLI when missing or out of date.",
+          "type": "boolean",
+          "default": true
+        },
         "coder.headerCommand": {
           "markdownDescription": "An external command that outputs additional HTTP headers added to all requests. The command must output each header as `key=value` on its own line. The following environment variables will be available to the process: `CODER_URL`. Defaults to the value of `CODER_HEADER_COMMAND` if not set.",
           "type": "string",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -63,6 +63,11 @@
           "type": "string",
           "default": ""
         },
+        "coder.enableDownloads": {
+          "markdownDescription": "Allow the plugin to download the CLI when missing or out of date.",
+          "type": "boolean",
+          "default": true
+        },
```
buenos-nachos, line 70: I'm still trying to get up to speed on VS Code extension stuff. I think I get the point of the contribution point properties, but aside from the default values, how do these properties get set over time?

[Looks like it's strictly the `config.update` method?](https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration)

Parkreiner (approved):
Had to go through the VSCode docs a little to see how things are set up, but the changes seem straightforward and make sense to me!!

package.json
code-asher: Yup, you can update it with `config.update()` from plugin code, although we have no need to update it ourselves in this case, we just let it get set via the settings window (which is presumably using `config.update()`).

Funny thing about the default is that it seems to have no actual effect on the code, it only changes what is displayed in the settings window.  In the code you just get `undefined`.

code-asher: Technically you could also munge the settings.json file and VS Code will pick that up but that is a pretty janky way to update a value.  I have seen it done though although I forget why.

Issue #241, opened 2024-04-17
Optionally open last used folder


stirby:
A customer requested that we add an option to open the last used directory in VSCode, rather than the one configured by the agent. 

If VSCode desktop is configured without the folder, it should default to the last-used directory. However, if the coder_agent is given a default directory to use for ssh connections, this will override that behavior.

It would be nice if in the terraform provider we could explicitly enforce opening the last-used directory.

We want to mimic the desktop behavior as closely as possible. 

```hcl
resource "coder_app" "vscode" {
  agent_id     = agent_id
  icon         = "/icon/code.svg"
  slug         = "vscode"
  display_name = "VS Code Desktop"
  order        = var.order
  ...
  open_recent = true # defaults to false
}
```

kylecarbs:
Good idea

bpmct:
This would be huge!

code-asher:
I added it to the VS Code desktop app module, do note though that it will not affect the builtin VS Code desktop app since we expose no options for configuring that.  At least, I think?

Does that seem sufficient or do you think we might need a way to configure this for the built-in app?  Or make opening recent the default for the built-in app?

stirby:
@code-asher that's great. I agree we should direct customers to the registry when possible. 

**However**, the requesting customer can't use our modules (airgapped), so we'd have to support something in the built-in app too. I initially imagined adding a new field under [`display_apps`](https://registry.terraform.io/providers/coder/coder/latest/docs/resources/agent#display_apps) in `coder_agent`; I now realize that's not possible. 

We should just make opening recent the default for the built-in app. 

Ideal behaviour:
- First connection uses the ssh folder configured in the agent
- Subsequent connections use last-used folder

Again, rationale is emulating local VSCode.

code-asher:
Sounds good, I will just make it the default!

PR #272, opened 2024-05-14
Add openRecent query parameter

code-asher:
This covers the plugin side of https://github.com/coder/vscode-coder/issues/241.  We will still need to make changes on the module side.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ There are a few ways you can test the "Open in VS Code" flow:
 
 - Use the "VS Code Desktop" button from a Coder dashboard.
 - Manually open the link with `Developer: Open URL` from inside VS Code.
-- Use `code --open-url <link>` on the command line.
+- Use `code --open-url` on the command line.
 
 The link format is `vscode://coder.coder-remote/open?${query}`. For example:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,28 @@ The `vscodessh` subcommand on the `coder` binary periodically flushes it's netwo
 
 Coder Remote periodically reads the `network-info-dir + "/" + matchingSSHPID` file to display network information.
 
+## Testing
+
+There are a few ways you can test the "Open in VS Code" flow:
+
+- Use the "VS Code Desktop" button from a Coder dashboard.
+- Manually open the link with `Developer: Open URL` from inside VS Code.
+- Use `code --open-url <link>` on the command line.
+
+The link format is `vscode://coder.coder-remote/open?${query}`. For example:
+
+```bash
+code --open-url 'vscode://coder.coder-remote/open?url=dev.coder.com&owner=my-username&workspace=my-ws&agent=my-agent'
+```
+
+There are some unit tests as well:
+
+```bash
+yarn test
+```
+
+However, fully testing the plugin is blocked on switching back to the standard VS Code extension testing framework.
+
 ## Development
 
 1. Run `yarn watch` in the background.
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -407,19 +407,14 @@ async function openWorkspace(
     if (opened.length === 1 || (opened.length > 1 && openRecent)) {
       folderPath = opened[0].folderUri.path
     } else if (opened.length > 1) {
-      const items = opened.map<vscode.QuickPickItem>((folder) => {
-        return {
-          label: folder.folderUri.path,
-        }
-      })
-      const item = await vscode.window.showQuickPick(items, {
+      const items = opened.map((f) => f.folderUri.path)
+      folderPath = await vscode.window.showQuickPick(items, {
         title: "Select a recently opened folder",
       })
-      if (!item) {
+      if (!folderPath) {
         // User aborted.
         return
       }
-      folderPath = opened[items.indexOf(item)].folderUri.path
     }
   }
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -407,7 +407,7 @@ async function openWorkspace(
     if (opened.length === 1 || (opened.length > 1 && openRecent)) {
       folderPath = opened[0].folderUri.path
     } else if (opened.length > 1) {
-      const items: vscode.QuickPickItem[] = opened.map((folder): vscode.QuickPickItem => {
+      const items = opened.map<vscode.QuickPickItem>((folder) => {
         return {
           label: folder.folderUri.path,
         }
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -31,6 +31,28 @@ The `vscodessh` subcommand on the `coder` binary periodically flushes it's netwo
 
 Coder Remote periodically reads the `network-info-dir + "/" + matchingSSHPID` file to display network information.
 
+## Testing
```
code-asher, line 34: I wanted this section below the development section, but when I do that Prettier goes haywire and starts injecting spaces all over the place, leading to insanity like this:

```
- Use the "VS Code Desktop" button from a Coder dashboard.
- Manually    open the link wit   h `Developer: Ope   n URL` from in
    sid   e VS Code.
    - Use    `code --    open-url    <
    link    >` on the     co   mm    and l    ine.

The        
        link format i    s `        v   sc    ode://co    der      
      .cod   er-    r        emote/    ope    n?$        {   qu    ery}
    `        .             For exam               ple:        
    
``     
       `bash
    cod               e                 --ope    n        -url    
             'vs            c           od    e        :/                /coder    
    .        cod                    e            r           -r    e        mo         
           te/ope        n                                            ?u        rl=               d                ev.cod    e                                    r            .c        om            &own                                   e        r=        my        -    u    s                    e        rn    am        e    &w                o               r    k    s                        pa    c    e                    =my    -                ws&    a            ge                                                                           nt=my-agent'
```

src/commands.ts
```diff
@@ -383,36 +387,39 @@ async function openWorkspace(
...
-          title: "Select a recently opened folder",
-        })
-        if (!item) {
-          return
+    // openRecent will always use the most recent.  Otherwise, if there are
+    // multiple we ask the user which to use.
+    if (opened.length === 1 || (opened.length > 1 && openRecent)) {
+      folderPath = opened[0].folderUri.path
+    } else if (opened.length > 1) {
+      const items: vscode.QuickPickItem[] = opened.map((folder): vscode.QuickPickItem => {
```
buenos-nachos: Not a huge deal, but maybe it could be helpful to pass the type parameter into the `.map` call to remove the need to type the array and the callback output?

```ts
const items = opened.map<vscode.QuickPickItem>((folder)  => {
  // Logic goes here
});
```

src/commands.ts
```diff
@@ -383,36 +387,39 @@ async function openWorkspace(
...
+        return {
+          label: folder.folderUri.path,
         }
-        selected = opened[items.indexOf(item)]
-      } else {
-        selected = opened[0]
+      })
+      const item = await vscode.window.showQuickPick(items, {
+        title: "Select a recently opened folder",
+      })
```
buenos-nachos: Making sure I'm understanding this part: is the setup basically that we pass in a list of items to the function, it displays options, and then `item` becomes whichever one was selected?

src/commands.ts
```diff
@@ -369,6 +372,7 @@ async function openWorkspace(
   workspaceName: string,
   workspaceAgent: string | undefined,
   folderPath: string | undefined,
+  openRecent: boolean | undefined,
```
buenos-nachos, line 375: Any particular reason for not using optional parameters via `?:`?

src/commands.ts
```diff
@@ -383,36 +387,39 @@ async function openWorkspace(
...
     const output: {
       workspaces: { folderUri: vscode.Uri; remoteAuthority: string }[]
     } = await vscode.commands.executeCommand("_workbench.getRecentlyOpened")
     const opened = output.workspaces.filter(
-      // Filter out `/` since that's added below.
+      // Remove recents that do not belong to this connection.  The remote
+      // authority maps to a workspace or workspace/agent combination (using the
+      // SSH host name).  This means, at the moment, you can have a different
+      // set of recents for a workspace versus workspace/agent combination, even
+      // if that agent is the default for the workspace.
```
buenos-nachos, line 401: Do we have any docs on how this works? The comment seems well-written, but I feel like I'm still not quite getting it because of missing context

CONTRIBUTING.md
buenos-nachos: I think all documentation should look like that. Make everything like House of Leaves

Parkreiner (approved):
Tentatively approving. I feel like there's some context around how the plugin works that I'm missing, and that could give me a bit more confidence about the changes, but everything _seems_ good

src/commands.ts
code-asher: Oo yeah I like this.

src/commands.ts
code-asher: Honestly I am not sure this is the best way (it was already like this, I just shuffled it around), but yup `showQuickPick` returns the item that was selected (or undefined if canceled).

code-asher: Oh the part I was not sure about, is that we then use the index of that item to get the original item, I think it has to be done this way because we have to give it `QuickPickItem[]`, but I did not check to see if maybe there is a way to use a generic instead, like `{label: "label", item: myItem}` could be nice, then we just return `selected.item.folderUri.path`, but pretty sure there is no option for this.

code-asher: Actually, since the label is the path, seems like we could just return that.  `item.label`.  :thinking:   Or it can take an array of strings instead of `QuickPickItem`, could just do that.  Hold please.

code-asher: Refactored, looks much better!

src/commands.ts
code-asher: Just to make sure it is specified.  I find sometimes when I add new arguments, if I make it optional with `?:` I can miss calls where I should have added it.  But doing it this way forces me to check every function call since they will error without being passed a value.

Once that is done, I might change it to `?:` after some consideration, but in this case since in all our calls to this function we have a value to pass, I did not bother.

src/commands.ts
code-asher: Nope, not that I know of.  I am honestly not really sure how to even write the comment, I rewrote it several times and still feel like I am not really explaining it well.

The problem is that both of these go to the same place:

```
ssh coder-vscode--my-workspace.my-agent
ssh coder-vscode--my-workspace # opens my-agent agent because my-agent is the default or only one
```

But VS Code has no idea these are the same, they are different strings after all.  So depending on which you use you could get a different set of recents.

When clicking from the dashboard we seem to use the second form (without the agent name).

I think probably what we should do is always append the agent name if it is missing, and never use the workspace-only form.  Then this will not be a problem at all.  But for now I just left the comment.

code-asher: And to further clarify, this only arises when using the sidebar in VS Code, where you can both connect to a workspace or directly to an agent.

src/commands.ts
buenos-nachos: Yeah, that makes sense. I think that I've favored the optional syntax for my own convenience up 'til now, but I can see how forcing all the parameters to be explicit would be handy, especially for code stability

PR #279, opened 2024-05-22
fix: escape variables in the header command

davo-canva:
The documentation for the header command setting says:

> The following environment variables will be available to the process: `CODER_URL`.

However, if the command line itself references `$CODER_URL`, when connecting via the VSCode extension the variable will get substituted _before_ the `coder` CLI can set the variable when running the header command, which is contrary to the behaviour of the JetBrains plugin. This is because the VSCode extension is currently writing the header command into the SSH config's `ProxyCommand` with double quotes, allowing the shell spawned by `ssh` to substitute `$CODER_URL` (with the empty string).

This ensures environment variable references within the header command are not substituted before the `coder` binary can receive them, by using single quotes rather than double quotes in the `ProxyCommand`.

Tested on a macOS client.

### DIFF
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -605,13 +605,21 @@ export class Remote {
       }
     }
 
-    const escape = (str: string): string => `'${str.replace(/'/g, "'\\''")}'`
+    const escape = (str: string): string => `"${str.replace(/"/g, '\\"')}"`
+    // Escape a command line to be executed by the Coder binary, so ssh doesn't substitute variables.
+    const escapeSubcommand: (str: string) => string =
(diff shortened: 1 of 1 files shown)


code-asher:
Thank you, and good catch!

I believe we will need to handle Windows differently like we are doing in the JetBrains plugin: https://github.com/coder/jetbrains-coder/blob/main/src/main/kotlin/com/coder/gateway/util/Escape.kt

davo-canva (commented):
Good call. Fixed, but I don't have a Windows machine to test on.

code-asher (approved):
Awesome, thank you!

PR #286, opened 2024-05-30
Refactor REST client

code-asher:
We support connecting to previous workspaces, but sometimes this breaks because the previous workspace belongs to a different deployment, and the plugin always uses the credentials of the current deployment (https://github.com/coder/vscode-coder/issues/173).

This is a first step to fixing that: separating the client into two: one used by the plugin as a whole (mainly used to list workspaces of the current deployment in the sidebar) and one used to connect to a workspace which may belong to a different deployment (which is used to start the workspace, get the version, etc).  Thankfully this is pretty chill since @Parkreiner's changes to Coder core!

For now that second client does still use the global url and token so functionally it is still the same, but I will fix that in a follow up PR by looking at the host name and figuring out the right credentials to use for it.

As part of this, I try to remove usage of a few globals, in particular we had stuff like `storage.getSessionToken()` all over the place, now I only call that once when creating a client and we use it for the lifetime of the task.

## Alternatives

Instead of creating a separate client, we could log the entire plugin in to the right deployment.  So if you are connected to deployment A, then connect to a workspace belonging to deployment B, the plugin would now entirely be logged into deployment B.

That felt a little too unexpected to me, so I went with the multiple client route.

But maybe it would be better to do a full login, because then the workspaces you see in the sidebar belong to the same deployment.

Still, I think these changes are generally good anyway because they are getting rid of a bunch of global stuff.  We can do a full login if we want to with these changes as well.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,0 +1,61 @@
+import { Api } from "coder/site/src/api/api"
+import fs from "fs/promises"
+import * as https from "https"
+import * as os from "os"
+import * as vscode from "vscode"
+import { CertificateError } from "./error"
+import { Storage } from "./storage"
+
+// expandPath will expand ${userHome} in the input string.
+const expandPath = (input: string): string => {
+  const userHome = os.homedir()
+  return input.replace(/\${userHome}/g, userHome)
+}
+
+/**
+ * Create an sdk instance using the provided URL and token and hook it up to
+ * configuration.  The token may be undefined if some other form of
+ * authentication is being used.
+ */
+export async function makeCoderSdk(baseUrl: string, token: string | undefined, storage: Storage): Promise<Api> {
+  const restClient = new Api()
+  restClient.setHost(baseUrl)
+  if (token) {
+    restClient.setSessionToken(token)
+  }
+
+  restClient.getAxiosInstance().interceptors.request.use(async (config) => {
+    // Add headers from the header command.
+    Object.entries(await storage.getHeaders(baseUrl)).forEach(([key, value]) => {
+      config.headers[key] = value
+    })
+
+    // Configure TLS.
+    const cfg = vscode.workspace.getConfiguration()
+    const insecure = Boolean(cfg.get("coder.insecure"))
+    const certFile = expandPath(String(cfg.get("coder.tlsCertFile") ?? "").trim())
+    const keyFile = expandPath(String(cfg.get("coder.tlsKeyFile") ?? "").trim())
+    const caFile = expandPath(String(cfg.get("coder.tlsCaFile") ?? "").trim())
+
+    config.httpsAgent = new https.Agent({
+      cert: certFile === "" ? undefined : await fs.readFile(certFile),
+      key: keyFile === "" ? undefined : await fs.readFile(keyFile),
+      ca: caFile === "" ? undefined : await fs.readFile(caFile),
+      // rejectUnauthorized defaults to true, so we need to explicitly set it to false
+      // if we want to allow self-signed certificates.
+      rejectUnauthorized: !insecure,
+    })
+
+    return config
+  })
+
+  // Wrap certificate errors.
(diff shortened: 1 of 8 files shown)


src/api.ts
```diff
@@ -0,0 +1,61 @@
...
+ * authentication is being used.
+ */
+export async function makeCoderSdk(baseUrl: string, token: string | undefined, storage: Storage): Promise<Api> {
+  const restClient = new Api()
+  restClient.setHost(baseUrl)
+  if (token) {
+    restClient.setSessionToken(token)
+  }
+
+  restClient.getAxiosInstance().interceptors.request.use(async (config) => {
```
buenos-nachos, line 27: Not a huge deal, but is there a reason why the Axios instance isn't a separate variable?

buenos-nachos: Though I guess another option would be to add both interceptors in one call of `interceptors.request.use`

src/commands.ts
```diff
@@ -97,125 +111,163 @@ export class Commands {
...
-                  message: err.x509Err || err.message,
-                  severity: vscode.InputBoxValidationSeverity.Error,
-                }
-              }
-              // This could be something like the header command erroring or an
-              // invalid session token.
-              const message =
-                err?.response?.data?.detail || err?.message || err?.response?.status || "no response from the server"
+        validateInput: async (value) => {
+          this.restClient.setSessionToken(value)
```
buenos-nachos: If `validateInput` fails, do you think it makes sense to set the token back to the previous value?

src/remote.ts
```diff
@@ -55,8 +49,45 @@ export class Remote {
     if (parts.length < 2 || parts.length > 3) {
```
buenos-nachos: Is there a reason why this isn't `parts.length !== 2`?

src/remote.ts
```diff
@@ -277,6 +303,7 @@ export class Remote {
       agent = matchingAgents[0]
     }
 
+    // Do some janky setting manipulation.
```
buenos-nachos, line 306: Embrace the jank

Parkreiner (approved):
This looks good! Happy to see that the core API refactoring did end up being helpful here

src/api.ts
code-asher: Oh like why not split out into `const axios = restClient.getAxiosInstance()` type of thing?

Could do that, I tend to like leaving things on their objects and accessing them that way, mostly just a preference thing I think.  I think maybe I like having them "namespaced" or something like that.

Edit: to add that this only applies to simple getters, if the function does work or the return value can change I would encapsulate into a var for sure

src/commands.ts
code-asher: I was thinking about this too haha and decided to be lazy

But yeah you are right, if this fails I should reset the token and probably the URL as well.

Would be cool if I could set it for just one request.

Actually maybe I should just create a new client just for this validation?  Then once it validates, then I set it on the main client.  Yeah I think that makes sense.

src/remote.ts
code-asher: Woah I did not even notice that.  No idea lol, someone else wrote this

It does seem weird.  I might as well change it while I am in here!

code-asher: Oh wait it is `2` or `3`.  Still seems better to check the range we want rather than the range we do not want though.

src/api.ts
code-asher: Oh and one is a `request` interceptor and the other a `response` interceptor so I think they have to be separate calls

buenos-nachos: Oh, wait – you're right. Totally missed that they were different names

src/remote.ts
buenos-nachos: Oh, wait, yeah, you're right. Guess I was tired, and forgot that it would be inclusive for [2, 3]

Issue #102, opened 2023-05-24
Support and list multiple deployments


matifali:
It would be nice to list and connect to workspaces from multiple deployments.
![image](https://github.com/coder/vscode-coder/assets/10648092/080e8f5b-da50-4065-9973-95fca084cda9)

code-asher:
In case it helps, the JetBrains plugin supports multiple deployments by using separate config directories for each one.  For example `~/.local/share/coder-gateway/dev.coder.com`.  Then in the SSH config we use the `--global-config` flag in the proxy command.

But it could be cool if the Coder CLI itself supported multiple deployments.

matifali:
But does the Coder's JetBrains Gateway plugin list and display multiple deployments?

code-asher:
Not at the same time, except in the recent connections window where connections from multiple deployments will be shown.

Would be cool to list all workspaces from multiple deployments though.  Code would be the same as the recent connections windows just without filtering for recently connected workspaces.

matifali:
Lack of this feature is causing issues like #173

matifali:
@EhabY I guess we can close this now.

PR #292, opened 2024-06-04
Add support for connections to multiple deployments

code-asher:
This is broadly three parts, as described by the commits, but the main gist is that the SSH blocks are now scoped (with the deployment URL in the host name), and the individual config files referenced in those config blocks are also scoped (with the deployment URL in the directory), and the connection process has been decoupled from the credentials in VS Code's memory and coupled to the credentials in the associated config directory instead.  More context in each commit message; it might even make sense to review by commit.

Existing hosts should not be affected, only new connections will be scoped by deployment.  Existing hosts (ones without the deployment in the host) will continue using the old config block and directories.  But it will not be possible to add new hosts using the old style; new connections will always be scoped. 

I should also clarify that this specifically is for supporting the "Recents" menu that may contain connections from multiple deployments and for reconnections (say you connect and then log into a different deployment while the other connection is still up).  It does not let you log into multiple deployments and see all the workspaces from all of them at once.

In other words it closes https://github.com/coder/vscode-coder/issues/173 but **not** https://github.com/coder/vscode-coder/issues/102.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -8,6 +8,7 @@ import { Commands } from "./commands"
 import { CertificateError, getErrorDetail } from "./error"
 import { Remote } from "./remote"
 import { Storage } from "./storage"
+import { toSafeHost } from "./util"
 import { WorkspaceQuery, WorkspaceProvider } from "./workspacesProvider"
 
 export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
@@ -108,6 +109,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
         // hit enter and move on.
         const url = await commands.maybeAskUrl(params.get("url"), storage.getUrl())
         if (url) {
+          restClient.setHost(url)
           await storage.setURL(url)
         } else {
           throw new Error("url must be provided or specified as a query parameter")
@@ -117,8 +119,12 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
         // prompted to sign in again, so we do not need to ensure it is set.
         const token = params.get("token")
         if (token) {
+          restClient.setSessionToken(token)
           await storage.setSessionToken(token)
         }
+
+        // Store on disk to be used by the cli.
+        await storage.configureCli(toSafeHost(url), url, token)
 
         vscode.commands.executeCommand("coder.open", owner, workspace, agent, folder, openRecent)
       }
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -115,7 +115,7 @@ export class Storage {
    * unable to download a working binary, whether because of network issues or
    * downloads being disabled.
    */
-  public async fetchBinary(restClient: Api, label: string | string): Promise<string> {
+  public async fetchBinary(restClient: Api, label: string): Promise<string> {
     const baseUrl = restClient.getAxiosInstance().defaults.baseURL
     this.output.appendLine(`Using deployment URL: ${baseUrl}`)
     this.output.appendLine(`Using deployment label: ${label || "n/a"}`)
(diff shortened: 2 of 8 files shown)


src/util.test.ts
```diff
@@ -0,0 +1,11 @@
+import { it, expect } from "vitest"
+import { toSafeHost } from "./util"
+
+it("escapes url host", async () => {
+  expect(toSafeHost("https://foobar:8080")).toBe("foobar")
+  expect(toSafeHost("https://ほげ")).toBe("xn--18j4d")
```
buenos-nachos, line 63: TIL that "hoge" is the Japanese equivalent of "foo"

src/extension.ts
```diff
@@ -117,9 +119,13 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
         // prompted to sign in again, so we do not need to ensure it is set.
         const token = params.get("token")
         if (token) {
+          restClient.setSessionToken(token)
           await storage.setSessionToken(token)
         }
 
+        // Store on disk to be used by the cli.
+        await storage.configureCli(toSafeHost(url), url, token)
```
buenos-nachos, line 127: I've noticed that all the calls to `storage.configureCli` have also required a call to `toSafeHost`. Do you think that `configureCli` could call `toSafeHost` internally?

src/storage.ts
```diff
@@ -109,16 +105,20 @@ export class Storage {
...
+   * the old deployment-unaware path instead.
+   *
    * If there is already a working binary and it matches the server version,
    * return that, skipping the download.  If it does not match but downloads are
    * disabled, return whatever we have and log a warning.  Otherwise throw if
    * unable to download a working binary, whether because of network issues or
    * downloads being disabled.
    */
-  public async fetchBinary(restClient: Api): Promise<string> {
+  public async fetchBinary(restClient: Api, label: string | string): Promise<string> {
```
buenos-nachos: Typo? `string | string` collapses down to `string`

Parkreiner (approved):
Looks good, and everything seems to make sense (I followed your recommendation and went through things commit-by-commit)

I'm gonna want to review more of how this codebase is set up, though. All the individual function calls seemed to make sense, but I feel like there's probably some finer nuances that I'm missing (especially with the VSCode APIs)

src/extension.ts
code-asher: That is a good question.  My thinking is that this storage interface is taking a directory name (or label as I am calling it), so from that perspective it does not make too much sense for it to be aware of URLs and how to encode them.

So, for similar reasons that, say, `fs.writeFile(fileName)` would not automatically encode the file name (er, at least, pretty sure it does not).

Similar to how `SSHConfig` takes a label rather than a URL as well.

src/storage.ts
code-asher: Whoops good catch lol will fix

You can pass a string OR A STRING!  DOUBLE STRING

src/extension.ts
code-asher: Probably it will always be a URL, but in the future there could be tests doing things like `storage.configureCli("temp-dir")` or some such.  But you could argue that the tests should just use URLs too, and the storage abstraction should change to take URLs rather than generic labels (and `SSHConfig` as well).

But, I am partial to the generic label abstraction personally, maybe because it feels less coupled/simpler to me.

I would have tests already except we need to switch away from vitest back to the VS Code testing framework.  :sob:

code-asher:
> I'm gonna want to review more of how this codebase is set up, though. All the individual function calls seemed to make sense, but I feel like there's probably some finer nuances that I'm missing (especially with the VSCode APIs)

I feel that, it is really tough to do code review in an unfamiliar repository.  I appreciate your reviews all the same though!

PR #308, opened 2024-06-25
Allow empty output from header command

davo-canva:
Allow no output from the header command, for parity with the Coder CLI, similarly to the JetBrains plugin in https://github.com/coder/jetbrains-coder/issues/404


code-asher (approved):
Thank you!

PR #322, opened 2024-07-19
Make log dir optional

bcpeinhardt:
This makes passing the --log-dir option to coder/coder optional and based on whether a proxyLogDirectory is set in the extensions settings.

The reasoning behind this is rather than make a unilateral decision about where and how long to store the logs, we simply give control to the user.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Changed
+
+- Make proxyLogDirectory a setting in the vscode extension rather than using the default for vscode,
+  and only send the --log-dir argument to coder/coder if the value is set.
+
 ## [v1.3.1](https://github.com/coder/vscode-coder/releases/tag/v1.3.1) (2024-07-15)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@
 
 ### Changed
 
-- Make proxyLogDirectory a setting in the vscode extension rather than using the default for vscode,
-  and only send the --log-dir argument to coder/coder if the value is set.
+- Previously, the extension would always log SSH proxy diagnostics to a fixed
+  directory. Now this must be explicitly enabled by configuring a new setting
+  `coder.proxyLogDirectory`. If you are having connectivity issues, please
+  configure this setting and gather the logs before submitting an issue.
 
 ## [v1.3.1](https://github.com/coder/vscode-coder/releases/tag/v1.3.1) (2024-07-15)
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Alternatively, manually install the VSIX from the
 Coder uses
 ${userHome} from VSCode's
 [variables reference](https://code.visualstudio.com/docs/editor/variables-reference).
-Use this when formatting paths in the coder extension settings rather than ~ or
+Use this when formatting paths in the Coder extension settings rather than ~ or
 $HOME.
 
 Example: ${userHome}/foo/bar.baz
--- a/README.md
+++ b/README.md
@@ -29,6 +29,10 @@ Alternatively, manually install the VSIX from the
 
 #### Variables Reference
 
-Coder uses ${userHome} from VSCode's [variables reference](https://code.visualstudio.com/docs/editor/variables-reference). Use this when formatting paths in the coder extension settings rather than ~ or $HOME.
+Coder uses
+${userHome} from VSCode's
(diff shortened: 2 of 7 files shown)


src/remote.ts
```diff
@@ -541,9 +541,29 @@ export class Remote {
...
+  /**
+   * Format's the --log-dir argument for the ProxyCommand
+   */
+  private async formatLogArg(): Promise<string> {
+    if (!supportsCoderAgentLogDirFlag(this.coderVersion)) {
+      return ""
+    }
+
+    // If the proxyLogDirectory is not set in the extension settings we don't send one.
+    // Question for Asher: How do VSCode extension settings behave in terms of semver for the extension?
```
code-asher, line 553:
```suggestion
```

There are no checks performed on the extension's version, it just depends on whether the setting exists in the `package.json` that is in that version.  If it were somehow missing I think it would just return undefined but I am not sure.  But someone would have to manually edit the `package.json` for that to happen.  :stuck_out_tongue:

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Changed
+
+- Make proxyLogDirectory a setting in the vscode extension rather than using the default for vscode,
+  and only send the --log-dir argument to coder/coder if the value is set.
```
code-asher: It could be helpful to add some description on why one might want to use the flag, and I think `--log-dir` could be considered an internal implementation detail so maybe we do not need to mention it.  Maybe something like:

```suggestion
- Previously, the extension would always log SSH proxy diagnostics to a fixed
  directory. Now this must be explicitly enabled by configuring a new setting
  `coder.proxyLogDirectory`. If you are having connectivity issues, please
  configure this setting and gather the logs before submitting an issue.
```

README.md
```diff
@@ -26,3 +26,13 @@ ext install coder.coder-remote
...
 Alternatively, manually install the VSIX from the
 [latest release](https://github.com/coder/vscode-coder/releases/latest).
+
+#### Variables Reference
+
+Coder uses
+${userHome} from VSCode's
+[variables reference](https://code.visualstudio.com/docs/editor/variables-reference).
+Use this when formatting paths in the coder extension settings rather than ~ or
+$HOME.
```
code-asher:
```suggestion
Use this when formatting paths in the Coder extension settings rather than ~ or
$HOME.
```

README.md
```diff
@@ -26,3 +26,13 @@ ext install coder.coder-remote
 
 Alternatively, manually install the VSIX from the
 [latest release](https://github.com/coder/vscode-coder/releases/latest).
+
+#### Variables Reference
+
+Coder uses
+${userHome} from VSCode's
```
code-asher: I think the official spelling has a space:

```suggestion
${userHome} from VS Code's
```

src/remote.ts
```diff
@@ -541,9 +541,29 @@ export class Remote {
...
+    const logDir = expandPath(String(vscode.workspace.getConfiguration().get("coder.proxyLogDirectory") ?? "").trim())
+    if (!logDir) {
+      return ""
+    }
+
+    await fs.mkdir(logDir, { recursive: true })
+    this.storage.writeToCoderOutputChannel(`Your logs are being written to ${logDir}`)
```
code-asher:
```suggestion
    this.storage.writeToCoderOutputChannel(`SSH proxy diagnostics are being written to ${logDir}`)
```

src/util.ts
```diff
@@ -58,3 +59,13 @@ export function toSafeHost(rawUrl: string): string {
   // should already have thrown in that case.
   return url.domainToASCII(u.hostname) || u.hostname
 }
+
+/**
+ * Expand a path with ${userHome} in the input string
+ * @param input string
+ * @returns string
```
code-asher, line 66: Very tiny nit/question: are these valuable with TypeScript?  My knee-jerk reaction is that at best it duplicates the info TypeScript gives and at worse it could get out of sync if someone changes the signature but not the comment.

But, maybe there is some IDE tooling that uses these.

code-asher (approved):
Looks great!  I only have some suggestions on wordings.  Feel free to merge whenever you feel it is ready.

src/util.ts
bcpeinhardt: I don't have strong opinions on this at all, my default was just to document the function however typescript folks usually do. I'll leave it for now so you don't have to re-review haha.

code-asher: Fair fair!  No worries about re-review for small things like this, by the way.  :)  For the most part once you have an approval you are free to merge.  If you make a big change or something though you might want a re-review.

PR #328, opened 2024-07-26
Use cli version for featureset detection

### LINKED ISSUE #323: Use cli version instead of Coder version
When connecting to the remote we use the Coder version to determine which flags are supported, but we should get the local cli version instead because it does not necessarily match (if the user has downloads disabled and is using an old cli).

Instead of all this:

https://github.com/coder/vscode-coder/blob/b35779a931637d176e21d39ef9b2baaec6a3a17d/src/remote.ts#L197-L218

We could do something like `supportsCoderAgentLogDirFlag(version(binPath))` where `version` comes from `cliManager.ts`, and make `version` return a `SemVer`.

bcpeinhardt:
This PR creates a `FeatureSet` object describing the features supported by a given coder version. It also changes the logic to use the CLI version if available (failing that we use the version from buildinfo like before). 

closes #323 

@code-asher when I go to test this the proxyLogDirectory setting doesn't show up in my extension settings, I must be doing something silly.


package.json
```diff
@@ -304,5 +304,6 @@
     "semver": "7.6.2",
     "trim": "0.0.3",
     "word-wrap": "1.2.5"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
code-asher, line 308: Oh interesting, never seen this before.  Seems useful.

src/featureSet.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Builds and returns a FeatureSet object for a given coder version.
+ */
+export function featureSetForVersion(version: semver.SemVer | null): FeatureSet {
+  return {
+    coderExtension: !(
+      version?.major === 0 &&
+      version?.minor <= 14 &&
+      version?.patch < 1 &&
+      version?.prerelease.length === 0
+    ),
```
code-asher: Since this indicates support for the `vscodessh` sub-command maybe we can call it `vscodessh` or something like that.

Issue #341, opened 2024-08-18
"Open Workspace" command pallette does nothing for me


bpmct:
https://github.com/user-attachments/assets/dff0494b-25cd-4362-94b0-7cbcf439437b

The only logs I see in "Extension Host" are

```
2024-08-18 10:08:49.890 [info] ExtensionService#_doActivateExtension coder.coder-remote, startup: false, activationEvent: 'onCommand:coder.openFromSidebar'
```

code-asher:
Oh interesting, it looks like the wrong command.  `coder.open` is the right one, `coder.openFromSidebar` should only be invoked from the sidebar.  We should hide this command from the palette, I think it never makes sense for a user to manually invoke it from the command palette.

bpmct:
Can we prioritize this @stirby?

code-asher:
Let me throw up a PR real swift, should just be a few lines

PR #364, opened 2024-09-24
Remove coder.openFromSidebar from command list

code-asher:
Two steps:

1. Hide the command
2. Make the command fall back to the regular open if it is somehow called

More details can be found in the commit messages.

Close https://github.com/coder/vscode-coder/issues/341

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,6 +192,12 @@
       }
     ],
     "menus": {
+      "commandPalette": [
+          {
+              "command": "coder.openFromSidebar",
+              "when": "false"
+          }
+      ],
       "view/title": [
         {
           "command": "coder.logout",
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -351,6 +351,10 @@ export class Commands {
         treeItem.workspaceFolderPath,
         true,
       )
+    } else {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -192,6 +192,12 @@
       }
     ],
     "menus": {
+      "commandPalette": [
+          {
+              "command": "coder.openFromSidebar",
+              "when": "false"
```
bcpeinhardt, line 198: Just real quick can you explain how this setting works? Why provide a command we don't want people to run?

code-asher: We need the command to run it in a context menu from the sidebar: https://github.com/coder/vscode-coder/blob/2d428eb88ca352519dc5be2666c6dd8eba276617/package.json#L216-L220

But, while we want it in the context menu where it has access to a selected workspace, we do not want it in the palette, where there is no selected workspace.

code-asher: Commands show up by default in the palette, so we have to explicitly prevent that with the `when`.  Feels janky to me but apparently this is the intended way lol

bcpeinhardt: Agree haha

bcpeinhardt (approved):

PR #378, opened 2024-10-09
Do not require token auth with mTLS

code-asher:
The strategy: when configuring the cli, write blank token files, since currently the `vscodessh` always expects a file.  Still need to validate if this works; if not we may need to write something like `unused` or modify the cli itself.

We configure in two places: login via command palette and dashboard link, so both those code paths are changed to write `""` when not using token auth.  Additionally, the login route is changed to not prompt for the token (failure to auth will be a hard error instead).

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -2,7 +2,7 @@ import { Api } from "coder/site/src/api/api"
 import { getErrorMessage } from "coder/site/src/api/errors"
 import { User, Workspace, WorkspaceAgent } from "coder/site/src/api/typesGenerated"
 import * as vscode from "vscode"
-import { makeCoderSdk } from "./api"
+import { makeCoderSdk, needToken } from "./api"
 import { extractAgents } from "./api-helper"
 import { CertificateError } from "./error"
 import { Storage } from "./storage"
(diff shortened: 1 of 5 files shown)


src/commands.ts
```diff
@@ -147,78 +147,33 @@ export class Commands {
...
-              severity: vscode.InputBoxValidationSeverity.Error,
-            }
-          }
-        },
-      })
-    }
-    if (!token || !user) {
-      return
+    // Try to get a token from the user, if we need one, and their user.
+    const res = await this.maybeAskToken(url, args[1])
```
bcpeinhardt: Nonblocking Nit: Validating the args passed in and giving them descriptive variable names should happen in one section at the top.

src/extension.ts
```diff
@@ -92,8 +92,12 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
 
         // If the token is missing we will get a 401 later and the user will be
-        // prompted to sign in again, so we do not need to ensure it is set.
-        const token = params.get("token")
+        // prompted to sign in again, so we do not need to ensure it is set now.
+        // For non-token auth, we write a blank token since the `vscodessh`
+        // command currently always requires a token file.  However, if there is
+        // a query parameter for non-token auth go ahead and use it anyway; all
+        // that really matters is the file is created.
+        const token = needToken() ? params.get("token") : (params.get("token") ?? "")
```
bcpeinhardt, line 100: I had to check and make sure params.get would return null and not undefined. Seems weird, in my mind a missing entry in a map of parameters should return undefined (at least that's my understanding of JavaScript logic).

src/storage.ts
```diff
@@ -435,8 +435,8 @@ export class Storage {
   /**
    * Configure the CLI for the deployment with the provided label.
    *
-   * Falsey values are a no-op; we avoid unconfiguring the CLI to avoid breaking
-   * existing connections.
+   * Falsey URLs and null tokens are a no-op; we avoid unconfiguring the CLI to
+   * avoid breaking existing connections.
    */
   public async configureCli(label: string, url: string | undefined, token: string | undefined | null) {
```
bcpeinhardt: We don't pass an `undefined` token to this function anywhere as far as I can tell, so I think we should remove it from the type signature to prevent a future headache where we think it should act like null but it actually acts like "", if that makes sense.

Honestly a rule around keeping undefined out of function parameter type signatures isn't a bad idea. For me undefined is for uninitialized values.

bcpeinhardt (approved):

src/commands.ts
code-asher: Definitely agree, this was just old code moved and indented but it is a good opportunity to update.

code-asher: ~Wait no I did actually introduce this??~ nvm was looking at the wrong part of the diff, both my statements were wrong lol

code-asher: Either way, fixed!

src/extension.ts
code-asher: Same!  Not really sure what the thinking was here.

src/storage.ts
code-asher: Removed!  Makes sense to limit the types to the bare minimum we need.

PR #380, opened 2024-10-14
Add default URL and autologin settings

code-asher:


package.json
```diff
@@ -97,6 +97,16 @@
           "markdownDescription": "If not set, will inherit from the `no_proxy` or `NO_PROXY` environment variables. `http.proxySupport` must be set to `on` or `off`, otherwise VS Code will override the proxy agent set by the plugin.",
           "type": "string",
           "default": ""
+        },
+        "coder.defaultUrl": {
```
bcpeinhardt, line 101: Could you add a note about auto login precedence/falling back to CODER_URL to both of these descriptions?

bcpeinhardt (approved):

package.json
code-asher: Will do for `defaultUrl`, the other already has it though, is there something specific you had in mind?  Or do you mean the `or` is not clear on precedence, and it would be nice to be explicit about it?

code-asher: Although, I am tempted to remove the autologin details from `defaultUrl` because it feels like a distraction/redundant, all that information should maybe just be on the `autologin` setting.  What do you think?

bcpeinhardt: Yeah I meant the or wasn't quite explicit, but I agree on removing it altogether. I would call it the "autologin_url" or something in the description instead of referring to the environment var specifically, so they go check that settings description and find the precedence logic.

package.json
```diff
@@ -99,12 +99,12 @@
           "default": ""
         },
         "coder.defaultUrl": {
-          "markdownDescription": "This will be shown in the URL prompt, along with CODER_URL if set, for the user to select when logging in. If autologin is enabled, this URL will be automatically logged into when the extension is activated.",
+          "markdownDescription": "This will be shown in the URL prompt, along with CODER_URL if set, for the user to select when logging in.",
```
bcpeinhardt: This line, I would change CODER_URL to "the default url" or something.

code-asher: My worry with that is there is nowhere else that describes the `CODER_URL` environment variable will be read and used in the URL prompt, but maybe it is superfluous to this setting, technically.  :thinking:

package.json
code-asher: It also is used to populate the default selection in the URL selector, so I was hesitant to call it specifically something related only to autologin, but there might be a better name somewhere that encapsulates both things.

package.json
bcpeinhardt: OH WAIT I get it now in my head this setting was describing the URL prompt not just the default URL yeah it's perfect as is.

PR #392, opened 2024-10-31
Fix the view logs command

code-asher:
The command was opening the old log path.


bcpeinhardt (approved):
LGTM

bcpeinhardt:
I think *maybe* we can flag this as something an E2E test would have caught?

code-asher:
> I think _maybe_ we can flag this as something an E2E test would have caught?

Definitely!

PR #400, opened 2024-11-18
Start workspaces by shelling out to CLI

aaronlehmann:
Replace the REST-API-based start flow with one that shells out to the coder CLI.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,4 +1,4 @@
-import { spawn, ChildProcessWithoutNullStreams } from "child_process"
+import { spawn } from "child_process"
 import { Api } from "coder/site/src/api/api"
 import { ProvisionerJobLog, Workspace } from "coder/site/src/api/typesGenerated"
 import fs from "fs/promises"
@@ -137,11 +137,7 @@ export async function startWorkspaceIfStoppedOrFailed(
   }
 
   return new Promise((resolve, reject) => {
-    const startProcess: ChildProcessWithoutNullStreams = spawn(binPath, [
-      "start",
-      "--yes",
-      workspace.owner_name + "/" + workspace.name,
-    ])
+    const startProcess = spawn(binPath, ["start", "--yes", workspace.owner_name + "/" + workspace.name])
 
     startProcess.stdout.on("data", (data: Buffer) => {
       data
--- a/src/api.ts
+++ b/src/api.ts
@@ -125,6 +125,7 @@ export async function makeCoderSdk(baseUrl: string, token: string | undefined, s
  */
 export async function startWorkspaceIfStoppedOrFailed(
   restClient: Api,
+  globalConfigDir: string,
   binPath: string,
   workspace: Workspace,
   writeEmitter: vscode.EventEmitter<string>,
@@ -137,7 +138,13 @@ export async function startWorkspaceIfStoppedOrFailed(
   }
 
   return new Promise((resolve, reject) => {
-    const startArgs = ["start", "--yes", workspace.owner_name + "/" + workspace.name]
+    const startArgs = [
+      "--global-config",
+      globalConfigDir,
+      "start",
+      "--yes",
+      workspace.owner_name + "/" + workspace.name,
+    ]
     const startProcess = spawn(binPath, startArgs)
 
     startProcess.stdout.on("data", (data: Buffer) => {
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,3 +1,4 @@
+import { spawn, ChildProcessWithoutNullStreams } from "child_process"
 import { Api } from "coder/site/src/api/api"
 import { ProvisionerJobLog, Workspace } from "coder/site/src/api/typesGenerated"
 import fs from "fs/promises"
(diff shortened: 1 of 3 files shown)


src/api.ts
```diff
@@ -122,29 +123,56 @@ export async function makeCoderSdk(baseUrl: string, token: string | undefined, s
...
   // Before we start a workspace, we make an initial request to check it's not already started
   const updatedWorkspace = await restClient.getWorkspace(workspace.id)
 
   if (!["stopped", "failed"].includes(updatedWorkspace.latest_build.status)) {
     return updatedWorkspace
   }
 
-  const latestBuild = await restClient.startWorkspace(updatedWorkspace.id, versionID)
+  return new Promise((resolve, reject) => {
+    const startProcess: ChildProcessWithoutNullStreams = spawn(binPath, [
```
code-asher: Was there an issue with the types?  I would expect this to have the same effect:

```suggestion
    const startProcess = spawn(binPath, [
```

src/api.ts
```diff
@@ -122,29 +123,56 @@ export async function makeCoderSdk(baseUrl: string, token: string | undefined, s
...
+      "start",
+      "--yes",
+      workspace.owner_name + "/" + workspace.name,
```
code-asher, line 146: This will be the slightly tricky part I think.  For most folks this will also need `--global-config` so the cli can pick up the url and token.

`--global-config` expects `url` and `session` files in the directory to which it points, so it may be slightly tricky because the `vscodessh` command that the plugin has used up to this point takes separate url and file config flags (not sure why), and I guess as a consequence the token file (perhaps unintentionally) ended up with the slightly different name `session_token` (again not sure why the difference).

So all that to say, I think we will have to also make sure the `session` file exists, either by migrating from `session_token` to `session` or writing both, and then add the `--global-config` flag.

src/api.ts
```diff
@@ -122,29 +123,56 @@ export async function makeCoderSdk(baseUrl: string, token: string | undefined, s
...
 
-  return {
-    ...updatedWorkspace,
-    latest_build: latestBuild,
-  }
+    startProcess.on("close", (code: number) => {
+      if (code === 0) {
+        resolve(restClient.getWorkspace(workspace.id))
+      } else {
+        reject(new Error(`"coder start" process exited with code ${code}`))
```
code-asher: Thinking out loud, I was hitting this (because of the `--global-config` stuff) and wondering if we can make the error clearer.  My thinking:

1. Could possibly put the full command we are running, so I can copy-paste to try it myself.
2. If we capture `stderr` separately perhaps we could add it to this error.  Or, if we avoid [disposing](https://github.com/coder/vscode-coder/blob/59ac05cf65198a1de26c4d704514759ed137c257/src/remote.ts#L151) the terminal until later (or maybe we could just not dispose it at all, could be nice to pull up the build log whenever you want instead of having it immediately vanish) so it can still be viewed while the popup is up, that could work!

src/api.ts
```diff
@@ -122,29 +123,56 @@ export async function makeCoderSdk(baseUrl: string, token: string | undefined, s
...
+      "--yes",
+      workspace.owner_name + "/" + workspace.name,
+    ])
+
+    startProcess.stdout.on("data", (data: Buffer) => {
+      data
+        .toString()
+        .split(/\r*\n/)
+        .forEach((line: string) => {
+          if (line !== "") {
```
code-asher, line 155: Are there a lot of blank lines we have to strip?  Does the API strip these out for us maybe so this is for parity?  

Not a big deal, my only thought is that maybe `start` uses blank lines to delineate sections for readability or something (no idea though).

code-asher (commented):
Thank you for putting this together!  Just to test, I hard-coded `--global-config` and it worked great, so that is the only thing we really need to figure out.

aaronlehmann:
Thanks for reviewing! I'm traveling right now but will come back to this in early December.

src/api.ts
aaronlehmann: I did see blank lines here, I believe due to the CLI's use of `\r` to overwrite existing lines (which doesn't work if passed through directly to VS Code's output).

code-asher: Aahhh right, I completely forgot it does that.  That makes perfect sense.  Since we display this in a (fake) terminal I wonder if we can pass through directly in a way that handles that case so it looks the same as when you run it directly on the command line, but this looks great to me for now.

src/api.ts
aaronlehmann: Fixed

src/api.ts
aaronlehmann: Fixed by migrating `session_token` to `session` and adding the `--global-config` flag to `coder start`. Please let me know if this approach looks good.

src/api.ts
aaronlehmann: Added the full command and stderr output to the error message.

src/remote.ts
```diff
@@ -156,6 +181,9 @@ export class Remote {
 
     const workspaceName = `${parts.username}/${parts.workspace}`
 
+    // Migrate "session_token" file to "session", if needed.
+    this.storage.migrateSessionToken(parts.label)
```
code-asher: We should await this to make sure the next line is not reading from a file that does not exist yet

```suggestion
    await this.storage.migrateSessionToken(parts.label)
```

src/storage.ts
```diff
@@ -488,6 +502,24 @@ export class Storage {
...
 
+  /**
+   * Migrate the session token file from "session_token" to "session", if needed.
+   */
+  public async migrateSessionToken(label: string) {
+    const oldTokenPath = this.getLegacySessionTokenPath(label)
+    try {
+      await fs.stat(oldTokenPath)
+      const newTokenPath = this.getSessionTokenPath(label)
+      await fs.rename(oldTokenPath, newTokenPath)
```
code-asher: Not a blocker at all, just wanted to mention that `stat` is often an anti-pattern; because `rename` can also throw `ENOENT` and there is technically no guarantee the file still exists between the `stat` and `rename`, it is always necessary to check the error of `rename` anyway, so might as well do it with one call.

code-asher (commented):
Looks and works great!  The only thing is that it asked me to log in again, which I think is just because of the missing `await`, the next line checked for `session` but it had not been renamed yet.  Subsequent runs are working wonderfully.

src/storage.ts
aaronlehmann: Right, that makes sense. Removed the `stat`.

code-asher (approved):
Thank you!  After merging I will cut a new release.

PR #409, opened 2024-12-12
Suppress login failure dialog on autologin

aaronlehmann:
This can be noisy and confusing if login attempts fail for users who don't have Coder accounts.


code-asher (approved):
Awesome thank you!

PR #432, opened 2025-02-04
changelog updates

bcpeinhardt:
Adding content to the changelog to describe @aaronlehmann 's updates which took place over several PR's before cutting a release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 - Recreate REST client after starting a workspace to ensure fresh TLS certificates.
 - Use `coder ssh` subcommand in place of `coder vscodessh`.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Unreleased
 
 - Recreate REST client after starting a workspace to ensure fresh TLS certificates.
+- Use `coder ssh` subcommand in place of `coder vscodessh`.
+- Simplify written SSH config to avoid excessive api calls when configuring the coder cli.
```
aaronlehmann: This was actually only an issue with the JetBrains plugin AFAIK.

bcpeinhardt: Ah right, removing, I shouldn't write docs after long meetings 💀

aaronlehmann (approved):

PR #431, opened 2025-02-03
Recreate REST client after starting a workspace

aaronlehmann:
`maybeWaitForRunning` can take indefinitely long because it prompts for confirmation, and in that time, client certificates can expire. Start setup fresh after this call to make sure the REST client uses the latest certificates.

PR #436, opened 2025-02-19
Recreate REST client after starting a workspace (part 2)

aaronlehmann:
This is a followup to #431. It addresses the need to also recreate the REST client used inside `maybeWaitForRunning`, since it makes API calls before returning to `setup`.


bcpeinhardt (approved):
LGTM

PR #439, opened 2025-02-27
Remove agent singleton

aaronlehmann:
This causes TLS certificates to be reread on every API request, so if they are refreshed on disk, the extension will pick up the new ones. It also avoids the need to recreating the REST client (which was an ultimately unsuccessful effort to make the extension pick up new certificates).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,20 @@
 
 ## Unreleased
 
-## [v1.4.1](https://github.com/coder/vscode-coder/releases/tag/v1.3.9) (2025-02-19)
+- Remove agent singleton so that client TLS certificates are reloaded on every API request.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,18 +2,22 @@
 
 ## Unreleased
 
-## [v1.4.1](https://github.com/coder/vscode-coder/releases/tag/v1.3.9) (2025-02-19)
+## [v1.4.2](https://github.com/coder/vscode-coder/releases/tag/v1.4.2)
```
bcpeinhardt: In terms of release flow we keep these under Unreleased until we're ready to make a separate tagged commit to cut the release :) We want to update that in the future so releasing is slightly less of a hassle, but for now would you mind sticking this under Unreleased?

aaronlehmann: Done

bcpeinhardt (approved):

PR #440, opened 2025-02-28
Use Axios client for EventSource

aaronlehmann:
This allows client TLS certificates to be used for event monitoring. Upgrade to a newer `eventsource` package that supports a custom `fetch` function, and provide a custom `fetch` function which wraps Axios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,29 +2,42 @@
 
 ## Unreleased
 
+### Fixed
+
 - Remove agent singleton so that client TLS certificates are reloaded on every API request.
 - Use Axios client to receive event stream so TLS settings are properly applied.
 
-### Fixed
-
 ## [v1.4.1](https://github.com/coder/vscode-coder/releases/tag/v1.4.1) (2025-02-19)
+
+### Fixed
 
 - Recreate REST client in spots where confirmStart may have waited indefinitely.
 
 ## [v1.4.0](https://github.com/coder/vscode-coder/releases/tag/v1.4.0) (2025-02-04)
 
+### Fixed
+
 - Recreate REST client after starting a workspace to ensure fresh TLS certificates.
+
+### Changed
+
 - Use `coder ssh` subcommand in place of `coder vscodessh`.
 
 ## [v1.3.10](https://github.com/coder/vscode-coder/releases/tag/v1.3.10) (2025-01-17)
+
+### Fixed
 
 - Fix bug where checking for overridden properties incorrectly converted host name pattern to regular expression.
 
 ## [v1.3.9](https://github.com/coder/vscode-coder/releases/tag/v1.3.9) (2024-12-12)
 
+### Fixed
+
 - Only show a login failure dialog for explicit logins (and not autologins).
 
 ## [v1.3.8](https://github.com/coder/vscode-coder/releases/tag/v1.3.8) (2024-12-06)
+
+### Changed
 
 - When starting a workspace, shell out to the Coder binary instead of making an
   API call. This reduces drift between what the plugin does and the CLI does. As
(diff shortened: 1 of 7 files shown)


src/api.ts
```diff
@@ -224,6 +280,7 @@ export async function waitForBuild(
             | undefined,
         },
         followRedirects: true,
+        agent: restClient.getAxiosInstance().defaults.httpAgent,
```
aaronlehmann: This fixes a similar issue with the WebSocket request not using TLS settings

CHANGELOG.md
```diff
@@ -2,18 +2,20 @@
 
 ## Unreleased
 
-## [v1.4.1](https://github.com/coder/vscode-coder/releases/tag/v1.3.9) (2025-02-19)
+- Use Axios client to receive event stream so TLS settings are properly applied.
 
 ### Fixed
```
bcpeinhardt, line 5: Looks like the `### Fixed` header accidentally got moved above the version header

src/api.ts
```diff
@@ -120,6 +122,60 @@ export async function makeCoderSdk(baseUrl: string, token: string | undefined, s
...
+        response.data.destroy()
+        return Promise.resolve()
+      },
+    })
+
+    const createReader = () => stream.getReader()
+
+    return {
+      body: {
+        getReader: () => createReader(),
```
bcpeinhardt: I think this could just be `getReader: stream.getReader`.

src/api-helper.ts
```diff
@@ -9,6 +10,8 @@ export function errToStr(error: unknown, def: string) {
     return error.response.data.message
   } else if (isApiErrorResponse(error)) {
     return error.message
+  } else if (error instanceof ErrorEvent) {
+    return error.code ? `${error.code}: ${error.message}` : error.message?.toString() || def
```
bcpeinhardt: Since error.message has type `string | undefined`, it seems like we should do 
```ts
return error.code ? `${error.code}: ${error.message || def}` : error.message || def
```
here. That gives us a fallback message in all cases and avoids the unnecessary `toString` call on something we already know is a string.

CHANGELOG.md
aaronlehmann: Tried to fix the existing entries - let me know if this looks right.

src/api.ts
aaronlehmann: Hmm, that seems to cause a compilation issue:

```
      TS2322: Type '(url: string | URL, init?: FetchLikeInit) => Promise<{ body: { getReader: ReadableStreamDefaultReader<any>; }; url: string; status: number; redirected: boolean; headers: { get: (name: string) => string | null; }; }>' is not assignable to type 'FetchLike'.
ts-loader-default_ab4862433a94faea
 @ ./src/extension.ts 47:29-60
```

But this compiles and seems to work:

```
getReader: () => stream.getReader(),
```

Updated.

src/api-helper.ts
aaronlehmann: Made this change, thanks

CHANGELOG.md
bcpeinhardt: Looks good, and I appreciate you fixing the previous entries too 😎

src/api.ts
bcpeinhardt: That's so weird I can't get `tsc` or VSCode to produce the error. What version of `tsc` do you have?

aaronlehmann: Can't reproduce, I think I must have tried `getReader: createReader()` by accident. Updated with your simpler version.

aaronlehmann: Nope, the simpler version doesn't work:
```
[2025-02-28T22:15:47.581Z] Value of "this" must be of type ReadableStream
```
Reverted back.

bcpeinhardt: So weird I can't get the error, even though the error message makes perfect sense. Just need to tweak my typescript setup I suppose.

bcpeinhardt (approved):

PR #444, opened 2025-03-03
feat: add support for coder inbox

DanielleMaywood:
Closes https://github.com/coder/internal/issues/337


src/inbox.ts
```diff
@@ -0,0 +1,92 @@
+import { Api } from "coder/site/src/api/api"
+import * as vscode from "vscode"
+import { WebSocket } from "ws"
+import { errToStr } from "./api-helper"
+import { Storage } from "./storage"
```
buenos-nachos: I want to say this could be turned into a `type import`

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+  notification: {
+    id: string
+    user_id: string
+    template_id: string
+    targets: string[]
+    title: string
+    content: string
+    actions: {
+      [key: string]: string
+    }
```
buenos-nachos: This can be simplified as `actions: Record<string, string>`

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+    read_at: string
+    created_at: string
+  }
+}
+
+const TemplateWorkspaceOutOfMemory = "a9d027b4-ac49-4fb1-9f6d-45af15f64e7a"
+const TemplateWorkspaceOutOfDisk = "f047f6a3-5713-40f7-85aa-0394cce9fa3a"
```
buenos-nachos: Nit: TypeScript generally defines true constants in SCREAMING_CASE
Beyond that, it wasn't immediately clear to me what these values are, or how they're used. I think it'd be fine if we just had a comment to explain that

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export class Inbox implements vscode.Disposable {
+  private disposed = false
+  private socket: WebSocket
+
+  constructor(
+    httpAgent: ProxyAgent,
+    restClient: Api,
+    private readonly storage: Storage,
```
buenos-nachos: Nit: I know this works, but I would prefer for the constructor to be explicit about embedding the `storage` parameter into the class. Plus, with the recent push from TypeScript to start "deprecating" TypeScript-only behavior, there's more of a risk that this pattern could break in the future

```ts
private readonly storage: Storage
constructor(
  httpAgent: ProxyAgent,
  restClient: Api,
  storage: Storage,
) {
  this.storage = storage;
  // Other code
}
```

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+    const socketProto = baseUrl.protocol === "https:" ? "wss:" : "ws:"
+    const socketUrlRaw = `${socketProto}//${baseUrl.host}/api/v2/notifications/watch`
+
+    this.socket = new WebSocket(new URL(socketUrlRaw), {
+      followRedirects: true,
+      agent: httpAgent,
+      headers: {
+        "Coder-Session-Token": restClient.getAxiosInstance().defaults.headers.common["Coder-Session-Token"] as
+          | string
+          | undefined,
```
buenos-nachos: It's been a while since I've touched the Axios API logic – what is the type of `common["Coder-Session-Token"]` without the type assertion?

buenos-nachos: One other thing you could do, if we know for a fact that the session token needs to stay in sync between Axios and the client, is this:
```ts
this.socket = new WebSocket(new URL(socketUrlRaw), {
  headers: {
    [coderSessionTokenHeader]: restClient.getAxiosInstance().defaults.headers.common[coderSessionTokenHeader]
  }
});
```

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    this.socket.on("message", (data) => {
+      try {
+        const inboxMessage = JSON.parse(data.toString()) as InboxMessage
+
+        if (
+          inboxMessage.notification.template_id === TemplateWorkspaceOutOfDisk ||
+          inboxMessage.notification.template_id === TemplateWorkspaceOutOfMemory
+        ) {
+          vscode.window.showWarningMessage(inboxMessage.notification.title)
```
buenos-nachos: I don't know enough about VSCode's API to say one way or the other, but my gut reaction to seeing this was wondering why it couldn't be `showInformationMessage` instead

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+        }
+      } catch (error) {
+        this.notifyError(error)
+      }
```
buenos-nachos, line 67: I'm wondering if we want to expose the full error to the user, or if we could show something more human readable if we know the error can only be triggered if the parsing fails

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    this.socket.on("open", () => {
+      this.storage.writeToCoderOutputChannel("Listening to Coder Inbox")
+    })
+
+    this.socket.on("error", (error) => {
+      this.notifyError(error)
+    })
+
+    this.socket.on("message", (data) => {
```
buenos-nachos: Just making sure, because I've only ever used the browser version of the `WebSocket` constructor: `data` won't be an event object, right? It'll be some kind of string or other "lower-level" value?

Parkreiner (commented):
Things look pretty good so far. Just had a few small bits of feedback

src/inbox.ts
buenos-nachos: Just pulled up the types, and [it looks like an event is expected](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6a2c9e76041cbb1fb46f22c120ac8ebaea6ebf01/types/ws/index.d.ts#L75)

src/inbox.ts
```diff
@@ -0,0 +1,92 @@
...
+      try {
+        const inboxMessage = JSON.parse(data.toString()) as InboxMessage
```
buenos-nachos: Assuming that data is an event, and isn't the data property in the event, data.toString() should work, but there's a good chance it'll be serialized as something like `[object Object]`. The `MessageEvent` type doesn't specify a specific constructor that it uses, so it could be a plain JS object

src/inbox.ts
DanielleMaywood: My IDE suggests it is of type `Data = string | Buffer | ArrayBuffer | Buffer[];`

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/47e9cf6d8947e46a82bb09bbcc3dd74a3ac0ffa1/types/ws/index.d.ts#L76

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/47e9cf6d8947e46a82bb09bbcc3dd74a3ac0ffa1/types/ws/index.d.ts#L105

src/inbox.ts
DanielleMaywood: In my local testing it hasn't caused any issues but will definitely test further

src/inbox.ts
DanielleMaywood: My thought process here is that Out Of Disk/Out Of Memory notifications are meant as a warning to the user so using `showWarningMessage` felt like the logical choice. I don't have a problem with choosing `showInformationMessage` though.

src/inbox.ts
DanielleMaywood: `this.notifyError` uses `errToStr` to attempt to make the string more human-readable

src/inbox.ts
```diff
@@ -0,0 +1,93 @@
...
+    user_id: string
+    template_id: string
+    targets: string[]
+    title: string
+    content: string
+    actions: Record<string, string>
+    read_at: string
+    created_at: string
+  }
+}
```
DanielleMaywood: note: this will probably be removed once `coder/coder` has the API defined, and will be replaced with a generated type.

src/inbox.ts
DanielleMaywood: It was discussed in a notifications meeting that for now all notifications should be `showInformationMessage` so will make that change

src/remote.ts
```diff
@@ -403,6 +404,11 @@ export class Remote {
     disposables.push(monitor)
     disposables.push(monitor.onChange.event((w) => (this.commands.workspace = w)))
 
+    // Watch coder inbox for messages
+    const httpAgent = await createHttpAgent()
```
buenos-nachos, line 408: Do you think it makes sense to add error-handling in case this call fails?

buenos-nachos: I'm seeing that we have a few calls to this without any try/catches, so I'm not sure if this is a major worry

src/inbox.ts
```diff
@@ -0,0 +1,80 @@
...
+// These are the template IDs of our notifications.
+// Maybe in the future we should avoid hardcoding
+// these in both coderd and here.
+const TEMPLATE_WORKSPACE_OUT_OF_MEMORY = "a9d027b4-ac49-4fb1-9f6d-45af15f64e7a"
+const TEMPLATE_WORKSPACE_OUT_OF_DISK = "f047f6a3-5713-40f7-85aa-0394cce9fa3a"
+
+export class Inbox implements vscode.Disposable {
+  private readonly storage: Storage
+  private disposed = false
+  private socket: WebSocket
```
buenos-nachos: One annoying thing about TypeScript is that there's a way to flag fields as private during compile-time only, and a way to mark it as private at both compile-time and runtime. `private` is the compile-time-only verison

If we're worried about that, we can [change these fields to start with `#`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties) and then get rid of the `private` modifier

src/inbox.ts
```diff
@@ -0,0 +1,80 @@
...
+      throw new Error("No base URL set on REST client")
+    }
+
+    const watchTemplates = [TEMPLATE_WORKSPACE_OUT_OF_DISK, TEMPLATE_WORKSPACE_OUT_OF_MEMORY]
+    const watchTemplatesParam = encodeURIComponent(watchTemplates.join(","))
+
+    const watchTargets = [workspace.id]
+    const watchTargetsParam = encodeURIComponent(watchTargets.join(","))
+
+    const baseUrl = new URL(baseUrlRaw)
```
buenos-nachos, line 36: Don't think this is inherently a problem, but want to flag it just in case: the `URL` constructor can throw. It's been a while since I've used Axios, but I'm guessing that if we get a URL from it, it should be valid and shouldn't have any problems. Maybe we could add a comment to say why this isn't an issue?

src/inbox.ts
```diff
@@ -0,0 +1,80 @@
...
+    const socketProto = baseUrl.protocol === "https:" ? "wss:" : "ws:"
+    const socketUrlRaw = `${socketProto}//${baseUrl.host}/api/v2/notifications/inbox/watch?templates=${watchTemplatesParam}&targets=${watchTargetsParam}`
```
buenos-nachos: Nit: I'm not sure why this has `raw` in the variable name? I would expect it to be fully-formatted at this point, especially since it's being fed directly into the socket instance

src/inbox.ts
```diff
@@ -0,0 +1,80 @@
...
+      },
+    })
+
+    this.socket.on("open", () => {
+      this.storage.writeToCoderOutputChannel("Listening to Coder Inbox")
+    })
+
+    this.socket.on("error", (error) => {
+      this.notifyError(error)
+    })
```
buenos-nachos: Do we want to close the socket if this triggers? The browser's version of the error event will auto-close the connection if this event comes through, but I don't know if that also applies to the `ws` version

src/inbox.ts
buenos-nachos: Oh wait, sorry. I made the mistake of trying to review this without pulling up an editor. I'm not making that mistake this time around – this should be good

Parkreiner (approved):
Changes look pretty straightforward to me. Only thing I'm wondering is how hard it'd be to add tests

PR #445, opened 2025-03-04
fix: remove unused date object

bcpeinhardt:
This date object doesn't get used for anything, I'm not sure why it's being passed.


Parkreiner (commented):
Is this right? I just pulled down the branch, and I'm getting feedback from the IDE that we need the date argument.

This is the version of the function I see on the PR branch:
```ts
getWorkspaceBuildLogs = async (
  buildId: string,
  before: Date,
): Promise<TypesGen.ProvisionerJobLog[]> => {
  const response = await this.axios.get<TypesGen.ProvisionerJobLog[]>(
    `/api/v2/workspacebuilds/${buildId}/logs?before=${before.getTime()}`,
  );

  return response.data;
};
```

bcpeinhardt:
*sigh* wtf mine doesn't have a date parameter. Hold on let me figure out if somethings wonky, I'll do a fresh clone.

bcpeinhardt:
On a completely fresh clone for me there's no Date parameter, something's weird here.

bcpeinhardt:
Okay so the difference appears to be in how `yarn install` vs `npm install` handles our vendored setup. `yarn install` leaves the date param there, whereas `npm install` removes the date param. In code/coder, the date param is not present.

Parkreiner (approved):

PR #446, opened 2025-03-05
Fix version comparison logic for checking wildcard support in "coder ssh"

aaronlehmann:
This meant to check that the version was at least 2.19.0, but the logic was faulty.


bcpeinhardt (approved):

aaronlehmann:
Fortunately it's a harmless error. `vscodessh` still works fine with Coder 2.19, and switching to `ssh` is really a cleanup.

PR #451, opened 2025-03-11
docs: document other editor support

bcpeinhardt:
Document the addition of Windsurf support and clarify that the extension now works with multiple editors beyond VS Code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,4 +135,4 @@ to make sure we're using up to date versions of the client.
 3. Push a tag matching the new package.json version.
 4. Update the resulting draft release with the changelog contents.
 5. Publish the draft release.
-6. Download the `.vsix` file from the release and upload to the marketplace (both Microsoft and OpenVSX).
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -135,4 +135,4 @@ to make sure we're using up to date versions of the client.
 3. Push a tag matching the new package.json version.
 4. Update the resulting draft release with the changelog contents.
 5. Publish the draft release.
-6. Download the `.vsix` file from the release and upload to the marketplace.
+6. Download the `.vsix` file from the release and upload to the marketplace (both Microsoft and OpenVSX).
```
buenos-nachos: Nit: "the marketplace" makes it sound like there should only be one, and I think that can lead to some confusion, when the rest of the sentence mentions two platforms. Also wondering if it's worth adding links to each one:
```suggestion
6. Download the `.vsix` file from the release and upload to both the [official VS Code Extension Marketplace](https://code.visualstudio.com/api/working-with-extensions/publishing-extension), and the [open-source VSX Registry](https://open-vsx.org/).
```

Parkreiner (approved):
I'm cool with changing the text – just had a suggestion for making it more clear

PR #453, opened 2025-03-20
fix: disable autostart on ssh command

f0ssel:
Previously we used the `coder vscodessh` command for the proxy command in the ssh config, which does not have the autostart-on-connect capabilities. This restores the previous behavior.


f0ssel:
I still need to read through the steps to build and test this, but I believe this is the solution.

aaronlehmann:
LGTM

bcpeinhardt (approved):

PR #463, opened 2025-04-01
feat: add ability to attach to devcontainers

DanielleMaywood:
Relates to https://github.com/coder/coder/issues/16426

Adds the ability to connect to a running devcontainer inside a Coder Workspace.

We add a new handler to handle the URI 
```
vscode://coder.coder-remote/openDevContainer
```

With the following query params:
- `url`: The URL of the deployment
- `token`: The API key
- `owner`: The workspace owner
- `workspace`: The workspace name
- `devContainerName`: The name of the dev container
- `devContainerFolder`: The folder inside the dev container

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -6,7 +6,7 @@ import { makeCoderSdk, needToken } from "./api"
 import { extractAgents } from "./api-helper"
 import { CertificateError } from "./error"
 import { Storage } from "./storage"
-import { AuthorityPrefix, toSafeHost } from "./util"
+import { toRemoteAuthority, toSafeHost } from "./util"
 import { OpenableTreeItem } from "./workspacesProvider"
 
 export class Commands {
@@ -500,6 +500,26 @@ export class Commands {
   }
 
   /**
+   * Open a devcontainer from a workspace belonging to the currently logged-in deployment.
+   *
+   * Throw if not logged into a deployment.
+   */
+  public async openDevContainer(...args: string[]): Promise<void> {
+    const baseUrl = this.restClient.getAxiosInstance().defaults.baseURL
+    if (!baseUrl) {
+      throw new Error("You are not logged in")
+    }
+
+    const workspaceOwner = args[0] as string
+    const workspaceName = args[1] as string
+    const workspaceAgent = undefined // args[2] is reserved, but we do not support multiple agents yet.
+    const devContainerName = args[3] as string
+    const devContainerFolder = args[4] as string
+
+    await openDevContainer(baseUrl, workspaceOwner, workspaceName, workspaceAgent, devContainerName, devContainerFolder)
+  }
+
+  /**
    * Update the current workspace.  If there is no active workspace connection,
    * this is a no-op.
    */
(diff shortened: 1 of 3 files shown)


src/commands.ts
```diff
@@ -499,6 +499,26 @@ export class Commands {
...
+  public async openDevContainer(...args: string[]): Promise<void> {
+    const baseUrl = this.restClient.getAxiosInstance().defaults.baseURL
+    if (!baseUrl) {
+      throw new Error("You are not logged in")
+    }
+
+    const workspaceOwner = args[0] as string
+    const workspaceName = args[1] as string
+    const workspaceAgent = undefined // args[2] is reserved, but we do not support multiple agents yet.
+    const devContainerName = args[3] as string
```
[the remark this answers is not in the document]
DanielleMaywood: We already have precedent for doing this with the `openWorkspace` flow

src/extension.ts
```diff
@@ -123,6 +178,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   vscode.commands.registerCommand("coder.login", commands.login.bind(commands))
   vscode.commands.registerCommand("coder.logout", commands.logout.bind(commands))
   vscode.commands.registerCommand("coder.open", commands.open.bind(commands))
+  vscode.commands.registerCommand("coder.openDevContainer", commands.openDevContainer.bind(commands))
```
code-asher, line 181: Would there be any value in adding this to the command palette via the `package.json` so it can be invoked from within VS Code like `coder.open` or through the sidebar?

I suppose that would also require that we fetch the dev container name, otherwise we would have to ask the user to supply it.  But, could be nice to support connecting from VS Code directly without using the dashboard.

**Edit**: hmm that would work for connecting to dev containers that are already running, but I am not seeing an obvious way to make the dev container extension build a container that does not yet exist via `vscode.openFolder`, which might mean we would have to check for and potentially build the container ourselves first, if we want to make this work.

src/commands.ts
```diff
@@ -598,3 +615,32 @@ async function openWorkspace(
     reuseWindow: !newWindow,
   })
 }
+
+async function openDevContainer(
```
code-asher, line 619: A nit, but I am curious why we break this out into a separate function?  It only seems to be called from `Commands.openDevContainer` at the moment, and it is fairly small.

Oh, actually, maybe one day we will have the ability to launch from the sidebar as well?

src/extension.ts
```diff
@@ -111,6 +111,61 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
         vscode.commands.executeCommand("coder.open", owner, workspace, agent, folder, openRecent)
+      } else if (uri.path === "/openDevContainer") {
+        const workspaceOwner = params.get("owner")
+        const workspaceName = params.get("workspace")
+        const workspaceAgent = params.get("agent")
+        const devContainerName = params.get("devContainerName")
+        const devContainerFolder = params.get("devContainerFolder")
+
+        if (!workspaceOwner) {
+          throw new Error("workspace owner must be specified as a query parameter")
```
code-asher, line 122: No need to change in this PR since this is a pattern that already existed, but at some point I wonder if we should catch handler errors and show them in an error notification.  Currently the only way to see them seems to be to go to the extension's features > runtime status > uncaught errors.

code-asher (approved):
Slick!  I only tested VS Code (not Cursor, etc), but it looks good.

src/extension.ts
code-asher: Actually maybe it is possible, using `dev-container+` if the container does not exist, and `attached-container+` otherwise?

DanielleMaywood: I wrote my findings about this in https://github.com/coder/coder/issues/16426. I'm open to supporting that functionality but probably in a future PR.

src/commands.ts
DanielleMaywood: I don't mind compacting it down into a single function. I was pretty much copying what we do with `openWorkspace`.

code-asher: That makes sense to me!  We will probably reuse it at some point.

src/extension.ts
code-asher: Works for me!

PR #465, opened 2025-04-01
chore: add additional logging around resource and memory useage

### LINKED ISSUE #429: Add debuggin' to the plugin
Recently we have been investigating a hard-to-reproduce integrated terminal latency issue. We should introduce some logging and diagnostics into the plugin so we can provide a build to the affected user and glean more information:

- [ ] add centralized logger (if there isn't one already)
- [ ] add debug calls to thorny areas of code as identified in our [walkthrough](https://www.notion.so/coderhq/VSCode-Bug-Bash-Notes-185d579be59280189438c42da74261fd), as well as to lifecycle events
- [ ] provide user with a debug build

Logging Ideas:
https://github.com/coder/vscode-coder/blob/main/src/remote.ts#L745
Is polling continuing indefinitely here?

https://github.com/coder/vscode-coder/blob/main/src/remote.ts#L785
are repeated file reads here slowing down performance?

https://github.com/coder/coder/blob/main/cli/vscodessh.go#L33
anything weird going on in here? This is the Go code on the Coder side. We wouldn't add logging here but might consider this interface on the plugin side considering VS Code is reading from it continuously to display connection info via a callback.
https://github.com/coder/coder/blob/main/cli/vscodessh.go#L199

We might want to log memory usage over time to see if we find any jump.

brettkolodny:
Adds a `MemoryLogger` class to the extension to log the memory and resource usage of the extension during various parts of the extension's life.

Closes #429 

> [!WARNING]  
> Various aspects of the code in this PR were written with the help of AI.
> I personally attest to the quality, but I am not as familiar with this code base so some extra scrutinizing would be appreciated.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -261,7 +261,6 @@ function extensionModeToString(mode: vscode.ExtensionMode): string {
   }
 }
 
-// Add deactivation handler to log memory usage on extension shutdown
 export function deactivate(): void {
   const logger = getMemoryLogger()
   logger.info("Coder extension deactivating")
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -15,7 +15,18 @@ import { getMemoryLogger } from "./memoryLogger"
 
 export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   // Initialize the memory logger right when the extension starts.
-  getMemoryLogger();
+  const logger = getMemoryLogger();
+  await logger.initLogFile(ctx.globalStorageUri.fsPath)
+
+  // Log initial memory usage and activation
+  logger.info("CODER extension activating")
+  logger.logMemoryUsage("EXTENSION_ACTIVATE")
+
+  // Register disposal of the logger when the extension deactivates
+  ctx.subscriptions.push({ dispose: () => logger.dispose() })
+
+  // Log extension mode
+  logger.info(`Extension mode: ${extensionModeToString(ctx.extensionMode)}`);
 
   // The Remote SSH extension's proposed APIs are used to override the SSH host
   // name in VS Code itself. It's visually unappealing having a lengthy name!
@@ -29,9 +40,13 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
     vscode.extensions.getExtension("codeium.windsurf-remote-openssh") ||
     vscode.extensions.getExtension("ms-vscode-remote.remote-ssh")
   if (!remoteSSHExtension) {
+    logger.error("Remote SSH extension not found, cannot activate Coder extension")
     vscode.window.showErrorMessage("Remote SSH extension not found, cannot activate Coder extension")
     throw new Error("Remote SSH extension not found")
   }
+  
+  logger.info(`Found Remote SSH extension: ${remoteSSHExtension.id}`);  
+  
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const vscodeProposed: typeof vscode = (module as any)._load(
     "vscode",
@@ -229,4 +244,26 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
       }
     }
   }
+
+  logger.info("Coder extension activation complete")
+}
+
+function extensionModeToString(mode: vscode.ExtensionMode): string {
+  switch (mode) {
+    case vscode.ExtensionMode.Development:
+      return "Development";
+    case vscode.ExtensionMode.Production:
+      return "Production";
+    case vscode.ExtensionMode.Test:
+      return "Test";
+    default:
+      return `Unknown (${mode})`;
+  }
+}
+
+// Add deactivation handler to log memory usage on extension shutdown
+export function deactivate(): void {
+  const logger = getMemoryLogger();
+  logger.info("Coder extension deactivating");
+  logger.logMemoryUsage("EXTENSION_DEACTIVATE");
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -11,8 +11,12 @@ import { Remote } from "./remote"
 import { Storage } from "./storage"
 import { toSafeHost } from "./util"
 import { WorkspaceQuery, WorkspaceProvider } from "./workspacesProvider"
+import { getMemoryLogger } from "./memoryLogger"
 
 export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
+  // Initialize the memory logger right when the extension starts.
+  getMemoryLogger();
+
   // The Remote SSH extension's proposed APIs are used to override the SSH host
   // name in VS Code itself. It's visually unappealing having a lengthy name!
   //
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -7,15 +7,15 @@ import { makeCoderSdk, needToken } from "./api"
 import { errToStr } from "./api-helper"
 import { Commands } from "./commands"
 import { CertificateError, getErrorDetail } from "./error"
+import { getMemoryLogger } from "./memoryLogger"
 import { Remote } from "./remote"
 import { Storage } from "./storage"
 import { toSafeHost } from "./util"
 import { WorkspaceQuery, WorkspaceProvider } from "./workspacesProvider"
-import { getMemoryLogger } from "./memoryLogger"
 
 export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   // Initialize the memory logger right when the extension starts.
-  const logger = getMemoryLogger();
+  const logger = getMemoryLogger()
   await logger.initLogFile(ctx.globalStorageUri.fsPath)
 
   // Log initial memory usage and activation
@@ -26,7 +26,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
   ctx.subscriptions.push({ dispose: () => logger.dispose() })
 
   // Log extension mode
-  logger.info(`Extension mode: ${extensionModeToString(ctx.extensionMode)}`);
+  logger.info(`Extension mode: ${extensionModeToString(ctx.extensionMode)}`)
 
   // The Remote SSH extension's proposed APIs are used to override the SSH host
   // name in VS Code itself. It's visually unappealing having a lengthy name!
@@ -44,9 +44,9 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
     vscode.window.showErrorMessage("Remote SSH extension not found, cannot activate Coder extension")
     throw new Error("Remote SSH extension not found")
   }
-  
-  logger.info(`Found Remote SSH extension: ${remoteSSHExtension.id}`);  
-  
+
+  logger.info(`Found Remote SSH extension: ${remoteSSHExtension.id}`)
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const vscodeProposed: typeof vscode = (module as any)._load(
     "vscode",
@@ -251,19 +251,19 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 function extensionModeToString(mode: vscode.ExtensionMode): string {
   switch (mode) {
     case vscode.ExtensionMode.Development:
-      return "Development";
+      return "Development"
     case vscode.ExtensionMode.Production:
-      return "Production";
+      return "Production"
     case vscode.ExtensionMode.Test:
-      return "Test";
+      return "Test"
     default:
-      return `Unknown (${mode})`;
+      return `Unknown (${mode})`
   }
 }
 
 // Add deactivation handler to log memory usage on extension shutdown
 export function deactivate(): void {
-  const logger = getMemoryLogger();
-  logger.info("Coder extension deactivating");
-  logger.logMemoryUsage("EXTENSION_DEACTIVATE");
+  const logger = getMemoryLogger()
+  logger.info("Coder extension deactivating")
+  logger.logMemoryUsage("EXTENSION_DEACTIVATE")
 }
--- a/src/inbox.ts
+++ b/src/inbox.ts
@@ -4,8 +4,8 @@ import { ProxyAgent } from "proxy-agent"
 import * as vscode from "vscode"
 import { WebSocket } from "ws"
 import { errToStr } from "./api-helper"
-import { type Storage } from "./storage"
 import { getMemoryLogger } from "./memoryLogger"
+import { type Storage } from "./storage"
 
 // These are the template IDs of our notifications.
 // Maybe in the future we should avoid hardcoding
--- a/src/inbox.ts
+++ b/src/inbox.ts
@@ -49,7 +49,6 @@ export class Inbox implements vscode.Disposable {
 
     logger.debug(`Connecting to inbox WebSocket at: ${socketUrl}`)
 
-    const coderSessionTokenHeader = "Coder-Session-Token"
     const token = restClient.getAxiosInstance().defaults.headers.common[coderSessionTokenHeader] as string | undefined
     this.#socket = new WebSocket(new URL(socketUrl), {
       agent: httpAgent,
--- a/src/inbox.ts
+++ b/src/inbox.ts
@@ -5,6 +5,7 @@ import * as vscode from "vscode"
 import { WebSocket } from "ws"
 import { errToStr } from "./api-helper"
 import { type Storage } from "./storage"
+import { getMemoryLogger } from "./memoryLogger"
 
 // These are the template IDs of our notifications.
 // Maybe in the future we should avoid hardcoding
@@ -16,9 +17,16 @@ export class Inbox implements vscode.Disposable {
   readonly #storage: Storage
   #disposed = false
   #socket: WebSocket
+  #messageCount = 0
+  #workspaceId: string
 
   constructor(workspace: Workspace, httpAgent: ProxyAgent, restClient: Api, storage: Storage) {
+    const logger = getMemoryLogger()
     this.#storage = storage
+    this.#workspaceId = workspace.id
+
+    logger.trackResourceCreated("InboxWebSocket", workspace.id)
+    logger.info(`Creating inbox for workspace: ${workspace.owner_name}/${workspace.name} (${workspace.id})`)
 
     const baseUrlRaw = restClient.getAxiosInstance().defaults.baseURL
     if (!baseUrlRaw) {
@@ -37,6 +45,8 @@ export class Inbox implements vscode.Disposable {
     const socketProto = baseUrl.protocol === "https:" ? "wss:" : "ws:"
     const socketUrl = `${socketProto}//${baseUrl.host}/api/v2/notifications/inbox/watch?format=plaintext&templates=${watchTemplatesParam}&targets=${watchTargetsParam}`
 
+    logger.debug(`Connecting to inbox WebSocket at: ${socketUrl}`)
+
     const coderSessionTokenHeader = "Coder-Session-Token"
     this.#socket = new WebSocket(new URL(socketUrl), {
       followRedirects: true,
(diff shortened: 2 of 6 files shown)


src/remote.ts
```diff
@@ -699,125 +711,218 @@ export class Remote {
...
+        if (matches.length < 2) {
+          return
+        }
+        const port = Number.parseInt(matches[1])
+        if (!port) {
+          return
+        }
+
+        logger.info(`Found SSH socks port: ${port}, searching for process`)
+        const processes = await find("port", port)
```
brettkolodny, line 853: In my testing I could not get this function call to return any processes using the given port. But running `lsof -i :<port>` showed that it was indeed in use. Not sure if this is relevant but figured it was worth mentioning

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Get or initialize the memory logger instance
+ */
+export function getMemoryLogger(): MemoryLogger {
+  if (!instance) {
+    instance = new MemoryLogger()
+  }
+  return instance
+}
```
buenos-nachos, line 217: I'm not sure if I'm missing something here, but do we need all this? This feels very Java-y to me
```suggestion
export const memoryLogger = new MemoryLogger();
```
If there's always going to be a singleton instance, I feel like we could export that instance directly, so that the rest of the file doesn't have to deal with `undefined` cases

Depending on what kinds of dependency injection we want to do, we could have a `getDefaultMemoryLogger` function that returns the instance?

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
+import * as fs from "fs/promises"
+import * as os from "os"
+import * as path from "path"
```
buenos-nachos, line 3: Biome is dinging these for not starting with the `node:` prefix (`node:fs/promise`, etc.)

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+ * A class for tracking memory usage and logging resource lifecycles
+ * to help identify memory leaks in the extension.
+ */
+export class MemoryLogger {
+  private outputChannel: vscode.OutputChannel
+  private logFile: string | undefined
+  private resourceCounts = new Map<string, number>()
+  private startTime: number = Date.now()
+  private logInterval: NodeJS.Timeout | undefined
+  private disposed: boolean = false
```
buenos-nachos: The annotation isn't necessary here

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+    this.outputChannel.show()
+
+    // Setup periodic logging of memory usage
+    this.startPeriodicLogging()
+  }
+
+  /**
+   * Start logging memory usage periodically
+   */
+  private startPeriodicLogging(intervalMs = 60000) {
```
buenos-nachos, line 29: I think I'd rather have this be extracted out into a constant so that it's easier to find and tweak down the line. I don't think it needs to exported, though

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+      }
+      this.logMemoryUsage("PERIODIC")
+      this.logResourceCounts()
+    }, intervalMs)
+  }
+
+  /**
+   * Initialize the log file for persistent logging
+   */
+  public async initLogFile(globalStoragePath: string): Promise<void> {
```
buenos-nachos, line 46: The current setup makes me worried, because we have a singleton for all the logging, but anybody can call `initLogFile` to change where the log output will go. If we need to output to multiple files, couldn't this risk crossing the streams a bit, especially since we would've already kicked off `startPeriodicLogging` at this point?

Wondering if this could be put in the constructor to enforce that it's initialization logic (whether that's moving the logic to the constructor directly, or making this a private method that the constructor calls)

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /**
+   * Log current memory usage
+   */
+  public logMemoryUsage(context: string): void {
+    try {
+      const memoryUsage = process.memoryUsage()
+      const nodeMemoryInfo = {
+        rss: `${(memoryUsage.rss / 1024 / 1024).toFixed(2)}MB`,
```
buenos-nachos, line 122: I think, just to make this a little more readable, this could be turned into something like a `bytesToMegabytes` helper

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+        heapTotal: `${(memoryUsage.heapTotal / 1024 / 1024).toFixed(2)}MB`,
+        heapUsed: `${(memoryUsage.heapUsed / 1024 / 1024).toFixed(2)}MB`,
+        external: `${(memoryUsage.external / 1024 / 1024).toFixed(2)}MB`,
+        uptime: formatDuration(process.uptime() * 1000),
```
buenos-nachos, line 126: Could you add a comment explaining that we're multiplying this by 1000 because `formatDuration` expects milliseconds, and `process.uptime` gives back seconds?

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+        totalUptime: formatDuration(Date.now() - this.startTime),
+      }
+
+      const systemMemoryInfo = {
+        totalMem: `${(os.totalmem() / 1024 / 1024 / 1024).toFixed(2)}GB`,
```
buenos-nachos, line 131: Similar comment for the formatting – feel like this could be a `bytesToGigabytes` helper

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+        freeMem: `${(os.freemem() / 1024 / 1024 / 1024).toFixed(2)}GB`,
+        loadAvg: os
+          .loadavg()
```
buenos-nachos: Looking at the description for `os.loadavg`, and it's making me wonder if we want to do anything special for Windows, since the value there will always be `[0, 0, 0]`

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+          .map((load) => load.toFixed(2))
+          .join(", "),
+      }
+
+      const memoryLog = `[MEMORY:${context}] Node: ${JSON.stringify(nodeMemoryInfo)} | System: ${JSON.stringify(systemMemoryInfo)}`
```
buenos-nachos, line 135: How important is it that the direct JSON output be human-readable out of the box, as opposed to throwing the entire output onto a single line? [`JSON.stringify` supports formatting the output to make it more readable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+      this.outputChannel.appendLine(memoryLog)
+      this.writeToLogFile(memoryLog)
+    } catch (err) {
+      this.outputChannel.appendLine(`[ERROR] Failed to log memory usage: ${err}`)
```
buenos-nachos, line 139: If you stringify the error directly, you'll only get the error type and the top-level error message. Do we want to output the `err.stack` property, too, if it exists?

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+    this.writeToLogFile(fullMessage)
+  }
+
+  /**
+   * Log current memory usage
+   */
+  public logMemoryUsage(context: string): void {
+    try {
+      const memoryUsage = process.memoryUsage()
+      const nodeMemoryInfo = {
```
buenos-nachos, line 121: Do we want to add type annotations to these objects?

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+  private resourceCounts = new Map<string, number>()
+  private startTime: number = Date.now()
+  private logInterval: NodeJS.Timeout | undefined
+  private disposed: boolean = false
+
+  constructor() {
+    this.outputChannel = vscode.window.createOutputChannel("Coder Memory Logging")
+    this.outputChannel.show()
+
+    // Setup periodic logging of memory usage
```
buenos-nachos, line 22: Feel like this comment isn't doing much

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+    const count = (this.resourceCounts.get(resourceType) || 0) + 1
+    this.resourceCounts.set(resourceType, count)
+    this.info(`RESOURCE_CREATED: ${resourceType}${id ? ":" + id : ""} (Total: ${count})`)
+  }
+
+  /**
+   * Log a resource disposal
+   */
+  public trackResourceDisposed(resourceType: string, id: string = ""): void {
+    const count = Math.max(0, (this.resourceCounts.get(resourceType) || 1) - 1)
```
buenos-nachos, line 76: Doesn't the `|| 1` ensure that the `Math.max` won't ever be necessary?

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+      this.outputChannel.appendLine(`[ERROR] Failed to log memory usage: ${err}`)
+    }
+  }
+
+  /**
+   * Log the current counts of active resources
+   */
+  private logResourceCounts(): void {
+    const counts = Array.from(this.resourceCounts.entries())
+      .map(([type, count]) => `${type}=${count}`)
```
buenos-nachos, line 148: Nit: this can be cleaned up a little bit. `Array.from` already has support for mapping built in via its second parameter, and will also auto-convert the map into an entries iterable

src/memoryLogger.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * Dispose of the logger
+   */
+  public dispose(): void {
+    this.disposed = true
+    if (this.logInterval) {
+      clearInterval(this.logInterval)
+      this.logInterval = undefined
+    }
```
buenos-nachos, line 188: nit: this if statement isn't necessary. `clearInterval` has support for no-ops if it's passed an undefined value

src/memoryLogger.ts
buenos-nachos: Actually, wait, this is an async method. We could still do some trickery to make sure things are initialized together, but the other thing that sticks out to me is that we have this init logic in a try/catch. If this method fails, the catch will suppress the error, and the `activate` function will still be allowed to move on as if though everything is fine

Do we want this method to return a boolean to indicate setup success?

src/extension.ts
```diff
@@ -225,4 +244,26 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+    case vscode.ExtensionMode.Production:
+      return "Production"
+    case vscode.ExtensionMode.Test:
+      return "Test"
+    default:
+      return `Unknown (${mode})`
+  }
+}
+
+// Add deactivation handler to log memory usage on extension shutdown
```
buenos-nachos: I feel like this comment isn't doing much, especially if we end up needing to add more cleanup logic that isn't logging-related down the line

src/inbox.ts
```diff
@@ -16,9 +17,16 @@ export class Inbox implements vscode.Disposable {
   readonly #storage: Storage
   #disposed = false
   #socket: WebSocket
+  #messageCount = 0
+  #workspaceId: string
 
   constructor(workspace: Workspace, httpAgent: ProxyAgent, restClient: Api, storage: Storage) {
+    const logger = getMemoryLogger()
```
buenos-nachos, line 26: I think it might be better to refactor this class to take a single init object. That way, we can dependency-inject the logger, and also make it easier to add new parameters down the line. Moving the logger directly to the constructor parameters would give us a five-argument function, which feels a bit unwieldy

src/inbox.ts
```diff
@@ -49,35 +59,72 @@ export class Inbox implements vscode.Disposable {
...
+    })
+
     this.#socket.on("message", (data) => {
+      this.#messageCount++
+
+      // Log periodic message stats
+      if (this.#messageCount % 10 === 0) {
+        logger.info(`Inbox received ${this.#messageCount} messages for workspace: ${workspace.id}`)
+        logger.logMemoryUsage("INBOX_WEBSOCKET")
+      }
```
buenos-nachos, line 88: Do we want a little more granularity here? The current setup means that we have to wait for the first 10 messages before we log anything. I don't know how quickly the messages are coming in, but I'm worried we'll never log anything. Wondering if we could have something more logarithmic, where maybe we could have:
- First message
- First five messages
- Every 10 messages after that

src/inbox.ts
```diff
@@ -49,35 +59,72 @@ export class Inbox implements vscode.Disposable {
...
+          logger.logMemoryUsage("INBOX_PERIODIC")
+        } else {
+          clearInterval(memoryInterval)
+        }
+      },
+      5 * 60 * 1000,
+    ) // Every 5 minutes
   }
 
   dispose() {
```
buenos-nachos, line 113: Should we move the interval to the class level, and clear it directly when we dispose the instance? Right now, the interval will continue to run up to 5 minutes after the instance is disposed before the interval is allowed to clean itself up

src/remote.ts
```diff
@@ -688,9 +689,20 @@ export class Remote {
   // showNetworkUpdates finds the SSH process ID that is being used by this
   // workspace and reads the file being created by the Coder CLI.
   private showNetworkUpdates(sshPid: number): vscode.Disposable {
+    const logger = getMemoryLogger()
```
buenos-nachos, line 692: I think I'd rather this be injected at the constructor level

src/memoryLogger.ts
buenos-nachos: Oh, actually, does `prettyBytes` take care of this for us already? Figured we might as well use it if it's already a dependency

src/remote.ts
```diff
@@ -699,125 +711,218 @@ export class Remote {
...
-            return
-          }
-          if (first) {
-            networkStatus.tooltip += `\n\nOther regions:`
-            first = false
-          }
-          networkStatus.tooltip += `\n${region}: ${Math.round(network.derp_latency[region] * 100) / 100}ms`
-        })
-      }
+      try {
```
buenos-nachos: I don't know if `prettyBytes` can throw, but the only things that jump out to me that can fail are the property accesses for `network.derpLatency`. I feel like it might be better to get rid of the wrapper, and beef up those accesses to deal with `undefined` cases

src/remote.ts
```diff
@@ -699,125 +711,218 @@ export class Remote {
...
             updateStatus(parsed)
           } catch (ex) {
-            // Ignore
+            logger.error(`Failed to update status from parsed network info`, ex)
           }
         })
-        .catch(() => {
-          // TODO: Log a failure here!
+        .catch((error) => {
+          // Replace empty catch with proper error logging
```
buenos-nachos: This comment feels like an AI artifact that we can remove everywhere it was added. Specifically, all the comments that end with "proper error logging"

src/remote.ts
```diff
@@ -699,125 +711,218 @@ export class Remote {
...
+          }
           return JSON.parse(content)
         })
         .then((parsed) => {
           try {
             updateStatus(parsed)
           } catch (ex) {
-            // Ignore
+            logger.error(`Failed to update status from parsed network info`, ex)
           }
```
buenos-nachos: It shouldn't be possible for this `try/catch` to trigger anymore, with how `updateStatus` does things now

src/remote.ts
```diff
@@ -699,125 +711,218 @@ export class Remote {
...
     }
   }
 
   // findSSHProcessID returns the currently active SSH process ID that is
   // powering the remote SSH connection.
   private async findSSHProcessID(timeout = 15000): Promise<number | undefined> {
+    const logger = getMemoryLogger()
+    logger.info(`Starting SSH process ID search with timeout: ${timeout}ms`)
+
+    let attempts = 0
```
buenos-nachos: Could this be renamed to something like `loopAttempts`? I feel like that would make it more clear how it's being used, since it's accessed in `search`, but only in a readonly way

src/remote.ts
```diff
@@ -699,125 +711,218 @@ export class Remote {
...
+        // this to find the SSH process that is powering this connection. That SSH
+        // process will be logging network information periodically to a file.
+        const text = await fs.readFile(logPath, "utf8")
+
+        if (attempts % 5 === 0) {
+          logger.debug(`SSH log file size: ${text.length} bytes`)
+        }
+
+        const matches = text.match(/-> socksPort (\d+) ->/)
+        if (!matches) {
```
buenos-nachos, line 841: Nit: this could be `/-> socksPort (\d+) ->/.test(text)` since we're not doing anything with the match results

src/remote.ts
```diff
@@ -699,125 +711,218 @@ export class Remote {
...
+          return
+        }
+        if (matches.length < 2) {
+          return
+        }
+        const port = Number.parseInt(matches[1])
```
buenos-nachos: I'd specify the radix so that the number can't randomly be parsed in the wrong way

src/workspacesProvider.ts
```diff
@@ -56,10 +57,21 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
   // Calling this while already refreshing or not visible is a no-op and will
   // return immediately.
   async fetchAndRefresh() {
-    if (this.fetching || !this.visible) {
+    const logger = getMemoryLogger()
```
buenos-nachos, line 60: Similar comment about injecting the logger at the class level, rather than doing it for each method

src/workspacesProvider.ts
```diff
@@ -157,45 +174,83 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
...
   refresh(item: vscode.TreeItem | undefined | null | void): void {
+    const logger = getMemoryLogger()
+    logger.debug(
+      `WorkspaceProvider(${this.getWorkspacesQuery}): Refreshing tree view${item ? " for specific item" : ""}`,
+    )
+
     this._onDidChangeTreeData.fire(item)
   }
 
+  dispose(): void {
```
buenos-nachos, line 235: Do we want to make sure this turns into a no-op if this is called multiple times?

src/workspacesProvider.ts
```diff
@@ -227,41 +282,80 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
...
   const metadataUrl = new URL(`${url}/api/v2/workspaceagents/${agentId}/watch-metadata`)
+
+  logger.info(`Starting metadata watcher for agent: ${agentId} at ${metadataUrl}`)
+
   const eventSource = new EventSource(metadataUrl.toString(), {
     fetch: createStreamingFetchAdapter(restClient.getAxiosInstance()),
   })
 
   let disposed = false
+  let eventCount = 0
```
buenos-nachos, line 299: Nit: could this be moved down to be right above the data listener?

src/workspacesProvider.ts
```diff
@@ -227,41 +282,80 @@ export class WorkspaceProvider implements vscode.TreeDataProvider<vscode.TreeIte
...
 
+  eventSource.addEventListener("open", () => {
+    logger.info(`Metadata EventSource connection opened for agent: ${agentId}`)
+  })
+
   eventSource.addEventListener("data", (event) => {
+    eventCount++
+
+    // Log periodic updates
+    if (eventCount % 50 === 0) {
```
buenos-nachos, line 322: How did we go about deciding the thresholds for when to log different things? I'm seeing a lot of different values, which I could see making sense, but I'm not 100% sure

Parkreiner (commented):
Tried to do the best job I could, but I think there might be a few VSCode-specific nuances I'm missing

I can circle around for a second, quick review later, but I don't know if @code-asher has the bandwidth to give a good rubber stamp

src/memoryLogger.ts
brettkolodny: This project uses prettier and eslint instead of biome, so doesn't look like we use the `node` prefix anywhere. I think it's probably better to keep it for consistency

src/workspacesProvider.ts
brettkolodny: The thresholds were decided somewhat arbitrarily based on how often it'd be logging to try and strike a balance of useful to spammy

src/memoryLogger.ts
brettkolodny: Not sure I follow since It's already pretty readable to me. Are you suggesting pretty printing the output, or changing the structure of the output completely to not be JSON?

src/memoryLogger.ts
brettkolodny: That's a good question. Windows may not be relevant for this implementation but it'd be a good idea to check

brettkolodny: Yup very important, good catch!

code-asher:
I have not reviewed the code yet, but I am happy to take a look.  I will hold off until the second round, unless y'all are OK with me potentially making the same comments hahaha

Tangentially, wondering if we have had folks try profiling (unfortunately it seems VS Code only has CPU profiling, although the process explorer does show current memory usage).  https://github.com/Microsoft/vscode/wiki/Performance-Issues 

I do have some concerns that this is an unusual way to profile a process and may not get us the data we need, but I have not had a chance to see if there is a way to toggle standard Node profiling on the extension host, so maybe this is as good as it gets for our circumstances.

Also, do we have any rotation on the new log file to prevent it from growing unbounded?  And do we need the separate file?  Could we just log everything to the output channel?  Possibly using `env.logLevel` to determine in how much detail we should log.

src/memoryLogger.ts
brettkolodny: Thinking about it I don't think the load avg is particularly useful for what we're trying to debug here regardless. It gives the 1, 5, and 15 minute load averages and we're looking for an issue that doesn't arise until the the extension has been running for hours

PR #466, opened 2025-04-02
Avoid using undefined token header

code-asher:
It seems that Node will throw an error when headers are undefined.


aaronlehmann (approved):
Thanks!

Issue #361, opened 2024-09-23
Removed connection status indicator?


ammario:
I'm on the latest extension and noticed that we removed the connections status indicator at some point.

I found it helpful, in particular, to know whether the connection is P2P and, if not, which DERP region I'm connecting through. It would also be nice to see an estimate of end-to-end latency. In general I have noticed a decay in responsiveness in my VS Code - Coder experience and I'd like to get to the bottom of it.

![image](https://github.com/user-attachments/assets/8cc7b8ef-63f6-4b70-9a4a-53fe949ba6bb)

redartera:
Been having the same issue - it's normally very useful for me to keep track of latency and P2P/Relay connection status

code-asher:
I think we do some janky overriding of the indicator, maybe some internal API changed and now it no longer works.

matifali:
Resolved in #480

Issue #413, opened 2024-12-30
VS Code keeps disconnecting if the client device sleeps


matifali:
> Hi Team! When using Cursor/VSCode desktop, whenever we close and reopen laptop even for a few seconds, we see a popup Cannot reconnect. Please reload the window. I was wondering if this is expected behavior and if you might be aware of some settings that would make VSCode reconnect more smoothly? I tried changing connect timeout and max reconnection attempts so far but didn't see a difference.

VSCode connection often disconnects even for a slight network issue or if the device sleeps. I am unsure if this is caused by how VSCode handles remote connections as `coder ssh <workspace>` survives device sleep.

f0ssel:
It is possible that https://github.com/coder/vscode-coder/pull/420 may have solved this issue. @bcpeinhardt said he's going to try to see if he can repro and confirm if this change fixes things.

bcpeinhardt:
@bpmct curious why this is marked as blocked?

bcpeinhardt:
Leaving as `todo` in Bugs Board because it doesn't seem to have a Blocked status available.

bpmct:
@matifali told me we are are unable to reproduce and waiting on a user. If we are able to repro, this feel pretty important so don't mean to arbitrarily "block" anything

bcpeinhardt:
After attempting and failing to reproduce, I'm moving it back to blocked (where it was in the first place) and reaching back out to the customer.

csibe17:
We are also seeing this problem constantly.
Usually it happens when the device goes to sleep and then wakes up and you try to continue working.
This does not happen with web based VSCode at all, that can immediately reconnect.

![Image](https://github.com/user-attachments/assets/9d7d60a3-72bd-4688-be05-f849e73ee0cb)

Mikescops:
I confirm I have the same issue for the past weeks, still going on on 1.5.0. Have to reload the window whenever I leave my computer 5 minutes.

matifali:
Hi @Mikescops, can you still reproduce the issue on the latest extension version 1.8.0?

Mikescops:
@matifali absolutely yes

matifali:
I can reliably reproduce the issue, and we will investigate this further. 

I am seeing a line like `[<unknown>][a81f0b5a][ManagementConnection] Unknown reconnection token (never seen).` whenever I see the `Reload Window` dialogue in `Remote - SSH` output logs. 

[...]

PR #480, opened 2025-04-15
fix: support finding ssh process by port on macos sequoia

ethanndickson:
Relates to #361.

This brings back the connection indicator, at least on macOS Sequoia.
image

The indicator is populated from a file named `<sshPID>.json`. The library we were using to find that PID, `find-process` hasn't been updated for Sequoia.

Full explanation of the necessary change can be found in an umerged PR for that library:
https://github.com/yibn2008/find-process/pull/78 

This change was never merged, so I've forked it into our org for now. When it gets upstreamed we can switch back.


ethanndickson:
* **#480** <a href="https://app.graphite.dev/github/pr/coder/vscode-coder/480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/vscode-coder/480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

yarn.lock
```diff
@@ -2657,6 +2657,21 @@ eventsource@*, eventsource@^3.0.6:
   dependencies:
     eventsource-parser "^3.0.1"
 
+execa@^8.0.1:
```
ethanndickson, line 2660: I notice this dependency is huge... Maybe we just fork `find-process` until it gets upstreamed?

matifali (approved):
LGTM. Thanks @ethanndickson

code-asher (approved):
Interesting that `find-process` execs `netstat` instead of reading `/proc` directly.  Although, maybe that is not possible on macOS?

In any case, looks good to me.  If we ever implement our own connection code we will be able to get the PID more easily.  :crossed_fingers:

matifali:
> If we ever implement our own connection code we will be able to get the PID more easily. 🤞

This is becoming more than a want now, and we may actually need this soon.

matifali:
@ethanndickson, I am noticing a better reconnection flow after this PR. When my device returns from sleep, VS Code auto-reconnects without me pressing the reload button.

This may probably solve #413.

matifali:
The upstream library fixed the issue, so we may consider switching to that again.

PR #482, opened 2025-04-17
feat: coder connect integration

### LINKED ISSUE #447: Coder Desktop Integration
Implement an enhancement making this extension aware of Coder Desktop. With Coder Desktop running, there is already an established tunnel to a workspace; thus, the extension does not need to create another tunnel.

The extension should prioritize the existing tunnel rather than creating a new one and should be able to display tunnel statistics related to the VPN within the editor.

Related : coder/jetbrains-coder#543

Once coder/internal#466 is closed, this work can begin.

The extension should:

1. Check that the workspace belongs to the logged-in user. (Coder Connect only supports workspaces actually owned by the user.) If not, fallback to current behavior.
2. Query `api/v2/deployment/ssh` to obtain the `HostnameSuffix`. If not set, fallback.
3. Do a DNS lookup for an IPv6 address (AAAA record) for the workspace and agent as `<agent>.<workspace>.me.<suffix>`. If the DNS lookup fails (NXDOMAIN) or returns an address **not** in the prefix Coder Service Prefix `fd60:627a:a42b::/48`, fallback. Checking the prefix ensures that captive DNS portals that return addresses for any query are not false positives.
4. If the DNS lookup was successful, this means Coder Connect is running, and the extension should connect to the workspace directly, rather than via the `coder` CLI, e.g. by directly initiating an SSH connection to `<agent>.<workspace>.me.<suffix>`.

ethanndickson:
Closes #447.

On both macOS and Windows, with Coder Connect running, I've tested that:
- Opening a workspace from the sidebar uses Coder Connect
- Opening a workspace from the command palette uses Coder Connect
- Opening a workspace from a `vscode://` URI uses Coder Connect
- Opening an unowned workspace fallsback to using the Coder CLI, as Connect doesn't support shared workspaces yet.
- Opening a stopped workspace fallsback to using the Coder CLI in order to prompt the user to start the workspace first. If Coder Connect is available after the build, it'll switch to using that tunnel.
- Opening a Coder extension workspace in the VSCode history, like so, image switches to the Connect tunnel.

Including the connection status indicator when using Connect is out of scope of the linked issue and this PR, as Coder Desktop does not currently expose those values in any way.

I recommend reviewing the PR with whitespace diffs disabled.

EDIT: PR closed in favor of just adding the integration to `coder ssh`, and incurring the minor performance cost.


ethanndickson:
* **#482** <a href="https://app.graphite.dev/github/pr/coder/vscode-coder/482?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/vscode-coder/482?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/commands.ts
```diff
@@ -491,12 +507,30 @@ export class Commands {
     } else {
       workspaceOwner = args[0] as string
       workspaceName = args[1] as string
-      // workspaceAgent is reserved for args[2], but multiple agents aren't supported yet.
+      workspaceAgent = args[2] as string
```
ethanndickson: The URI handler does indeed populate the agent name in `args[2]`, but it may be `undefined`.
```
vscode.commands.executeCommand("coder.open", owner, workspace, agent, folder, openRecent)
```

The original comment is years out of date.

src/workspacesProvider.ts
```diff
@@ -353,7 +353,7 @@ export class WorkspaceTreeItem extends OpenableTreeItem {
       showOwner ? vscode.TreeItemCollapsibleState.Collapsed : vscode.TreeItemCollapsibleState.Expanded,
       workspace.owner_name,
       workspace.name,
-      undefined,
+      agents[0]?.name,
```
ethanndickson, line 356: This saves us needing to fetch the agent in `openFromSidebar` for a single-agent workspace. 
`openFromSidebar` isn't callable on workspaces with multiple agents, so the `undefined` there is fine.
if the workspace is stopped, this will be undefined, prompting `openWorkspace` to always use the CLI instead of Connect.

src/commands.ts
```diff
@@ -1,6 +1,10 @@
+import { isAxiosError } from "axios"
 import { Api } from "coder/site/src/api/api"
 import { getErrorMessage } from "coder/site/src/api/errors"
 import { User, Workspace, WorkspaceAgent } from "coder/site/src/api/typesGenerated"
+import { lookup } from "dns"
+import { inRange } from "range_check"
```
deansheather: If we're going to import a library for this, we should probably get the most popular one: https://www.npmjs.com/package/ip-range-check

also has the benefit of not requiring any deps

src/commands.ts
```diff
@@ -392,14 +396,26 @@ export class Commands {
...
+          this.restClient,
+          baseUrl,
+          treeItem.workspaceOwner,
+          treeItem.workspaceName,
+          treeItem.workspaceAgent,
+          treeItem.workspaceFolderPath,
+          true,
+        )
+      } catch (err) {
+        const message = getErrorMessage(err, "no response from the server")
```
deansheather: Can you just make `openWorkspace` do this rather than having it twice?

ethanndickson: I wasn't sure if `openWorkspace` was deliberately not a member of the `Commands` class. I guess I can move it in?

src/commands.ts
ethanndickson: Yeah, I tried that package but it just straight up exports one function, and none of the stuff in the readme. Am I missing something obvious?

deansheather:
```
import ipRangeCheck from "ip-range-check";
// ...
ipRangeCheck(ip, cidr);
```

matifali:
> Including the connection status indicator when using Connect is out of scope of the linked issue and this PR, as Coder Desktop does not currently expose those values in any way.

This may be possible after we have https://github.com/coder/coder-desktop-macos/issues/64

> Opening an unowned workspace fallsback to using the Coder CLI, as Connect doesn't support shared workspaces yet.

Does this mean we only use the CLI to start the workspace and later again connect via Coder Connect?

ethanndickson:
> > Including the connection status indicator when using Connect is out of scope of the linked issue and this PR, as Coder Desktop does not currently expose those values in any way.
> 
> This may be possible after we have [coder/coder-desktop-macos#64](https://github.com/coder/coder-desktop-macos/issues/64)

Unfortunately, getting connection stats from the network extension to the desktop app is just one part. We also need to expose them programmatically in order for the VSCode extnesion to read them.

> > Opening an unowned workspace fallsback to using the Coder CLI, as Connect doesn't support shared workspaces yet.
> 
> Does this mean we only use the CLI to start the workspace and later again connect via Coder Connect?

As of my last commit, yep! You can start a workspace, watch it build in VSCode, and then once the build is finished VSCode will switch over to using the Connect tunnel.

DanielleMaywood:
Connecting to a dev container inside a Coder workspace also makes use of SSH. Should that also be updated?

matifali:
What will happen if the Coder Connect tunnel breaks due to an upgrade, or is turned off, or whatever? DO we gracefully fall back to CLI?
Ideally, the user should not have to reconnect manually or reload the VS Code window.

src/api.ts
```diff
@@ -280,3 +280,34 @@ export async function waitForBuild(
...
+      case 404: {
+        // Very old deployment that doesn't support SSH config
+        return {
+          hostname_prefix: "coder",
+          hostname_suffix: "coder",
+          ssh_config_options: {},
+        }
+      }
+      case 401: {
+        vsc.window.showErrorMessage("Your session expired...")
```
code-asher: I know this already existed, but do we need this notification still?  It looks like everywhere we call this we already show a notification.

src/commands.ts
```diff
@@ -540,107 +547,167 @@ export class Commands {
...
+    // For 2. `Remote.setup` will call `maybeAskAgent` and then migrate us to Coder Connect
+    if (workspaceAgent) {
+      let sshConfig
+      try {
+        // Fetch (or get defaults) for the SSH config.
+        sshConfig = await fetchSSHConfig(this.restClient, this.vscodeProposed)
+      } catch (error) {
+        const message = getErrorMessage(error, "no response from the server")
+        this.storage.writeToCoderOutputChannel(`Failed to open workspace: ${message}`)
+        this.vscodeProposed.window.showErrorMessage("Failed to open workspace", {
```
code-asher: If the error was a 401, ideally we add a "log in" button on the notification that can be used to provide a new token (if token auth is being used) and try again, similar to what `Remote.setup` does.

src/commands.ts
```diff
@@ -540,107 +547,167 @@ export class Commands {
...
+    // 2. A `vscode://coder.coder-remote` URI does not include the agent as a
+    // query parameter.
+    //
+    // For 1. `Remote.setup` will migrate us to Coder Connect once the workspace is built.
+    // For 2. `Remote.setup` will call `maybeAskAgent` and then migrate us to Coder Connect
+    if (workspaceAgent) {
+      let sshConfig
+      try {
+        // Fetch (or get defaults) for the SSH config.
+        sshConfig = await fetchSSHConfig(this.restClient, this.vscodeProposed)
```
code-asher: The extension has an SSH config section which I assume is not going apply to Coder Connect, is that right? As well as anything the user has added to their SSH config manually.

Wondering if we should update the setting description to say the variables will not apply if there is a Coder Connect session we can use.

Speaking of settings, should using Coder Connect be behind a setting as well?  Or a prompt?  "We see Coder Connect, do you want to use that?" or similar.

src/commands.ts
```diff
@@ -540,107 +547,167 @@ export class Commands {
...
+    workspaceOwner: string,
+    workspaceName: string,
+    workspaceAgent: string | undefined,
+    devContainerName: string,
+    devContainerFolder: string,
+  ) {
+    let remoteAuthority = toRemoteAuthority(baseUrl, workspaceOwner, workspaceName, workspaceAgent)
+
+    if (workspaceAgent) {
+      let sshConfig
```
code-asher, line 665: Same thoughts here about the SSH config and 401s.

src/commands.ts
```diff
@@ -540,107 +547,167 @@ export class Commands {
...
+      }
+
+      const coderConnectAddr = await maybeCoderConnectAddr(
+        workspaceAgent,
+        workspaceName,
+        workspaceOwner,
+        sshConfig.hostname_suffix,
+      )
+      if (coderConnectAddr) {
+        remoteAuthority = `ssh-remote+${coderConnectAddr}`
```
code-asher, line 689: Probably a dumb question, but this can result in multiple host names for the same workspace/agent?  One for Connect, one for the fallback.  Will that be potentially confusing, for example in the recents list I imagine it could show both?  I know for the fallback we reopen with Connect if we can, but it does seem unfortunate to have both.

Is there any world where we use Connect if possible as part of the proxy command or does it have to be in the extension itself?  That would make it trivial to implement for other IDEs and would let us reuse the same host name (and I suppose it would make it work for `ssh <vscode host>` on the command line too, although idk why someone would do that except for debugging).  I would dread having to do this for Toolbox, if we even can.  :fearful:

src/util.ts
```diff
@@ -21,14 +25,26 @@ export const AuthorityPrefix = "coder-vscode"
  * Throw an error if the host is invalid.
  */
 export function parseRemoteAuthority(authority: string): AuthorityParts | null {
-  // The authority looks like: vscode://ssh-remote+<ssh host name>
-  const authorityParts = authority.split("+")
+  // The Dev Container authority looks like: vscode://attached-container+containerNameHex@ssh-remote+<ssh host name>
```
code-asher, line 28: Good catch!

src/util.ts
```diff
@@ -21,14 +25,26 @@ export const AuthorityPrefix = "coder-vscode"
...
+  // The SSH authority looks like: vscode://ssh-remote+<ssh host name>
+  const authorityParts = authority.split("@")
```
code-asher: I wonder if `new URL(authority)` would be more robust here for parsing?  The `attached-container+containerNameHex` part would be in `url.username`, and `ssh-remote+<ssh host name>` would be in `url.host`.

src/util.ts
```diff
@@ -53,14 +69,32 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
...
+  workspace: string,
+  owner: string,
+  hostnameSuffix: string,
+): Promise<string | undefined> {
+  const coderConnectHostname = `${agent}.${workspace}.${owner}.${hostnameSuffix}`
+  try {
+    const res = await promisify(lookup)(coderConnectHostname)
+    // Captive DNS portals may return an unrelated address, so we check it's
+    // within the Coder Service Prefix.
+    return res.family === 6 && ipRangeCheck(res.address, "fd60:627a:a42b::/48") ? coderConnectHostname : undefined
```
code-asher, line 86: idk anything about Connect; is it not possible for it to be using ipv4?

code-asher (commented):
I have not ran it yet but left some thoughts!

src/commands.ts
code-asher: I think it was just because it was not using any internal properties, no problem moving it imo

src/util.ts
```diff
@@ -53,14 +69,32 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
...
 
+export async function maybeCoderConnectAddr(
+  agent: string,
+  workspace: string,
+  owner: string,
+  hostnameSuffix: string,
+): Promise<string | undefined> {
+  const coderConnectHostname = `${agent}.${workspace}.${owner}.${hostnameSuffix}`
+  try {
+    const res = await promisify(lookup)(coderConnectHostname)
```
code-asher, line 83: The PR description mentioned Connect will not be used for other users' workspaces, is this where that happens?  The response here is some kind of error?

src/commands.ts
ethanndickson: >I would dread having to do this for Toolbox, if we even can.

It looks like Toolbox & regular Jetbrains don't use wildcard SSH configs? I think that makes implementing this on those a lot easier:

```
Match host coder-jetbrains-toolbox-foo--test.coder.bar exec "/path/to/coder connect exists agent1.foo.owner.coder"
  HostName agent1.foo.owner.coder
Host coder-jetbrains-toolbox-foo--test.coder.bar
  ProxyCommand /path/to/coder ssh owner/foo.agent1
  ConnectTimeout 0
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
  LogLevel ERROR
  SetEnv CODER_SSH_SESSION_TYPE=JetBrains
```

This is similar to how[ the new `coder config-ssh` works](https://github.com/coder/coder/pull/17419#issuecomment-2812080120), we only use the ProxyCommand if the DNS record doesn't exist, which we check using `coder connect exists`.


Since we use wildcard configs for the VSCode extension, using just the SSH config is pretty complicated. We'd need `coder connect exists` to be able to handle an argument of the form `<owner>--<workspace>.<agent>`. Then, we'd need another command to translate `<owner>--<workspace>.<agent>` to `<agent>.<workspace>.<owner>.<hostnamesuffix>` if `coder connect exists` returned 0.

I assume it would look something like:
```
Host coder-vscode.dev.coder.com--* exec "/path/to/coder connect exists %h"
  ProxyCommand ssh "$(coder connect convert %h)
```

In both `exists` and `convert`, we'd need to auth with `coderd` to retrieve the hostname suffix, and then if the agent is missing from the authority string, we'd also need to resolve that somehow. 

I think this approach is just too slow; two CLI invocations that need to init a `codersdk.Client` after reading from disk. Right now `coder connect exists` doesn't need to do that.

Alternatively, we could change the portion after `coder-vscode.dev.coder.com--`, to `<agent>.<workspace>.<owner>.<hostnamesuffix>`, but `coder ssh` doesn't support that right now, and we'd still need to fetch the hostnamesuffix in the extension. 
This might be worth doing? We just only set the new hostname in the remote authority if the Coder CLI would support it. 

However, it might be a big lift to ensure each plugin knows what agent it's connecting to *before* calling SSH. Right now the Coder CLI handles that if it's not supplied in the remote authority slug.

>  Will that be potentially confusing, for example in the recents list I imagine it could show both?

But yeah, it does show both unfortunately. WDYT about the alternatives?

src/util.ts
ethanndickson: Yep, workspaces are only assigned IPs in that specific range.

src/util.ts
ethanndickson: Yeah, if Coder Connect hasn't configured DNS records for a workspace into it's resolver, the lookup will fail and the extension will just use `coder ssh`. There's no error message or anything, presumably the user will know it won't work because that workspace doesn't appear in Coder Desktop.

src/commands.ts
ethanndickson: > Wondering if we should update the setting description to say the variables will not apply if there is a Coder Connect session we can use.

Will add. 

>  As well as anything the user has added to their SSH config manually.

These will apply, as long as they're on the Coder Connect hostname (so, if they just specify those options when running `coder config-ssh` it'll get added to `*.<hostnamesuffix>`).


> Speaking of settings, should using Coder Connect be behind a setting as well? Or a prompt? "We see Coder Connect, do you want to use that?" or similar.

I don't think this is necessary, I can't imagine a scenario where the user would have Coder Connect on and not want to use it?

ethanndickson:
> What will happen if the Coder Connect tunnel breaks due to an upgrade, or is turned off, or whatever? DO we gracefully fall back to CLI?
> Ideally, the user should not have to reconnect manually or reload the VS Code window.

We don't currently. We'd need to watch for disconnect events from VSCode (assuming it exposes those via it's extension API), and react accordingly.

src/commands.ts
matifali: > I don't think this is necessary, I can't imagine a scenario where the user would have Coder Connect on and not want to use it?

I agree.

src/commands.ts
matifali: Toolbox does use wildcard hostnames by default, with an option to opt out. see the sample config from my system

```config
# --- START CODER JETBRAINS TOOLBOX dev.coder.com
Host coder-jetbrains-toolbox-dev.coder.com--*
  ProxyCommand "/Users/matifali/Library/Application Support/coder-toolbox/dev.coder.com/coder-darwin-arm64" --global-config "/Users/matifali/Library/Application Support/coder-toolbox/dev.coder.com/config" --url https://dev.coder.com ssh --stdio --disable-autostart --usage-app=jetbrains --ssh-host-prefix coder-jetbrains-toolbox-dev.coder.com-- %h
  ConnectTimeout 0
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
  LogLevel ERROR
  SetEnv CODER_SSH_SESSION_TYPE=JetBrains

Host coder-jetbrains-toolbox-dev.coder.com-bg--*
  ProxyCommand "/Users/matifali/Library/Application Support/coder-toolbox/dev.coder.com/coder-darwin-arm64" --global-config "/Users/matifali/Library/Application Support/coder-toolbox/dev.coder.com/config" --url https://dev.coder.com ssh --stdio --disable-autostart --usage-app=disable --ssh-host-prefix coder-jetbrains-toolbox-dev.coder.com-bg-- %h
  ConnectTimeout 0
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
  LogLevel ERROR
  SetEnv CODER_SSH_SESSION_TYPE=JetBrains
# --- END CODER JETBRAINS TOOLBOX dev.coder.com
```

matifali:
> We don't currently. We'd need to watch for disconnect events from VSCode (assuming it exposes those via it's extension API), and react accordingly.

I think this should be next in our priority, given how important it is to provide a seamless connection.

src/commands.ts
code-asher: I think the tricky part is that the user might not realize it will use Connect.  So, they have `X11Forwarding yes` in their VS Code config or in their SSH config under `Host coder-vscode--*`, they are trying out Connect, one day they update the plugin, and suddenly no more forwarding.  Some days it does forward, some days does not, depending on whether Connect is up (idk if this is realistic usage/behavior though).

So there should be clear messaging what the plugin is going to do IMO, somewhere.  Silently changing behavior feels bad to me.  But, I am neither a VS Code nor Connect user, so take what I say with a grain of salt.  :laughing:  Maybe the behavior change will be obvious to users.

If we could make the same SSH settings apply either way, then that would be a different story.

src/commands.ts
code-asher: Yeah, we do use wildcards in Gateway and Toolbox now.  I do not think we will be able to switch the hostname from out of under Toolbox like we can here in VS Code unfortunately (at least not the way it is currently implemented, maybe it can if we implement our own connector instead of using their SSH stuff but idk if that is possible).  :cry:   If we can implement purely in the SSH config and CLI, that would be ideal.

IDK the best way to do it, but my initial thought would be to make `coder ssh` handle it.  Maybe it should be behind a `--use-connect` flag or something, but basically it would just do whatever `coder connect convert` is doing, right?  It could even do the `exists` part as well, if we want to avoid writing new blocks.

IMO remote editing from the IDE's point of view should be more or less just `ssh <hostname>` and we can hide away and reuse all this logic in the CLI.

code-asher: But, I do like the example you have as well.  How slow are we talking?  Some extra seconds to connect may not be the worst thing, if it reduces complexity on our end and makes it possible to support Connect with Toolbox.

matifali: Not directly related but with toolbox JB uses the system ssh and support all SSH options. In gateway they had a built-in SSH implementation that didn't support all options.

ethanndickson: We decided against making it part of `coder ssh` because of the performance cost of a bicopy on the SSH connection. That's what led us to add `coder connect exists` for `config-ssh`.

> How slow are we talking? 

Honestly, it might be negligible on macOS and Linux. On the other hand, Spike discovered that `coder connect exists` takes almost 1.5s on Windows, and that's literally just to execute the binary and do a DNS lookup, so I'd expect this approach to be significantly slower there.

src/commands.ts
ethanndickson: > If we could make the same SSH settings apply either way, then that would be a different story.

With the updated `coder ssh-config`, any options applied to that will also apply to Coder Connect via VSCode, at least.

Though if we change the approach to what I discussed below that solves this issue too ig

src/commands.ts
code-asher: Ahhh, fair.  Was the performance cost of the bicopy measured?  It does sound unfortunate.

1.5s for running `exists` is wild lol, I guess that is Windows for ya.

IMO we have a few paths forward, ideally we figure out which one before merging since this may inform how we should do it here as well:

1. Do it somehow in the SSH config to support Toolbox
2. Decide not to support Toolbox
3. Figure out if there is a different way we can support Toolbox

But also do not let me just be a blocker if we need to move forward for now haha.  I do think the doubled-up host names and the SSH settings have the potential to cause some discomfort though.

It occurs to me that if we had our own connection code instead of delegating to Microsoft's plugin, we could transparently use a different host name under the connection without doubling up on them.  We could name the recent connections anything we wanted at that point.  We could even add settings at that point too, I think??  :thinking:

src/commands.ts
code-asher: > With the updated coder ssh-config, any options applied to that will also apply to Coder Connect via VSCode, at least.

That is quite nice!

PR #488, opened 2025-04-21
Bcpeinhardt/ai agent session in vscode

bcpeinhardt:
https://github.com/user-attachments/assets/e9823178-f140-4666-8be3-681c406052da


matifali (approved):
Can we also test with Cursor and WindSurf just to make sure it works as expected?

Issue #504, opened 2025-05-15
SSH config overwritten with VS Code Coder config


aaronlehmann:
Both myself and and another user experienced `~/.ssh/config` being completely overwritten to include just a `START CODER VSCODE` block with the corresponding wildcard `Host` entry. This happened to me using version 1.8.0 of the extension with multiple remote development sessions active.

johnstcn:
Hey Aaron, 

I found that, a missing `#--- END CODER VSCODE .* ---` comment in combination with more than one `#--- START CODER VSCODE .* ---` comment could cause any content between the first START and the first END to be removed.

Example:
```
Host beforeconfig
  HostName before.config.tld
  User before

# --- START CODER VSCODE dev.coder.com ---
Host coder-vscode.dev.coder.com--*
  ConnectTimeout 0
  LogLevel ERROR
  ProxyCommand some-command-here
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null

Host donotdelete
  HostName dont.delete.me
  User please

# --- START CODER VSCODE dev.coder.com ---
Host coder-vscode.dev.coder.com--*
  ConnectTimeout 0
  LogLevel ERROR
  ProxyCommand some-command-here
  StrictHostKeyChecking no
  UserKnownHostsFile /dev/null
# --- END CODER VSCODE dev.coder.com ---

Host afterconfig
  HostName after.config.tld
  User after 
```

With `coder config-ssh`, the above will currently (correctly) cause an error. However, the VSCode extension will silently remove `Host donotdelete`. 

https://github.com/coder/vscode-coder/pull/510 should fix this.

Does this sound plausible as a root cause to your issue?

Sidenote: there's also a potential issue with concurrent file writes. Since you mentioned having multiple remote development sessions it's also possible there was some kind of race condition with writing the updated config. `config-ssh` does do an atomic write, but currently the VSCode extension doesn't. Filed https://github.com/coder/vscode-coder/pull/511 for this separately.

aaronlehmann:
My guess is that the issue was non-atomicity. This happened seemingly randomly, so I wouldn't expect a deterministic failure mode like a missing `#--- END CODER VSCODE .* ---` to be the issue. Also, the other person this happened to tends to have a lot of VS Code windows open at once.

aaronlehmann:
Somewhat related: we've seen a race condition in the past updating the `coder` binary: https://github.com/coder/vscode-coder/issues/347

[...]

PR #510, opened 2025-05-20
fix: SSHConfig: check for multiple start/end blocks

johnstcn:
Relates to https://github.com/coder/vscode-coder/issues/504

# What this does:

Ports over the existing checks in `coder config-ssh` to handle malformed blocks.

Now, if your SSH config is malformed, you will see an error similar to the below upon attempting to connect:

* If one of the automatically generated blocks is missing a start or end comment, you will see the below error:
  Screenshot 2025-05-21 at 17 15 04

* If there is more than one block with a given label, you will see the below error:
Screenshot 2025-05-21 at 17 14 39

In both cases, it will unfortunately be necessary to edit the SSH config and fix the issue manually.

### DIFF
--- a/src/sshConfig.ts
+++ b/src/sshConfig.ts
@@ -126,25 +126,26 @@ export class SSHConfig {
     const raw = this.getRaw()
     const startBlock = this.startBlockComment(label)
     const endBlock = this.endBlockComment(label)
+
     const startBlockCount = countSubstring(startBlock, raw)
     const endBlockCount = countSubstring(endBlock, raw)
-    const startBlockIndex = raw.indexOf(startBlock)
-    const endBlockIndex = raw.indexOf(endBlock)
-    const hasBlock = startBlockIndex > -1 && endBlockIndex > -1
-
-    if (!hasBlock) {
-      return
-    }
-
     if (startBlockCount !== endBlockCount) {
       throw new SSHConfigBadFormat(
         `Malformed config: Unterminated START CODER VSCODE ${label ? label + " " : ""}block: Each START block must have an END block.`,
       )
     }
+
     if (startBlockCount > 1 || endBlockCount > 1) {
       throw new SSHConfigBadFormat(
         `Malformed config: ssh config has ${startBlockCount} START CODER VSCODE ${label ? label + " " : ""}sections, please remove all but one.`,
       )
+    }
+
+    const startBlockIndex = raw.indexOf(startBlock)
+    const endBlockIndex = raw.indexOf(endBlock)
+    const hasBlock = startBlockIndex > -1 && endBlockIndex > -1
+    if (!hasBlock) {
+      return
     }
 
     if (startBlockIndex === -1) {
--- a/src/sshConfig.ts
+++ b/src/sshConfig.ts
@@ -138,12 +138,12 @@ export class SSHConfig {
 
     if (startBlockCount !== endBlockCount) {
       throw new SSHConfigBadFormat(
-        `Malformed config: ssh config has ${startBlockCount} ${label} START comments but ${endBlockCount} ${label} END comments. Each START must have a matching END.`,
+        `Malformed config: Unterminated START CODER VSCODE ${label ? label + " " : ""}block: Each START block must have an END block.`,
       )
     }
     if (startBlockCount > 1 || endBlockCount > 1) {
       throw new SSHConfigBadFormat(
-        `Malformed config: ssh config has ${startBlockCount} ${label} sections, please remove all but one.`,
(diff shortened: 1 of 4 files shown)


src/util.ts
```diff
@@ -120,3 +120,22 @@ export function expandPath(input: string): string {
   const userHome = os.homedir()
   return input.replace(/\${userHome}/g, userHome)
 }
+
+/**
+ * Return the number of times a substring appears in a string.
+ * @param needle string
+ * @param haystack string
+ * @returns number
```
code-asher: Not too important, but the types seem redundant to me with TypeScript, idk why we had these on the above function, but I think generally we have not used these annotations

src/sshConfig.ts
```diff
@@ -123,14 +124,29 @@ export class SSHConfig {
...
     }
 
+    if (startBlockCount !== endBlockCount) {
+      throw new SSHConfigBadFormat(
+        `Malformed config: ssh config has ${startBlockCount} ${label} START comments but ${endBlockCount} ${label} END comments. Each START must have a matching END.`,
+      )
+    }
+    if (startBlockCount > 1 || endBlockCount > 1) {
+      throw new SSHConfigBadFormat(
+        `Malformed config: ssh config has ${startBlockCount} ${label} sections, please remove all but one.`,
```
code-asher: I wonder if we should just say `${label} sections` without the `startBlockCount`, or the max of the start and end, or write both counts like the above error, since for example if there are zero start blocks but 1+ end blocks it would say "zero sections, please remove".

code-asher (approved):
Nice find!

src/sshConfig.ts
johnstcn: Won't `startBlockCount` and `endBlockCount` be different then, meaning we won't even hit this code path in that case? You'll get an error like

```
Malformed config: ssh config has 87 dev.coder.com sections, please remove all but one.
```

code-asher: Oh duh yup!

src/sshConfig.ts
```diff
@@ -123,14 +124,29 @@ export class SSHConfig {
...
+    const endBlockIndex = raw.indexOf(endBlock)
     const hasBlock = startBlockIndex > -1 && endBlockIndex > -1
 
     if (!hasBlock) {
       return
     }
 
+    if (startBlockCount !== endBlockCount) {
+      throw new SSHConfigBadFormat(
+        `Malformed config: ssh config has ${startBlockCount} ${label} START comments but ${endBlockCount} ${label} END comments. Each START must have a matching END.`,
```
bcpeinhardt: Suggestion: the word that I'm most used to/understand the quickest in this context is "unterminated", as in "error: unterminated string, string must have a closing quote."
Maybe something like "Unterminated START CODER VSCODE Block: Each START block must have an END block". I think leading with the count is a bit weird because that's more how we detected the error than the error itself.
It may also be nice to provide the full `START CODER VSCODE dev.coder.com` text to search for. They might have multiple blocks in their ssh config produced by coder or using the START/END format if they use more than one editor, coder instance, and/or the cli

src/util.ts
```diff
@@ -120,3 +120,22 @@ export function expandPath(input: string): string {
...
+  if (needle.length < 1 || haystack.length < 1) {
+    return 0
+  }
+  let count = 0
+  let pos = haystack.indexOf(needle)
+  while (pos !== -1) {
+    count++
+    pos = haystack.indexOf(needle, pos + needle.length)
+  }
+  return count
```
mafredri, line 137: I think this could be simplified, like so?

```ts
return [...haystack.matchAll(needle)].length
```

mafredri (approved):
Small suggestion but otherwise LGTM 👍🏻

src/util.ts
johnstcn: It looks way better, but `matchAll` wants a `RegExp`, which means `needle` would need to be properly un-escaped first.

mafredri: Oh yeah, forgot not all these methods are string | RegExp. That can be simply amended with `RegExp.escape(needle)`.

johnstcn: Only available in Node 24 and current version of VSCode is on 23.11.0 if I'm not mistaken 🥲

src/sshConfig.ts
johnstcn: Updated

src/util.ts
johnstcn: Removed type annotations

bcpeinhardt:
Note to self: add changelog updates in separate PR so we don't force reapproval.

PR #511, opened 2025-05-20
fix: SSHConfig: atomically write ssh config

johnstcn:
Relates to https://github.com/coder/vscode-coder/issues/504

To avoid race conditions, write to a tempfile and then rename.

Smoke-tested by:

- Setting immutable flag on `/Users/cian/.ssh/config` (`sudo chflags uchg $PATH`) and validating that the rename failed:
Screenshot 2025-05-21 at 20 13 14

- Setting immutable flag on `/Users/cian/.ssh` and validating that writing the temp file failed:
Screenshot 2025-05-21 at 20 13 56

- Once immutable flag removed (`sudo chflags nouchg $PATH`), connecting to workspace is successful.

### DIFF
--- a/src/sshConfig.ts
+++ b/src/sshConfig.ts
@@ -239,7 +239,9 @@ export class SSHConfig {
       recursive: true,
     })
     const randSuffix = Math.random().toString(36).substring(8)
-    const tempFilePath = `${this.filePath}.${randSuffix}`
+    const fileName = path.basename(this.filePath)
+    const dirName = path.dirname(this.filePath)
+    const tempFilePath = `${dirName}/.${fileName}.vscode-coder-tmp.${randSuffix}`
     await this.fileSystem.writeFile(tempFilePath, this.getRaw(), {
       mode: existingMode,
       encoding: "utf-8",
--- a/src/sshConfig.ts
+++ b/src/sshConfig.ts
@@ -1,4 +1,4 @@
-import { mkdir, readFile, writeFile, rename } from "fs/promises"
+import { mkdir, readFile, rename, stat, writeFile } from "fs/promises"
 import path from "path"
 import { countSubstring } from "./util"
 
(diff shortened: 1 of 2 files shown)


src/sshConfig.ts
```diff
@@ -223,10 +225,13 @@ export class SSHConfig {
       mode: 0o700, // only owner has rwx permission, not group or everyone.
       recursive: true,
     })
-    return this.fileSystem.writeFile(this.filePath, this.getRaw(), {
+    const randSuffix = Math.random().toString(36).substring(8)
+    const tempFilePath = `${this.filePath}.${randSuffix}`
+    await this.fileSystem.writeFile(tempFilePath, this.getRaw(), {
```
code-asher: I am probably overthinking this, but if there is somehow a file with this suffix already, would we want to error instead of overwrite?  We could use `flag: "wx"` on the `writeFile` call to error if the path already exists.

Vanishingly unlikely, but maybe there could be a freak collision where a user has a `~/.ssh/config.conf` and the random suffix was also `conf`.

johnstcn: 1) We could just write to `/tmp`
2) We could instead name it something like `config.vscode-coder.${timestamp}.${randSuffix}` to minimize the likelihood of collisions

Thoughts?

code-asher: I like 2 personally because I think it gives us a better chance to get the temp files for debugging, with `/tmp` they could be long gone.

But I could see an argument for `/tmp` to avoid "clutter".

bcpeinhardt: What about a ULID? It would be sufficiently random and you could sort them by time to determine write order when debugging.

mafredri: If we want to guarantee atomicity, same folder is best (`/tmp` for instance could be on another filesystem).

To hide clutter in the case of error, I'd suggest a `.` prefix on the filename, perhaps `.tmp.[filename].[rand]`. At this point guaranteeing uniqueness beyond this would seem like overkill.

I.e. my suggestion is just `/home/user/.ssh/config -> /home/user/.ssh/.tmp.config.XXXXXX`.

One use-case that won't work here is if `/home/user/.ssh/config` is bind-mounted. In that situation we either error out or try to write the file directly. Erroring is obviously safest so perhaps we go with that.

johnstcn: > What about a ULID?

I like the idea in theory, but it would be another import.

mafredri: Forgot to write down one more thing, we should also respect the existing file modes and permissions and make sure they're copied over.

johnstcn: Done. I've also adjusted the tempfile naming convention to `~/.ssh/.config.vscode-coder-tmp.abc123` to make it abundantly clear that
1) This refers to `~/.ssh/config`
2) It was created by `vscode-coder`

mafredri (approved):
Nice work!

bcpeinhardt (approved):

Issue #430, opened 2025-01-29
Proxy log directory doesn't handle Windows style \


spikecurtis:
I set my Proxy Log Directory to `C:\Users\micha\logs` in settings.

Then, when connecting to a workspace, I get:

```
[17:15:11.240] > vdollar_percent_expand: unknown key %3
> percent_expand: failed
> The process tried to write to a nonexistent pipe.
```

in ~\.ssh\config I see that this is rendered: `--log-dir C%3A%5CUsers%5Cmicha%5Clogs` which is presumably what OpenSSH is choking on.

PR #512, opened 2025-05-22
feat: handle windows specific paths in log-path parameter

ibetitsmike:
Fixes: #430

### DIFF
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -677,7 +677,7 @@ export class Remote {
 		this.storage.writeToCoderOutputChannel(
 			`SSH proxy diagnostics are being written to ${logDir}`,
 		);
-		return ` --log-dir ${escape(logDir)}`;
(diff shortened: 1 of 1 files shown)


package-lock.json
matifali, line 1: This looks like mistakenly committed. We use `yarn`, which produces a `yarn.lock` file.

bcpeinhardt:
I suspect the issue was that calude nuked your lockfile. Just reverting the lockfile fixed the build. I'll leave passing the test and lint checks up to you :)

code-asher (approved):

ibetitsmike:
> Oh duh it was using `escape` before.  Looks good to me.  I do think we should test this stuff eventually.

@code-asher as in 'unit test'?

code-asher:
Unit test would be ideal yeah!  But I think `updateSSHConfig()` would need to be broken out of `Remote`.  Might not be worth the effort right now, feel free to merge.

PR #526, opened 2025-06-02
fix: specify `--header-command` when running `coder start`

### LINKED ISSUE #499: Bug: When using AppProxy in Azure, headers are not passed
Similar to : https://github.com/coder/vscode-coder/issues/76 but different source.

This is for an Issue a customer raised in Coder directly: https://github.com/coder/coder/issues/17687, after debugging this affects the VS Code extension and using coder cli has different behaviour.

It seems that the Coder Extension misses to pass the --header-command to the CLI while coder cli on it's own will work.

**Issue:**
 I can not start a stopped workspace from a local VSCode instance when Coder is behind a Azure AppProxy (using --header)

**Expected behaviour:**
Workspace should start

**Actual behaviour:**
Failed to open workspace

"--global config
...
coder-config-xyz exited with code 1: Encountered an error running "coder start", see "coder start --help" for more information
error: invalid character '<' looking for beginning of value"

**Steps to Reproduce**
Wait until a workspace is stopped in local VSCode
Reload Window while network was down (has nothing to do with the issue)
When VSCode asks to start the stopped workspace say yes

Coder: 2.20.2
OS: Linux

aslilac:
Closes #499

The `coder start` command can't always work properly without specifying these args (eg., getting past proxies). This refactors the code a little to make it easier to get the right argument values, and passes them to `coder start`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Multiple open instances of the extension could potentially clobber writes to
   `~/.ssh/config`. Updates to this file are now atomic.
 - Add support for `anysphere.remote-ssh` Remote SSH extension.
+- Use `--header-command` properly when starting a workspace.
 
 ## [v1.9.0](https://github.com/coder/vscode-coder/releases/tag/v1.9.0) 2025-05-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Use `--header-command` properly when starting a workspace.
+
 ## [v1.9.1](https://github.com/coder/vscode-coder/releases/tag/v1.9.1) 2025-05-27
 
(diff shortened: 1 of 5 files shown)


bcpeinhardt (approved):

bcpeinhardt:
@aslilac would you mind adding a not to the changelog as well?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Multiple open instances of the extension could potentially clobber writes to
   `~/.ssh/config`. Updates to this file are now atomic.
 - Add support for `anysphere.remote-ssh` Remote SSH extension.
+- Use `--header-command` properly when starting a workspace.
```
bcpeinhardt: I think you meant to put this in the unreleased section!

aslilac: it's dark and scary in here, I got lost 😭

PR #528, opened 2025-06-03
chore: simplify prettier config

aslilac:
Pretty much all of the settings in the prettier config here are either in conflict with coder/coder, or are just the defaults now. We can just remove the whole file to get closer to matching coder/coder.


aslilac:
I would also like to switch this repository to tabs, but that will be a much bigger diff.

matifali:
We don't use prettier in coder/coder anymore. Maybe move to biome here too.

PR #529, opened 2025-06-03
chore: downgrade ua-parser-js version

aslilac:
This version must match the version in coder/coder. @aqandrew and dependabot mistakenly upgraded this version. (not mad, just letting you know)

I've updated dependabot's config to avoid it updating it to begin with, and added a build check to CI so that we can have better confidence in the build-ability of the main branch.


bcpeinhardt (approved):

aqandrew (approved):
My bad / thanks for the heads up!

PR #536, opened 2025-06-16
Add package scripts and cli library, enable integration testing

jaggederest:
A working vscode-test integration test that validates that the module will load and add commands, at least.


.eslintignore
```diff
@@ -0,0 +1 @@
+vitest.config.ts
```
aslilac: nit: missing a newline at the end. most editors have a setting to prevent this, and it's specified in our .editorconfig file here if your editor supports that.

src/extension.ts
```diff
@@ -21,25 +21,36 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
 	if (!remoteSSHExtension) {
-		vscode.window.showErrorMessage(
-			"Remote SSH extension not found, cannot activate Coder extension",
+		if (!isTestMode) {
+			vscode.window.showErrorMessage(
+				"Remote SSH extension not found, cannot activate Coder extension",
+			);
+			throw new Error("Remote SSH extension not found");
+		}
+		// In test mode, use regular vscode API
```
code-asher: Nice!  I wonder if we can avoid the special testing behavior.  Maybe we should always try to activate, but if we are unable to get this special API we skip the remote registration (and possibly log a warning)?  Similar to what you already have except we always do it, not just in testing.

That would make it nicer for use cases where you have not installed the remote extension yet, for whatever reason, while reducing drift between production and testing.

src/remote.ts
```diff
@@ -6,7 +6,9 @@ import * as fs from "fs/promises";
 import * as jsonc from "jsonc-parser";
 import * as os from "os";
 import * as path from "path";
-import prettyBytes from "pretty-bytes";
+// Dynamic import for ESM module
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let prettyBytes: any;
```
code-asher: I know you said this was temporary, just leaving a note here to acknowledge we need to update this and other references before merge.  :smile:

src/test/extension.test.ts
```diff
@@ -0,0 +1,56 @@
+import * as assert from "assert";
+import * as vscode from "vscode";
+
+suite("Extension Test Suite", () => {
```
code-asher, line 4: Exciting to have this working!!

code-asher (commented):
Leaving off an approve until we sort `prettyBytes` but this is great!

src/extension.ts
jaggederest: Yeah I suspect this is the kind of behavior we want, so 👍 to logging a warning and handling it gracefully

src/extension.ts
```diff
@@ -21,25 +21,31 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+		);
+	} else {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		vscodeProposed = (module as any)._load(
+			"vscode",
+			{
+				filename: remoteSSHExtension.extensionPath,
+			},
+			false,
 		);
```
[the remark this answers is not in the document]
jaggederest: Yeah I get it buddy I don't like it either but we're polymorphic on the SSH extensions.

src/extension.ts
```diff
@@ -21,25 +21,31 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
 		vscode.extensions.getExtension("anysphere.remote-ssh") ||
 		vscode.extensions.getExtension("ms-vscode-remote.remote-ssh");
+
+	let vscodeProposed: typeof vscode = vscode;
+
 	if (!remoteSSHExtension) {
 		vscode.window.showErrorMessage(
-			"Remote SSH extension not found, cannot activate Coder extension",
+			"Remote SSH extension not found, this may not work as expected.\n" +
+				// NB should we link to documentation or marketplace?
```
code-asher, line 36: In the VS Code case, it is a dependency so it should automatically download, but for the other forks I could see value in trying to determine which it is and print the right extension they need to install.  Something we can think about later, I think.

src/extension.ts
```diff
@@ -278,7 +284,13 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 	// Since the "onResolveRemoteAuthority:ssh-remote" activation event exists
 	// in package.json we're able to perform actions before the authority is
 	// resolved by the remote SSH extension.
-	if (vscodeProposed.env.remoteAuthority) {
+	//
+	// In addition, if we don't have a remote SSH extension, we skip this
+	// activation event. This may allow the user to install the extension
+	// after the Coder extension is installed, instead of throwing a fatal error
+	// (this would require the user to uninstall the Coder extension and
+	// reinstall after installing the remote SSH extension, which is annoying)
```
code-asher, line 292: :tada:

src/remote.ts
```diff
@@ -841,7 +841,7 @@ export class Remote {
 			`${sshPid}.json`,
 		);
 
-		const updateStatus = (network: {
+		const updateStatus = async (network: {
```
code-asher: We can revert this back to non-async yeah?

tsconfig.json
```diff
@@ -1,14 +1,17 @@
 {
 	"compilerOptions": {
 		"module": "commonjs",
-		"target": "es6",
+		"target": "ES2022",
```
code-asher: Do we know whether our minimum version of VS Code (1.73.0) supported es2022? Looks like it was Electron 19.0.17 based on Node 16.14.2 but I am not sure how to find the compatibility.

tsconfig.json
```diff
@@ -1,14 +1,17 @@
...
+		"moduleResolution": "node",
```
code-asher, line 5: I was reading that `node` was renamed to `node10` and `node` is a deprecated alias now.

This is a change from the default, right?  Supposedly `node10` is not meant to be used anymore but idk exactly what the consequences are.

tsconfig.json
```diff
@@ -1,14 +1,17 @@
...
 		"outDir": "out",
 		// "dom" is required for importing the API from coder/coder.
-		"lib": ["es6", "dom"],
+		"lib": ["ES2022", "dom"],
 		"sourceMap": true,
-		"rootDirs": ["node_modules", "src"],
 		"strict": true,
-		"esModuleInterop": true
+		"esModuleInterop": true,
+		"skipLibCheck": true,
```
code-asher, line 12: This loosens some type checking, right?  Do we need to disable it?

src/extension.ts
jaggederest: Oooh, interesting, let me put a ticket in for that, I think that's a wise idea.

tsconfig.json
jaggederest: I will check on that, thank you for pointing that out.

tsconfig.json
jaggederest: I don't believe we do, I'll revert that.

src/extension.ts
jaggederest: #537

package.json
```diff
@@ -278,8 +278,11 @@
 		"package:prerelease": "npx vsce package --pre-release",
 		"lint": "eslint . --ext ts,md",
 		"lint:fix": "yarn lint --fix",
-		"test": "vitest ./src",
-		"test:ci": "CI=true yarn test"
+		"test": "vitest",
+		"test:ci": "CI=true yarn test",
+		"test:integration": "vscode-test",
+		"pretest": "yarn run compile-tests && yarn run build && yarn run lint",
+		"compile-tests": "tsc -p . --outDir out"
```
aslilac: even if we only need it for compiling tests, this doesn't just compile tests. bit of a weird script name.

package.json
```diff
@@ -311,7 +314,8 @@
 		"vitest": "^0.34.6",
 		"vscode-test": "^1.5.0",
 		"webpack": "^5.99.6",
-		"webpack-cli": "^5.1.4"
+		"webpack-cli": "^5.1.4",
+		"@vscode/test-cli": "^0.0.10"
```
aslilac: ordering

src/workspacesProvider.ts
```diff
@@ -156,7 +156,7 @@ export class WorkspaceProvider
 
 		// Create tree items for each workspace
 		const workspaceTreeItems = await Promise.all(
```
aslilac: this is no longer an array of promises, so you can remove the `Promise.all` too

tsconfig.json
```diff
@@ -1,14 +1,17 @@
...
+		"forceConsistentCasingInFileNames": true
 	},
-	"exclude": ["node_modules", ".vscode-test"]
+	"exclude": ["node_modules", "vitest.config.ts"],
```
aslilac: nit: this is all outside of src/ and would never be picked up anyway

tsconfig.json
aslilac: we should probably set this to `"node16"`

.eslintrc.json
```diff
@@ -17,7 +18,10 @@
 	"overrides": [
 		{
 			"files": ["*.md"],
-			"parser": "markdown-eslint-parser"
+			"parser": "markdown-eslint-parser",
+			"rules": {
+				"@typescript-eslint/require-await": "off"
```
aslilac: why are we disabling this rule for markdown files? we shouldn't be linting those with eslint

tsconfig.json
jaggederest: Looks like vscode itself as of 1.73.0 uses a `target` of `ES2021`, which is also recommended by the tsconfig for node 16 that version uses. https://github.com/tsconfig/bases/blob/main/bases/node16.json - I'll swap over to that.

tsconfig.json
jaggederest: `moduleResolution` has dependencies on `module`, in that it cannot be node16 without module also being node16 which would cause the prior `pretty-bytes` import issue. I think leaving it for now and fixing it in a future PR is probably fine, given that's the case. We could also kick the `pretty-bytes` lib to the curb as well if it's not playing nice.

jaggederest: (and the default value when it's not set when `module` is commonjs is `node10` which is the synonym)

.eslintrc.json
jaggederest: `"lint": "eslint . --ext ts,md",` is the current version, I'm happy to disable it, I was just trying not to change behavior :)

tsconfig.json
jaggederest: Ah, we do, for the `pretty-bytes` issue. It's the recommended config from typescript mothership so I'm fine with it.

tsconfig.json
code-asher: Ooo this repo looks useful.  es2021 sounds like it should work, although, weirdly I see es2021 in `src` but es2020 in `build` for 1.73.0, no idea what is going on there.  

https://github.com/microsoft/vscode/blob/1.73.0/src/tsconfig.json
https://github.com/microsoft/vscode/blob/1.73.0/build/tsconfig.json

tsconfig.json
code-asher: Ohhhhh I see, so this was the default already, not a change to the setting, just making it explicit.  :+1:   Something we can take a look at again in the future, perhaps.

tsconfig.json
code-asher: Ahh fair, I get nervous because the docs say something like it reduces type accuracy :stuck_out_tongue:  Maybe we can take a look at it again later if we come back to updating the module resolution in the future since `pretty-bytes` is a blocker there as well.

.eslintrc.json
code-asher: Ah yeah prettier is ran through eslint, which runs on markdown.  It does seem weird to me as well, maybe they should be split.

Does this mean if we have any other rules we have to disable them here?  I wonder if there is some configuration we are missing that is supposed to make this work more smoothly.

code-asher: Or, would it work better if we enabled the rule only in `.ts` files maybe?

jaggederest: Yeah that makes sense, I'll swap that out since now I'm skipping it on md and json both, which is 🤦 , so I'll explicitly enable it on `.ts` files only instead. the joys of whitelist/blacklist

code-asher (approved):
Looks good to me!  Good call on the async rule.  If there is an easy way to configure it more cleanly that would be awesome, otherwise maybe we could opt to split linting and formatting at some point.

src/remote.ts
```diff
@@ -910,9 +910,9 @@ export class Remote {
 				.then((content) => {
 					return JSON.parse(content);
 				})
-				.then((parsed) => {
+				.then(async (parsed) => {
 					try {
-						updateStatus(parsed);
+						await updateStatus(parsed);
```
code-asher: `updateStatus` is not async anymore so I think we can revert this.

package.json
```diff
@@ -1,38 +1,36 @@
...
+	"type": "commonjs",
 	"main": "./dist/extension.js",
+	"scripts": {
+		"build": "webpack",
+		"fmt": "prettier --write .",
+		"lint": "eslint . --ext ts,md,json",
+		"lint:fix": "yarn lint --fix",
+		"package": "webpack --mode production --devtool hidden-source-map",
+		"package:prerelease": "npx vsce package --pre-release",
+		"pretest": "tsc -p . --outDir out && yarn run build && yarn run lint",
```
code-asher, line 27: Do we have to compile separately from Webpack because vscode-test requires that or something?  Is there a reason it cannot use the Webpack bundle?

jaggederest: I'm not sure, I'll see if I can ditch it in my next PR 👍

PR #538, opened 2025-06-24
fix: handle agent parameter in URIs

ethanndickson:
This fixes a bug where the `agent` query parameter on the extension URI was ignored.

We were previously doing:
```ts
vscode.commands.executeCommand(
	"coder.openDevContainer",
	workspaceOwner,
	workspaceName,
	workspaceAgent,
	devContainerName,
	devContainerFolder,
);
```
where `args[2]` was the agent name, but we were discarding it:
```ts
const workspaceOwner = args[0] as string;
const workspaceName = args[1] as string;
const workspaceAgent = undefined; // args[2] is reserved, but we do not support multiple agents yet.
const devContainerName = args[3] as string;
const devContainerFolder = args[4] as string;
```

The same was true for the `coder.open` command. Presumably due to the comment saying multiple agents aren't supported, which hasn't been true for years.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -748,7 +748,7 @@ async function openDevContainer(
 	baseUrl: string,
 	workspaceOwner: string,
 	workspaceName: string,
-	workspaceAgent: string | undefined,
+	workspaceAgent: string,
 	devContainerName: string,
 	devContainerFolder: string,
 ) {


ethanndickson:
* **#538** <a href="https://app.graphite.dev/github/pr/coder/vscode-coder/538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/coder/vscode-coder/538?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/commands.ts
```diff
@@ -628,7 +628,7 @@ export class Commands {
 
 		const workspaceOwner = args[0] as string;
 		const workspaceName = args[1] as string;
-		const workspaceAgent = undefined; // args[2] is reserved, but we do not support multiple agents yet.
+		const workspaceAgent = args[2] as string;
```
ethanndickson, line 631: `openDevcontainer` URIs must include the agent name to work properly, I believe, so this is `string` and not `string | undefined`.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,5 @@
+# If you would like `git blame` to ignore commits from this file, run:
+# git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# chore: simplify prettier config (#528)
+f785902f3ad20d54344cc1107285c2a66299c7f6
```
mafredri, line 5: What's the purpose of this?

mafredri (approved):
LGTM, thanks for fixing this!

.git-blame-ignore-revs
ethanndickson: We have the same file on `coder/coder` - it's just a convenient way to remove a formatting commit from `git blame` results. I noticed every line was blamed on #528. No longer the case after running `git config blame.ignoreRevsFile .git-blame-ignore-revs`

PR #541, opened 2025-06-26
Bulk refactor - increase test coverage and add integration tests, extract some difficult to test methods

jaggederest:
Intended to shore up test coverage in preparation for more logging and disconnect work.

Tests are AI generated so they're not the clearest, but I haven't found any places I'd like to obviously improve them so I'll leave them be for now. Production code is largely unchanged, besides being broken up a bit from e.g. extension.ts#activate function into subsidiary objects and methods. Once again not the perfect breakup or nomenclature but more than adequate for the purpose of the moment.


jaggederest:
Apologies in advance for the huge PR, I decided after splitting it up a few different ways that it was best to just rip the bandaid off all at once. Particularly looking for reviews of the changes from the existing tests and production code, the new tests are mock-heavy and thus don't need as intensive a review given the number of lines of code involved. As is always my policy, also interested in any linter rules we might like to turn on - the async-without-await rules caught a few useful things, so more in that vein is beneficial.

aslilac (changes requested):
I understand the appeal of a bulk refactor, but +11k is way too large of a change to actually review effectively. I'm also incredibly suspicious of the fact that the diff is only -2k. yes, a lot of that diff is tests, but by your own admission...

> Tests are AI generated so they're not the clearest

tests provide value when they're clearly readable, and written with specific behaviors to test in mind. I don't have much confidence that that's what is happening here. from my own experience claude can be pretty good at writing code these days but it is still terrible at writing tests. even claude 4 still doesn't seem to have any understanding of what a test is really for. it just knows what they look like on average.

in general we try to keep pull requests under 500 lines. on rare occasion when it makes sense _and_ is coordinated with the reviewers we'll let pull requests through that are a bit over 1000 lines. over 10000 absolute _must_ be broken down further. 😅

jaggederest:
Cool, will do, appreciate the commentary!

Issue #532, opened 2025-06-13
Add debug logging around requests


code-asher:
We should log the start and stop of all requests, especially the long-lived ones like `/watch` (which we call for the current workspace before connecting to a remote) and `/watch-metadata` (which we call for every running agent on every user-owned workspace in the sidebar, although this should only happen while the sidebar is displaying).

So probably a request ID, endpoint, method, and payload for the start log, and then a request ID, status code, response body, and duration for the end log.

Also I think our logs are not timestamped right now?

This should help us narrow down reports of leaks, since it looks like these connections may not be disposing when they should.

PR #545, opened 2025-07-03
Logging adapter and usage

jaggederest:
#532

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -15,6 +15,6 @@ export default defineConfig({
 		environment: "node",
 		env: {
 			NODE_ENV: "test",
-		  },
+		},
 	},
 });


src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+export class OutputChannelAdapter implements LogAdapter {
+	constructor(private outputChannel: vscode.OutputChannel) {}
+
+	write(message: string): void {
+		try {
+			this.outputChannel.appendLine(message);
+		} catch {
+			// Silently ignore - channel may be disposed
```
code-asher, line 21: We could perhaps `console.log` as a last-ditch effort

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+
+	clear(): void {
+		try {
+			this.outputChannel.clear();
+		} catch {
+			// Silently ignore - channel may be disposed
```
code-asher, line 29: Same here

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+	private configListener: vscode.Disposable | null = null;
+
+	constructor() {
+		this.updateLogLevel();
+		this.setupConfigListener();
+	}
+
+	private setupConfigListener(): void {
+		// In test environment, vscode.workspace might not be available
+		if (!vscode.workspace?.onDidChangeConfiguration) {
```
code-asher, line 73: Would it be possible to always mock these?  I feel like that would be preferable so we can minimize the number of test-specific branches in the production code, which increases the chance we miss something in tests.

Also, thinking out loud, this is only a problem because of our custom test setup right?  When you scaffold an extension by default, even in unit tests `vscode` is available I think?  Or am I wrong about that lol

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+		this.configListener = vscode.workspace.onDidChangeConfiguration((e) => {
+			if (e.affectsConfiguration("coder.verbose")) {
+				this.updateLogLevel();
+			}
+		});
+	}
+
+	private updateLogLevel(): void {
+		// In test environment, vscode.workspace might not be available
+		if (!vscode.workspace?.getConfiguration) {
```
code-asher, line 85: Same question here

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+			this.level = LogLevel.INFO;
+			return;
+		}
+		const config = vscode.workspace.getConfiguration("coder");
+		const verbose = config.get<boolean>("verbose", false);
```
code-asher, line 90: Do you think it would make sense to add a setting for levels instead?  I could see one day adding a level setting, then we would have both levels and verbose.

Which is itself not that strange, I think plenty of tools have both a verbose and log level flag, but it would match the remote extensions which have dropdowns for level settings.

At a minimum I think we would want trace, debug, info, warn, and error (even if we only use info right now, just so we have no problems with the numbering in the future where a setting of `2` used to mean `none` but would now mean `info`).

Also, we need to add the setting to the `package.json` as well right?

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+		this.level = verbose ? LogLevel.DEBUG : LogLevel.INFO;
+	}
+
+	private formatMessage(message: string, level: LogLevel): string {
+		const levelStr = LogLevel[level].toLowerCase();
+		const timestamp = new Date().toISOString();
+		let formatted = `[${levelStr}] ${timestamp} ${message}`;
+
+		// Add source location for debug messages when verbose is enabled
+		if (level === LogLevel.DEBUG && this.level === LogLevel.DEBUG) {
```
code-asher, line 100: In coder/coder at least, we use `debug` like a noisier `info`, and stack traces would be overwhelming.  I could see adding it to `trace` though, the name does imply a stack trace, and it would mirror `console.trace` which outputs a stack trace.

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+			const stack = new Error().stack;
+			if (stack) {
+				const lines = stack.split("\n");
+				// Find the first line that's not from the logger itself
+				for (let i = 2; i < lines.length; i++) {
+					const line = lines[i];
+					if (!line.includes("logger.ts") && !line.includes("Logger.")) {
```
code-asher, line 107: The stack frames are deterministic, right?  The first line is "Error", second is this function, third is the caller, I think?  Could we just skip the first stack frame/second line?

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+						const match =
+							line.match(/at\s+(.+)\s+\((.+):(\d+):(\d+)\)/) ||
+							line.match(/at\s+(.+):(\d+):(\d+)/);
+						if (match) {
+							const location =
+								match.length === 5
+									? `${match[1]} (${match[2]}:${match[3]})`
+									: `${match[1]}:${match[2]}`;
+							formatted += `\n  at ${location}`;
+						}
```
code-asher, line 117: Is it intentional that this drops the column number?  Do we need to mess with the stack frames though?  I imagine everything the stack contains is useful debugging info.

I suppose we also risk dropping lines that happen not to match, although idk if there are cases that fall outside these regular expressions.

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+	log(message: string, severity: LogLevel = LogLevel.INFO): void {
+		if (!this.adapter || severity < this.level) {
+			return;
+		}
+
+		const formatted = this.formatMessage(message, severity);
+		this.adapter.write(formatted);
+	}
+
+	debug(message: string): void {
```
code-asher, line 136: Love to see we are getting proper log levels :tada:

src/logger.ts
```diff
@@ -0,0 +1,198 @@
+import * as vscode from "vscode";
```
code-asher, line 1: Not a mandate, but wanted to make aware that we have a logger we use in some other projects: https://github.com/coder/ts-logger in case you wanted to make use of it.  It matches the logging style in our Go code and it also supports tracking how long things take, which might come in handy for request logging.  And it has color support, although that does not matter for the output channel.

It would look something like this I think:

```typescript
import { logger, Argument, Formatter, Level } from "@coder/logger";

export class OutputChannelFormatter extends Formatter {
  constructor() {
    super("", false) // Disable color.
  }

  // Output channel does not support colors.
  // Really we should make the formatter not require these functions unless colors
  // are enabled or something though.
  protected style(color?: string, weight?: string): string {
    return "";
  }

  // Output channel does not support colors.
  protected reset(): string {
    return ""
  }

  protected doWrite(level: Level, message: MessageFormat): void {
    const timestamp = new Date().toISOString()
    // Would need to implement sprintf
    let output = `${$timestamp} ${sprintf(message.format, message.args)}`
    // Output channel has no grouping; serialize fields and display on one line.
    const obj: { [key: string]: Argument } = {}
    if (message.fields.length > 0) {
      message.fields.forEach((field) => (obj[field.identifier] = field.value))
      output += " " + JSON.stringify(obj);
    }
    this.outputChannel.appendLine(output);
    if (level <= Level.Debug) {
      // output stack trace
    }
  }
}

logger.formatter = new OutputChannelFormatter()
```

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+	clear(): void {
+		try {
+			this.outputChannel.clear();
+		} catch {
+			// Silently ignore - channel may be disposed
+		}
+	}
+}
+
+export class ArrayAdapter implements LogAdapter {
```
code-asher, line 34: Could we move this into tests?  From what I gather this is purely for testing.

Or, we could mock the output channel.  Something like:

```typescript
let logs: string[] = []
const mockChannel = {
  appendLine: vi.fn().mockImplementation((line) => {
  logs.push(line)
}),
  clear: vi.fn().mockImplementation(() => logs = []),
} as unknown as vscode.OutputChannel
logger.setAdapter(new OutputChannelAdapter(mockChannel))
// do a thing
expect(logs).toBe(..)
```

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+	clear(): void {
+		this.logs = [];
+	}
+
+	getSnapshot(): readonly string[] {
+		return [...this.logs];
+	}
+}
+
+export class NoOpAdapter implements LogAdapter {
```
code-asher, line 50: Is logger performance really something we need to worry about?  If so, could we create a separate benchmarking suite, maybe with mitata or something?  I think to settle on hard performance thresholds in a unit test we would need to see the performance data first, although maybe we would want a separate pipeline for enforcing performance constraints.  

But also, the current test is comparing the difference between doing nothing and pushing to an array, which I am not sure tells us anything we could act on, and neither adapter is used in production anyway.

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+		this.adapter = null;
+		this.level = LogLevel.INFO;
+		if (this.configListener) {
+			this.configListener.dispose();
+			this.configListener = null;
+		}
+		// Re-setup config listener for next test
+		this.updateLogLevel();
+		this.setupConfigListener();
+	}
```
code-asher, line 183: If this is all purely for tests, could we refactor in a way that does not add unused API to the production code?  Maybe by extending the logger class or something?

I also see that the only places we use `withAdapter` is to test `withAdapter`, so it exists only for itself and seems like we could just delete it entirely.

For reset, we could create new instances instead, might be better to do that anyway to ensure unit tests are isolated.

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+
+	// Initialize with OutputChannel for production use
+	initialize(outputChannel: vscode.OutputChannel): void {
+		if (this.adapter !== null) {
+			throw new Error("Logger already initialized");
+		}
+		this.adapter = new OutputChannelAdapter(outputChannel);
+	}
```
code-asher, line 191: It feels a little backwards to me that the logger is built around a generic adapter, but also has knowledge about a specific adapter, if that makes sense?  Could we do something like `new Logger(new OutputChannelAdapter(output))` or something like that (or allow setting the adapter outside of tests).

src/logger.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Export singleton instance
+export const logger = new LoggerImpl();
+
+// Export types for testing
+export type Logger = typeof logger;
```
code-asher, line 198: If we only need it for testing, we should move it to the test file!

src/storage.ts
```diff
@@ -507,12 +508,12 @@ export class Storage {
...
-		// We don't want to focus on the output here, because the
-		// Coder server is designed to restart gracefully for users
-		// because of P2P connections, and we don't want to draw
-		// attention to it.
+	/**
+	 * Compatibility method for Logger interface used by headers.ts and error.ts.
+	 * Delegates to the logger module.
+	 */
+	public writeToCoderOutputChannel(message: string): void {
+		logger.info(message);
```
code-asher, line 516: Oh wait I thought you changed `headers.ts` and `error.ts`.  Can we delete this function now?

src/logger.ts
jaggederest: Cool, that sounds great. It looks like (based on my limited understanding) that it'll show up in the dev tools panel in VSCode, which is better than nothing. I'll make that change.

src/logger.ts
jaggederest: Yes, let me spike out an attempt to convert new tests to the mocha environment, using the vscode test cli runner maybe. I don't think we need electron full stack testing for these but I'm not sure how it'll work before a spike, but I definitely don't want to put a bunch of work into expanding and revamping the tests only to change them.

src/logger.ts
jaggederest: Yes on both fronts, I'll rejigger this branch to use a full log level setup, I wasn't sure about it but after working with it for a bit, it's not overkill. The spike I did using the log levels made it really clear that we're going to want a full stack anyway in short order. Thank you for the thorough review.

src/logger.ts
jaggederest: Yep I will pare this back. Sometimes Claude gets overly frisky.

src/storage.ts
jaggederest: Yep, as per usual Claude is overly conservative in what it retains for backwards compatibility. I'll get out my pruning shears manually 👍

Issue #121, opened 2023-08-10
Opening workspace or agent opens two different workspaces (even for a single-agent workspace)


mafredri:
Depending on if I open my workspace or agent explicitly, I get two different workspaces:

<img width="276" alt="image" src="https://github.com/coder/vscode-coder/assets/147409/5e934712-7e15-459f-a51b-76f3e5afbab8">
<img width="274" alt="image" src="https://github.com/coder/vscode-coder/assets/147409/484bef70-c90b-4541-a202-5a0f4f10e451">

In the former, VS Code remembers/re-opens files from last time, and in the latter VS Code opens another set of files, from last time I explicitly opened the agent.

Ideally opening coder workspace would translate to the same VS Code workspace as the primary agent.

EhabY:
This has been fixed by #552

PR #552, opened 2025-07-15
Use primary agent for the remote authority when opening a workspace

EhabY:
#121 

This change updates the behavior when opening a workspace from the sidebar: the first agent is now opened automatically using the same remote authority URI as when opening the agent directly.

If a workspace has multiple agents, we always open the first one, skipping the agent selection QuickPick when the folder is opened. I implemented it this way since we're already opening the `folderPath` of the first agent by default.

If this behavior should only apply when there's a single agent, it's easy to adjust. But in that case, should we also avoid passing the `folderPath` when multiple agents are present?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
   and configFile are provided.
 - Fix opening a workspace restores the VS Code session (files, tabs, context)
   of the selected agent.
+- Consistently use the same session for each agent. Previously,
+  depending on how you connected, it could be possible to get two
+  different sessions for an agent. Existing connections may still
+  have this problem, only new connections are fixed.
 
 ## [v1.9.2](https://github.com/coder/vscode-coder/releases/tag/v1.9.2) 2025-06-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Update `/openDevContainer` to support all dev container features when hostPath
   and configFile are provided.
+- Fix opening a workspace restores the VS Code session (files, tabs, context)
+  of the selected agent.
 
 ## [v1.9.2](https://github.com/coder/vscode-coder/releases/tag/v1.9.2) 2025-06-25
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -436,6 +436,9 @@ export class Commands {
 			if (!baseUrl) {
 				throw new Error("You are not logged in");
 			}
+			if (treeItem.primaryAgentName === undefined) {
+				return;
+			}
 			await openWorkspace(
 				baseUrl,
 				treeItem.workspaceOwner,
@@ -524,6 +527,8 @@ export class Commands {
 		let folderPath: string | undefined;
 		let openRecent: boolean | undefined;
 
+		let workspace: Workspace | undefined;
+
 		const baseUrl = this.restClient.getAxiosInstance().defaults.baseURL;
 		if (!baseUrl) {
 			throw new Error("You are not logged in");
@@ -570,7 +575,7 @@ export class Commands {
 					});
 			});
 			quickPick.show();
-			const workspace = await new Promise<Workspace | undefined>((resolve) => {
+			workspace = await new Promise<Workspace | undefined>((resolve) => {
 				quickPick.onDidHide(() => {
 					resolve(undefined);
 				});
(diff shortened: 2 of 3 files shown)


code-asher:
I think it is reasonable to open the first agent when you click on the workspace.  I could see an argument for making it pop open the agent select dialog, but if a user wants a different agent they could just click the right one directly on the sidebar instead.  But, idk if users feel differently.  We could always change it if we get feedback, this is already an improvement IMO.

That said, I think we should make an additional fix, in `openWorkspace` we should make `workspaceAgent` non-undefined so we always have to pass one in, and we always get a consistent remote authority, no matter what calls the function.

EhabY:
>  Ideally we do it before, so we are guaranteed one stable remote authority per agent.

It is now done when calling `open(...)` so it happens before we open the workspace folder.

> I think we should keep it in case someone uses the open command from the command palette. When going that route, I think it makes sense to show the agent picker still.

Actually for line 410 in `remote.ts#setup`: `const gotAgent = await this.commands.maybeAskAgent(workspace, parts.agent);` seems useless now because we never open without an agent (from the tree view or the command palette). I guess it's useful to keep if users manually input their remote authority (or use one from an older version)

code-asher:
> Actually for line 410 in remote.ts#setup: const gotAgent = await this.commands.maybeAskAgent(workspace, parts.agent); seems useless now because we never open without an agent (from the tree view or the command palette). I guess it's useful to keep if users manually input their remote authority (or use one from an older version)

I was thinking the same thing, it would be nice to remove it but I guess we need it for existing connections from the recents menu which might lack the agent.  I think we do not need it for the manual input case though, because that would trigger the URI handler which calls `open` which now ensures it gets an agent.

Possibly we could rewrite these entries to add the agent, then we could delete that code, but not sure how hard that would be.

src/commands.ts
```diff
@@ -597,12 +617,6 @@ export class Commands {
 			}
 			folderPath = agent.expanded_directory;
```
code-asher: This would override any folder passed via query param, so I think this will have to be:

```suggestion
      if (!folderPath) {
        folderPath = agent.expanded_directory;
      }
```

src/commands.ts
```diff
@@ -589,6 +594,21 @@ export class Commands {
...
 			workspaceName = workspace.name;
+		} else {
+			workspaceOwner = args[0] as string;
+			workspaceName = args[1] as string;
+			workspaceAgent = args[2] as string | undefined;
+			folderPath = args[3] as string | undefined;
+			openRecent = args[4] as boolean | undefined;
+		}
+
+		if (workspaceAgent === undefined) {
```
code-asher: I think maybe we want `if (!workspaceAgent)` in case someone has `agent=` in the URI (so a blank string).

code-asher:
I just realized you were talking about the quick pick in `remote.ts` this whole time; I thought the first mention was the one in `commands.ts` which is why I was talking about the command palette, my bad :facepalm:

So:

- remote.ts picker: now only used when opening a recent connection
- commands.ts picker: used when handling a URI that lacks an agent or when launching from the command palette

src/commands.ts
EhabY: Fixed

src/commands.ts
EhabY: Fixed

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Update `/openDevContainer` to support all dev container features when hostPath
   and configFile are provided.
+- Fix opening a workspace restores the VS Code session (files, tabs, context)
```
code-asher: My suggestion here might actually be more confusing but maybe we could add a bit more detail like:

> When clicking on a workspace to connect from the sidebar, connect to and use the session (files, tabs, context) from the first agent. Previously, it would have asked for which agent to connect (if there was more than one), but then it would have its own session separate from the one you get when clicking directly on the agent in the sidebar.

code-asher (approved):
:tada:

CHANGELOG.md
code-asher: Oh I guess we also changed the URI handling flow a bit.

> Additionally, when launching from the Coder dashboard, use the same session for the agent after asking for the agent (if there is more than one) rather than using that same separate session.

code-asher: Maybe this is better:

> Consistently use the same session for each agent. Previously, depending on how you connected, it could be possible to get two different sessions for an agent. Existing connections may still have this problem, only new connections are fixed.

EhabY: I like the last one, I'll apply it!

Issue #257, opened 2024-05-07
Include timestamps in Coder extension logs


gtaylor:
To assist with troubleshooting, it'd be helpful to have timestamps in the Coder extension logs.

kylecarbs:
Agreed. @stirby for the v1.0

EhabY:
I can work on this.

Which solutions would you prefer? We can use VS Code `LogOutputChannel` which already outputs the timestamps (not ISO):

<img width="1962" height="112" alt="Image" src="https://github.com/user-attachments/assets/5bfc9f82-d8de-48b5-8cf6-e0977223b8b5" />

This has the benefit of choosing the logging level (Trace, Debug, Info, Warning, Error) and you can then filter those in the `Coder` `Output` channel by severity or text which could be very useful. It also looks better because there's some coloring being applied by VS Code :)

Or perhaps just prepend the timestamp (in whatever format - possibly ISO) so it would look something like this:

<img width="1893" height="114" alt="Image" src="https://github.com/user-attachments/assets/03070aef-e971-49fa-b4b2-81520f4e222c" />

matifali:
@EhabY Thanks for showing interest. I would like the first option given its a standard and allows log level selection.

PR #553, opened 2025-07-15
Use VS Code's LogOutputChannel for logging

EhabY:
#257

In this PR I create a `LogOutputChannel` instead of the old `OutputChannel`. By default, `appendLine` uses the `info` logging level so I used `info` to log most messages, except errors where the `error` level was used.

I think all of the info can be downgraded into `debug` level actually, and the errors could be downgraded into `warn` since they are handled or thrown right away. What do you think?

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -14,7 +14,7 @@ const MAX_URLS = 10;
 
 export class Storage {
 	constructor(
-		private readonly output: vscode.OutputChannel,
+		private readonly output: vscode.LogOutputChannel,
 		private readonly memento: vscode.Memento,
 		private readonly secrets: vscode.SecretStorage,
 		private readonly globalStorageUri: vscode.Uri,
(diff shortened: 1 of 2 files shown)


src/storage.ts
```diff
@@ -508,7 +499,7 @@ export class Storage {
 	}
 
 	public writeToCoderOutputChannel(message: string) {
-		this.output.appendLine(`[${new Date().toISOString()}] ${message}`);
+		this.output.info(message);
```
EhabY: This could be `info` while the rest of internal `info` logging here can be made `debug` so it's easier to filter out messages

code-asher: I think `info` is the right call for all the logging here :+1: 

Long term I think it would be nice to delete `writeToCoderOutputChannel` and call `info` on the output directly, maybe something like `storage.output.info()`.

code-asher (approved):
I had no idea this existed!  Awesome.

code-asher:
Ah and I mentioned `info` already but to this point:

> the errors could be downgraded into warn since they are handled or thrown right away. What do you think?

I think you are right, looking at the `error` calls, `warn` feels appropriate to me for these

src/storage.ts
```diff
@@ -223,20 +220,18 @@ export class Storage {
...
+		this.output.info(`Got status code ${resp.status}`);
 
 		switch (resp.status) {
 			case 200: {
 				const rawContentLength = resp.headers["content-length"];
 				const contentLength = Number.parseInt(rawContentLength);
 				if (Number.isNaN(contentLength)) {
-					this.output.appendLine(
+					this.output.error(
 						`Got invalid or missing content length: ${rawContentLength}`,
```
aslilac:
```suggestion
						"Got invalid or missing content length",
						rawContentLength,
```

The methods on `LogOutputChannel` all take `...args: any[]`, so we can take advantage of that instead of continuing to use string interpolation

src/storage.ts
```diff
@@ -187,9 +184,9 @@ export class Storage {
 		const removed = await cli.rmOld(binPath);
 		removed.forEach(({ fileName, error }) => {
 			if (error) {
-				this.output.appendLine(`Failed to remove ${fileName}: ${error}`);
+				this.output.error(`Failed to remove ${fileName}: ${error}`);
```
aslilac:
```suggestion
				this.output.error(`Failed to remove ${fileName}`, error);
```

aslilac (approved):

src/storage.ts
EhabY: Should I replace all messages that have a dynamic value at the end and remove the trailing `:` from the message? 

The only issue here is that we cannot guarantee that VS Code is going to log those in the same way in future versions. Is it simply a `message + args.join(" ")`?

src/storage.ts
EhabY: I'll look into removing `writeToCoderOutputChannel` and replacing it with `storage.output.X` :pray:

EhabY: Done

src/storage.ts
EhabY: It seems this is how VS Code formats the args, primitives are kept as is, objects are stringified and errors made into human readable errors:

```ts
function format(args: any, verbose: boolean = false): string {
	let result = '';

	for (let i = 0; i < args.length; i++) {
		let a = args[i];

		if (a instanceof Error) {
			a = toErrorMessage(a, verbose);
		}

		if (typeof a === 'object') {
			try {
				a = JSON.stringify(a);
			} catch (e) { }
		}

		result += (i > 0 ? ' ' : '') + a;
	}

	return result;
}
```

src/storage.ts
EhabY: Done

matifali:
HI @EhabY. Thanks for your contribution. Could you also add the change to https://github.com/coder/vscode-coder/blob/main/CHANGELOG.md
I am ssuming that the usuall way to do it @code-asher

code-asher:
Ah right yeah, adding a line to the changelog in the PR is the best way yeah.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Update `/openDevContainer` to support all dev container features when hostPath
   and configFile are provided.
+- Use `LogOutputChannel` for the Coder output channel and ensure messages are logged
+  with the appropriate log level.
```
code-asher: I would word it to be more user-centric, like how does this affect the user and what can/should they do about it?

Something like this maybe:

```
Respect log levels.  In the Coder output panel, you can now select the log level to filter messages.
```

src/storage.ts
aslilac: that's really seems like the "right" way to do it, I'm pretty convinced they won't change it majorly

EhabY: Yeah that makes sense

CHANGELOG.md
EhabY: Added the following: 

> Coder output panel enhancements: All log entries now include timestamps, and you can filter messages by log level in the panel.

code-asher: Wonderful!

Issue #191, opened 2024-01-20
move agent metadata from sidebar to bottom status bar


matifali:
Given the scarcity of space on the status bar, we can only show the first entry by default to use space, with the rest expandable when clicked efficiently. Originally requested in coder/coder#11720

matifali:
Something that shows up on hover. It would be nice if we could color-code it based on the value.

For example, the current values for CPU and RAM are X/32 GB, etc. It would be difficult to paste these values so precisely if we didn't do it.

<img width="344" height="177" alt="Image" src="https://github.com/user-attachments/assets/74758953-4d87-4fe9-b707-ca3654b7b874" />

EhabY:
Could be something like this, from what I see we cannot really extract the usage percentage unless we use Regex with something generic like `<number>/<number>`. If so, what percentage should a warning (yellow) be used and for what percentage should an error (red) be used?

Here is how it would look without coloring: 

<img width="691" height="558" alt="Image" src="https://github.com/user-attachments/assets/909e3bff-307d-4125-b1a0-9d9936657238" />

matifali:
Looks clean.

EhabY:
I can submit a PR then :)

* I am guessing no colors for now since it might produce false negatives?
* Should the label of the status bar item should just be the value of the first item as depicted?

matifali:
> I am guessing no colors for now since it might produce false negatives?

Yes I agree.

> Should the label of the status bar item should just be the value of the first item as depicted?

Sounds fine to me.

PR #555, opened 2025-07-16
Add agent metadata statusbar to monitor resource usage

EhabY:
#191

Looks like the following:

image

The statusbar creation happens in `remote.ts#setup` which AFAIK seems to be called only once per session/extension activation. The creation of this statusbar happens at the very end when everything goes smoothly.


src/workspacesProvider.ts
```diff
@@ -42,7 +37,8 @@ export class WorkspaceProvider
 {
 	// Undefined if we have never fetched workspaces before.
 	private workspaces: WorkspaceTreeItem[] | undefined;
-	private agentWatchers: Record<WorkspaceAgent["id"], AgentWatcher> = {};
+	private agentWatchers: Record<WorkspaceAgent["id"], AgentMetadataWatcher> =
+		{};
```
code-asher, line 41: Did the formatter add this newline?  It looks kinda wacky :smile:

src/remote.ts
```diff
@@ -974,6 +982,50 @@ export class Remote {
 		return loop();
 	}
 
+	/**
+	 * Creates and manages a status bar item that displays metadata information for a given workspace agent.
+	 * The status bar item updates dynamically based on changes to the agent's metadata,
+	 * and hides itself if no metadata is available or an error occurs.
+	 */
+	private createAgentMetadataStatusBar(
```
code-asher, line 984: Thinking out loud, not sure if this is a good move or not, but would it make sense to make this part of `WorkspaceMonitor`?  Such that `new WorkspaceMonitor()` would monitor both the workspace and the agent metadata and update the status bar with both pieces of info.

But on the other hand, I could also see an argument that monitoring the workspace and monitoring the agent are really two different things and should not be colocated.

src/remote.ts
```diff
@@ -974,6 +982,50 @@ export class Remote {
...
+		agent: WorkspaceAgent,
+		restClient: Api,
+		disposables: vscode.Disposable[],
```
code-asher: I know it is chill in JavaScript to pass an array in and modify it like this, but I wonder if we should return an array of disposables instead?  Only because that matches the style of the other disposables we have.

src/remote.ts
```diff
@@ -974,6 +982,50 @@ export class Remote {
...
+		const agentWatcher = createAgentMetadataWatcher(agent.id, restClient);
+		disposables.push(agentWatcher);
+
+		agentWatcher.onChange(
+			() => {
+				if (agentWatcher.error) {
+					this.storage.writeToCoderOutputChannel(
+						formatMetadataError(agentWatcher.error),
+					);
+					statusBarItem.hide();
```
code-asher: Would it make sense to show the error here?  Or just a warning icon or something with the error in the tooltip maybe?

src/workspacesProvider.ts
EhabY: Yes it just barely exceeds the limit :sob:

src/remote.ts
EhabY: I had thought of that but indeed I thought this is not really a `Workspace` item, it's purely an `Agent` item

src/remote.ts
EhabY: Ah fair enough, made it return a disposable array

src/remote.ts
EhabY: Hmmm tbh I am not entirely sure how frequent an error like this can happen.

If we show an error, it could be annoying if it keeps popping up frequently. Perhaps adding an error icon with some label and the tooltip can include the error message is good (while still logging of course). Though I don't love having a "useless" status bar item since the status bar is already crowded :thinking: 

Which is why I favor the current approach, what do you think?

EhabY:
This is just using the logo SVG:

image

It seems a bit big tbh, maybe the `C` alone or the dash can be smaller?

EhabY:
> I prefer the no logo, or if we can make the logo smaller, such as 70% of its original size.

Agreed, the logo is a bit too big, here's how it looks if the logo is scaled down by 70%:

image

Ideally there should be a more "square" form of the Coder logo that would fit the VS Code environment more

matifali:
@rachelmullenax can we have a square logo for use cases like this?

src/workspacesProvider.ts
code-asher: formatters lol

src/remote.ts
code-asher: yeah fair enough

src/remote.ts
code-asher: My thinking is that if we think it is important enough to show metadata in the status bar, it is important enough to show errors getting that metadata.

Definitely agree there should be no popups or anything like that, that would be super annoying.

Thinking about it from the original request, the user wanted to be able to see disk space in the status bar.  Suppose one day there was an error and we could not fetch the metadata, likely the user wants to be aware of that (I am projecting of course lol)

code-asher:
I wonder if some generic icon indicating a remote resource would work better than a logo?  Like a cloud icon or something.

EhabY:
> I wonder if some generic icon indicating a remote resource would work better than the logo? Like a cloud icon or something.

I think this could work well since the VS Code codions fit well within the ecosystem.

image

image

image

The only worry here is that just having an icon won't give enough of an indication that this is `Coder's` stats but could be fine since other extensions use those so it's consistent :thinking:

src/remote.ts
EhabY: Hmmm yeah I see your point now. In that case, I think I can show the something like a warning logo with a label then when hovering we show the error details (the same one that is logged). We can even change the background color so it's yellow-ish for example.

So my proposal is something like 

```ts
statusBarItem.text = "$(warning) Agent Status Unavailable";
statusBarItem.tooltip = errMessage;
statusBarItem.color = new vscode.ThemeColor(
	"statusBarItem.warningForeground",
);
statusBarItem.backgroundColor = new vscode.ThemeColor(
	"statusBarItem.warningBackground",
);
```

code-asher: I like it!

code-asher:
> The only worry here is that just having an icon won't give enough of an indication that this is Coder's stats but could be fine since other extensions use those so it's consistent

True!  But yeah I like the idea of consistent-looking icons in the status bar.  I think mostly the logo being so thick and bold compared to the other icons is what throws me off.  But I will defer to what you and @matifali think.

EhabY:
Alright, I'll keep the 70% one but I did have to write a `fantasticon` script to generate the font from the logo. Should I push everything there and possibly make it part of the script? Or just push the font that contains the logo?. (There is no way to use custom logos in the statusbar without adding a custom font...)

The former will make sure we can easily add more icons but requires some discussion since it should be added to the build workflow, while the latter is just a single `woff2` file :thinking:

src/remote.ts
EhabY: Ah I almost forgot, we still hide the status bar if there is no metadata. Do we want that still?

code-asher: Oh good point hmmm...I wonder if it would make sense to show "no metadata" and in the popup explain that if you want metadata you have to add it to your workspace template?

But on the other hand, maybe if the template has no metadata it is because the user explicitly did not want to see metadata, and this could be annoying to them.

Honestly not sure, maybe the best move is to add a toggle to hide/disable the metadata?

EhabY:
This is how it looks like when it is failing (but with an error message):

image

When it works it looks like this:

image

If there are no errors but the metadata is empty, then we simply hide the status bar item.

src/remote.ts
EhabY: You mean always show the status bar item, and if there are no data then we simply say "No metadata available" and the user can toggle the status bar item in a setting :thinking: ?

Yeah not sure as well... For reference, in the tree view, if there is no metadata then nothing is shown to the user about that

code-asher: Yup that was what I was thinking.

I think for now we just hide it, like you say we hide it in the sidebar as well.  It could be nice to show a message for discoverability reasons; users may not even realize they can have metadata to show.  But, I think it would make sense to revisit that together with the sidebar at a later point.

code-asher:
Looks awesome thank you!

matifali:
If there is a template with no `metadata` block in `coder_agent`, how do we handle it? Do we hide it completely or show an error? If it's the latter, then we should probably look into hiding it from the status bar.

EhabY:
> If there is a template with no `metadata` block in `coder_agent`, how do we handle it? Do we hide it completely or show an error? If it's the latter, then we should probably look into hiding it from the status bar.

Currently if the `metadata` is undefined (or an empty list), then nothing is shown in the tree view nor the status bar item. See this thread: https://github.com/coder/vscode-coder/pull/555#discussion_r2229746489

PR #558, opened 2025-07-23
Add binary signature verification

code-asher:
I extracted the download function (since I needed to reuse it to download signatures) to a separate commit so it is easier to review the signature additions separately, if that is of interest.

This downloads the detached signature from Coder if available or releases.coder.com if not, then verifies the binary using that detached signature and the bundled public key.  The check is performed only when the binary is first download.

- [x] Add option to disable
- [x] Skip headers when requesting releases.coder.com
- [x] Tests (for signing, no tests for the UI changes, ideally we get something set up to do that at some point, although we could mock the VS Code APIs in the meantime...)


pgp-public.key
```diff
@@ -0,0 +1,99 @@
...
+hvI3eC7DBC8sXjJUxsrgFAkxQxSx/njP8f4HdUwhNnB1YA2/5IY5bk8QrXxzrAK1
+sbIAjpJdtPYOrZByyyj4ZpRcSm3ngV2n8yd1muJ5u+oRIQoGCdEIaweCj598jNFa
+k378ZA11hCyNFHjpPIKnF3tfsQ8vjDatoq4Asy+HXFuo1GA/lvNgNb3Nv4FUozuv
+JYJ0KaW73FZXlFBIBkMkRQE8TspHy2v/IGyNXBwKncmkszaiiozBd+T+1NUZgtk5
+9o5uKQwLHVnHIU7r/w/oN5LvLawLg2dP/f2u/KoQXMxjwLZncSH4+5tRz4oa/GMn
+k4F84AxTIjGfLJeXigyP6xIPQbvJy+8iLRaCpj+v/EPwAedbRV+u0JFeqqikca70
+aGN86JBOmwpU87sfFxLI7HdI02DkvlxYYK3vYlA6zEyWaeLZ3VNr6tHcQmOnFe8Q
+26gcS0AQcxQZrcWTCZ8DJYF+RnXjSVRmHV/3YDts4JyMKcD6QX8s/3aaldk=
+=dLmT
+-----END PGP PUBLIC KEY BLOCK-----
```
aslilac:
```suggestion
-----END PGP PUBLIC KEY BLOCK-----

```

all of the test files have trailing new lines but this one does not

src/pgp.ts
```diff
@@ -0,0 +1,92 @@
...
+	) {
+		super(message);
+	}
+
+	summary(): string {
+		switch (this.code) {
+			case VerificationErrorCode.Invalid:
+				return "Signature does not match";
+			default:
+				return "Failed to read signature";
```
aslilac, line 30:
```suggestion
				return "Failed to verify signature";
```

since this is the default, it might not necessarily be about reading I guess?

src/storage.ts
```diff
@@ -202,122 +204,38 @@ export class Storage {
...
-				"Accept-Encoding": "gzip",
-				"If-None-Match": `"${etag}"`,
-			},
-			decompress: true,
-			// Ignore all errors so we can catch a 404!
-			validateStatus: () => true,
+		// Download the binary to a temporary file.
+		await fs.mkdir(path.dirname(binPath), { recursive: true });
+		const tempFile =
+			binPath + ".temp-" + Math.random().toString(36).substring(8);
```
aslilac, line 210: this is such a silly way to get a couple random characters lol. I love it.

src/storage.ts
```diff
@@ -402,6 +320,236 @@ export class Storage {
...
+	 */
+	private async download(
+		client: AxiosInstance,
+		source: string,
+		writeStream: WriteStream,
+		headers?: AxiosRequestConfig["headers"],
+	): Promise<number> {
+		const baseUrl = client.defaults.baseURL;
+
+		const controller = new AbortController();
```
aslilac, line 335: I've known about this api for years and literally never seen someone use it in production. 🥹 so proud!!

src/pgp.ts
```diff
@@ -0,0 +1,92 @@
...
+			const error = `Unable to verify the authenticity of the binary: ${errToStr(e)}. The binary may have been tampered with.`;
+			logger?.warn(error);
+			return new VerificationError(VerificationErrorCode.Invalid, error);
+		}
+	} catch (e) {
+		const error = `Failed to read signature or binary: ${errToStr(e)}.`;
+		logger?.warn(error);
+		return new VerificationError(VerificationErrorCode.Read, error);
+	}
+	return true;
```
aslilac: I feel weird about returning errors in js. if these were `throw`n instead we could do something like...

```ts
try {
  await verifySignature(...);
  // more happy path here
} catch (error) {
  if (error instanceof VerificationError) {
    // unhappy path here
  }
}
```

which is maybe a bit noisier than what you're currently doing, but is also a bit more "when in rome", y'know?

aslilac (approved):
I'm gonna need you to sign a waiver indicating that you know I'm not a cryptography expert and will not be held responsible for any glaring security issues before you merge this 😝 but it looks good!

pgp-public.key
code-asher: This one is the real public key!  I figured I would leave it as it came to me :smile:

src/pgp.ts
code-asher: Ah true, there are only two error codes and the other is `read` so as it currently is this is accurate, but if in the future a code was added and someone did not update the switch statement, we could possibly return the wrong message.  I think I can make use of `never` instead in some way to force the switch to be exhaustive, will fix

src/storage.ts
code-asher: Ahaha I wish I could take credit but this was written by none other than Kyle himself (three years ago!  :open_mouth:)

pgp-public.key
aslilac: but I'm picky 😭 I swear, I'm gonna make a code formatter that just makes text files end with a new line

src/storage.ts
code-asher: lol I cannot lay claim to the silliness sadly, I think this came from Joe's era

I like it too though!

src/pgp.ts
code-asher: You are probably right, initially I did it that way but the added nesting felt like too much, although I did break things up after that so it would not be quite so bad now.  **Edit**: also the type checking is unfortunate, we have to add code to check this is a verification error but it can never be anything but that, actually I think this was the real reason.

Do you think I can get away with it if I call it `VerificationResult` instead lol

code-asher: Ehhhh nvm on second thought I will throw

pgp-public.key
code-asher: hahahahaha fair OK I will add a newline, I am just being paranoid about editing it lol

src/pgp.ts
code-asher: Waaaaiiit looks like TypeScript is smart about this now??  No need to do anything special.  When did that happen lol

So I can just replace `default` with `VerificationErrorCode.Read` and all is well.

Issue #330, opened 2024-07-30
Add search filter to sidebar


code-asher:
If you have many workspaces, it can be difficult to find the one you want.

One workaround is to run the `Coder: Open Workspace` (`coder.open`) command which should pop up something that can be searched.

yelnatscoding:
I've observed a delay in how new workspaces appear in the "All Workspaces" section compared to "My Workspaces." The refresh functionality in this PR addresses this by allowing the extension to promptly check for newly added workspaces.

matifali:
Can the search bar be moved to the sidebar too?

yelnatscoding:
Regarding a search bar or custom input fields, my research indicates the TreeView API is limited. It's not possible to embed or position such elements directly within the TreeView or immediately above it using native components, as the API is designed solely for hierarchical data display, not custom UI.

Consequently, the only viable approach is a Webview View.

This offers full HTML, CSS, and JavaScript control, custom event handling, and dynamic content integration. However, this signifies a fundamental shift: TreeViews rely on a TreeDataProvider for VS Code to render, while a Webview View requires manually building and rendering the entire UI (including all interactions) within its sandboxed environment.

Therefore, implementing this would necessitate a major overhaul of the current TreeView usage in the codebase.

yelnatscoding:
Once this PR is merged, you'll see the search functionality is quite intuitive: clicking the search icon directly focuses the search bar at the top of VS Code.

aslilac:
you can create the pr whenever you're ready

PR #562, opened 2025-07-29
feat: Add search filter to sidebar

yelnatscoding:
See - https://github.com/coder/vscode-coder/issues/330

Features added:
- Added these buttons "All workspaces" section (Clear, Refresh, and Search) See Image below - 
Image

- Clicking on 'Search icon' (Magnifying glass) toggles the Global search
   - VS Codes file tree system does not allow creating Input fields directly within the "All workspaces" section
Image

- Search uses a fuzzy search/substring method users can search via name, owner, template, or status of their workspaces
   - it prioritizes exact word matches first
   - fallback - substring search

## Before search

Image

## After search term typed

Image


src/workspacesProvider.ts
```diff
@@ -300,7 +310,90 @@ export class WorkspaceProvider
...
+
+		// Check for exact word matches (higher priority)
+		const hasExactWordMatch =
+			searchWords.length > 0 &&
+			searchWords.some((word) => {
+				// Escape special regex characters to prevent injection
+				const escapedWord = word.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+				const wordBoundaryRegex = new RegExp(`\\b${escapedWord}\\b`, "i");
+				return wordBoundaryRegex.test(allText);
+			});
```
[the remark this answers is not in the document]
yelnatscoding: Thanks for the feedback! I'll implement the regex pre-compilation optimization.

src/workspacesProvider.ts
```diff
@@ -300,7 +310,90 @@ export class WorkspaceProvider
...
+			templateName,
+			status,
+			...agentNames,
+		].join(" ");
+
+		// Check for exact word matches (higher priority)
+		const hasExactWordMatch =
+			searchWords.length > 0 &&
+			searchWords.some((word) => {
+				// Escape special regex characters to prevent injection
```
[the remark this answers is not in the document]
yelnatscoding: Regarding the DoS concern - since this runs locally in VS Code, I'm not sure if we need to worry about malicious attacks, but I can add a reasonable length limit (e.g., 200 chars) to prevent accidental performance issues from very long search terms. (or if the user errantly pastes a very long string into the input). Does that sound reasonable?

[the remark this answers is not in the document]
yelnatscoding: Added a 200 char limit in this commit - https://github.com/coder/vscode-coder/pull/562/commits/3e340bfef4f5d22b1431ede606fd625180304f86

src/workspacesProvider.ts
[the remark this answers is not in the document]
yelnatscoding: addressed in this commit - https://github.com/coder/vscode-coder/pull/562/commits/3e340bfef4f5d22b1431ede606fd625180304f86

src/workspacesProvider.ts
```diff
@@ -300,7 +310,90 @@ export class WorkspaceProvider
...
+		const hasMatchingAgent = agentNames.some((agentName) =>
+			agentName.includes(searchTerm),
+		);
+
+		// Check if any agent metadata contains the search term
+		const hasMatchingMetadata = agents.some((agent) => {
+			const watcher = this.agentWatchers[agent.id];
+			if (watcher?.metadata) {
+				return watcher.metadata.some((metadata) => {
+					const metadataStr = JSON.stringify(metadata).toLowerCase();
```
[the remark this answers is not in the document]
yelnatscoding: addressed in this commit - https://github.com/coder/vscode-coder/pull/562/commits/3e340bfef4f5d22b1431ede606fd625180304f86

src/workspacesProvider.ts
```diff
@@ -300,7 +310,90 @@ export class WorkspaceProvider
...
+			searchWords.length > 0 &&
+			searchWords.some((word) => {
+				// Escape special regex characters to prevent injection
+				const escapedWord = word.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+				const wordBoundaryRegex = new RegExp(`\\b${escapedWord}\\b`, "i");
+				return wordBoundaryRegex.test(allText);
+			});
+
+		// Check for substring matches (lower priority) - only if no exact word match
+		const hasSubstringMatch =
```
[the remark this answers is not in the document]
yelnatscoding: Addressed this in commit - https://github.com/coder/vscode-coder/pull/562/commits/3e340bfef4f5d22b1431ede606fd625180304f86

CHANGELOG.md
```diff
@@ -24,6 +24,12 @@
 - Add binary signature verification. This can be disabled with
   `coder.disableSignatureVerification` if you purposefully run a binary that is
   not signed by Coder (for example a binary you built yourself).
+- Add search functionality to the "All Workspaces" view with performance optimizations.
+  - Implement smart search that prioritizes exact word matches over substring matches.
+  - Add debounced search input (150ms delay) to improve responsiveness during typing.
+  - Pre-compile regex patterns and cache metadata strings for better performance.
+  - Include comprehensive error handling for malformed data and edge cases.
+  - Add input validation to prevent performance issues from long search terms.
```
aslilac: CHANGELOGs are primarily user-facing documentation. They're don't really care about the details of how the input is debounced, etc. A terse summary of the user impact is sufficient.

```suggestion
- Add search functionality to the "All Workspaces" view.
```

src/workspacesProvider.ts
```diff
@@ -52,6 +55,43 @@ export class WorkspaceProvider
 		// No initialization.
 	}
 
+	setSearchFilter(filter: string) {
+		// Validate search term length to prevent performance issues
+		if (filter.length > 200) {
+			filter = filter.substring(0, 200);
+		}
```
aslilac, line 62: what performance issue? why 200 characters?

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+				const watcher = this.agentWatchers[agent.id];
+				if (watcher?.metadata) {
+					watcher.metadata.forEach((metadata) => {
+						try {
+							metadataStrings.push(JSON.stringify(metadata).toLowerCase());
+						} catch (error) {
+							// Handle JSON serialization errors gracefully
+							this.storage.output.warn(
+								`Failed to serialize metadata for agent ${agent.id}: ${error}`,
+							);
```
aslilac, line 415: but now the metadata still gets cached even with a missing entry, right? this seems like a cache corruption bug

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+		const status = (
+			workspace.workspace.latest_build?.status || ""
+		).toLowerCase();
+
+		// Extract agent names with null safety
+		const agents = extractAgents(
+			workspace.workspace.latest_build?.resources || [],
+		);
+		const agentNames = agents
+			.map((agent) => (agent.name || "").toLowerCase())
```
aslilac:
```suggestion
			.map((agent) => agent.name.toLowerCase())
```

`agent.name` is `string`, this does nothing

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+	private extractSearchableFields(workspace: WorkspaceTreeItem): {
+		workspaceName: string;
+		ownerName: string;
+		templateName: string;
+		status: string;
+		agentNames: string[];
+		agentMetadataText: string;
+	} {
+		// Handle null/undefined workspace data safely
+		const workspaceName = (workspace.workspace.name || "").toLowerCase();
```
aslilac: same here

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+		const ownerName = (workspace.workspace.owner_name || "").toLowerCase();
```
aslilac: same here

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+		const templateName = (
+			workspace.workspace.template_display_name ||
+			workspace.workspace.template_name ||
+			""
```
aslilac: this first `||` does actually do something, because `template_display_name` can be empty, and this will fall back to `template_name`, but if both are empty then it'll be `""` anyway so the second `||` is redundant

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+		).toLowerCase();
+		const status = (
+			workspace.workspace.latest_build?.status || ""
```
aslilac: I'm not sure why there's a `?.` on `latest_build`, it isn't optional

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+		).toLowerCase();
+
+		// Extract agent names with null safety
+		const agents = extractAgents(
+			workspace.workspace.latest_build?.resources || [],
```
aslilac: same here, and the `|| []` does nothing because `resources` is also not optional

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+				// Escape special regex characters to prevent injection
+				const escapedWord = word.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+				regexPatterns.push(new RegExp(`\\b${escapedWord}\\b`, "i"));
+			} catch (error) {
+				// Handle invalid regex patterns
+				this.storage.output.warn(
+					`Invalid regex pattern for search word "${word}": ${error}`,
+				);
+				// Fall back to simple substring matching for this word
+				continue;
```
aslilac: `continue` at the end of a loop doesn't really do anything

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+			...fields.agentNames,
+			fields.agentMetadataText,
+		].join(" ");
+
+		// Check for exact word matches (higher priority)
+		const hasExactWordMatch =
+			regexPatterns.length > 0 &&
+			regexPatterns.some((pattern) => {
+				try {
+					return pattern.test(allText);
```
aslilac: does `test` ever throw an error? it's only documented as returning a `boolean` on MDN and I can't even get it to throw a `TypeError`; it just coerces the value to bool.

src/commands.ts
```diff
@@ -684,6 +684,46 @@ export class Commands {
...
+
+		// Get current search filter to show in the input
+		const currentFilter = (await vscode.commands.executeCommand(
+			"coder.getWorkspaceSearchFilter",
+		)) as string;
+		if (currentFilter) {
+			quickPick.value = currentFilter;
+		}
+
+		quickPick.ignoreFocusOut = true; // Keep open when clicking elsewhere
```
aslilac, line 706: this seems like an odd thing to set. can you explain your rationale?

package.json
```diff
@@ -244,6 +256,21 @@
...
+				{
+					"command": "coder.searchWorkspaces",
+					"when": "coder.authenticated && coder.isOwner && view == allWorkspaces",
+					"group": "navigation"
+				},
+				{
+					"command": "coder.refreshWorkspaces",
+					"when": "coder.authenticated && coder.isOwner && view == allWorkspaces",
+					"group": "navigation"
+				},
```
aslilac, line 269: feels a little weird to put this in the middle

src/workspacesProvider.ts
```diff
@@ -300,7 +348,178 @@ export class WorkspaceProvider
...
+		const searchWords = searchTerm
+			.split(/\s+/)
+			.filter((word) => word.length > 0);
+
+		const regexPatterns: RegExp[] = [];
+		for (const word of searchWords) {
+			try {
+				// Escape special regex characters to prevent injection
+				const escapedWord = word.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+				regexPatterns.push(new RegExp(`\\b${escapedWord}\\b`, "i"));
```
aslilac: it's a bit odd to me to try escaping these symbols instead of interpreting them as word boundaries

package.json
yelnatscoding: where do you suggest we move it to? or should we remove it entirely?

src/workspacesProvider.ts
yelnatscoding: image

it was mainly implemented due to this comment from Copilot

src/workspacesProvider.ts
yelnatscoding: Yes you are correct here is my understanding of what it does - 

Converts the argument to a string using ToString()
Returns true if the regex matches, false otherwise
Never throws an error - even with invalid inputs

src/workspacesProvider.ts
yelnatscoding: You're totally right! I was trying to solve: when searching "container", it should show Docker workspaces (since Docker is related to containers). So I wanted fuzzy matching rather than strict exact word matching. The regex approach gives us that flexibility - users can search for "container" and it'll match "Docker" workspaces. I can change this to interpret word boundaries.

src/commands.ts
yelnatscoding: Yes, here was my thought process on user behavior - 

When a user clicks outside the Quickpick (like on the workspace list or elsewhere in VS Code), the quickpick automatically closes

1. They type "docker", see the filtered results, but accidentally click outside -> Quickpick closes
2. User clicks on a workspace to see details, but that closes the search input
3. User has to reopen the search every time they want to redefine their search

With the current approach it solves the above scenarios. The context is still preserved in state (private searchFilter) so the filtered results will not change if the input closes - its more so just a means for the user to not have to click on the search button again if they are not satisfied or have not completed their search

src/workspacesProvider.ts
yelnatscoding: Also, i'd like to note: 

- each word in the search term gets compiled into a regex pattern even though we are moving to a word boundary search, we are then testing each word against every workspace
- we also run string operations .toLowerCase(), .split(), and multiple .includes() checks on concatenated text
- We're storing all of these regex patterns in memory

The 200 char limit was just an arbitrary number to limit the length of the search input

We could do the following - 

```
   const searchWords = searchTerm.split(/\s+/).filter(word => word.length > 0);
   if (searchWords.length > 10) {
       // Limit to first 10 words
   }
```

Let me know how we should proceed or what you suggest

src/workspacesProvider.ts
yelnatscoding: Yes, it you're right it never throws an error. That was just a bit of over engineering here. Will be removing this in next commit

src/workspacesProvider.ts
yelnatscoding: fixed this - https://github.com/coder/vscode-coder/pull/562/commits/de2dd41f600e73e1f8aa1bc6a961586451726501. We now only cache the data if all metadata is serialized successfully

src/workspacesProvider.ts
yelnatscoding: removed - https://github.com/coder/vscode-coder/pull/562/commits/de2dd41f600e73e1f8aa1bc6a961586451726501

src/workspacesProvider.ts
yelnatscoding: removed - https://github.com/coder/vscode-coder/pull/562/commits/de2dd41f600e73e1f8aa1bc6a961586451726501

src/workspacesProvider.ts
yelnatscoding: removed - https://github.com/coder/vscode-coder/pull/562/commits/de2dd41f600e73e1f8aa1bc6a961586451726501

src/workspacesProvider.ts
yelnatscoding: removed - https://github.com/coder/vscode-coder/pull/562/commits/de2dd41f600e73e1f8aa1bc6a961586451726501

EhabY:
Closing this in favor of https://github.com/coder/vscode-coder/pull/603 since it uses VS Code's own mechanism

PR #574, opened 2025-08-13
Use truncated version with signature fallback

### LINKED ISSUE #573: Adjust the fallback path
We need to adjust the fallback path so that only `MAJOR.MINOR.PATCH` version is included:
https://github.com/coder/vscode-coder/blob/c26018f96e12f7e7bb3c210fd83aef937307e851/src/storage.ts#L235

Signature downloads from [releases.coder.com](https://releases.coder.com/) fail due to a mismatch between the value returned by the [buildinfo](https://coder.com/docs/reference/api/schemas#codersdkbuildinforesponse) endpoint and the object names in the underlying GCS bucket.

Image

### Example:

Currently the handler would attempt to fetch the signature from:
https://releases.coder.com/coder-cli/2.24.1+f97bd76/{detached_signature} 🔴 

Should be:
https://releases.coder.com/coder-cli/2.24.1/{detached_signature} 🟢

code-asher:
Fixes #573

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -142,7 +142,9 @@ export class Storage {
 		this.output.info("Got server version", buildInfo.version);
 		const parsedVersion = semver.parse(buildInfo.version);
 		if (!parsedVersion) {
-			throw new Error(`Got invalid version from deployment: ${buildInfo.version}`)
+			throw new Error(
+				`Got invalid version from deployment: ${buildInfo.version}`,
+			);
 		}
 
 		// Check if there is an existing binary and whether it looks valid.  If it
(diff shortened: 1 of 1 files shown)


src/storage.ts
```diff
@@ -139,6 +140,10 @@ export class Storage {
 		// and to log for debugging.
 		const buildInfo = await restClient.getBuildInfo();
 		this.output.info("Got server version", buildInfo.version);
+		const parsedVersion = semver.parse(buildInfo.version);
```
jdomeracki-coder, line 143: Do I get it right that it will handle the leading `v`?
https://github.com/npm/node-semver?tab=readme-ov-file#versions

jdomeracki-coder: Nvm, I get it ✔️

code-asher: Yup exactly, I tested with `v2.25.1-devel+92d505c52` (from dev.coder.com).

jdomeracki-coder (approved):

Issue #500, opened 2025-05-07
feat: support global CLI flag configurations


ericpaulsen:
coder's CLI has global flags, some of which impact the nature of an SSH connection, such as `--disable-direct-connections`.

our VS Code extension should support the ability to set abritrary CLI flags.

PR #577, opened 2025-09-02
Add support for CLI global flag configurations

EhabY:
#500 

Add the settings `coder.globalFlags` to add global flags to all `coder` CLI invocations in VS Code. However, since we already have `coder.headerCommand` which sets the `--header-command`, this setting takes precedence over manually entered `--header-command` in the global flags settings. Same thing happens for the `--global-config` though that one is set internally and not through a setting

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -195,7 +195,7 @@ export async function startWorkspaceIfStoppedOrFailed(
 			startArgs.push(...["--reason", "vscode_connection"]);
 		}
 
-		const startProcess = spawn(binPath, startArgs);
+		const startProcess = spawn(binPath, startArgs, { shell: true });
 
 		startProcess.stdout.on("data", (data: Buffer) => {
 			data


src/globalFlags.ts
```diff
@@ -0,0 +1,29 @@
...
+		filteredGlobalFlags = globalFlags.filter(
+			(flag) => !flag.startsWith("--header-command"),
+		);
+	}
+
+	if (globalConfigArgs.length > 0) {
+		filteredGlobalFlags = globalFlags.filter(
+			(flag) => !flag.startsWith("--global-config"),
+		);
+	}
```
EhabY: This is passed internally and not from user input, should the user input take precedence over this?

src/remote.ts
```diff
@@ -758,19 +758,15 @@ export class Remote {
...
 			? `${AuthorityPrefix}.${label}--`
 			: `${AuthorityPrefix}--`;
 
+		const globalConfigs = this.globalConfigs(featureSet, label);
+
 		const proxyCommand = featureSet.wildcardSSH
-			? `${escapeCommandArg(binaryPath)}${headerArgList} --global-config ${escapeCommandArg(
-					path.dirname(this.storage.getSessionTokenPath(label)),
-				)} ssh --stdio --usage-app=vscode --disable-autostart --network-info-dir ${escapeCommandArg(this.storage.getNetworkInfoPath())}${await this.formatLogArg(logDir)} --ssh-host-prefix ${hostPrefix} %h`
-			: `${escapeCommandArg(binaryPath)}${headerArgList} vscodessh --network-info-dir ${escapeCommandArg(
```
EhabY, line 773: Is it deliberate that `vscodessh` does not take `--global-config` while `ssh` does? I tried to keep the same behavior but not sure if it's an oversight

src/commands.ts
```diff
@@ -494,12 +495,16 @@ export class Commands {
...
+						this.storage.getSessionTokenPath(toSafeHost(url)),
+					);
+					const globalFlags = getGlobalFlags(
+						vscode.workspace.getConfiguration(),
+						configDir,
+					);
 					terminal.sendText(
-						`${escape(binary)} ssh --global-config ${escape(
-							path.dirname(this.storage.getSessionTokenPath(toSafeHost(url))),
-						)} ${app.workspace_name}`,
```
EhabY, line 511: `coder.headerCommand` was ignored in this request, might be an oversight

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Add support for CLI global flag configurations.
```
code-asher: Could be good to add a reference to the setting name (`coder.globalFlags`)

package.json
```diff
@@ -119,6 +119,13 @@
 					"markdownDescription": "Disable Coder CLI signature verification, which can be useful if you run an unsigned fork of the binary.",
 					"type": "boolean",
 					"default": false
+				},
+				"coder.globalFlags": {
+					"markdownDescription": "Global flags to pass to every Coder CLI invocation. Enter each flag as a separate array item; values are passed verbatim and in order. Do **not** include the `coder` command itself. See the [CLI reference](https://coder.com/docs/reference/cli) for available global flags.\n\nNote that the `#coder.headerCommand#` setting **takes precedence** and will override any `--header-command` value specified here.",
```
code-asher: Could be good to call this out on the header command setting as well.

Also what do you think about deprecating `coder.headerCommand`?  We could ask them  use `coder.globalFlags` instead in the description.

For example maybe something like:

> An external command that outputs additional HTTP headers added to all requests. The command must output each header as `key=value` on its own line. The following environment variables will be available to the process: `CODER_URL`. Defaults to the value of `CODER_HEADER_COMMAND` if not set.  Equivalent to and deprecated by setting `--header-command` in `#coder.globalFlags#`.  If both are set, this setting takes precedence.

src/globalFlags.ts
code-asher: The user could shoot themselves in the foot because if they set `--global-config` it bypasses all the setup we do for the token and url so I think ignoring any `--global-config` from the user is the right move (we may want to call this out in the setting description as well).

src/commands.ts
code-asher: I think you are right

src/globalFlags.ts
```diff
@@ -0,0 +1,29 @@
...
+	configs: WorkspaceConfiguration,
+	configDir?: string,
+): string[] {
+	const globalFlags = configs.get<string[]>("coder.globalFlags") || [];
+	const headerArgs = getHeaderArgs(configs);
+	const globalConfigArgs = configDir
+		? ["--global-config", escapeCommandArg(configDir)]
+		: [];
+
+	// Precedence of "coder.headerCommand" is higher than "coder.globalConfig" with the "--header-command" flag
```
code-asher: Actually, since `coder` already has precedence in the order of arguments, we could just do something like this I think?  (or spread instead of concat)

```javascript
return (configs.get<string[]>("coder.globalFlags") || [])
  .concat(["--global-config", escapeCommandArg(configDir)]
  .concat(getHeaderArgs)
```

And the stuff at the end would automatically take precedence by the cli.

src/remote.ts
code-asher: Yeah, `vscodessh` is a legacy command and it does not support `--global-config`.  It would be safe to pass the flag, but it would be ignored.

We could set it anyway, just to reduce the number of code paths.  Either way seems reasonable to me.

src/globalFlags.ts
```diff
@@ -0,0 +1,29 @@
+import { WorkspaceConfiguration } from "vscode";
+import { getHeaderArgs } from "./headers";
+import { escapeCommandArg } from "./util";
+
+export function getGlobalFlags(
+	configs: WorkspaceConfiguration,
+	configDir?: string,
```
code-asher: At first I was not sure why this was optional, then realized it was because of `vscodessh`.  Maybe we could drop a comment here saying something like "A config directory is only optional for the vscodessh sub-command, otherwise one should always be provided".

code-asher (commented):
Looking good!

CHANGELOG.md
EhabY: Done

package.json
EhabY: Currently, my code follows this precedence for determining the header command:

1. `coder.headerCommand` setting
2. `CODER_HEADER_COMMAND` environment variable
3. `--header-command` in the `coder.globalFlags`

The challenge with deprecating `coder.headerCommand` is that `headers.ts#getHeaders` actually **executes** the command and extracts the headers when initializing the Coder SDK. This makes the setting valuable since it's explicit about this behavior.

Alternatively, we *could* deprecate the setting and instead read `--header-command` directly from `coder.globalFlags`. This means that even `header.ts` would read from the `coder.globalFlags`. 

If we go down that path, I think the following precedence makes the most sense:

1. `--header-command` in the `coder.globalFlags`
2. `coder.headerCommand` setting
3. `CODER_HEADER_COMMAND` environment variable

This way, global flags take priority, while older configurations continue to work as expected since global flags are empty by default.

src/globalFlags.ts
EhabY: You mentioned above that `vscodessh` possibly ignores unknown flags. If that's the case then I'll make this required. Do you know where the documentation for that legacy script?

src/globalFlags.ts
EhabY: We still have to filter `--global-config` for the reasons you mentioned bas this is neat!

src/globalFlags.ts
EhabY: I'm filtering out any `--global-config` and mentioning that in the description

src/remote.ts
EhabY: Done

package.json
code-asher: Ohhh right right we still need the command outside of passing it to the cli.  I think it is better to have the explicit setting rather than parse it out of `coder.globalFlags`, so no change necessary!

One day, if we end up replacing all the API calls with cli calls, then it might make sense to deprecate.

src/globalFlags.ts
code-asher: That sounds perfect.  There is no documentation really since it was an internal command meant to be used by the plugin and not users but you can see the source code for it here: https://github.com/coder/coder/blob/main/cli/vscodessh.go#L1

When it was made, `vscodessh` was basically a copy of the `ssh` command with some extra things.  Since then we have copied those extra things to the regular `ssh` command so we were able to switch to it instead of using the special `vscodessh` command.

By nature of how the commands are structured, `vscodessh` inherits all global flags like any other command, so it should not throw a parsing error or anything like that to include them, even if they end up being unused.

src/globalFlags.ts
code-asher: If we end up making the config non-optional then we will always include a `--global-config` which will overwrite any the user might have set, which means we could even remove that filter. :open_mouth:

package.json
EhabY: Ah yeah that's valid, there's a ticket for that but requires a big refactor if we down that path.

Also I updated the description of the settings to mention the correct precedence

src/globalFlags.ts
EhabY: Indeed, I removed the filters and relied instead on the order

src/globalFlags.ts
EhabY: Aha, interesting, it does seem to be still used. I guess for old Coder instances

code-asher: Yeah, hopefully everyone has upgraded by now, but who knows.  One of these days maybe we can make a breaking change and remove it.

src/commands.ts
```diff
@@ -503,12 +504,16 @@ export class Commands {
...
+						`${escapeCommandArg(binary)}${globalFlags.length === 0 ? "" : ` ${globalFlags.join(" ")}`} ssh ${app.workspace_name}`,
```
code-asher: Now that we know there is always a `--global-config` flag at least, we could skip the length check.

src/remote.ts
```diff
@@ -828,6 +824,15 @@ export class Remote {
 		return sshConfig.getRaw();
 	}
 
+	private globalConfigs(label: string): string {
+		const vscodeConfig = vscode.workspace.getConfiguration();
+		const args: string[] = getGlobalFlags(
+			vscodeConfig,
+			path.dirname(this.storage.getSessionTokenPath(label)),
+		);
+		return args.length === 0 ? "" : ` ${args.join(" ")}`;
```
code-asher: Similar thought here about always having at least one item so the length check will never matter.

But, maybe being defensive is better, could be we change this in the future, so feel free to leave the checks.  Or if we can somehow guarantee there is at least one item using the types, maybe that would be optimal, not sure if Typescript can do that though.

src/remote.ts
```diff
@@ -828,6 +824,15 @@ export class Remote {
 		return sshConfig.getRaw();
 	}
 
+	private globalConfigs(label: string): string {
+		const vscodeConfig = vscode.workspace.getConfiguration();
+		const args: string[] = getGlobalFlags(
```
code-asher: Is the type annotation necessary?

code-asher (approved):
Looks great!

src/commands.ts
EhabY: Yes, you are right. Removed the check here and elsewhere

src/remote.ts
EhabY: Wouldn't an empty list mean that we just do something like `coder  ssh <other args>`? (notice the double space). 

This is allowed in all shells/terminal AFAIK. So I'll just remove the checks for now

src/remote.ts
EhabY: It is not, removed

src/api.ts
```diff
@@ -195,7 +195,7 @@ export async function startWorkspaceIfStoppedOrFailed(
 			startArgs.push(...["--reason", "vscode_connection"]);
 		}
 
-		const startProcess = spawn(binPath, startArgs);
+		const startProcess = spawn(binPath, startArgs, { shell: true });
```
EhabY, line 198: This would run the given command in the native shell which would mean it's similar to the other ways we invoke the CLI (in 2 other places IIRC). Since I noticed that if a variable is escaped then this breaks because by default `spawn` executes in a cross-platform way without interpreting the args (so `CMD "arg1"` is interpreted literally while if `shell: true` then it's interpreted as the local shell does which is equivalent to `CMD arg1`)

src/remote.ts
code-asher: Yup exactly it would just be a double space, definitely safe.  It would just look a bit unprofessional is all, if someone was looking at the process listing lol

But since it should never happen, and would not break even if it did, I think removing it is chill

src/api.ts
code-asher: I see two `execFile`s but those do not spawn with a shell by default, I think?  Are you referring to somewhere else where we spawn with a shell?

But, this makes sense to me.

EhabY: You are correct for the usages of `execFile`, those do not spawn a shell on *nix systems (but do on Windows). But the only usage there is execute `coder version` which requires no special escaping nor any global configs/headers.

However, for some other usages like `terminal.sendText` and through the `proxyCommand`, they are executed in the native shell and not through NodeJS like this one used to do.

code-asher: > but do on Windows

Even for exe files?

> only usage there is execute coder version which requires no special escaping

Yup yup

> However, for some other usages like terminal.sendText and through the proxyCommand

Oh yeah good point on `terminal.sendText`, forgot about that one.

Great points, everywhere we have quoted variables we are using a shell, makes sense.  I think some people also use environment variables (like in the header command), so we need a shell for expansion as well.

EhabY: > Even for exe files?

Apparently no, but for `bat` and `cmd` files...

> so we need a shell for expansion as well.

Indeed

Issue #542, opened 2025-06-30
log files in the proxy log directory are empty


spikecurtis:
After setting the Proxy Log Directory setting, log files are created in the directory, but they are all empty.

![Image](https://github.com/user-attachments/assets/b2893671-3ab2-4f83-b764-7218ae44d8f8)

Tested on Windows 11, plugin v1.9.2

EhabY:
I tested this on both Windows and Linux, and the logs are being created correctly. The issue is that the CLI isn’t very chatty, so when nothing goes wrong, the logs often end up empty.

@code-asher suggested adding a `-v`/`--verbose` flag automatically when `coder.proxyLogDirectory` is set, since users who enable logging are likely interested in debug logs as well. Thoughts?

spikecurtis:
yeah, we almost certainly want `-v` if the log directory is set. I'm not even sure if you get INFO logs without it set.

EhabY:
Fixed in #578

PR #578, opened 2025-09-04
Enable verbose output when setting a log directory

EhabY:
#542

### DIFF
--- a/src/remote.ts
+++ b/src/remote.ts
@@ -676,7 +676,7 @@ export class Remote {
 			"SSH proxy diagnostics are being written to",
 			logDir,
 		);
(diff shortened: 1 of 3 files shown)


src/remote.ts
```diff
@@ -676,7 +676,7 @@ export class Remote {
 			"SSH proxy diagnostics are being written to",
 			logDir,
 		);
-		return ` --log-dir ${escapeCommandArg(logDir)}`;
+		return ` --log-dir ${escapeCommandArg(logDir)} -v`;
```
EhabY, line 679: This will essentially make `-v` (global flag) always on if we pass `--log-dir`. Related to https://github.com/coder/vscode-coder/issues/500 which allows setting global flags

PR #580, opened 2025-09-09
Add logging on all RESTful and WebSocket requests in VS Code

### LINKED ISSUE #532: Add debug logging around requests

EhabY:
Fixes #532 

This PR introduces several enhancements to improve logging and consistency:

1. **Axios Logging Interceptor**
Added a logging interceptor to the Axios instance during its creation. Each request is tagged with a unique UUID, allowing us to correlate it with its corresponding response and measure the total in-flight time. To avoid excessive noise, response bodies are deliberately excluded from the logs as they were too verbose and difficult to read.

2. **Unified WebSocket Creation**
Consolidated WebSocket creation logic by standardizing on a `OneWayWebSocket`. For VS Code, this uses the `ws` Node library. Previously, there were two separate places creating their own WebSocket connections and another two relying on SSE. These SSE implementations were converted to one-way WebSockets for improved consistency and performance. Additional logging was also added to track WebSocket creation, errors, and closure events.

For now, the logging level is set to `trace` for non-error logs and `error` for errors. This can easily be adjusted later. I have also isolated the logging logic in `logging` so it's all in one place. Let me know if we want to add more details or possibly remove some.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 - Always enable verbose (`-v`) flag when a log directory is configured (`coder.proxyLogDir`).
 - Automatically start a workspace without prompting if it is explicitly opened but not running.
-- Replaced SSE paths with a one-way WebSocket and centralized creation on `OneWayWebSocket`, and unified socket handling.
 
 ### Added
 
--- a/src/logging/httpLogger.ts
+++ b/src/logging/httpLogger.ts
@@ -1,8 +1,4 @@
-import type {
-	AxiosError,
-	AxiosResponse,
-	InternalAxiosRequestConfig,
-} from "axios";
+import type { AxiosError, AxiosResponse } from "axios";
 import { isAxiosError } from "axios";
 import { getErrorMessage } from "coder/site/src/api/errors";
 import { getErrorDetail } from "../error";
@@ -22,6 +18,9 @@ import {
 } from "./types";
 import { createRequestId, shortId } from "./utils";
 
+/**
+ * Creates metadata for tracking HTTP requests.
+ */
 export function createRequestMeta(): RequestMeta {
 	return {
 		requestId: createRequestId(),
@@ -29,35 +28,33 @@ export function createRequestMeta(): RequestMeta {
 	};
 }
 
+/**
+ * Logs an outgoing HTTP RESTful request.
+ */
 export function logRequest(
 	logger: Logger,
-	requestId: string,
-	config: InternalAxiosRequestConfig,
+	config: RequestConfigWithMeta,
 	logLevel: HttpClientLogLevel,
 ): void {
 	if (logLevel === HttpClientLogLevel.NONE) {
 		return;
 	}
 
-	const method = formatMethod(config.method);
-	const url = formatUri(config);
+	const { requestId, method, url } = parseConfig(config);
 	const len = formatContentLength(config.headers, config.data);
 
-	let msg = `→ ${shortId(requestId)} ${method} ${url} ${len}`;
-	if (logLevel >= HttpClientLogLevel.HEADERS) {
-		msg += `\n${formatHeaders(config.headers)}`;
-	}
-
-	if (logLevel >= HttpClientLogLevel.BODY) {
-		msg += `\n${formatBody(config.data)}`;
-	}
-
-	logger.trace(msg);
+	const msg = [
+		`→ ${shortId(requestId)} ${method} ${url} ${len}`,
+		...buildExtraLogs(config.headers, config.data, logLevel),
+	];
+	logger.trace(msg.join("\n"));
 }
 
+/**
+ * Logs an incoming HTTP RESTful response.
+ */
 export function logResponse(
 	logger: Logger,
-	meta: RequestMeta,
 	response: AxiosResponse,
 	logLevel: HttpClientLogLevel,
 ): void {
@@ -65,66 +62,98 @@ export function logResponse(
 		return;
 	}
 
-	const method = formatMethod(response.config.method);
-	const url = formatUri(response.config);
-	const time = formatTime(Date.now() - meta.startedAt);
+	const { requestId, method, url, time } = parseConfig(response.config);
 	const len = formatContentLength(response.headers, response.data);
 
-	let msg = `← ${shortId(meta.requestId)} ${response.status} ${method} ${url} ${len} ${time}`;
-
-	if (logLevel >= HttpClientLogLevel.HEADERS) {
-		msg += `\n${formatHeaders(response.headers)}`;
-	}
-
-	if (logLevel >= HttpClientLogLevel.BODY) {
-		msg += `\n${formatBody(response.data)}`;
-	}
-
-	logger.trace(msg);
+	const msg = [
+		`← ${shortId(requestId)} ${response.status} ${method} ${url} ${len} ${time}`,
+		...buildExtraLogs(response.headers, response.data, logLevel),
+	];
+	logger.trace(msg.join("\n"));
 }
 
-export function logRequestError(
+/**
+ * Logs HTTP RESTful request errors and failures.
+ *
+ * Note: Errors are always logged regardless of log level.
+ */
+export function logError(
 	logger: Logger,
 	error: AxiosError | unknown,
+	logLevel: HttpClientLogLevel,
 ): void {
 	if (isAxiosError(error)) {
 		const config = error.config as RequestConfigWithMeta | undefined;
-		const meta = config?.metadata;
-		const method = formatMethod(config?.method);
-		const url = formatUri(config);
-		const requestId = meta?.requestId || "unknown";
-		const time = meta ? formatTime(Date.now() - meta.startedAt) : "?ms";
+		const { requestId, method, url, time } = parseConfig(config);
 
-		const msg = getErrorMessage(error, "");
+		const errMsg = getErrorMessage(error, "");
 		const detail = getErrorDetail(error) ?? "";
+		const errorParts = [errMsg, detail]
+			.map((part) => part.trim())
+			.filter(Boolean);
 
+		let logPrefix: string;
+		let extraLines: string[];
 		if (error.response) {
-			const msgParts = [
-				`← ${shortId(requestId)} ${error.response.status} ${method} ${url} ${time}`,
-				msg,
-				detail,
-			];
-			if (msg.trim().length === 0 && detail.trim().length === 0) {
-				const responseData =
+			if (errorParts.length === 0) {
+				errorParts.push(
 					error.response.statusText ||
-					String(error.response.data).slice(0, 100) ||
-					"No error info";
-				msgParts.push(responseData);
+						String(error.response.data).slice(0, 100) ||
+						"No error info",
+				);
 			}
 
-			const fullMsg = msgParts.map((str) => str.trim()).join(" - ");
-			const headers = formatHeaders(error.response.headers);
-			logger.error(`${fullMsg}\n${headers}`, error);
+			logPrefix = `← ${shortId(requestId)} ${error.response.status} ${method} ${url} ${time}`;
+			extraLines = buildExtraLogs(
+				error.response.headers,
+				error.response.data,
+				logLevel,
+			);
 		} else {
-			const reason = error.code || error.message || "Network error";
-			const errorInfo = [msg, detail, reason].filter(Boolean).join(" - ");
-			const headers = formatHeaders(config?.headers ?? {});
-			logger.error(
-				`✗ ${shortId(requestId)} ${method} ${url} ${time} - ${errorInfo}\n${headers}`,
-				error,
+			if (errorParts.length === 0) {
+				errorParts.push(error.code || "Network error");
+			}
+			logPrefix = `✗ ${shortId(requestId)} ${method} ${url} ${time}`;
+			extraLines = buildExtraLogs(
+				error?.config?.headers ?? {},
+				error.config?.data,
+				logLevel,
 			);
 		}
+
+		const msg = [[logPrefix, ...errorParts].join(" - "), ...extraLines];
+		logger.error(msg.join("\n"));
 	} else {
 		logger.error("Request error", error);
 	}
+}
+
+function buildExtraLogs(
+	headers: Record<string, unknown>,
+	body: unknown,
+	logLevel: HttpClientLogLevel,
+) {
+	const msg = [];
+	if (logLevel >= HttpClientLogLevel.HEADERS) {
+		msg.push(formatHeaders(headers));
+	}
+	if (logLevel >= HttpClientLogLevel.BODY) {
+		msg.push(formatBody(body));
+	}
+	return msg;
+}
+
+function parseConfig(config: RequestConfigWithMeta | undefined): {
+	requestId: string;
+	method: string;
+	url: string;
+	time: string;
+} {
+	const meta = config?.metadata;
+	return {
+		requestId: meta?.requestId || "unknown",
+		method: formatMethod(config?.method),
+		url: formatUri(config),
+		time: meta ? formatTime(Date.now() - meta.startedAt) : "?ms",
+	};
 }
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -14,7 +14,7 @@ import { getHeaderCommand, getHeaders } from "../headers";
 import {
 	createRequestMeta,
 	logRequest,
-	logRequestError,
+	logError,
 	logResponse,
 } from "../logging/httpLogger";
 import { Logger } from "../logging/logger";
@@ -211,27 +211,24 @@ function addLoggingInterceptors(
 ) {
 	client.interceptors.request.use(
 		(config) => {
-			const meta = createRequestMeta();
-			(config as RequestConfigWithMeta).metadata = meta;
-			logRequest(logger, meta.requestId, config, getLogLevel(configProvider()));
+			const configWithMeta = config as RequestConfigWithMeta;
+			configWithMeta.metadata = createRequestMeta();
+			logRequest(logger, configWithMeta, getLogLevel(configProvider()));
 			return config;
 		},
 		(error: unknown) => {
-			logRequestError(logger, error);
+			logError(logger, error, getLogLevel(configProvider()));
 			return Promise.reject(error);
 		},
 	);
 
 	client.interceptors.response.use(
 		(response) => {
-			const meta = (response.config as RequestConfigWithMeta).metadata;
-			if (meta) {
-				logResponse(logger, meta, response, getLogLevel(configProvider()));
-			}
+			logResponse(logger, response, getLogLevel(configProvider()));
 			return response;
 		},
 		(error: unknown) => {
-			logRequestError(logger, error);
+			logError(logger, error, getLogLevel(configProvider()));
 			return Promise.reject(error);
 		},
 	);


src/api.ts
```diff
@@ -113,57 +116,75 @@ export function makeCoderSdk(
...
+		(response) => {
+			const { requestId, startedAt } =
+				(response.config as RequestConfigWithMetadata).metadata ?? {};
+			const ms = startedAt ? Date.now() - startedAt : undefined;
+
+			logger.trace(
+				`res ${requestId ?? "n/a"}: ${response.status}${
+					ms !== undefined ? ` in ${ms}ms` : ""
+				}`,
+				// { responseBody: response.data }, // TODO too noisy
```
EhabY: Alternatively we can try logging part of the message?

src/websocket/oneWayCodeWebSocket.ts
```diff
@@ -0,0 +1,138 @@
+/**
```
EhabY: This is adapted from the `OneWayWebSocket` but uses `ws` instead of the default `WebSocket` that ships with all browsers/Node applications

src/workspaceMonitor.ts
```diff
@@ -38,19 +38,19 @@ export class WorkspaceMonitor implements vscode.Disposable {
...
+		socket.addEventListener("open", () => {
+			this.storage.output.info(`Monitoring ${this.name}...`);
 		});
 
-		eventSource.addEventListener("data", (event) => {
+		socket.addEventListener("message", (event) => {
 			try {
-				const newWorkspaceData = JSON.parse(event.data) as Workspace;
+				// Perhaps we need to parse this and validate it.
+				const newWorkspaceData = event.parsedMessage!.data as Workspace;
```
EhabY: This is `unknown` type. (Kept it as is for now)

src/extension.ts
```diff
@@ -67,16 +68,25 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 		storage,
 	);
 
+	// TODO this won't get updated when users change their settings; Listen to changes and update this
+	const httpAgent = await createHttpAgent();
```
EhabY: It seems we have an axios interceptor that creates this on every request. I'm not sure that's efficient given that this might do some IO (read 3 files). I'm thinking of refactoring this into a singleton module that includes the proxy, whenever the user modifies their configurations (or possibly also modify one of the read files) then we invalidate the object and create a new one again (possibly also have listeners so we can recreate a new websocket connections when changing configurations). Thoughts about this?

I can create another ticket to refactor this in a follow up

src/api.ts
EhabY: Revised

src/api.ts
```diff
@@ -83,6 +90,9 @@ export function makeCoderSdk(
 		restClient.setSessionToken(token);
 	}
 
+	// Logging interceptor
```
code-asher: Nit: this comment just duplicates the function name, could remove

src/agentMetadataHelper.ts
```diff
@@ -61,6 +53,11 @@ export function createAgentMetadataWatcher(
 		}
 	});
 
+	socket.addEventListener("error", (error) => {
+		watcher.error = error;
+		onChange.fire(null);
+	});
+
 	return watcher;
```
code-asher: Unrelated, but I wonder if we should also handle the `close` event?  Maybe something like `watcher.error = new Error("web socket closed unexpectedly")`.

src/api.ts
```diff
@@ -242,70 +231,44 @@ export async function startWorkspaceIfStoppedOrFailed(
...
+		};
+		socket.addEventListener("close", closeHandler);
+		socket.addEventListener("message", (data) => {
+			const log = data.parsedMessage!;
+			writeEmitter.fire(log.output + "\r\n");
+		});
+		socket.addEventListener("error", (error) => {
+			// Do not want to trigger the close handler.
+			socket.removeEventListener("close", closeHandler);
+			socket.close();
```
code-asher: Does an errored socket need to be closed?  I think an error will automatically destroy the socket.

src/api.ts
```diff
@@ -242,70 +231,44 @@ export async function startWorkspaceIfStoppedOrFailed(
...
-					),
-				);
-			});
-			socket.on("close", () => {
-				resolve();
-			});
-		} catch (error) {
-			// If this errors, it is probably a malformed URL.
-			reject(
+		const rejectError = (error: unknown) => {
```
code-asher: Is `rejectError` meant to be reused at some point?  Curious as to why it was broken out into a function instead of being directly in the error handler, since it is so small.

src/api.ts
```diff
@@ -242,70 +231,44 @@ export async function startWorkspaceIfStoppedOrFailed(
...
+		};
+
+		const socket = webSocketClient.watchBuildLogsByBuildId(
+			workspace.latest_build.id,
+			logs,
+		);
+		const closeHandler = () => {
+			resolve();
+		};
+		socket.addEventListener("close", closeHandler);
```
code-asher: Just a matter of preference so feel free to ignore but we could directly call `socket.addEventListener("close", resolve)` since `closeHandle` has no functionality.

src/api.ts
```diff
@@ -242,70 +231,44 @@ export async function startWorkspaceIfStoppedOrFailed(
...
+		socket.addEventListener("message", (data) => {
+			const log = data.parsedMessage!;
```
code-asher: I would feel safer without the `!`.  This happens when there is a parsing error, right?  So we could do:

```typescript
if (data.parseError) {
  writeEmitter.fire(errToStr(data.parseError, "failed to parse message") + "\r\n");
} else {
  writeEmitter.fire(data.parsedMessage.output + "\r\n");
}
```

src/api.ts
```diff
@@ -242,70 +231,44 @@ export async function startWorkspaceIfStoppedOrFailed(
...
+			writeEmitter.fire(log.output + "\r\n");
+		});
+		socket.addEventListener("error", (error) => {
+			// Do not want to trigger the close handler.
+			socket.removeEventListener("close", closeHandler);
```
code-asher: Is this just for good hygiene or is there a practical reason?  Technically I think it would be fine to let it trigger since you can only reject/resolve a promise once and the resolve would just be ignored.

Not saying we should remove it, but we could perhaps explain more in the comment.

src/extension.ts
code-asher: I think we will want to create it on every request in this PR because otherwise this is technically a breaking change.

We used to have a singleton that watched for setting updates but the problem was that the certificates can change.  Rather than use a file watcher we decided to just read them every time.  I am not sure if the optimization is worth it, hopefully we are not doing *that* many requests.

For now, what do you think about making `CoderWebSocketClient` call `createHttpAgent()` each time it makes a web socket connection?

src/inbox.ts
```diff
@@ -19,68 +16,39 @@ const TEMPLATE_WORKSPACE_OUT_OF_DISK = "f047f6a3-5713-40f7-85aa-0394cce9fa3a";
...
-				const inboxMessage = JSON.parse(
-					data.toString(),
-				) as GetInboxNotificationResponse;
-
-				vscode.window.showInformationMessage(inboxMessage.notification.title);
-			} catch (error) {
-				this.notifyError(error);
-			}
+		this.#socket.addEventListener("message", (data) => {
+			const inboxMessage = data.parsedMessage!;
```
code-asher: Same comment here about the `!`.

src/logging/netLog.ts
```diff
@@ -0,0 +1,176 @@
...
+		return data.byteLength;
+	}
+	if (
+		typeof data === "object" &&
+		"size" in data &&
+		typeof data.size === "number"
+	) {
+		return data.size;
+	}
+	return 0;
```
code-asher: Zero here feels misleading to me, really the size is unknown right?

src/logging/netLog.ts
```diff
@@ -0,0 +1,176 @@
...
+export function logRequestStart(
+	logger: vscode.LogOutputChannel,
+	requestId: string,
+	config: InternalAxiosRequestConfig,
+): void {
+	const method = (config.method ?? "GET").toUpperCase();
+	const url = config.url || "";
+	const len = config.headers?.["content-length"] as string | undefined;
+	const lenStr = len ? ` (${len}b)` : "";
+	logger.trace(`→ ${shortId(requestId)} ${method} ${url}${lenStr}`);
```
code-asher: I like the arrows and formatting!

I know your description mentioned not logging request/response bodies due to verbosity but I do think sometimes seeing the bodies can be helpful.  Maybe we could have a toggle for it?

src/logging/netLog.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+export function logRequestStart(
+	logger: vscode.LogOutputChannel,
+	requestId: string,
+	config: InternalAxiosRequestConfig,
+): void {
+	const method = (config.method ?? "GET").toUpperCase();
+	const url = config.url || "";
+	const len = config.headers?.["content-length"] as string | undefined;
```
code-asher: I think blank strings risks making it look like we are dropping info, what do you think about making them explicit like: `config.url || "<no url>"` and `config.headers?.["content-length"] || "unknown"`?  (And same for the others below.)

src/logging/netLog.ts
```diff
@@ -0,0 +1,176 @@
...
+export function logRequestError(
+	logger: vscode.LogOutputChannel,
+	error: AxiosError | unknown,
+): void {
+	if (isAxiosError(error)) {
+		const config = error.config as RequestConfigWithMeta | undefined;
+		const meta = config?.metadata;
+		const method = (config?.method ?? "GET").toUpperCase();
+		const url = config?.url || "";
+		const requestId = meta?.requestId ?? "unknown";
```
code-asher: Maybe `||` would be better than `??` for these, like if `requestId` is blank for some reason we might still want to show `unknown`.

src/logging/netLog.ts
```diff
@@ -0,0 +1,176 @@
...
+		if (this.byteCount > 0) {
+			stats.push(`${this.byteCount}b`);
+		}
+
+		const codeStr = code ? ` (${code})` : "";
+		const reasonStr = reason ? ` - ${reason}` : "";
+		const statsStr = stats.length > 0 ? ` [${stats.join(", ")}]` : "";
+
+		this.logger.trace(
+			`✗ WS ${shortId(this.id)} closed${codeStr}${reasonStr}${statsStr}`,
```
code-asher: To me the `✗` implies error (as used above and below), but `close` could be intentional.  Maybe the left arrow would work here (like with responses)?

src/logging/netLog.ts
```diff
@@ -0,0 +1,176 @@
...
+): void {
+	if (isAxiosError(error)) {
+		const config = error.config as RequestConfigWithMeta | undefined;
+		const meta = config?.metadata;
+		const method = (config?.method ?? "GET").toUpperCase();
+		const url = config?.url || "";
+		const requestId = meta?.requestId ?? "unknown";
+		const ms = meta ? Date.now() - meta.startedAt : "?";
+
+		if (error.response) {
```
code-asher: There could be an error message and detail in the response we should show.  See `getErrorMessage` and `getErrorDetail` from `coder/site/src/api/errors` (probably easiest to just import them).

src/websocket/oneWayCodeWebSocket.ts
code-asher: Should this be named `OneWayCoderWebSocket`? (missing `r`)  Or is it `Code` as in `VS Code`?

src/websocket/webSocketClient.ts
```diff
@@ -0,0 +1,131 @@
+import { Api } from "coder/site/src/api/api";
```
code-asher: What do you think about merging this with the rest client? That would match how all the web socket and regular API request methods are together in `coder/coder`.  Then we can pass around just the one rest client and do `restClient.watchWorkspace()`, etc (we would have to extend `Api` of course, overriding the related functions).

src/websocket/webSocketClient.ts
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * WebSocket client for Coder API connections.
+ *
+ * Automatically configures logging for WebSocket events:
+ * - Connection attempts and successful opens at the trace level
+ * - Connection closes at the trace level
+ * - Errors at the error level
+ */
+export class CoderWebSocketClient {
```
code-asher: Really nice how the web socket client is abstracting things, so much better than what we had. :+1: 

Just to make sure I am on the right page, these (or at least some of these) already exist in `coder/coder`, right?  But we have to duplicate them because they use browser web socket right?

At some point we may want to modify `coder/coder`'s `OneWayWebSocket` so we can pass in a web socket constructor and use it externally as well.

src/workspaceMonitor.ts
code-asher: Same thoughts here about `!`.

src/websocket/webSocketClient.ts
```diff
@@ -0,0 +1,131 @@
...
+		webSocket.addEventListener("message", (event) => {
+			wsLogger.logMessage(event.sourceEvent.data);
+		});
+
+		webSocket.addEventListener("close", (event) => {
+			wsLogger.logClose(event.code, event.reason);
+		});
+
+		webSocket.addEventListener("error", (event) => {
+			wsLogger.logError(event?.error ?? event);
```
code-asher: Could `event.error` ever be blank?  Wonder if we should prefer `||`.

code-asher (commented):
I have not had a chance to run it yet (hopefully tomorrow) but gave it a first pass and it is looking fantastic!

src/workspaceMonitor.ts
EhabY: Ah I think I misunderstood the API a bit, it's either that `parsedMessage` or `parseError` is defined. I'll add proper error handling to check if it's one or the other!

src/websocket/oneWayCodeWebSocket.ts
EhabY: Yes `Code` as in `VS Code`. I get the confusion though :sweat:

src/api.ts
EhabY: Removed

src/api.ts
EhabY: I think I used it twice before but you are right. Just inlined it now

src/api.ts
EhabY: Technically this won't work here if we want to detach the close handler when an error is encountered (so we close without resolving THEN reject the enclosing promise).

src/api.ts
EhabY: Indeed much cleaner, got confused with the API!

src/api.ts
EhabY: If we do not remove the close handler then we trigger the enclosing `resolve`, thinking that we completed normally before we can reject. The nature of async process in JS means we have to remove the close listener before we trigger it lest we get unexpected behavior.

Changed to "Do not want to trigger the close handler and resolve the promise normally."

src/api.ts
EhabY: [The answer here](https://stackoverflow.com/questions/38181156/websockets-is-an-error-event-always-followed-by-a-close-event) quotes the RFC6455, but in essence it is likely that it will be closed but we cannot be sure about this :/

src/inbox.ts
EhabY: Fixed and added error handling

src/logging/netLog.ts
EhabY: Indeed yes, this is the part of the websocket logging metadata that we can collect. I tested it out a bit and it seems to record good enough info.

Should this return `undefined` and we handle it in the caller (in this case it might just be ignored anyway)

src/logging/netLog.ts
EhabY: I think that makes sense, I'll test this out and see how it looks :)!

src/logging/netLog.ts
EhabY: Yeah fair enough, changed!

src/logging/netLog.ts
EhabY: Ah thanks, I've used something similar to how `extension.ts` logs request error messages and details. Though the logger here is more verbose!

src/logging/netLog.ts
EhabY: I suppose the close event can come from either the server or client, so a direction won't make sense here :thinking: 

I changed it to `▣`

src/websocket/webSocketClient.ts
EhabY: Do you mean merging it in the other repo or here by overriding the type?

If you mean here then yes, that seems reasonable and would make it one less object to carry around. We can even apply the same `createHttpAgent` logic there (create it for every connection)

src/websocket/webSocketClient.ts
EhabY: > Just to make sure I am on the right page, these (or at least some of these) already exist in coder/coder, right? But we have to duplicate them because they use browser web socket right?

Yes indeed, it's two things:

1. VS Code uses `ws` (npm library) as compared to `WebSocket` (native)
2. In the `coder/coder` it always assumes the client is a browser and so it uses stuff like `window.location`

Unifying everything here also makes it easy to port this into a singular WebSocket implementation once we fix the two problems I mentioned above (and make use of the `baseUrl` in WebSockets).

src/websocket/webSocketClient.ts
EhabY: I think `event.error` should be an object. I'll pass `event.message` and `event.error` actually though. Since we seem to already have the message

src/extension.ts
EhabY: > Rather than use a file watcher we decided to just read them every time

Indeed that was my suggestion but wanted to ask if there's a reason it's not implemented here. I think this can be implemented in the future and could prove to be better for users with certificate files. I'll ignore for now though :pray: 

> For now, what do you think about making `CoderWebSocketClient` call `createHttpAgent()` each time it makes a web socket connection?

That makes perfect sense, I didn't do it because `createHttpAgent()` is async while the websocket creation is sync (it's used within constructors). I could change some `readFile` into `readFileSync` so it doesn't break a lot of things :thinking:

src/agentMetadataHelper.ts
EhabY: Wouldn't this cause an error when we also close from the client (by disposing the watcher)?

src/extension.ts
EhabY: Done, now `createHttpAgent` is only used in requests and websocket creations (created every time)

src/api.ts
EhabY: Actually it might work but doing something like `socket.removeEventHandler("close", resolve)` seems strange to me and might introduce bugs if we modify the `close` behavior later on

src/logging/netLog.ts
EhabY: That is possible, do you mean adding something like `coder.logging.verbose` (or similar key)?

src/logging/netLog.ts
EhabY: It seems that content-length is usually undefined so we get something like this:

```
2025-09-11 13:29:00.199 [info] Remote setup complete
2025-09-11 13:29:00.199 [info] Logged in to XYZ; checking credentials
2025-09-11 13:29:00.205 [trace] → 79d6e4cc GET /api/v2/users/me <unknown size>
2025-09-11 13:29:00.767 [trace] ← 79d6e4cc 200 GET /api/v2/users/me 562ms (1405b)
2025-09-11 13:29:00.767 [info] Credentials are valid
2025-09-11 13:29:00.782 [trace] ← WS f63c6ef4 connected 583ms
2025-09-11 13:29:32.642 [trace] → d9e90f15 GET /api/v2/workspaces?q=owner%3Ame <unknown size>
2025-09-11 13:29:33.463 [trace] ← d9e90f15 200 GET /api/v2/workspaces?q=owner%3Ame 822ms <unknown size>
2025-09-11 13:29:33.467 [trace] → WS bf1e59bd /api/v2/workspaceagents/<UUID>/watch-metadata-ws
2025-09-11 13:29:34.032 [trace] ← WS bf1e59bd connected 565ms
2025-09-11 13:29:38.468 [trace] → 43ca5d03 GET /api/v2/workspaces?q=owner%3Ame <unknown size>
2025-09-11 13:29:39.262 [trace] ← 43ca5d03 200 GET /api/v2/workspaces?q=owner%3Ame 794ms <unknown size>
2025-09-11 13:29:44.262 [trace] → 9bd5d0e7 GET /api/v2/workspaces?q=owner%3Ame <unknown size>
```

src/websocket/webSocketClient.ts
EhabY: Implemented an extended API (`CodeApi`) that does everything `Api` does but with WebSockets. This made it easier to refactor some things out of `api.ts`

src/websocket/oneWayCodeWebSocket.ts
EhabY: There's also now `CodeApi` as in VS Code's version of the `Api` but yeah maybe needs a change :thinking:

src/logging/netLog.ts
```diff
@@ -0,0 +1,189 @@
...
+		const reasonStr = reason ? ` - ${reason}` : "";
+		const statsStr = stats.length > 0 ? ` [${stats.join(", ")}]` : "";
+
+		this.logger.trace(
+			`▣ WS ${shortId(this.id)} closed${codeStr}${reasonStr}${statsStr}`,
+		);
+	}
+
+	logError(error: unknown, message: string): void {
+		const ms = Date.now() - this.startedAt;
```
EhabY: Note to self, this might be very large, might make sense to format this

src/websocket/oneWayCodeWebSocket.ts
code-asher: Maybe `VSCode` would make it the most clear.

Although, no harm in just also calling our version `OneWayWebSocket` right?  Nothing about it is specific to VS Code.

src/extension.ts
code-asher: Oooh right yeah that makes sense.

Awesome!

src/agentMetadataHelper.ts
code-asher: Good point, I guess we would need to distinguish between an abnormal close and an expected close.  Something we can leave for later, if we decide to do anything, I think.

src/api.ts
code-asher: Yeah the close *event* may not fire, but the underlying connection should be closed if it is still open, so calling `close` should be a no-op.

> the client MUST Close the WebSocket Connection

Not that this really matters!  It just felt like something we could leave to the library to handle, to me.

src/api.ts
code-asher: Yup `socket.removeEventHandler("close", resolve)` is what I was thinking.  I see your reasoning though that someone might change the `close` handler and not realize they should also change the removal further down.

src/api.ts
code-asher: `socket.close()` does not trigger the close event synchronously does it?  Which means `rejectError()` would always run first.

If we are worried, we could always do:

```
rejectError(...)
socket.removeEventListener(...)
socket.close()
```

But imo there is no need for the `socket.close()` anyway.

src/logging/netLog.ts
code-asher: Yup I think returning `undefined` is the right move.  Maybe we want to have some kind of indication when we get at least one undefined size, like we say "at least x bytes" or ">= x bytes", idk.

src/logging/netLog.ts
code-asher: Yup something like that.  For inspiration in the JetBrains plugin we have a separate HTTP logging setting called `httpClientLogLevel` which can be set to `none`, `basic` (logs method, URL, response status), `headers` (also logs sanitized request and response headers), or `body` (also logs body contents).

https://github.com/coder/coder-jetbrains-toolbox/commit/5af07afbc41caab4b2726b084b8760395c110832#diff-0085cecd6b95c447edce684689b0c2a079d44affd1430400d8d7a15b603c2876

src/logging/netLog.ts
code-asher: Hahaha yeah I noticed the same thing when I logged the content size of the binary, seems to always be unset.

I think this is good though, it is better IMO to know the size is unknown then to wonder if the plugin is just not logging it for some reason.

We could maybe shorten to `?b`, maybe that is too unclear though  idk

src/logging/netLog.ts
code-asher: Ooo nice I like that.

src/websocket/webSocketClient.ts
code-asher: Yes exactly!!  Awesome.

I think in the future this means if we make the socket constructor configurable in `coder/coder`'s `Api`, all we have to do is remove our extension and the rest of the code remains the same.

I know `codeApi` stands for the VS Code version of `Api` but maybe `coderApi` would be more accurate?  Since it is an API that accesses Coder?  Like if we added an API for accessing Stripe or something, it would be `StripeApi`.

src/websocket/webSocketClient.ts
code-asher: > Unifying everything here also makes it easy to port this into a singular WebSocket implementation once we fix the two problems I mentioned above (and make use of the baseUrl in WebSockets).

:heart_eyes:

src/websocket/oneWayCodeWebSocket.ts
EhabY: You are right and we do not mix and match so it's not really an issue. Done!

src/agentMetadataHelper.ts
EhabY: The simplest way here is to consider `1000` as normal closure and anything else as abnormal (though not necessarily errors)

Added the following:

```ts
if (event.code !== 1000) {
	handleError(
		new Error(
			`WebSocket closed unexpectedly: ${event.code} ${event.reason}`,
		),
	);
}
```

src/api.ts
EhabY: Ah I see your point, the Mozilla documentation:

> The error event is fired when a connection with a WebSocket has been closed due to an error (some data couldn't be sent for example).

I honestly I'm not sure if the `ws` implementation follows the same standard here or is there a chance an `error` is triggered while keeping the socket open. I quickly looked at the implementation there and it seems to always call close after an error so I'll remove this for now

EhabY: Yet in `coder/coder` there is a piece of code that does this:

```ts
socket.addEventListener("error", () => {
	onError(new Error("Connection for logs failed."));
	socket.close();
});
```

I am confused :sob:

src/api.ts
EhabY: Removed the `removeEventListener` and `socket.close` for now, but see other comment :pray:

src/logging/netLog.ts
EhabY: Added a `>=` indication if there's at least one unknown size!

src/logging/netLog.ts
EhabY: Done

src/logging/netLog.ts
EhabY: Shortened for `?b`

src/logging/netLog.ts
EhabY: Oooh interesting, I'll take some inspiration from that!

For reference, this should only apply to HTTP requests and not WS requests? Like for websockets it would just be the current logging vs. logging every message which could be very noisy

EhabY: Added the logging setting similar to the Jetbrains Toolbox, also separated `netLog` into separate util files more separation of concerns

src/agentMetadataHelper.ts
code-asher: Oh sick that looks great.

src/api.ts
code-asher: Ah yeah I think this is a common mistake.

> is there a chance an error is triggered while keeping the socket open.

I think, if the library did decide to keep the connection open (assuming it even can), that implies the connection is still usable, so we should keep using it.  Or in other words, we can defer to the library's judgement, hopefully.

src/logging/netLog.ts
code-asher: Technically web sockets are also over HTTP :laughing: 

Hm but yeah I think you are right, for now at least we can skip logging every web socket message, if we end up needing this we can add another level or something.

src/api.ts
EhabY: Makes sense, I removed the explicit `close` from the `onerror` handler

src/logging/netLog.ts
EhabY: Hahaha fair enough, in mind I was thinking the protocol `http(s)` vs `ws(s)`

The setting `coder.httpRestClientLogLevel` seems a bit verbose than say `coder.httpClientLogLevel` which matches the name used by JetBrains

src/api/auth.ts
```diff
@@ -0,0 +1,50 @@
+import fs from "fs";
```
code-asher: I know most of the stuff here is auth-related but some (like proxy settings) are not.  What do you think about calling this `middleware.ts` maybe?  Although `needToken` does not fit "middleware" well either.  `util.ts`?  Or we split into `auth.ts` and `agent.ts` maybe.

src/api/codeApi.ts
```diff
@@ -0,0 +1,248 @@
...
+
+const coderSessionTokenHeader = "Coder-Session-Token";
+
+type WorkspaceConfigurationProvider = () => WorkspaceConfiguration;
+
+/**
+ * Unified API class that includes both REST API methods from the base Api class
+ * and WebSocket methods for real-time functionality.
+ */
+export class CodeApi extends Api {
```
code-asher: What do you think about calling this `CoderApi`?  `CodeApi` to me feels like an API to Code rather than to Coder.

src/api/codeApi.ts
```diff
@@ -0,0 +1,248 @@
...
+
+		return socket;
+	};
+
+	private createWebSocket<TData = unknown>(configs: {
+		apiRoute: string;
+		protocols?: string | string[];
+		searchParams?: Record<string, string> | URLSearchParams;
+		options?: ClientOptions;
+	}) {
```
code-asher: I can see arguments either way, but I had the thought we could use `Omit<OneWayWebSocketInit, "location">` here instead of recreating the type.

src/commands.ts
```diff
@@ -239,10 +240,12 @@ export class Commands {
 		token: string,
 		isAutologin: boolean,
 	): Promise<{ user: User; token: string } | null> {
-		const restClient = makeCoderSdk(url, token, this.storage);
-		if (!needToken()) {
+		const client = CodeApi.create(url, token, this.storage.output, () =>
+			vscode.workspace.getConfiguration(),
```
code-asher, line 244: Does this need to be a function because the configuration object can go stale?  I thought (with no evidence lol, just an assumption I had) that `get()` would always return the latest value.

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@
 ### Changed
 
 - Always enable verbose (`-v`) flag when a log directory is configured (`coder.proxyLogDir`).
+- Replaced SSE paths with a one-way WebSocket and centralized creation on `OneWayWebSocket`, and unified socket handling.
```
code-asher: I think this is too implementation-specific to deserve a callout in a user-facing changelog.  Probably users do not particularly care if we use SSE or web sockets.  Since the functionality from the user perspective is identical in this case and there are no actions the user can take, we can just omit the change IMO.

src/logging/httpLogger.ts
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if (logLevel >= HttpClientLogLevel.BODY) {
+		msg += `\n${formatBody(response.data)}`;
+	}
+
+	logger.trace(msg);
+}
+
+export function logRequestError(
```
code-asher: Wondering if it might be valuable to print headers for error responses as well (assuming we have access to them, anyway).

src/logging/httpLogger.ts
```diff
@@ -0,0 +1,119 @@
...
+		const requestId = meta?.requestId || "unknown";
+		const time = meta ? formatTime(Date.now() - meta.startedAt) : "?ms";
+
+		const msg = getErrorMessage(error, "No error message");
+		const detail = getErrorDetail(error) ?? "";
+
+		if (error.response) {
+			const responseData =
+				error.response.statusText || String(error.response.data).slice(0, 100);
+			const errorInfo = [msg, detail, responseData].filter(Boolean).join(" - ");
```
code-asher: Oooo `.filter(Boolean)` is a neat idea

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@
...
 
 ### Added
 
 - Add support for CLI global flag configurations through the `coder.globalFlags` setting.
+- Add logging for all REST and some WebSocket traffic, with REST verbosity configurable via `coder.httpClientLogLevel` (`none`, `basic`, `headers`, `body`).
```
code-asher: We mention the web socket logs in a point below so I think mentioning here is redundant. 

```suggestion
- Add logging for all REST traffic. Verbosity is configurable via `coder.httpClientLogLevel` (`none`, `basic`, `headers`, `body`).
```

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@
...
+- An Axios interceptor that tags each request with a UUID to correlate with its response and measure duration.
```
code-asher: I feel like the interceptor is too much of an implementation detail, what do you think about something like:

> Both REST and WebSocket logs will include a UUID for correlating requests/responses and opens/closes.

Possibly we also move it below the following line so the order goes: note about REST logs, note about WebSocket logs, note about both.

src/logging/httpLogger.ts
```diff
@@ -0,0 +1,119 @@
...
+		const url = formatUri(config);
+		const requestId = meta?.requestId || "unknown";
+		const time = meta ? formatTime(Date.now() - meta.startedAt) : "?ms";
+
+		const msg = getErrorMessage(error, "No error message");
+		const detail = getErrorDetail(error) ?? "";
+
+		if (error.response) {
+			const responseData =
+				error.response.statusText || String(error.response.data).slice(0, 100);
```
code-asher: After experimenting a bit, I wonder if it is too much to show both the error detail and the status text?  Maybe we only show status text if the msg and detail are blank.

For example I can get this somewhat redundant line:

> 2025-09-15 13:37:29.711 [error] ← 9253a837 401 GET /api/v2/users/me 328ms - You are signed out or your session has expired. Please sign in again to continue. - API key expired at "2025-09-09 17:03:07.82664 +0000 UTC". - Unauthorized Request failed with status code 401

code-asher (approved):
Tried it out and it looks glorious

code-asher:
One other minor note: The comment for `WorkspaceMonitor` mentions `SSE` still.

code-asher:
I have to admit the `(?b)` does look kinda confusing, what do you think?  Maybe `??? B` to match for example `533 B` would be better?  Or...`>= 0 B`, idk.  Maybe we should try estimating the size by looking at the body?

src/logging/wsLogger.ts
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	logConnecting(): void {
+		this.logger.trace(`→ WS ${shortId(this.id)} ${this.url}`);
+	}
+
+	logOpen(): void {
+		this.openedAt = Date.now();
+		const time = formatTime(this.openedAt - this.startedAt);
+		this.logger.trace(`← WS ${shortId(this.id)} connected ${time}`);
```
code-asher: I know we can correlate with the IDs to find the URL but could we throw the URL into the web socket logs?  We have it on "connecting" but not the others.  That could make it a bit easier to know which web socket is being logged without having to cross-reference.

src/api/auth.ts
EhabY: I've renamed to `utils.ts`

src/api/codeApi.ts
EhabY: You are correct, nothing about this is `Code` specific. I called it `CodeApi` because `Api` is also technically a "Coder API" but this one is a mix of REST + WS

Renamed!

src/api/codeApi.ts
EhabY: I kinda like this simplification, done

src/commands.ts
EhabY: Yep that's exactly it, when you do `vscode.workspace.getConfiguration()`, you get a snapshot of the settings. I think I'll abstract this in my refactoring work using adapters. But for now it's a provider :/

CHANGELOG.md
EhabY: Fair enough, I suppose the changelog is for user visible stuff

src/logging/httpLogger.ts
EhabY: Simple but effective :)!

CHANGELOG.md
EhabY: Done

CHANGELOG.md
EhabY: I changed it to the following:

```
- Add logging for all REST traffic. Verbosity is configurable via `coder.httpClientLogLevel` (`none`, `basic`, `headers`, `body`).
- Add lifecycle logs for WebSocket creation, errors, and closures.
- Include UUIDs in REST and WebSocket logs to correlate events and measure duration.
```

src/logging/wsLogger.ts
EhabY: Added the URL to all loggings!

src/logging/httpLogger.ts
EhabY: Ah yes that is redundant, I made it log `msg - detail` and only `status text` if both are blank

EhabY:
> I have to admit the `(?b)` does look kinda confusing, what do you think? Maybe `??? B` to match for example `533 B` would be better? Or...`(>= 0 B)`, idk. Maybe we should try estimating the size by looking at the body?

With estimation it looks like this (I simply use `JSON.stringify(body)`):

```
2025-09-16 12:03:27.474 [trace] → 79d177f5 GET /api/v2/workspaces?q=owner%3Ame (??? B)
2025-09-16 12:03:28.308 [trace] ← 79d177f5 200 GET /api/v2/workspaces?q=owner%3Ame (~5.48 kB) 834ms
```

Or possibly default to `0`:

```
2025-09-16 12:09:31.970 [trace] → 238d9059 GET /api/v2/workspaces?q=owner%3Ame (0 B)
2025-09-16 12:09:32.862 [trace] ← 238d9059 200 GET /api/v2/workspaces?q=owner%3Ame (~5.48 kB) 892ms
```

But I just realized that showing this info for `GET` requests is a bit useless since those do not have a body anyway (always 0).

src/logging/httpLogger.ts
EhabY: We do have access to them, should they be printed if the log level is >= headers or all the time :thinking: ?

I would think that for errors we do not give the user a choice here unless it's a different settings. Since we always want to log errors for example

EhabY: It seems printing the msg/detail and the error is a bit redundant as well:

`reason` is `EAI_AGAIN`, while the message also include `EAI_AGAIN`. Then at the end we print the full error which is just the message again. Should I just make it print `msg/detail` only? 

```
2025-09-16 12:34:04.616 [error] ✗ 2c46fd5b GET /api/v2/workspaces?q=owner%3Ame 94ms - getaddrinfo EAI_AGAIN dev.coder.com - EAI_AGAIN
Accept: application/json, text/plain, */*
Content-Type: undefined
Coder-Session-Token: <redacted>
User-Agent: axios/1.8.4
Accept-Encoding: gzip, compress, deflate, br
 A system error occurred (getaddrinfo EAI_AGAIN dev.coder.com)
 ```

src/api/codeApi.ts
code-asher: Sweet!  `Api` is also a mix of REST and WS, btw, except we are not using the WS parts.

src/commands.ts
code-asher: Wild decision on VS Code's part lol, but makes sense!

CHANGELOG.md
code-asher: wunderbar!

src/logging/httpLogger.ts
code-asher: My vote is only if log level is >= headers.

Good point, I think it makes sense to prefer `msg/detail` only when they are available.

EhabY: Done!

For consistency, I also log the body if it's available when the log level is sufficient.

EhabY:
> I like it with the estimation!
> 
> > But I just realized that showing this info for GET requests is a bit useless since those do not have a body anyway (always 0).
> 
> Ohhhhh yeah good point. Personally I like the consistency of always seeing bytes, but I will defer to you if you want to remove it for get requests or show zero. 👌

For consistency I'll show 0 for now, I was just worried it'll cause confusion since people might assume that a `0` byte is an error

Issue #309, opened 2024-06-26
Prevent workspace being kept alive when it should not


code-asher:
There is a problem (which might only exist on Linux) where if the workspace shuts off, VS Code tries to reconnect (which it actually seems to do by reloading the window, from what I have seen), and then as part of that the Coder plugin starts the workspace again, so a workspace stays up forever which is not always what a user wants.

If we can prevent the reload/reconnect to begin with, using a "Do you want to reconnect" popup, that would be great.  

However, we might not be able to stop VS Code from reloading the window.  If that is the case, we can instead show a "Your workspace is off, should we start it?" popup on the connect phase when the workspace is down, which has a similar effect but a bit delayed since it requires the window to reload and the plugin to get started up again first, and it means you have to answer yes for a first connect as well (unless there is a way for us to tell the difference between a first connect and reconnect).

Also probably worth looking into whether there are existing settings to control this behavior.

Also curious to see if the behavior is different between a workspace shutting down and temporary network instability.  Does VS Code reload for the latter case, too?  Maybe it makes a few attempts before reloading?  Maybe in the first case our proxy command is exiting but in the latter case it stays up?  Yeah it might be ssh exiting that is the trigger for the reload.

code-asher:
Another observation.  This issue may or may not happen depending on the timing as we only start the workspace if it is in the `stopped` state.

So if VS Code reconnects, and the workspace is still shutting down, the start will be skipped and it will wait for the stop build to finish, then say "the workspace is stopped" with an option to restart.

If VS Code reconnects, and the workspace shut down quickly enough, once the plugin starts running the workspace will be already `stopped` and so it will then be started, and the plugin waits for the start build to finish.

To make this less timing-dependendant, maybe we put this in a loop where we keep trying to get the workspace into a connectable state (in addition to the prompt to actually start the workspace).

Kira-Pilot:
https://coder.zendesk.com/agent/tickets/2970 corresponding zendesk ticket should be updated when this goes out

PR #583, opened 2025-09-12
Automatically start workspace if started and not running

### LINKED ISSUE #582: Automatically start workplace if started and not currently running
If the workspace is stopped and I press the play/start button: 

Image

A new page opens but I am prompted by the modal below to choose whether I want to start my workspace back up. 

Image

Because I have already pressed play/start in order to get to this page, it is unnecessary friction to asked the user to confirm if they want to start up the workspace. Instead, if the user presses play/start, the workspace should immediately start up without having the modal pop-up and ask for input.

EhabY:
Closes #582

Now it'll attempt to start the workspace when clicking on the "play" button or reloading a window that is already connected.

Currently when building we show a terminal with the build progress, this is good, but we can potentially also show a progress indicator between the different steps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
 - Always enable verbose (`-v`) flag when a log directory is configured (`coder.proxyLogDir`).
+- Automatically start a workspace if it is opened but not running.
 
 ### Added
 


EhabY:
> I presume it will be hard to cover with unit tests since it is mainly in `remote.ts`?

Indeed, and this is heavily using the VS Code API where there are no existing tests for (yet). Currently working a design that would allows us to mock this for tests :)

code-asher:
And the original issue that added the dialog: https://github.com/coder/vscode-coder/issues/309

PR #589, opened 2025-09-22
Refactor storage.ts into testable modules and add unit tests

### LINKED ISSUE #588: Refactor storage.ts into testable modules and add unit tests
`storage.ts` is a large, tightly coupled file that mixes core logic with direct `vscode` usage, making it hard to test. Split it into small, focused modules and add unit tests for the core logic using a thin `vscode` adapter/mocks.

### Proposal

* Extract **core** storage logic (try to remove `vscode` imports or simply mock them in tests).
* Introduce simple interfaces (e.g., `BinaryManager`, `CliConfigManager`, `MementoManager`, `SecretsManager`).
* Write unit tests for core; adapter tests can use a minimal mock.

EhabY:
Closes #588


src/__mocks__/vscode.runtime.ts
```diff
@@ -0,0 +1,142 @@
+import { vi } from "vitest";
```
EhabY, line 1: Sadly we have to duplicate this "vscode" definition manually for tests because we cannot just import VS Code types

src/commands.ts
```diff
@@ -35,8 +40,14 @@ export class Commands {
 	public constructor(
 		private readonly vscodeProposed: typeof vscode,
 		private readonly restClient: Api,
-		private readonly storage: Storage,
-	) {}
+		private readonly logger: Logger,
+		private readonly pathResolver: PathResolver,
+		private readonly mementoManager: MementoManager,
+		private readonly secretsManager: SecretsManager,
+		private readonly binaryManager: BinaryManager,
```
EhabY: Potentially I can introduce something like a `ServiceContainer` that contains all our services/data objects

src/remote.ts
```diff
@@ -646,6 +652,22 @@ export class Remote {
 		};
 	}
 
+	/**
+	 * Migrate the session token file from "session_token" to "session", if needed.
+	 */
+	private async migrateSessionToken(label: string) {
```
EhabY, line 654: This util will be cleaned up as part of the effort for refactoring the `remote.ts`

src/remote.ts
```diff
@@ -978,6 +997,29 @@ export class Remote {
 		return loop();
 	}
 
+	/**
+	 * Returns the log path for the "Remote - SSH" output panel.  There is no VS
+	 * Code API to get the contents of an output panel.  We use this to get the
+	 * active port so we can display network information.
+	 */
+	private async getRemoteSSHLogPath(): Promise<string | undefined> {
```
EhabY, line 1001: Same here

src/__mocks__/testHelpers.ts
```diff
@@ -0,0 +1,219 @@
...
+	 */
+	clear(): void {
+		this.config.clear();
+	}
+
+	/**
+	 * Setup the vscode.workspace.getConfiguration mock to return our values
+	 */
+	private setupVSCodeMock(): void {
+		vi.mocked(vscode.workspace.getConfiguration).mockImplementation(
```
code-asher, line 43: You mentioned that `getConfiguration()` returns a static set of values, could we do the same here for parity with VS Code?

If I understand correctly, our mock would do this:

```typescript
const c = vscode.workspace.getConfiguration()
vscode.workspace.getConfiguration().set("coder.enableDownloads", false)
c.get("coder.enableDownloads") // would be false
```

While in VS Code it will do this:

```typescript
const c = vscode.workspace.getConfiguration()
vscode.workspace.getConfiguration().set("coder.enableDownloads", false)
c.get("coder.enableDownloads") // would still be undefined, or whatever the previous value was
```

I assume you do get the updated value in VS Code if you use the same config to update it though? 

```typescript
const c = vscode.workspace.getConfiguration()
c.set("coder.enableDownloads", false)
c.get("coder.enableDownloads") // would be false
```

src/__mocks__/testHelpers.ts
```diff
@@ -0,0 +1,219 @@
...
+	private externalUrls: string[] = [];
+
+	constructor() {
+		this.setupVSCodeMock();
+	}
+
+	/**
+	 * Set a response for a specific message or set a default response
+	 */
+	setResponse(response: string | undefined): void;
```
code-asher: I wonder if we get rid of `default`, it seems too easy for a dev to call `setResponse("whatever")` and accidentally respond to an unintended notification.  Making which notification is being responded to explicit could be a good move.

src/__mocks__/vscode.runtime.ts
code-asher: Honestly less mocking than I was expecting, not too bad at all.  👍

src/core/binaryManager.test.ts
```diff
@@ -0,0 +1,607 @@
...
+			// Disable signature verification for these tests
+			mockConfig.set("coder.disableSignatureVerification", true);
+		});
+
+		it("reuses matching binary without downloading", async () => {
+			withExistingBinary(TEST_VERSION);
+			const result = await manager.fetchBinary(mockApi, "test");
+			expect(result).toBe(BINARY_PATH);
+			expect(mockAxios.get).not.toHaveBeenCalled();
+			expect(mockLogger.info).toHaveBeenCalledWith(
```
code-asher: Something about testing log messages feels off to me, what do you think?  It feels less like we are testing what we want directly, and more testing an internal side effect of what we want.

Could we test this via the file mock instead (or by using an in-memory filesystem)?  Check that the binary was untouched?  Similar for all the log message tests below.

If we absolutely need to, we could return sentinel values from `fetchBinary` to indicate what happened in a formal way.

Or, if you feel strongly about testing the logs then I think we should do both, test the logs and the resulting behavior.

Still, this is already really great, we could leave this change for a separate PR to not block this one, as it is still a massive improvement.

src/core/binaryManager.test.ts
```diff
@@ -0,0 +1,607 @@
...
+			mockUI.setResponse("Download signature");
+			const result = await manager.fetchBinary(mockApi, "test");
+			expect(result).toBe(BINARY_PATH);
+			expect(vscode.window.showWarningMessage).toHaveBeenCalledWith(
+				"Signature not found",
+				expect.any(Object),
+				expect.any(String),
+				expect.any(String),
+			);
+			// First download and when verfiying twice (404 then 200)
```
code-asher: some typos

```suggestion
			// First download and then verifying twice (404 then 200)
```

src/core/binaryManager.test.ts
```diff
@@ -0,0 +1,607 @@
...
+			withSuccessfulDownload();
+			await manager.fetchBinary(mockApi, "test");
+			expect(cli.version).toHaveBeenCalledWith(BINARY_PATH);
+			expect(mockLogger.info).toHaveBeenCalledWith(
+				"Downloaded binary version is",
+				TEST_VERSION,
+			);
+		});
+
+		it("logs file sizes for debugging", async () => {
```
code-asher: I suppose in this case the point of the test *is* to test the log message, but it still feels off to me to be testing logs.

Maybe another way of thinking about it is, should the edit or removal of this log entry be considered a regression?

But, if you end up deciding to leave the other log tests in, then it would make sense to me to keep this one as well.

src/core/binaryManager.test.ts
```diff
@@ -0,0 +1,607 @@
...
+			get: vi.fn(),
+		} as unknown as AxiosInstance;
+		return {
+			getBuildInfo: vi.fn().mockResolvedValue({ version }),
+			getAxiosInstance: () => axios,
+		} as unknown as Api;
+	}
+
+	function withExistingBinary(version: string) {
+		vi.mocked(cli.stat).mockReset();
```
code-asher: Thinking out loud, could we use an in-memory fs library like memfs instead of mocking `cliManager`?  It feels like we miss out on testing that files and directories are read and written out correctly.

That does make `version` a bit tricky because it needs to be an actual executable, in the Jetbrains plugin we used a script: https://github.com/coder/jetbrains-coder/blob/b1a32ff39d3121167f39cf3d5dc9d0c7f4d4c151/src/test/kotlin/com/coder/gateway/cli/CoderCLIManagerTest.kt#L49-L62

src/core/binaryManager.ts
```diff
@@ -508,7 +393,7 @@ export class Storage {
 				options.push("Download signature");
 			}
 			options.push("Run without verification");
-			const action = await this.vscodeProposed.window.showWarningMessage(
+			const action = await vscode.window.showWarningMessage(
```
code-asher: Will this not fail?  Only the proposed version of the API is allowed to use `useCustom` in notifications, I thought.

src/core/binaryManager.ts
```diff
@@ -562,7 +447,7 @@ export class Storage {
 					this.output,
 				);
 			} catch (error) {
-				const action = await this.vscodeProposed.window.showWarningMessage(
+				const action = await vscode.window.showWarningMessage(
```
code-asher: ditto

src/core/cliConfig.test.ts
```diff
@@ -0,0 +1,139 @@
...
+	const mockFs = vi.mocked(fs);
+	const writtenFiles = new Map<string, string>();
+
+	beforeEach(() => {
+		vi.resetAllMocks();
+		writtenFiles.clear();
+		pathResolver = new PathResolver("/test/base", "/test/log");
+		cliConfigManager = new CliConfigManager(pathResolver);
+
+		mockFs.mkdir.mockResolvedValue(undefined);
```
code-asher: Similarly here I wonder if an in-memory filesystem would be more flexible compared to mocking.  It would be much the same, but with the advantage of full behavioral compatibility.

src/core/cliConfig.ts
```diff
@@ -0,0 +1,81 @@
+import fs from "fs/promises";
+import path from "path";
+import { PathResolver } from "./pathResolver";
+
+export class CliConfigManager {
```
code-asher: Just an idle thought, but it kinda feels to me like this could be part of `CliManager`.

src/core/mementoManager.test.ts
```diff
@@ -0,0 +1,105 @@
+import { describe, it, expect, beforeEach } from "vitest";
+import type { Memento } from "vscode";
+import { MementoManager } from "./mementoManager";
+
+// Simple in-memory implementation of Memento
+class InMemoryMemento implements Memento {
```
code-asher: Could we move this into the file with the other VS Code mocks?  Or does this need to be here?

src/core/pathResolver.test.ts
```diff
@@ -0,0 +1,74 @@
+import * as path from "path";
+import { describe, it, expect, beforeEach } from "vitest";
+import { MockConfigurationProvider } from "../__mocks__/testHelpers";
+import { PathResolver } from "./pathResolver";
+
+describe("PathResolver", () => {
```
code-asher, line 6: A bunch of these tests appear to be exact copies of the functions they are testing, so they are not really testing anything.  Do we need those?  I suppose coverage is a thing but in theory these lines get hit by other tests.

src/core/secretsManager.test.ts
```diff
@@ -0,0 +1,75 @@
+import { describe, it, expect, beforeEach } from "vitest";
+import type { SecretStorage, Event, SecretStorageChangeEvent } from "vscode";
+import { SecretsManager } from "./secretsManager";
+
+// Simple in-memory implementation of SecretStorage
+class InMemorySecretStorage implements SecretStorage {
```
code-asher: Could this go into the file with the other VS Code mocks?

src/commands.ts
code-asher: I do like that idea

src/core/binaryManager.ts
```diff
@@ -3,141 +3,26 @@ import globalAxios, {
...
-const MAX_URLS = 10;
+import * as vscode from "vscode";
+import { errToStr } from "../api/api-helper";
+import * as cli from "../cliManager";
+import { Logger } from "../logging/logger";
+import * as pgp from "../pgp";
+import { PathResolver } from "./pathResolver";
 
-export class Storage {
+export class BinaryManager {
```
code-asher: Not a strong opinion, but I do kinda feel like `BinaryManager` could be part of `CliManager` (or vice versa).  It is not too clear to me what the difference is, even, just from the names alone.

src/error.test.ts
```diff
@@ -2,260 +2,275 @@ import axios from "axios";
...
+	// Node (for now I do this manually by shimming Node).
+	const isElectron =
+		(process.versions.electron || process.env.ELECTRON_RUN_AS_NODE) &&
+		!process.env.VSCODE_PID; // Running from the test explorer in VS Code
 
-// TODO: Remove the vscode mock once we revert the testing framework.
-beforeAll(() => {
-	vi.mock("vscode", () => {
-		return {};
+	// TODO: Remove the vscode mock once we revert the testing framework.
```
code-asher: Could delete this comment now IMO

src/__mocks__/vscode.runtime.ts
EhabY: This is not extensive but yeah not so bad :)

src/__mocks__/testHelpers.ts
EhabY: Huh, I just tested this and the results are a bit surprising.

When we do `get(CONFIG)` we just query this snapshot of configurations, however, updating updates the underlying persisted storage ONLY. So if you query the snapshot it'll ALWAYS return the same value even if you `update` some values through it (essentially the `update` is a `static` method while the `get` uses the internal data).

I'll fix my adapter so it returns an immutable data essentially.

src/__mocks__/testHelpers.ts
EhabY: True, done!

src/error.test.ts
EhabY: True..

src/core/binaryManager.test.ts
EhabY: I like the `memfs` idea! I will use it to check that the binary has been written correctly or that folders are created etc. But I think we should still mock `cli.stat` and `cli.version` since this is a unit test that is only testing the binary download (unless I merge `cliManager` with `binaryManager`?)

src/core/cliConfig.test.ts
EhabY: Used `memfs`

src/core/binaryManager.test.ts
EhabY: I removed all log tests and instead checked the binary content, this feels more real than just checking messages

src/core/binaryManager.test.ts
EhabY: Removed this test :pray:

src/core/mementoManager.test.ts
EhabY: Does not need to, done!

src/core/secretsManager.test.ts
EhabY: Done!

src/core/pathResolver.test.ts
EhabY: Valid, I removed 2 tests there are just a bunch of `path.join`. I kept stuff that accesses the configs or normalizes

src/core/binaryManager.ts
EhabY: I do understand your point, we currently have the following:

* BinaryManager (Maybe I can rename it to `BinaryDownloader` if I it'll just do that)
* CliManager (a bunch of utils for CLI)
* CliConfig (Used to configure CLI - helper not util)

Let me try to remove the confusion here

src/core/cliConfig.test.ts
code-asher: glorious

src/core/binaryManager.test.ts
code-asher: True but that means some of these tests mock the filesystem (`cli.stat`, etc) and some do not (use `memfs`) and to me it feels like it will be easy to make a mistake if something is not mocked perfectly correctly.

src/core/pathResolver.test.ts
code-asher: perfection

src/core/binaryManager.ts
code-asher: Another angle of my thinking is that it would be a nice dev UX to do things like `cli.download()` and `cli.configure()` rather than have two/three separate classes in order to deal with the cli.

code-asher: Another is that `binary` implies a generic class for downloading binaries (or at least this was my first impression), but this is very much specific to the cli only.

src/__mocks__/testHelpers.ts
code-asher: trippy!

src/core/binaryManager.test.ts
EhabY: Hmmm true, I had come up with something like this:

```
// Only mock the platform detection functions from CLI manager
vi.mock("../cliManager", async () => {
	const actual =
		await vi.importActual<typeof import("../binaryUtils")>("../cliManager");
	return {
		...actual,
		// No need to test script execution here
		version: vi.fn(),
		// Only mock platform detection
		name: vi.fn(),
		goos: vi.fn(),
		goarch: vi.fn(),
	};
});
```

The issue is that if we want script execution we need to either mock the `child_process` to use `memfs`, and then mock `os`. Or we can live with partial mocks. I tried mocking the `child_process` but it was a bit verbose and felt error prone (esp. if we want to run it on both Linux and Windows)

src/core/binaryManager.ts
EhabY: You are correct, nice catch!

src/core/binaryManager.test.ts
code-asher: Ah right good point, mocking `child_process` sounds like a nightmare.  Sounds like we might have no good options except to mock `cli.version` for the time being.  :cry:

src/core/binaryManager.ts
EhabY: I think the `BinaryManager` and `CliConfig` can be merged, but the `CliManager` is technically **just** utils, I just don't want a very monolithic huge file. I'll try to balance that out

src/commands.ts
EhabY: I'll do it in the follow up, I have it almost ready :eyes:

src/core/binaryManager.test.ts
EhabY: I mocked the `os` so now only `cliUtils.version` is mocked. We can explore just adding the right process mocks if needed be in the future

src/core/binaryManager.ts
EhabY: Fixed

src/core/binaryManager.ts
EhabY: Fixed

src/core/cliConfig.ts
EhabY: Merged this with `BinaryManager` but kept `cliUtils` separate since those are only utils with no path specific info needed

src/core/binaryManager.ts
EhabY: Done, merged into `CliManager` and renamed `cliManager.ts` into `cliUtils`

src/core/cliConfig.ts
code-asher: dope! :rocket:

PR #590, opened 2025-09-23
Improve consistency in client-side login/logout experience

### LINKED ISSUE #498: Improve consistency in client-side logout experience
## Problem Statement
As of today, it's confusing when a coder user is logged out of a connected VScode window. When a user is connected to multiple workspaces, each window may show different behavior after logout. 

One VSCode window might let you know that you signed out and that you need to re-login. Another will become unusable and disconnected, even after re-auth'ing Coder. The user must then close and reopen it. Another account had the user try to update their workspace from Coder's 'update' button in VSCode, but Coder was already logged out, so they got a 401 error.

## Proposal

To reduce confusion, we should provide a consistent experience across all windows when the user is signed out. Communicating early and clearly that the user is no longer authenticated. The user should be prompted to re-authenticate when this advisory is presented.

EhabY:
Fixes #498

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -219,12 +219,12 @@ export class Commands {
 		this.restClient.setHost(url);
 		this.restClient.setSessionToken(res.token);
 
-		// Store on disk to be used by the cli.
-		await this.cliManager.configure(label, url, res.token);
-
 		// Store these to be used in later sessions.
 		await this.mementoManager.setUrl(url);
 		await this.secretsManager.setSessionToken(res.token);
+
+		// Store on disk to be used by the cli.
+		await this.cliManager.configure(label, url, res.token);
 
 		// These contexts control various menu items and the sidebar.
 		this.contextManager.set("coder.authenticated", true);
@@ -247,9 +247,9 @@ export class Commands {
 				}
 			});
 
+		await this.secretsManager.triggerLoginStateChange("login");
 		// Fetch workspaces for the new deployment.
 		vscode.commands.executeCommand("coder.refreshWorkspaces");
-		this.secretsManager.triggerLoginStateChange("login");
 	}
 
 	/**
@@ -403,9 +403,9 @@ export class Commands {
 				}
 			});
 
+		await this.secretsManager.triggerLoginStateChange("logout");
 		// This will result in clearing the workspace list.
 		vscode.commands.executeCommand("coder.refreshWorkspaces");
-		this.secretsManager.triggerLoginStateChange("logout");
 	}
 
 	/**
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -197,14 +197,11 @@ export class Commands {
 		// It is possible that we are trying to log into an old-style host, in which
 		// case we want to write with the provided blank label instead of generating
 		// a host label.
-		const label = args?.label === undefined ? toSafeHost(url) : args?.label;
+		const label = args?.label === undefined ? toSafeHost(url) : args.label;
 
 		// Try to get a token from the user, if we need one, and their user.
-		const res = await this.maybeAskToken(
-			url,
-			args?.token,
-			args?.autoLogin === true,
-		);
+		const autoLogin = args?.autoLogin === true;
+		const res = await this.maybeAskToken(url, args?.token, autoLogin);
 		if (!res) {
 			return; // The user aborted, or unable to auth.
 		}
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -211,12 +211,12 @@ export class Commands {
 		this.restClient.setHost(url);
 		this.restClient.setSessionToken(res.token);
 
+		// Store on disk to be used by the cli.
+		await this.cliManager.configure(label, url, res.token);
+
 		// Store these to be used in later sessions.
 		await this.mementoManager.setUrl(url);
 		await this.secretsManager.setSessionToken(res.token);
-
-		// Store on disk to be used by the cli.
-		await this.cliManager.configure(label, url, res.token);
 
 		// These contexts control various menu items and the sidebar.
 		await vscode.commands.executeCommand(
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -263,14 +263,14 @@ export class Commands {
 	): Promise<{ user: User; token: string } | null> {
 		const client = CoderApi.create(url, token, this.logger);
 		const needsToken = needToken(vscode.workspace.getConfiguration());
-		try {
-			const user = await client.getAuthenticatedUser();
-			// For non-token auth, we write a blank token since the `vscodessh`
-			// command currently always requires a token file.
-			// For token auth, we have valid access so we can just return the user here
-			return { token: needsToken && token ? token : "", user };
-		} catch (err) {
-			if (!needToken(vscode.workspace.getConfiguration())) {
+		if (!needsToken || token) {
+			try {
+				const user = await client.getAuthenticatedUser();
+				// For non-token auth, we write a blank token since the `vscodessh`
+				// command currently always requires a token file.
+				// For token auth, we have valid access so we can just return the user here
+				return { token: needsToken && token ? token : "", user };
+			} catch (err) {
 				const message = getErrorMessage(err, "no response from the server");
 				if (isAutoLogin) {
 					this.logger.warn("Failed to log in to Coder server:", message);
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -179,19 +179,17 @@ export class Commands {
 	}
 
 	/**
-	 * Log into the provided deployment.  If the deployment URL is not specified,
+	 * Log into the provided deployment. If the deployment URL is not specified,
 	 * ask for it first with a menu showing recent URLs along with the default URL
 	 * and CODER_URL, if those are set.
 	 */
-	public async login(...args: string[]): Promise<void> {
-		// Destructure would be nice but VS Code can pass undefined which errors.
-		const inputUrl = args[0];
-		const inputToken = args[1];
-		const inputLabel = args[2];
-		const isAutologin =
-			typeof args[3] === "undefined" ? false : Boolean(args[3]);
-
-		const url = await this.maybeAskUrl(inputUrl);
+	public async login(args?: {
+		url?: string;
+		token?: string;
+		label?: string;
+		autoLogin?: boolean;
+	}): Promise<void> {
+		const url = await this.maybeAskUrl(args?.url);
 		if (!url) {
 			return; // The user aborted.
 		}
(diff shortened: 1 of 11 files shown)


src/commands.ts
```diff
@@ -237,21 +238,23 @@ export class Commands {
...
-				// command currently always requires a token file.
-				return { token: "", user };
-			} catch (err) {
+		const needsToken = needToken(vscode.workspace.getConfiguration());
+		try {
+			const user = await client.getAuthenticatedUser();
+			// For non-token auth, we write a blank token since the `vscodessh`
+			// command currently always requires a token file.
+			// For token auth, we have valid access so we can just return the user here
+			return { token: needsToken && token ? token : "", user };
```
EhabY: This changes makes it that we try to use the passed URL + Token combo so that we do not need to prompt the user and login automatically when another window changes the login state

The more I think about it the more I think it should be `token || !needsToken` (it used to be just `!needsToken`) :thinking:

src/commands.ts
```diff
@@ -286,6 +289,9 @@ export class Commands {
 			value: token || (await this.storage.getSessionToken()),
 			ignoreFocusOut: true,
 			validateInput: async (value) => {
+				if (!value) {
+					return null;
+				}
 				client.setSessionToken(value);
 				try {
 					user = await client.getAuthenticatedUser();
```
EhabY, line 314: Should we attempt to get the authenticated user when the `token` is a blank string (like when this first opens)?

EhabY:
> I tried out this PR and while I see a "you've been logged in" popup in multiple windows, the other windows are still stuck on this screen and there is no attempt to re-connect to the workspace

Can you provide a reproducible for this? I am not sure what's special about some windows that they do not respond to the change :thinking:

mafredri:
> Can you provide a reproducible for this? I am not sure what's special about some windows that they do not respond to the change 🤔

@EhabY Sure, I'll write down what I did.

Open two VS Code windows to the same workspace (different folders in the same workspace, whilst logged in):

```
coder open vscode my-workspace.dev /home/coder/foo
```

and

```
coder open vscode my-workspace.dev /home/coder/bar
```

Now you should have two windows that are both. Make sure VS Code is set to restore previous windows after quitting.

Then:

1. Quit VS Code
2. Delete `~/Library/Application Support/Code/.../session` (it's in globalStorage, not at the machine right now so I can't give you full path)
3. Open VS Code
4. Both windows should show the ^aforementioned dialogue
5. Perform log in on one of the windows
6. Notice the dialogue does not disappear on the second

Does this help?

EhabY:
Yes, thanks for the detailed report! I can see the issue now: VS Code modals aren't dismissible without user action, which is by design (see [https://github.com/Microsoft/vscode/issues/2732](https://github.com/Microsoft/vscode/issues/2732)). I can adjust our flow so we don't close the window if the user signs in in the meantime. The logs in the other window show that the sign-in succeeded, so the real problem is our fallback when a login completes after the modal appears.

Do keep in mind that we cannot change the content of the modal once it's shown, so we either go for something less intrusive or just handle the fallback correctly. What do you think?

mafredri:
Oh, that's unfortunate, but understandable (that the modals can't be dismissed).

> Using the status bar for messages which need to be updated / dismissed is suggested there as the recommended approach in the issue I mentioned. What do you think?

It's one option but I don't think it's a good one. It's fairly hidden for something that's required for your editor to start working.

I think updating the modal is the way forward. It could say something like "If you already performed login in another window, please use the cancel button."

image

This is what it looks like, after logging in from another window. See how it says "Opening remote" at the bottom, but nothing happens? I think it would be nice if we could make it successfully load the workspace in the background. Without looking at the actual code, perhaps we could do something along the lines of `await Promise.race([loginmodal, loginevent])`. If the workspace has loaded in the background, that's further confirmation for the user they can just hit `cancel`. Wdyt?

EhabY:
I like this idea. Updating the modal to clearly indicate that a login happens in another window, plus adding background login capability, would be a solid improvement. I think this is possible but I'll be careful to avoid any racey conditions.

I think the following makes sense:

Description:
> You must log in to access ${workspaceName}. If you already performed login in another window, please close this popup.

Then it's a race between two conditions as you said, 1. Login event, 2. Modal interaction. In the case of a login event, if the user clicks anything (`x`, `Cancel`, `Login`) it will just dismiss the modal and show the already connected workspace. The flow for the modal interaction stays the same as before.

src/remote.ts
```diff
@@ -217,39 +241,54 @@ export class Remote {
...
 
-		// It could be that the cli config was deleted.  If so, ask for the url.
-		if (
-			!baseUrlRaw ||
-			(!token && needToken(vscode.workspace.getConfiguration()))
-		) {
-			const result = await this.vscodeProposed.window.showInformationMessage(
-				"You are not logged in...",
+		const showLoginDialog = async (message: string) => {
+			const dialogPromise = this.vscodeProposed.window.showInformationMessage(
```
EhabY: I could move the `createLoginDetectionPromise` into the `showLoginDialog` but we have a window of opportunity between creating the promise and the user ALREADY logging in. Not sure if this could ever happen

EhabY: Moved there for now so we do not forget to set the promise before potentially resolving

src/remote/remote.ts
```diff
@@ -222,34 +245,48 @@ export class Remote {
...
 
-		// It could be that the cli config was deleted.  If so, ask for the url.
-		if (
-			!baseUrlRaw ||
-			(!token && needToken(vscode.workspace.getConfiguration()))
-		) {
-			const result = await this.vscodeProposed.window.showInformationMessage(
-				"You are not logged in...",
+		const showLoginDialog = async (message: string) => {
+			this.createLoginDetectionPromise();
```
mafredri, line 276: Do we need to guard against multiple calls to this or `createLoginDetectionPromise`? In theory, we could overwrite the previous one and never resolve it.

mafredri (commented):
Nice work @EhabY, the happy-path is now a pretty good user experience, thanks for implementing my proposal!

I noticed a situation where the login-into-other-windows functionality breaks.

0. [Initial steps from previous repro]
1. Select login in one (of two) windows
2. Enter the _wrong_ token (`asdf`), see a message like: "Failed to authenticate: You are ..."
3. Enter the _correct_ token -> Welcome in only _one_ window
4. In the other window, press login -> token prompt, press cancel -> close window, etc.

It's a bit strange to me that authentication is performed immediately after you've typed in your token, not when pressing [Enter]. (What if you're typing it manually.)

I'm also not sure why this breaks the login flow, perhaps a login attempt (even if failed) is considered a resolution of the promise.

EhabY:
> It's a bit strange to me that authentication is performed immediately after you've typed in your token, not when pressing [Enter]. (What if you're typing it manually.)

I think this is to provide feedback as soon as possible, VS Code allows this "validateInput" callback. If we want it on "Enter" only it might look a bit more awkward.

> I'm also not sure why this breaks the login flow, perhaps a login attempt (even if failed) is considered a resolution of the promise.

Hmmm, I tried the reproducer and it seems to work (no breakage), we should only show the "Welcome" message if the token is correct anyway, otherwise it's like nothing has changed. Did you do something else in the meantime? :thinking:

src/remote/remote.ts
EhabY: This should never happen since we cannot create this promise twice currently, but I suppose we can resolve the previous promise (or possibly even reject it) if it has not been resolved and we try to create a new one.

EhabY: Added a reject as a precaution so that we if it ever happens we fail fast

EhabY:
Hmmmm, to me this means it's one of two things: either we have a race where the 2nd window does not read the correctly entered configurations fast enough or that the session token has not changed (the former is impossible if you deleted the `session` file).

Can I bother you one more time to test the latest push so we can rule out (1)? I sadly cannot reproduce so I'm not sure if this fixes it

src/core/contextManager.ts
```diff
@@ -0,0 +1,33 @@
...
+		});
+	}
+
+	public set(key: CoderContext, value: boolean): void {
+		this.context.set(key, value);
+		vscode.commands.executeCommand("setContext", key, value);
+	}
+
+	public get(key: CoderContext): boolean {
+		return this.context.get(key) ?? CONTEXT_DEFAULTS[key];
```
mafredri, line 27: I'm unfamiliar with VS Code contexts, so just checking: is there any risk our local context could become out-of-sync with the VS Code context? I'm guessing this is a per-window context, in which case the only "risk" I see would be setting/accessing the context directly rather than going through this manager. Is that correct?

src/core/secretsManager.ts
```diff
@@ -19,11 +25,43 @@ export class SecretsManager {
...
+		return this.secrets.onDidChange(async (e) => {
+			if (e.key === LOGIN_STATE_KEY) {
+				const state = await this.secrets.get(LOGIN_STATE_KEY);
+				if (state?.startsWith("login")) {
+					listener("login");
+				} else if (state?.startsWith("logout")) {
+					listener("logout");
+				} else {
+					// Secret was deleted or is invalid
+					listener(undefined);
```
mafredri: Considering this can emit three states, login/logout/undefined, but we only handle login and assume logout otherwise in the rest of the code. Is it worth keeping this third state, should it be explicitly handled?

src/commands.ts
```diff
@@ -221,13 +227,9 @@ export class Commands {
...
 		// These contexts control various menu items and the sidebar.
-		await vscode.commands.executeCommand(
-			"setContext",
-			"coder.authenticated",
-			true,
-		);
+		this.contextManager.set("coder.authenticated", true);
 		if (res.user.roles.find((role) => role.name === "owner")) {
-			await vscode.commands.executeCommand("setContext", "coder.isOwner", true);
+			this.contextManager.set("coder.isOwner", true);
```
mafredri, line 232: Previously we awaited the result of `setContext` command, but that is not done in the context manager. Potentially problematic or fine?

src/extension.ts
```diff
@@ -327,6 +329,27 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+
+			if (state === "login") {
+				const token = await secretsManager.getSessionToken();
+				const url = mementoManager.getUrl();
+				// Should login the user directly if the URL+Token are valid
+				await commands.login({ url, token });
+				// Resolve any pending login detection promises
+				remote.resolveLoginDetected();
+			} else {
+				await commands.forceLogout();
```
mafredri: Do we need to resolve/reject on this code path?

EhabY: Should be an impossibility with the authenticated check in the logout command implementation, but even if it happens then I do not think we should resolve/reject because if the user logs in after that it should just work

src/core/secretsManager.ts
EhabY: We do handle the 3 cases but I can see the confusion now, I do:

```
if (state === undefined) {
	return
}

if (state === "login") {
	...
} else {
	...
}
```

I'll just make it an `enum` with `LOGIN`/`LOGOUT`/`INVALID` and handle it in a switch so we never forget this if we add a new state.

src/core/contextManager.ts
EhabY: There is no way to access a context, that's why I added the manager. But yes it could fall out of sync if we set the context directly without the manager. I might be able to add a linting rule to prevent that :thinking:

src/commands.ts
EhabY: Official doc says that `setContext` has no return value (no promise): https://code.visualstudio.com/api/references/commands

Also looking at the current implementation it's seems to be fully synchronous and a lot of other extensions and VS Code doc never use `await` with this exact command.

See VS Code impl: https://github.com/microsoft/vscode/blob/8b72f26d634d6d732b1f38fd41a29bb3b093c2bb/src/vs/platform/contextkey/browser/contextKeyService.ts#L641

src/core/contextManager.ts
EhabY: Added a simple `no-restricted-syntax` rule

PR #597, opened 2025-09-29
Add support to CODER_BINARY_DESTINATION environment variable

### LINKED ISSUE #256: write Coder binary to a fixed (not-user-specific) path
In order to get direct network connections working, a customer needs to add a firewall exception for `coder.exe` on Windows.  They manage their firewall rules centrally, and the problem is that the IDE plugin writes to the directory

```
c:\Users\%USERNAME%\AppData\Roaming\Code\User\globalStorage\coder.coder-remote
```

Since the directory is different for every person, they can't push out a central firewall rule.

It would be better if we could write to somewhere that doesn't depend on the username, and/or write somewhere centrally configurable.

EhabY:
Fixes #256

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -102,10 +102,13 @@ export class Commands {
 	 * CODER_URL or enter a new one.  Undefined means the user aborted.
 	 */
 	private async askURL(selection?: string): Promise<string | undefined> {
-		const defaultURL =
-			vscode.workspace.getConfiguration().get<string>("coder.defaultUrl") ?? "";
+		const defaultURL = vscode.workspace
+			.getConfiguration()
+			.get<string>("coder.defaultUrl")
+			?.trim();
 		const quickPick = vscode.window.createQuickPick();
-		quickPick.value = selection || defaultURL || process.env.CODER_URL || "";
+		quickPick.value =
+			selection || defaultURL || process.env.CODER_URL?.trim() || "";
 		quickPick.placeholder = "https://example.coder.com";
 		quickPick.title = "Enter the URL of your Coder deployment.";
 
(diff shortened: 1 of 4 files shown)


src/core/pathResolver.test.ts
```diff
@@ -41,7 +43,28 @@ describe("PathResolver", () => {
 		it("should normalize custom paths", () => {
 			mockConfig.set("coder.binaryDestination", "/custom/../binary/./path");
 			expect(pathResolver.getBinaryCachePath("deployment")).toBe(
-				path.normalize("/custom/../binary/./path"),
```
EhabY, line 44: I think this might fail on Windows due to the normalization to `\`. I'll test on Windows and create a follow up for making sure all tests run correctly on Windows

EhabY: Created a ticket for this since it's out of scope here: https://github.com/coder/vscode-coder/issues/604

src/core/pathResolver.ts
```diff
@@ -28,11 +28,14 @@ export class PathResolver {
...
 	 */
 	public getBinaryCachePath(label: string): string {
-		const configPath = vscode.workspace
+		const settingPath = vscode.workspace
 			.getConfiguration()
-			.get<string>("coder.binaryDestination");
-		return configPath && configPath.trim().length > 0
-			? path.normalize(configPath)
+			.get<string>("coder.binaryDestination")
+			?.trim();
```
code-asher, line 34: Non-blocking: I think we are not trimming `coder.headerCommand`, `coder.defaultUrl`, and their env vars.  Do you know if VS Code's configuration already trims values?  If so, we can remove this, if not maybe we should add `trim()` to the others as well (eventually).

code-asher (approved):
Awesome!

src/core/pathResolver.ts
EhabY: I just tested it out and VS Code does not trim, for now I trimmed the usages of `coder.headerCommand` and `coder.defaultUrl` (and their env variables)

code-asher: Awesome, thank you!

PR #603, opened 2025-09-30
Add search filter to Coder Workspaces tree views

### LINKED ISSUE #330: Add search filter to sidebar

EhabY:
Fixes #330

Adds a search filter for both views "My Workspaces" and "All Workspaces", this depends on the VS Code command `list.find` so support fuzzy match and filtering but does not differentiate between nodes (so we filter on everything not just workspaces or agents... etc).

Also, I have add `id`s to all tree items, this guarantees that the expansion state stays the same between refreshes.

One downside to using this method is that whenever the tree view refreshes (which is every ~1s) the focus is shifted to the tree instead of the input. I tried to implement selective refreshing such that we refresh only the affected nodes but that didn't fix it :/. See:

https://github.com/user-attachments/assets/ede5b005-a19b-4d48-8fe5-7db6df83652b

### DIFF
--- a/package.json
+++ b/package.json
@@ -204,6 +204,7 @@
 			{
 				"command": "coder.createWorkspace",
 				"title": "Create Workspace",
+				"category": "Coder",
 				"when": "coder.authenticated",
 				"icon": "$(add)"
 			},
@@ -226,7 +227,8 @@
 			},
 			{
 				"command": "coder.refreshWorkspaces",
-				"title": "Coder: Refresh Workspace",
+				"title": "Refresh Workspace",
+				"category": "Coder",
 				"icon": "$(refresh)",
 				"when": "coder.authenticated"
 			},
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -262,12 +284,22 @@
...
 				{
 					"command": "coder.refreshWorkspaces",
 					"when": "coder.authenticated && view == myWorkspaces",
-					"group": "navigation"
+					"group": "navigation@2"
+				},
+				{
+					"command": "coder.searchMyWorkspaces",
+					"when": "coder.authenticated && view == myWorkspaces",
+					"group": "navigation@3"
```
EhabY, line 297: Adding the `@` after the group gives us a deterministic order. Not sure what the right order is for those buttons though (Create/Refresh/Search?)

package.json
```diff
@@ -226,7 +227,8 @@
 			},
 			{
 				"command": "coder.refreshWorkspaces",
-				"title": "Coder: Refresh Workspace",
+				"title": "Refresh Workspace",
+				"category": "Coder",
```
EhabY, line 231: Doing this means that the hovers on the tree view show `Refresh Workspace` (Without the `Coder:` prefix) but in the command palette the prefix is there. It's redundant to include the category in the title and we should let VS Code display it where it makes sense (this is a common practice in VS Code extensions)

aslilac: definitely

aslilac (approved):

PR #605, opened 2025-10-01
Fix subscriptions not disposed on extension deactivation

EhabY:
* Fix subscriptions not collected
* Use `vscode.workspace.getConfiguration` instead of passing a provider
* Smol simplification of the `tsconfig.js` for tests


src/extension.ts
```diff
@@ -253,59 +268,89 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+	ctx.subscriptions.push(uriHandler);
 
 	// Register globally available commands.  Many of these have visibility
 	// controlled by contexts, see `when` in the package.json.
 	const commands = new Commands(serviceContainer, client);
-	vscode.commands.registerCommand("coder.login", commands.login.bind(commands));
-	vscode.commands.registerCommand(
-		"coder.logout",
-		commands.logout.bind(commands),
+	ctx.subscriptions.push(
```
code-asher, line 276: Not a big deal at all, but with cases like these where the disposables are all lined up, we could push these all in one call right?  Like:

```
	ctx.subscriptions.push(
		vscode.commands.registerCommand("coder.logout", commands.logout.bind(commands)),
		vscode.commands.registerCommand("coder.open", commands.open.bind(commands)),
		// etc, etc
	);
```

It would make the diff a bit nicer too.

src/remote/remote.ts
```diff
@@ -398,256 +393,262 @@ export class Remote {
...
 						remoteAuthority,
 						workspace.owner_name,
 						workspace.name,
 						agent.name,
-					),
-				);
-			}),
-		);
+					);
+					disposables.push(labelFormatterDisposable);
```
code-asher: Probably not a big deal, but in theory we could collect an unbounded number of disposed label formatters in the disposables array.  Maybe we should store a single function that disposes the current formatter?  Something like:

```javascript
disposables.push({
  dispose: () => labelFormatterDisposable.dispose()
})
```

src/api/coderApi.ts
```diff
@@ -33,17 +33,12 @@ import { createHttpAgent } from "./utils";
 
 const coderSessionTokenHeader = "Coder-Session-Token";
 
-type WorkspaceConfigurationProvider = () => WorkspaceConfiguration;
```
code-asher, line 36: Was removing the provider related to the disposing in any way?  Or just kinda a "while I am here" sort of thing?

code-asher (approved):
Awesome!!

src/extension.ts
EhabY: You are right yeah, combined those since they are all lined up

src/remote/remote.ts
EhabY: True, done!

src/api/coderApi.ts
EhabY: Unrelated, this was a follow up to the previous PR but I didn't want to add more to it. Essentially what we talked about that VS Code API should be mocked and not injected in unit tests

PR #607, opened 2025-10-02
Fix JSON stringify circular structure when logging

### LINKED ISSUE #606: `TypeError: Converting circular structure to JSON` when downloading Coder binary
We're seeing this exception when the extension tries to download a new version of the Coder CLI in Cursor.

```
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'TLSSocket'
    |     property 'parser' -> object with constructor 'HTTPParser'
```

It seems like this only affects older versions of VSCode, as I can't reproduce this in VSCode 1.104.2. The bug was introduced in extension version 1.11.0. The download works when downgrading the extension to version 1.10.1.

The output logs preceding this:

```
2025-10-02 14:14:18.402 [info] Downloading since existing binary does not match the server version
2025-10-02 14:14:18.403 [info] Removed coder-darwin-arm64.temp-bs1fi
2025-10-02 14:14:18.403 [info] Downloading binary from /bin/coder-darwin-arm64
2025-10-02 14:14:18.437 [info] Using ETag 93ba9ec75313948cc85baadc9914917c70e4350e
2025-10-02 14:14:18.862 [warning] Converting circular structure to JSON
    --> starting at object with constructor 'TLSSocket'
    |     property 'parser' -> object with constructor 'HTTPParser'
    --- property 'socket' closes the circle
```

The call stack looks like this, when I run in debug mode:

<details><summary>Full stack trace</summary>

```
formatContentLength (/Users/davidv/work/vscode-coder/src/logging/formatters.ts:40)
logResponse (/Users/davidv/work/vscode-coder/src/logging/httpLogger.ts:68)
<anonymous> (/Users/davidv/work/vscode-coder/src/api/coderApi.ts:232)
process.processTicksAndRejections (internal/process/task_queues:95)
Promise.then (Unknown Source:0)
_request (/Users/davidv/work/vscode-coder/node_modules/axios/dist/node/axios.cjs:4599)
request (/Users/davidv/work/vscode-coder/node_modules/axios/dist/node/axios.cjs:4478)
Axios.<computed> (/Users/davidv/work/vscode-coder/node_modules/axios/dist/node/axios.cjs:4647)
wrap (/Users/davidv/work/vscode-coder/node_modules/axios/dist/node/axios.cjs:31)
download (/Users/davidv/work/vscode-coder/src/core/cliManager.ts:253)
fetchBinary (/Users/davidv/work/vscode-coder/src/core/cliManager.ts:133)
await (Unknown Source:0)
processTicksAndRejections (internal/process/task_queues:95)
await (Unknown Source:0)
processTicksAndRejections (internal/process/task_queues:95)
await (Unknown Source:0)
processTicksAndRejections (internal/process/task_queues:95)
await (Unknown Source:0)
setup (/Users/davidv/work/vscode-coder/src/remote/remote.ts:301)
await (Unknown Source:0)
```

</details>

https://github.com/coder/vscode-coder/blob/67e85e6106000d6729f20cac5d35b3db1e58df9c/src/logging/formatters.ts#L40
https://github.com/coder/vscode-coder/blob/67e85e6106000d6729f20cac5d35b3db1e58df9c/src/logging/httpLogger.ts#L68
https://github.com/coder/vscode-coder/blob/67e85e6106000d6729f20cac5d35b3db1e58df9c/src/api/coderApi.ts#L232
https://github.com/coder/vscode-coder/blob/67e85e6106000d6729f20cac5d35b3db1e58df9c/src/core/cliManager.ts#L253
https://github.com/coder/vscode-coder/blob/67e85e6106000d6729f20cac5d35b3db1e58df9c/src/core/cliManager.ts#L133

#### IDE version information

Cursor Version: 1.7.28 (Universal)
VSCode Version: 1.99.3
Commit: adb0f9e3e4f184bba7f3fa6dbfd72ad0ebb8cfd0
Date: 2025-10-01T02:45:21.769Z
Electron: 34.5.8
Chromium: 132.0.6834.210
Node.js: 20.19.1
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.6.0

EhabY:
Closes #606

### DIFF
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -28,7 +28,7 @@ import {
 	type RequestConfigWithMeta,
 	HttpClientLogLevel,
 } from "../logging/types";
-import { serializeValue, sizeOf } from "../logging/utils";
+import { sizeOf } from "../logging/utils";
 import { WsLogger } from "../logging/wsLogger";
 import {
 	OneWayWebSocket,
@@ -261,7 +261,7 @@ function wrapRequestTransform(
 				? transformer
 				: [transformer];
 
-			// Transform the request first then estimate the size
+			// Transform the request first then get the size (measure what's sent over the wire)
 			const result = transformerArray.reduce(
 				(d, fn) => fn.call(config, d, headers),
 				data,
@@ -280,7 +280,7 @@ function wrapResponseTransform(
 ): AxiosResponseTransformer[] {
 	return [
 		(data: unknown, headers: AxiosResponseHeaders, status?: number) => {
-			// estimate the size before transforming the response
+			// Get the size before transforming the response (measure what's sent over the wire)
 			config.rawResponseSize = getSize(headers, data);
 
 			const transformerArray = Array.isArray(transformer)
@@ -301,14 +301,7 @@ function getSize(headers: AxiosHeaders, data: unknown): number | undefined {
 		return parseInt(contentLength, 10);
 	}
 
-	const size = sizeOf(data);
-	if (size !== undefined) {
-		return size;
-	}
-
-	// Fallback
-	const stringified = serializeValue(data);
-	return stringified === null ? undefined : Buffer.byteLength(stringified);
+	return sizeOf(data);
 }
 
 function getLogLevel(): HttpClientLogLevel {
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -1,4 +1,9 @@
-import { type AxiosInstance } from "axios";
+import {
+	type AxiosResponseHeaders,
+	type AxiosInstance,
+	type AxiosHeaders,
+	type AxiosResponseTransformer,
+} from "axios";
 import { Api } from "coder/site/src/api/api";
 import {
 	type GetInboxNotificationResponse,
@@ -23,6 +28,7 @@ import {
 	type RequestConfigWithMeta,
 	HttpClientLogLevel,
 } from "../logging/types";
+import { serializeValue, sizeOf } from "../logging/utils";
 import { WsLogger } from "../logging/wsLogger";
 import {
 	OneWayWebSocket,
@@ -207,7 +213,24 @@ function addLoggingInterceptors(client: AxiosInstance, logger: Logger) {
 		(config) => {
 			const configWithMeta = config as RequestConfigWithMeta;
 			configWithMeta.metadata = createRequestMeta();
-			logRequest(logger, configWithMeta, getLogLevel());
+
+			config.transformRequest = [
+				...wrapRequestTransform(
+					config.transformRequest || client.defaults.transformRequest || [],
+					configWithMeta,
+				),
+				(data) => {
+					// Log after setting the raw request size
+					logRequest(logger, configWithMeta, getLogLevel());
+					return data;
+				},
+			];
+
+			config.transformResponse = wrapResponseTransform(
+				config.transformResponse || client.defaults.transformResponse || [],
+				configWithMeta,
+			);
+
 			return config;
 		},
 		(error: unknown) => {
@@ -226,6 +249,66 @@ function addLoggingInterceptors(client: AxiosInstance, logger: Logger) {
 			return Promise.reject(error);
 		},
 	);
+}
+
+function wrapRequestTransform(
+	transformer: AxiosResponseTransformer | AxiosResponseTransformer[],
+	config: RequestConfigWithMeta,
+): AxiosResponseTransformer[] {
+	return [
+		(data: unknown, headers: AxiosHeaders) => {
+			const transformerArray = Array.isArray(transformer)
+				? transformer
+				: [transformer];
+
+			// Transform the request first then estimate the size
+			const result = transformerArray.reduce(
+				(d, fn) => fn.call(config, d, headers),
+				data,
+			);
+
+			config.rawRequestSize = getSize(config.headers, result);
+
+			return result;
+		},
+	];
+}
+
+function wrapResponseTransform(
+	transformer: AxiosResponseTransformer | AxiosResponseTransformer[],
+	config: RequestConfigWithMeta,
+): AxiosResponseTransformer[] {
+	return [
+		(data: unknown, headers: AxiosResponseHeaders, status?: number) => {
+			// estimate the size before transforming the response
+			config.rawResponseSize = getSize(headers, data);
+
+			const transformerArray = Array.isArray(transformer)
+				? transformer
+				: [transformer];
+
+			return transformerArray.reduce(
+				(d, fn) => fn.call(config, d, headers, status),
+				data,
+			);
+		},
+	];
+}
+
+function getSize(headers: AxiosHeaders, data: unknown): number | undefined {
+	const contentLength = headers["content-length"];
+	if (contentLength !== undefined) {
+		return parseInt(contentLength, 10);
+	}
+
+	const size = sizeOf(data);
+	if (size !== undefined) {
+		return size;
+	}
+
+	// Fallback
+	const stringified = serializeValue(data);
+	return stringified === null ? undefined : Buffer.byteLength(stringified);
 }
 
 function getLogLevel(): HttpClientLogLevel {
--- a/src/logging/formatters.ts
+++ b/src/logging/formatters.ts
@@ -3,7 +3,7 @@ import prettyBytes from "pretty-bytes";
 
 import { sizeOf } from "./utils";
 
-import type { InternalAxiosRequestConfig } from "axios";
+import type { AxiosRequestConfig } from "axios";
 
 const SENSITIVE_HEADERS = ["Coder-Session-Token", "Proxy-Authorization"];
 
@@ -21,7 +21,7 @@ export function formatTime(ms: number): string {
 }
 
 export function formatMethod(method: string | undefined): string {
-	return (method ?? "GET").toUpperCase();
+	return (method ? method : "GET").toUpperCase();
 }
 
 /**
@@ -55,9 +55,7 @@ export function formatContentLength(
 	return "(? B)";
 }
 
-export function formatUri(
-	config: InternalAxiosRequestConfig | undefined,
-): string {
+export function formatUri(config: AxiosRequestConfig | undefined): string {
 	return config?.url || "<no url>";
 }
 
--- a/src/logging/formatters.ts
+++ b/src/logging/formatters.ts
@@ -1,7 +1,6 @@
-import util from "node:util";
 import prettyBytes from "pretty-bytes";
 
-import { sizeOf } from "./utils";
+import { serializeValue } from "./utils";
 
 import type { AxiosRequestConfig } from "axios";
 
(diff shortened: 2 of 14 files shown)


src/logging/formatters.ts
```diff
@@ -66,8 +85,38 @@ export function formatHeaders(headers: Record<string, unknown>): string {
...
+				return "[Symbol]";
+			}
+			if (typeof value === "bigint") {
+				return value.toString();
+			}
+
+			return value;
+		});
+	} catch {
+		return null;
```
ibetitsmike: No error handling?

EhabY: I am not sure if we want error handling here since this is purely for estimating the `data`/`body` size and printing the body. Should it log the error here on `warn`/`error`?

ibetitsmike: if the path is non-critical, ie. no estimated size doesn't result in unexpected behavior on the client - I'm ok with no logs; 

but if the consequence is that something breaks down the road - a warn would be nice for troubleshooting

EhabY: The two calls to `safeStringify` are for logging (estimating body size and printing the actual body if the log level is sufficient) so def non-critical. Could still be useful to log since I do not think it's normal that we pass a recursive object here (`response.data`/`request.data` should both be simple JSON objects or primitives)

src/logging/formatters.ts
```diff
@@ -66,8 +85,38 @@ export function formatHeaders(headers: Record<string, unknown>): string {
...
 }
+
+function safeStringify(data: unknown): string | null {
+	try {
+		const seen = new WeakSet();
+		return JSON.stringify(data, (_key, value) => {
+			// Handle circular references
+			if (typeof value === "object" && value !== null) {
+				if (seen.has(value)) {
+					return "[Circular]";
```
code-asher: Oh woah `JSON.stringify` does not handle circular references??  Huh yeah

```
> t = {}
> t.t = t
> JSON.stringify(t)
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    --- property 't' closes the circle
    at JSON.stringify (<anonymous>)
```

I wonder if we should use `util.inspect`?

```
> t.foo = "bar"
> t.fn = () => undefined
> t.sy = Symbol("foo")
> t.b = BigInt("0x1fffffffffffff")
> util.inspect(t, { breakLength: Infinity })
"<ref *1> { t: [Circular *1], foo: 'bar', fn: [Function (anonymous)], sy: Symbol(foo), b: 9007199254740991n }"

Comment (EhabY) [reply src/logging/formatters.ts]: Oooh I kind of like `util.inspect` since it can handle node objects well and doesn't break easily. Since we want to print the whole body (or estimate based on that), I can configure it as such:

```ts
try {
	return util.inspect(data, {
		showHidden: false,
		depth: Infinity,
		maxArrayLength: Infinity,
		maxStringLength: Infinity,
		breakLength: Infinity,
		compact: true,
		getters: false, // avoid side-effects
	});
} catch {
	// Should rarely happen but just in case
	return null;
}
```

Let me write some tests and see how it looks.

My only concern here is performance, if we get big objects and we have to estimate the size often then it might be slow... At least with the `body` log level we have to enable it to print it, here we always estimate

src/logging/formatters.ts
```diff
@@ -18,7 +21,7 @@ export function formatTime(ms: number): string {
 }
 
 export function formatMethod(method: string | undefined): string {
-	return (method ?? "GET").toUpperCase();
+	return (method ? method : "GET").toUpperCase();
```
code-asher: Could be `method || "GET"`

src/logging/formatters.ts
```diff
@@ -66,8 +75,25 @@ export function formatHeaders(headers: Record<string, unknown>): string {
...
+			showHidden: false,
+			depth: Infinity,
+			maxArrayLength: Infinity,
+			maxStringLength: Infinity,
+			breakLength: Infinity,
+			compact: true,
+			getters: false, // avoid side-effects
+		});
+	} catch {
+		// Should rarely happen but just in case
```
code-asher: Is this just to be defensive or are there known ways `util.inspect` can throw?

code-asher (approved):
Very nice!

src/logging/formatters.ts
code-asher: True, maybe we could get the size in another way, like iterate through the object and tally up the individual key and value sizes.  That would also fix how `undefined` would be printed as a string and take up bytes even though it really takes up zero bytes (I assume).  Actually I guess that would depend on whether the response literally returned `undefined` in the response or not.  Is there a way to get the raw unparsed response size maybe?  Should be a byte stream or array somewhere I would imagine, but not sure we have access to it.

src/logging/utils.ts
```diff
@@ -5,17 +5,25 @@ export function shortId(id: string): string {
...
+ * otherwise returns undefined (e.g., for plain objects and arrays that would require serialization).
+ */
 export function sizeOf(data: unknown): number | undefined {
 	if (data === null || data === undefined) {
 		return 0;
 	}
-	if (typeof data === "string") {
-		return Buffer.byteLength(data);
+	if (typeof data === "number" || typeof data === "boolean") {
+		return 8;
```
code-asher, line 25: Ah wait actually is a boolean eight bytes in v8?  I thought maybe it was 4 bytes

src/logging/formatters.ts
EhabY: Apparently it could still throw sometimes like if a getter throws, getters are not traversed here so it's more defensive I suppose

src/logging/utils.ts
EhabY: Both are kinda estimations and depend on the system but yes 4 is more accurate than 8. Changed

src/logging/formatters.ts
EhabY: Right, I keep confusing `??` (for undefined only) with `||` (all falsey  values)

code-asher: Undefined and null as well, I think!

src/logging/formatters.ts
EhabY: I tried to use https://www.npmjs.com/package/object-sizeof but it just returns an error in case of recursive objects.

> Node.js version uses the Buffer.from(objectToString) method to convert the object's string representation to a buffer, and then it uses the byteLength property to obtain the buffer size in bytes.

So this is very similar anyway.

> Is there a way to get the raw unparsed response size maybe?

I think this is the way, unparsed responses are always `string` and thus much simpler to get exact sizes, same for the request before sending it to the server. I've added axios request and response transformers that just read the raw response/request body and gets the size from that and seems to work well!

src/api/coderApi.ts
```diff
@@ -228,6 +251,66 @@ function addLoggingInterceptors(client: AxiosInstance, logger: Logger) {
...
+	}
+
+	const size = sizeOf(data);
+	if (size !== undefined) {
+		return size;
+	}
+
+	// Fallback
+	const stringified = serializeValue(data);
+	return stringified === null ? undefined : Buffer.byteLength(stringified);
```
EhabY: I think this is not needed and is not even executed while I was testing. Could be there just in case 🤔

src/logging/formatters.ts
code-asher: Oh woah that is amazing!!  Nice work figuring that out :tada:

src/api/coderApi.ts
```diff
@@ -228,6 +251,66 @@ function addLoggingInterceptors(client: AxiosInstance, logger: Logger) {
...
+	config: RequestConfigWithMeta,
+): AxiosResponseTransformer[] {
+	return [
+		(data: unknown, headers: AxiosHeaders) => {
+			const transformerArray = Array.isArray(transformer)
+				? transformer
+				: [transformer];
+
+			// Transform the request first then estimate the size
+			const result = transformerArray.reduce(
```
code-asher, line 265: Do we need to manually call the other transformer(s)?  Since these already take an array I assume they call them in sequence so we could do something like:

```typescript
config.transformRequest = [
  (data, headers) => {
    config.rawRequestSIze = getSize(headers, data);
    return data;
  },
  ...(config.transformRequest || client.defaults.transformResponse || []),
]
```

That way ours gets called first with the original data, we record the size, then it goes on through the regular transform chain.

src/api/coderApi.ts
```diff
@@ -228,6 +251,66 @@ function addLoggingInterceptors(client: AxiosInstance, logger: Logger) {
...
+	];
+}
+
+function getSize(headers: AxiosHeaders, data: unknown): number | undefined {
+	const contentLength = headers["content-length"];
+	if (contentLength !== undefined) {
+		return parseInt(contentLength, 10);
+	}
+
+	const size = sizeOf(data);
```
code-asher: You mentioned the raw data is a string right?  Could we just do something like:

```typescript
// This must be called before any other transforms to ensure we operate on the original string.
return typeof data === "string" ? Buffer.byteLength(data) : undefined;
```

That way, no need to maintain `sizeOf`

src/logging/utils.ts
```diff
@@ -36,6 +44,19 @@ export function sizeOf(data: unknown): number | undefined {
 	return undefined;
 }
 
-export function createRequestId(): string {
-	return crypto.randomUUID().replace(/-/g, "");
+export function serializeValue(data: unknown): string | null {
```
code-asher: I liked the `stringify` name better personally, to me serialize sounds like this should convert to some format like yaml or protobuf or whatever, while this is just the original format but in printable string form.

src/api/coderApi.ts
EhabY: For requests we have to call ours last because we want to measure what's sent over the wire, also the transformers here could be doing the serializing anyway. It's the opposite for the response, where we have to get the size first thing since that is also what's sent over the wire

src/api/coderApi.ts
EhabY: Apparently it could be other types, for example, for binary data it could be `ArrayBuffer`. `sizeOf` is also used to get the size of websocket data as well. I prefer keeping it since it just handles the primitives, buffers, and objects with `size` property (very fast and simple)

src/logging/utils.ts
EhabY: Changed back to `safeStringify`

src/api/coderApi.ts
EhabY: I removed the stringification so it's something like this:

```ts
function getSize(headers: AxiosHeaders, data: unknown): number | undefined {
	const contentLength = headers["content-length"];
	if (contentLength !== undefined) {
		return parseInt(contentLength, 10);
	}

	return sizeOf(data);
}
```

Where `sizeOf` handles primitive sizes, buffers, and objects with `size`

src/api/coderApi.ts
EhabY: Removed this for now

src/api/coderApi.ts
code-asher: Ohhhh right good point we would want the reverse for requests.  We can still avoid calling the transformers ourself right?  Not a huge deal either way, just slightly more idiomatic.

src/api/coderApi.ts
code-asher: Makes sense to me!  I like that.

PR #617, opened 2025-10-08
Fix unit tests on windows

### LINKED ISSUE #604: Fix unit tests on Windows
It seems that some unit tests fail on Windows due to a few reasons:

1. The path separator is `\` instead of `/`, this applies to a lot of tests that might use `path.normalize` or `path.join`
2. The escaping of environment variables uses `'` instead of `"`
3. Some tests in `headers.test.ts` for some reason (executes a bash script perhaps?)
4. ~~`cliManager.test.ts`: `validates downloaded binary version` seems to be flakey on Windows, might have to do with the mock stream not flushing the content fully (expects `mock-binary-v1.2.3` but actual is `mock-bin` and sometimes `mock-binary-v1.2`)~~ - Fixed in https://github.com/coder/vscode-coder/pull/607/commits/4d9345620c7efce9f3f4d9f54a7263f5d4143883

EhabY:
Closes #604

### DIFF
--- a/test/utils/platform.ts
+++ b/test/utils/platform.ts
@@ -2,7 +2,9 @@ import os from "node:os";
 import path from "node:path";
 import { expect } from "vitest";
 
-export const isWindows = os.platform() === "win32";
+export function isWindows(): boolean {
+	return os.platform() === "win32";
+}
 
 /**
  * Returns a platform-independent command that outputs the given text.
@@ -39,15 +41,6 @@ export function expectPathsEqual(actual: string, expected: string) {
 	expect(normalizePath(actual)).toBe(normalizePath(expected));
 }
 
-export function quoteCommand(value: string): string {
-	const quote = getPlatformQuote();
-	return `${quote}${value}${quote}`;
-}
-
 function normalizePath(p: string): string {
-	return p.replaceAll(path.sep, "/");
-}
-
-function getPlatformQuote(): string {
-	return isWindows ? '"' : "'";
+	return p.replaceAll(path.sep, path.posix.sep);
 }
--- a/test/unit/core/cliUtils.test.ts
+++ b/test/unit/core/cliUtils.test.ts
@@ -29,7 +29,7 @@ describe("CliUtils", () => {
 		expect((await cliUtils.stat(binPath))?.size).toBe(4);
 	});
 
-	it.skipIf(isWindows)("version", async () => {
+	it.skipIf(isWindows())("version", async () => {
 		const binPath = path.join(tmp, "version");
 		await expect(cliUtils.version(binPath)).rejects.toThrow("ENOENT");
 
--- a/test/unit/globalFlags.test.ts
+++ b/test/unit/globalFlags.test.ts
@@ -3,7 +3,7 @@ import { type WorkspaceConfiguration } from "vscode";
 
 import { getGlobalFlags } from "@/globalFlags";
 
-import { quoteCommand } from "../utils/platform";
+import { isWindows } from "../utils/platform";
 
 describe("Global flags suite", () => {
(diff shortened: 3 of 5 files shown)


test/utils/platform.ts
```diff
@@ -0,0 +1,53 @@
...
+export function printCommand(output: string): string {
+	const escaped = output
+		.replace(/\\/g, "\\\\") // Escape backslashes first
+		.replace(/'/g, "\\'") // Escape single quotes
+		.replace(/"/g, '\\"') // Escape double quotes
+		.replace(/\r/g, "\\r") // Preserve carriage returns
+		.replace(/\n/g, "\\n"); // Preserve newlines
+
+	return `node -e "process.stdout.write('${escaped}')"`;
+}
```
mtojek, line 22: Hm... I know it is mainly test helper, but should we add a few unit tests to make sure it is escaping correctly?

EDIT:

actually, that comment applies to all exported functions

test/utils/platform.ts
```diff
@@ -0,0 +1,53 @@
+import os from "node:os";
+import path from "node:path";
+import { expect } from "vitest";
+
+export const isWindows = os.platform() === "win32";
```
mtojek: nit: no win64 :) ?

mafredri: 64-bit Windows is a lie, the proof is in the pudding: https://nodejs.org/api/os.html#osplatform

😆

mtojek: wat? TIL!

test/utils/platform.ts
```diff
@@ -0,0 +1,53 @@
...
+	expect(normalizePath(actual)).toBe(normalizePath(expected));
+}
+
+export function quoteCommand(value: string): string {
+	const quote = getPlatformQuote();
+	return `${quote}${value}${quote}`;
+}
+
+function normalizePath(p: string): string {
+	return p.replaceAll(path.sep, "/");
```
mafredri:
```suggestion
	return p.replaceAll(path.sep, path.posix.sep);
```
Just a minor suggestion, feel free to omit.

test/utils/platform.ts
```diff
@@ -0,0 +1,53 @@
...
+export function printEnvCommand(key: string, varName: string): string {
+	return `node -e "process.stdout.write('${key}=' + process.env.${varName})"`;
+}
+
+export function expectPathsEqual(actual: string, expected: string) {
+	expect(normalizePath(actual)).toBe(normalizePath(expected));
+}
+
+export function quoteCommand(value: string): string {
+	const quote = getPlatformQuote();
```
mafredri: This function doesn't seem to perform safe quoting and it's not immediately clear why platform quote is needed (perhaps a comment would be good?).

Perhaps a module like [shell-quote](https://www.npmjs.com/package/shell-quote) could be useful here?

EhabY: I can see why this is misleading but this is to test what our code does in prod, currently it escapes using `"` vs. `'` for Linux or Windows. I'll add a comment explaining this weird escaping behavior :sweat:

test/utils/platform.ts
EhabY: There is only one Windows and it's 32-bit :triumph:!

test/utils/platform.ts
EhabY: I added a comment and moved this into the `globalFlags.test.ts` so it's not as confusing

test/utils/platform.ts
EhabY: Ah nice, done!

test/utils/platform.ts
EhabY: Added some simple unit tests

PR #619, opened 2025-10-08
Pass the header command output to WebSocket creation

### LINKED ISSUE #618: Propagate `header-command` output to WebSocket creation
> We are behind an Azure App Proxy, WAF and a firewall. This means we are using **--header-command** to inject the needed token. This warning is consistent and does not go away when the VSCode connection is reloaded and pops up immediately .
> 
> Currently we are a bit behind regarding the server version. We are using **v2.22.1**.
> 
> There were also some undefined **error** logs:
> 
> ```
> 2025-10-08 10:02:04.127 [error] Unexpected server response: 200
> ```
> 
> Let me know if I can help you further 

 _Originally posted by @MrPeacockNLB in [#595](https://github.com/coder/vscode-coder/issues/595#issuecomment-3381118553)_

EhabY:
Closes #618

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixed
+
+- Fixed WebSocket connections not receiving headers from the configured header command
+  (`coder.headerCommand`), which could cause authentication failures with remote workspaces.
+
 ## [v1.11.2](https://github.com/coder/vscode-coder/releases/tag/v1.11.2) 2025-10-07
 
 ### Changed


src/api/coderApi.ts
```diff
@@ -127,14 +128,23 @@ export class CoderApi extends Api {
...
 			location: baseUrl,
 			...configs,
 			options: {
 				agent: httpAgent,
 				followRedirects: true,
 				headers: {
+					...headers,
 					...(token ? { [coderSessionTokenHeader]: token } : {}),
 					...configs.options?.headers,
 				},
```
EhabY, line 150: Does this order make sense? (From lowest to highest)

* Header args
* Token taken from the axios instance
* Explicitly passed headers

src/api/coderApi.ts
```diff
@@ -127,14 +128,23 @@ export class CoderApi extends Api {
 			coderSessionTokenHeader
 		] as string | undefined;
 
-		const httpAgent = createHttpAgent(vscode.workspace.getConfiguration());
+		const headers = await getHeaders(
+			baseUrlRaw,
+			getHeaderCommand(vscode.workspace.getConfiguration()),
+			this.output,
+		);
```
EhabY, line 135: Actual fix, we now read the headers and pass them to the websocket. The rest of the changes just propagate the non-async -> async change here

src/remote/remote.ts
```diff
@@ -135,9 +135,7 @@ export class Remote {
 		let attempts = 0;
 
 		function initWriteEmitterAndTerminal(): vscode.EventEmitter<string> {
-			if (!writeEmitter) {
-				writeEmitter = new vscode.EventEmitter<string>();
-			}
+			writeEmitter ??= new vscode.EventEmitter<string>();
```
code-asher, line 138: Ooo I like it, I forgot this construct existed

src/api/coderApi.ts
code-asher: I think this makes sense.

src/api/utils.ts
```diff
@@ -32,6 +34,12 @@ export function createHttpAgent(cfg: WorkspaceConfiguration): ProxyAgent {
 	const caFile = expandPath(String(cfg.get("coder.tlsCaFile") ?? "").trim());
 	const altHost = expandPath(String(cfg.get("coder.tlsAltHost") ?? "").trim());
 
+	const [cert, key, ca] = await Promise.all([
```
code-asher, line 37: nice 🤌

src/api/coderApi.ts
EhabY: After re-reading the original issue:

> We are behind an Azure App Proxy, WAF and a firewall. This means we are using --header-command to inject the needed token.

Would this mean we want the headers to have a higher priority over the token from the axios instance :thinking: ?

From highest to lowest:

* Explicit headers
* Header args (from the command)
* Implicitly derived headers like token from the axios instance

src/remote/remote.ts
EhabY: Haha same, SonarQube offers these tips and tricks by default :)!

src/api/coderApi.ts
EhabY: I just checked the RESTful client behavior and it's the following (highest to lowest):

* Header args (from the command)
* Explicitly passed headers (options)
* Implicitly derived headers like token from the axios instance (defaults)

I think I'll make the WebSocket follow the same pattern for consistency since the RESTful API has been used for so long.

EhabY: Done

code-asher: Oh yeah I think it was that way because we know the token is good, so no point in allowing the header command to override it.

On the other hand, having the ability to override might be nice, but not sure when someone would use that.  :thinking:   Keeping the status quo makes sense to me.

PR #623, opened 2025-10-13
Add SSE fallback to some one way WS connections

### LINKED ISSUE #620: Add SSE fallback for the new WebSocket API endpoints
In https://github.com/coder/vscode-coder/pull/580, we got rid of the two SSE connections (workspace `watch` and agent `watch-metadata`). However, this requires users to update their Coder Server to `>=2.22.0`  or otherwise would fail with 404. We should instead fallback to SSE when we get a 404 for those two requests so users don't get broken watching behavior simply because they use a new VS Code extension.

EhabY:
Add SSE fallback to some WS connections:

* `/api/v2/workspaces/${workspace.id}/watch-ws` -> `/api/v2/workspaceagents/${agentId}/watch-metadata`
* `/api/v2/workspaceagents/${agentId}/watch-metadata-ws` -> `/api/v2/workspaceagents/${agentId}/watch-metadata`

Restored the previous code regarding `createStreamingFetchAdapter` to stream in SSE events.

* Implemented a unified interface for WS and SSE to be similar to the `OneWayWebSocket`.
* Added unified logging for WS and SSE.
* Fixed issue with headers order precedence
* Add tests for `CoderApi` 

Closes #620

### DIFF
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -144,6 +144,12 @@ export class CoderApi extends Api {
 			vscode.workspace.getConfiguration(),
 		);
 
+		/**
+		 * Similar to the REST client, we want to prioritize headers in this order (highest to lowest):
+		 * 1. Headers from the header command
+		 * 2. Any headers passed directly to this function
+		 * 3. Coder session token from the Api client (if set)
+		 */
 		const headers = {
 			...(token ? { [coderSessionTokenHeader]: token } : {}),
 			...configs.options?.headers,
@@ -189,7 +195,9 @@ export class CoderApi extends Api {
 	}
 
 	/**
-	 * Create a WebSocket connection with SSE fallback on 404
+	 * Create a WebSocket connection with SSE fallback on 404.
+	 *
+	 * Note: The fallback on SSE ignores all passed client options except the headers.
 	 */
 	private async createWebSocketWithFallback<TData = unknown>(configs: {
 		apiRoute: string;
@@ -209,6 +217,7 @@ export class CoderApi extends Api {
 			return this.createSseFallback<TData>(
 				configs.fallbackApiRoute,
 				configs.searchParams,
+				configs.options?.headers,
 			);
 		}
 
@@ -216,6 +225,7 @@ export class CoderApi extends Api {
 			this.createSseFallback<TData>(
 				configs.fallbackApiRoute,
 				configs.searchParams,
+				configs.options?.headers,
 			),
 		);
 	}
@@ -260,6 +270,7 @@ export class CoderApi extends Api {
 	private async createSseFallback<TData = unknown>(
 		apiRoute: string,
 		searchParams?: Record<string, string> | URLSearchParams,
+		optionsHeaders?: Record<string, string>,
 	): Promise<UnidirectionalStream<TData>> {
 		this.output.warn(`WebSocket failed, using SSE fallback: ${apiRoute}`);
 
@@ -274,6 +285,8 @@ export class CoderApi extends Api {
 			apiRoute,
 			searchParams,
 			axiosInstance: this.getAxiosInstance(),
+			optionsHeaders: optionsHeaders,
+			logger: this.output,
 		});
 
 		this.attachStreamLogger(sseConnection);
--- a/src/api/streamingFetchAdapter.ts
+++ b/src/api/streamingFetchAdapter.ts
@@ -8,6 +8,7 @@ import { type IncomingMessage } from "http";
  */
 export function createStreamingFetchAdapter(
 	axiosInstance: AxiosInstance,
+	configHeaders?: Record<string, string>,
 ): (url: string | URL, init?: FetchLikeInit) => Promise<FetchLikeResponse> {
 	return async (
 		url: string | URL,
@@ -18,7 +19,7 @@ export function createStreamingFetchAdapter(
 		const response = await axiosInstance.request<IncomingMessage>({
 			url: urlStr,
 			signal: init?.signal,
-			headers: init?.headers,
+			headers: { ...init?.headers, ...configHeaders },
 			responseType: "stream",
 			validateStatus: () => true, // Don't throw on any status code
 		});
@@ -26,11 +27,19 @@ export function createStreamingFetchAdapter(
 		const stream = new ReadableStream({
 			start(controller) {
 				response.data.on("data", (chunk: Buffer) => {
-					controller.enqueue(chunk);
+					try {
+						controller.enqueue(chunk);
+					} catch {
+						// Stream already closed or errored, ignore
+					}
 				});
 
 				response.data.on("end", () => {
-					controller.close();
+					try {
+						controller.close();
+					} catch {
+						// Stream already closed, ignore
+					}
 				});
 
 				response.data.on("error", (err: Error) => {
--- a/src/websocket/sseConnection.ts
+++ b/src/websocket/sseConnection.ts
@@ -4,6 +4,7 @@ import { type WebSocketEventType } from "coder/site/src/utils/OneWayWebSocket";
 import { EventSource } from "eventsource";
 
 import { createStreamingFetchAdapter } from "../api/streamingFetchAdapter";
+import { type Logger } from "../logging/logger";
 
 import { getQueryString } from "./utils";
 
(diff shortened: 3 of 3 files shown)


src/api/streamingFetchAdapter.ts
```diff
@@ -0,0 +1,62 @@
+import { type AxiosInstance } from "axios";
+import { type FetchLikeInit, type FetchLikeResponse } from "eventsource";
+import { type IncomingMessage } from "http";
+
+/**
+ * Creates a fetch adapter using an Axios instance that returns streaming responses.
+ * This is used by EventSource to make authenticated SSE connections.
+ */
+export function createStreamingFetchAdapter(
```
EhabY, line 9: This was deleted and has now been restored as is. See https://github.com/coder/vscode-coder/blob/v1.10.1/src/api.ts#L120

src/websocket/sseConnection.ts
```diff
@@ -0,0 +1,191 @@
+import { type AxiosInstance } from "axios";
```
EhabY, line 1: This is not a websocket but it's only used as a websocket fallback and implements the same interfaces so that's why it's in this directory. Eventually, it'll be removed once we decide to not support older Coder servers

src/api/coderApi.ts
```diff
@@ -137,43 +143,141 @@ export class CoderApi extends Api {
...
+	}): Promise<UnidirectionalStream<TData>> {
+		let webSocket: OneWayWebSocket<TData>;
+		try {
+			webSocket = await this.createWebSocket<TData>({
+				apiRoute: configs.apiRoute,
+				searchParams: configs.searchParams,
+				options: configs.options,
+			});
+		} catch {
+			// Failed to create WebSocket, use SSE fallback
```
mtojek, line 216: Do you think there should be a warning when switching to the fallback option?

src/api/coderApi.ts
```diff
@@ -137,43 +143,141 @@ export class CoderApi extends Api {
...
+
+			const handleError = (event: ErrorEvent) => {
+				cleanup();
+				const is404 =
+					event.message?.includes("404") ||
+					event.error?.message?.includes("404");
+
+				if (is404 && onNotFound) {
+					connection.close();
+					onNotFound().then(resolve).catch(reject);
```
mtojek, line 256: is there a risk that `is404` is true, but `onNotFound` promise won't be resolved yet?

src/workspace/workspaceMonitor.ts
```diff
@@ -93,10 +93,12 @@ export class WorkspaceMonitor implements vscode.Disposable {
...
-				const newWorkspaceData = event.parsedMessage.data as Workspace;
-				monitor.update(newWorkspaceData);
-				monitor.maybeNotify(newWorkspaceData);
-				monitor.onChange.fire(newWorkspaceData);
+				const newWorkspaceData = event.parsedMessage.data as Workspace | null;
+				if (newWorkspaceData) {
+					monitor.update(newWorkspaceData);
+					monitor.maybeNotify(newWorkspaceData);
+					monitor.onChange.fire(newWorkspaceData);
+				}
```
mtojek, line 101: should it warn on `newWorkspaceData` being null?

src/api/coderApi.ts
```diff
@@ -137,43 +143,141 @@ export class CoderApi extends Api {
 		const httpAgent = await createHttpAgent(
 			vscode.workspace.getConfiguration(),
 		);
+
+		const headers = {
```
mtojek, line 153: It would be beneficial to add a short comment justifying the order of headers

src/api/coderApi.ts
```diff
@@ -84,8 +86,9 @@ export class CoderApi extends Api {
 	};
 
 	watchWorkspace = async (workspace: Workspace, options?: ClientOptions) => {
-		return this.createWebSocket<ServerSentEvent>({
+		return this.createWebSocketWithFallback<ServerSentEvent>({
```
mtojek, line 89: should we pass/process ClientOptions in the fallback or is it not necessary?

src/websocket/sseConnection.ts
```diff
@@ -0,0 +1,191 @@
...
+					callback as EventHandler<ServerSentEvent, "open">,
+				);
+				break;
+		}
+	}
+
+	private parseMessage(
+		event: MessageEvent,
+	): ParsedMessageEvent<ServerSentEvent> {
+		const wsEvent = { data: event.data } as WsMessageEvent;
```
mtojek, line 162: can `event.data` or `event` be null?

src/websocket/sseConnection.ts
```diff
@@ -0,0 +1,191 @@
...
+				this.callbacks.open.delete(
+					callback as EventHandler<ServerSentEvent, "open">,
+				);
+				break;
+		}
+	}
+
+	public close(code?: number, reason?: string): void {
+		this.eventSource.close();
+		this.callbacks.close.forEach((cb) =>
```
mtojek: is it an issue if any of `cb(...)` throws an error?

src/api/streamingFetchAdapter.ts
```diff
@@ -0,0 +1,62 @@
...
+			signal: init?.signal,
+			headers: init?.headers,
+			responseType: "stream",
+			validateStatus: () => true, // Don't throw on any status code
+		});
+
+		const stream = new ReadableStream({
+			start(controller) {
+				response.data.on("data", (chunk: Buffer) => {
+					controller.enqueue(chunk);
```
mtojek: any error handling?

src/api/streamingFetchAdapter.ts
```diff
@@ -0,0 +1,62 @@
...
+				});
+
+				response.data.on("end", () => {
+					controller.close();
```
mtojek: any error handling?

src/workspace/workspaceMonitor.ts
EhabY: In some cases, it can be `null` because the `ServerSentEvent` sends a `ping`, I would say that's expected in those cases. Note that if this is `null` it means the request was in the form of `{ type: "data" | "ping", data: null }` so would be deliberate since errors are handled differently

src/api/coderApi.ts
EhabY: Oh, `onNotFound` is not a promise but a callback that returns a promise. We only execute it if `is404 === true`

src/api/coderApi.ts
EhabY: We do that already in `createSseFallback`: 

```ts
this.output.warn(`WebSocket failed, using SSE fallback: ${apiRoute}`);
```

src/api/coderApi.ts
EhabY: True.. There are tests that document this behavior now for both WS  and HTTP requests

src/api/coderApi.ts
EhabY: `ClientOptions` are defined on the WS level, while SSE events just go through the same RESTful API through the Axios instance. I do not think we can pass info from there :thinking:

src/websocket/sseConnection.ts
EhabY: `event` itself cannot be `null`, I suppose there is nothing preventing the `data` from being null here if the server decides to send `null`. It shouldn't break anything though

src/api/coderApi.ts
EhabY: Done

src/api/coderApi.ts
EhabY: Actually, I could pass the headers at least, but the rest can be discarded. I've passed the headers in the same order as the headers for all other requests.

src/websocket/sseConnection.ts
EhabY: Oooh nice catch, all those callbacks could throw and lead to other callbacks to not be called (or clean up to not happen correct). I've surrounded it with `try/catch` and logged the error on the error level.

src/api/streamingFetchAdapter.ts
EhabY: I added a simple `try/catch` and ignored the error this can only throw if an error already occurred or if it was closed (which would be logged/thrown elsewhere AFAIK)

src/api/streamingFetchAdapter.ts
EhabY: Same here

mtojek (approved):
LGTM, but I didn't dive too deep. I don't see anything concerning, but it might be feasible to ask @code-asher for another look.

PR #624, opened 2025-10-14
Automatically publish extension and add pre-release version

### LINKED ISSUE #97: Automate extension releases to VS Code Marketplace and Open VSX Regsitry
Currently, I believe GitHub release artifacts have to be manually submitted to the VS Code Marketplace. Can we automate this so they are always in sync?

EhabY:
~~- Automatically publish a pre-release version on every merge to main with the following versioning: `MAJOR.(MINOR+1).YYYYMMDDNN`. It means we can have up to 100 releases (merges to main) in a day. This is similar to how VS Code's Remote SSH does preview releases.~~
- Automatically publish the pre-release when pushing a `v*-pre` tag to `main`.
- Automatically publish the stable release when pushing a `v*` tag to `main`.
- Both release workflows use the same `publish-extension.yaml` workflow for all publishing tasks (VS Marketplace, Open VSX, and GitHub releases), with the `isPreRelease` flag controlling the behavior.
- Package VSIX on every PR and merge to main. For PRs, it is retained for 7 days, but for main it's retained indefinitely.

## Notes
- Set `VSCE_PAT` and `OVSX_PAT` secrets in repo, if not set, then the publishing will be skipped.

Closes #97

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,7 @@ jobs:
       - uses: actions/setup-node@v6
         with:
           node-version: "22"
+          cache: "yarn"
 
       - run: yarn
 
@@ -39,6 +40,7 @@ jobs:
       - uses: actions/setup-node@v6
         with:
           node-version: "22"
+          cache: "yarn"
 
       - run: yarn
 
@@ -54,6 +56,7 @@ jobs:
       - uses: actions/setup-node@v6
         with:
           node-version: "22"
+          cache: "yarn"
 
       - name: Install dependencies
         run: |


.github/workflows/ci.yaml
```diff
@@ -41,3 +41,58 @@ jobs:
...
+          name: extension-pr-${{ github.event.pull_request.number }}
+          path: ${{ steps.setup.outputs.packageName }}
+          if-no-files-found: error
+          retention-days: 7
+
+      - name: Upload artifact (main)
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        uses: actions/upload-artifact@v4
+        with:
+          name: extension-main-${{ github.sha }}
```
EhabY, line 98: Could use better artifact name, does Coder have a standard for this?

code-asher: No standard that I know of.  Do we need to name this differently?  We could just upload the vsix as it is already named.  I like the coder-remote-$version-$sha.vsix scheme you have, we can use that name on all uploads IMO.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,67 @@
...
+          npm install -g @vscode/vsce
+
+      - name: Extract version from tag
+        id: version
+        run: |
+          # Extract version from tag (remove 'v' prefix and '-pre' suffix)
+          TAG_NAME=${GITHUB_REF#refs/tags/v}
+          VERSION=${TAG_NAME%-pre}
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Pre-release version: $VERSION"
```
code-asher, line 32: Had a thought, what if we check the `package.json` version and abort if it does not match?  Just to prevent accidentally tagging something without first updating the `package.json`.  (And same in the release workflow.)

.github/workflows/publish-extension.yaml
```diff
@@ -0,0 +1,97 @@
...
+        run: |
+          echo "Publishing version ${{ inputs.version }} to Open VSX"
+          if [ "${{ inputs.isPreRelease }}" = "true" ]; then
+            ovsx publish "./${{ inputs.packageName }}" --pre-release -p ${{ secrets.OVSX_PAT }}
+          else
+            ovsx publish "./${{ inputs.packageName }}" -p ${{ secrets.OVSX_PAT }}
+          fi
+
+  publishGH:
+    name: Create GitHub ${{ inputs.isPreRelease && 'Pre-' || '' }}Release
```
code-asher, line 110: Sooo....I know I said I liked the idea of symmetrical releases but since we might have to duplicate pre-releases to work around VS Code's jank I wonder if we only publish previews to the marketplaces and skip them on GH to keep GH free of dupes at least.  Thoughts?

.github/workflows/publish-extension.yaml
```diff
@@ -0,0 +1,97 @@
...
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+        description: "Version to publish"
+      packageName:
+        required: true
+        type: string
+        description: "Package filename"
```
code-asher: Could we derive the package name from the version and pre-release inputs?  I doubt this would actually happen, but it could be weird if you are able to pass `1.1.2` and false with the package name `extension-5.4.3-pre` for example.

.github/workflows/ci.yaml
```diff
@@ -41,3 +41,58 @@ jobs:
...
 
       - run: yarn test:ci
+
+  package:
+    name: Package
+    runs-on: ubuntu-22.04
+    needs: [lint, test]
+    outputs:
+      packageName: ${{ steps.setup.outputs.packageName }}
+      version: ${{ steps.version.outputs.version }}
```
code-asher: I think these outputs are unused.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: extension-${{ steps.version.outputs.version }}
+          path: ${{ steps.setup.outputs.packageName }}
+          if-no-files-found: error
+
+  publish:
+    name: Publish Extension and Create Pre-Release
+    needs: package
+    uses: ./.github/workflows/publish-extension.yaml
```
code-asher, line 72: Oh neat I did not know about `uses`

code-asher (commented):
Awesome!

.github/workflows/ci.yaml
EhabY: This is the artifact name, apparently we cannot have individual files as artifacts (you can checkout the PR artifact here as well)

.github/workflows/pre-release.yaml
EhabY: Ooooh yes that could be nice actually, the pattern should be as followed:

Say we want to release `v1.0.0`: 
* Bump `package.json` version to `1.0.0`
* Push `v1.0.0` tag
* Bump `package.json` version to `1.0.1`  so that all VSIXs built on the PR/main CI have a higher version (we can skip this of course but then it would maintain the same version which is confusing)

.github/workflows/ci.yaml
EhabY: Oh right, removed!

.github/workflows/publish-extension.yaml
EhabY: We could yes but that means `publish-extension.yaml` has to be aware of the format of the package name, right now it just takes the name as is so there's no implicit info that is deduced.

.github/workflows/pre-release.yaml
EhabY: ikr :recycle:

.github/workflows/publish-extension.yaml
EhabY: I was mistaken because of some caching issues, updating works fine from `pre-release` -> `release` -> `pre-release` (if I'm on the pre-release channel)

.github/workflows/ci.yaml
code-asher: Ah right I forgot it forces you to zip it up.

.github/workflows/publish-extension.yaml
code-asher: If we derive the name, we make it impossible for the template to be used incorrectly.  It seems bad that the template allows you to publish a mismatched version and package (or pre-release a non-prerelease, etc).

We could do it the other way, extract the version and whether it is a pre-release from the package.  But, if we ever have to kick this off manually, it would be nicer to type in the version to publish and check a box rather than type out the full package name.

EhabY: Yeah fair enough, didn't think of manually kicking it off but could be nice. Anyway constructed the package name from the version and flag!

PR #625, opened 2025-10-17
Add SSE Connection and streamingFetchAdapter tests

EhabY:


src/websocket/sseConnection.ts
```diff
@@ -109,11 +109,10 @@ export class SseConnection implements UnidirectionalStream<ServerSentEvent> {
...
 	private createErrorEvent(event: Event | ErrorEvent): WsErrorEvent {
-		const errorMessage =
-			event instanceof ErrorEvent && event.message
-				? event.message
-				: "SSE connection error";
-		const error = event instanceof ErrorEvent ? event.error : undefined;
+		// Check for properties instead of instanceof to avoid browser-only ErrorEvent global
+		const eventWithMessage = event as { message?: string; error?: unknown };
+		const errorMessage = eventWithMessage.message || "SSE connection error";
+		const error = eventWithMessage.error;
```
buenos-nachos, line 115: Granted, I'm not super familiar with all the conventions of non-browser JS, but I guess I'm trying to understand the setup we have right now. It sounds like the `ErrorEvent` class doesn't exist at runtime (which would be in line with normal Node.js), yet we're still receiving values that are structured like that. 

Do we know where those values are coming from? I would understand if we were receiving anonymous objects that just happened to have those properties defined on them, but it seems like the runtime `instanceof` check was working fine before. `instanceof` is a nominal typing check, so the runtime class had to have come from somewhere

I think your changes are probably the safest way to address a mismatch that we can't control, but I can't help but wonder if there's a "deeper" fix we can make

buenos-nachos: From looking at the rest of the code, it looks like the `ErrorEvent` is getting generated via an `EventSource` instance, and that isn't necessarily different from the built-in browser one. But then, even though the callback for `eventSource.addEventListener("error", ...)` only receives `ErrorEvent` instances, we broadened it to accept generic events as well?

test/unit/api/streamingFetchAdapter.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+describe("createStreamingFetchAdapter", () => {
+	let mockAxios: AxiosInstance;
+
+	beforeEach(() => {
+		vi.resetAllMocks();
+		mockAxios = {
+			request: vi.fn(),
+		} as unknown as AxiosInstance;
+	});
```
buenos-nachos: Do we want this? This setup doesn't seem especially complex, and we're not using any Vitest helpers to handle the mocking, so by definition, all of the test cases lose isolation

I think I'd rather have the initialization happen individually for each test case. If that setup process is involved enough, we can make a helper function to produce a new `AxiosInstance`. I think it's okay to only implement the functions we need for the tests for the moment, but at some point, I think it'd be good to have a full mock implementation that doesn't rely on type assertions

But especially in Vitest, not having test isolation means that the tests are no longer thread-safe in Vitest's Go runtime, so we can't do concurrent tests via this:

```ts
describe.concurrent("Main description", () => {
  // In concurrent mode, we have to to use the expect that's
  // scoped to the specific test run
  it("does something", async ({ expect }) => {
    const mockAxios = createMockAxiosInstance();
    // Rest of the setup and assertions go here
  });
});
```

test/unit/api/streamingFetchAdapter.test.ts
```diff
@@ -0,0 +1,214 @@
...
+			expect(mockAxios.request).toHaveBeenCalledWith({
+				url: TEST_URL,
+				signal, // correctly passes signal
+				headers: {},
+				responseType: "stream",
+				validateStatus: expect.any(Function),
+			});
+		});
+
+		it("applies headers in correct precedence order (config > init)", async () => {
```
buenos-nachos: Nit: I think there's a little bit of ambiguity around the `>`. It _seems_ like it's being used in the "greater than" context, but that wasn't obvious to me at first. I initially read it as "init overrides config because it comes later, and we're using an arrow to describe that relationship"

Maybe updating the description to say `(config overrides init)` would be better?

test/unit/api/streamingFetchAdapter.test.ts
```diff
@@ -0,0 +1,214 @@
...
+				}),
+			);
+
+			// Test 3: Config headers override init headers
+			const adapter3 = createStreamingFetchAdapter(mockAxios, {
+				"X-Header": "config-value",
+			});
+			await adapter3(TEST_URL, {
+				headers: { "X-Header": "init-value" },
+			});
```
buenos-nachos, line 72: Maybe I'm missing some context for why we're doing things this way, but I guess I'm surprised that something specified later is getting overridden by something specified when the adaptor is first created

I would've expected the later ones to be persisted, or at least for them to be merged in some way. I feel like I would run the risk of having values get unexpectedly dropped when I'm using the adaptor

test/unit/api/streamingFetchAdapter.test.ts
```diff
@@ -0,0 +1,214 @@
...
+				mockStream,
+			);
+
+			const adapter = createStreamingFetchAdapter(mockAxios);
+			const response = await adapter(TEST_URL);
+
+			expect(response.url).toBe(TEST_URL);
+			expect(response.status).toBe(200);
+			expect(response.headers.get("content-type")).toBe("text/event-stream");
+			expect(response.headers.get("CoNtEnT-TyPe")).toBe("text/event-stream");
```
buenos-nachos, line 100: I feel like we could have a comment explaining what kind of key aliasing we expect to happen here. Without looking at the source code, it's not clear what kinds of transformations happen

And for a test like this, I feel like ideally the case should function as a form of documentation, and shouldn't require that someone go through the code

test/unit/api/streamingFetchAdapter.test.ts
```diff
@@ -0,0 +1,214 @@
...
+			expect(response.body?.getReader).toBeDefined();
+		});
+
+		it("detects redirected requests", async () => {
+			const mockStream = createMockStream();
+			const mockResponse = {
```
buenos-nachos, line 107: Is there a type annotation we can add here? It's not immediately clear to me how this is supposed to be structured, but it seems like it might be a `FetchLikeResponse`?

test/unit/api/streamingFetchAdapter.test.ts
```diff
@@ -0,0 +1,214 @@
...
+	async function setupReaderTest(): Promise<{
+		mockStream: IncomingMessage;
+		reader: ReaderLike | ReadableStreamDefaultReader<Uint8Array<ArrayBuffer>>;
+	}> {
+		const mockStream = createMockStream();
+		setupAxiosResponse(mockAxios, 200, {}, mockStream);
+
+		const adapter = createStreamingFetchAdapter(mockAxios);
+		const response = await adapter(TEST_URL);
+		const reader = response.body!.getReader();
```
buenos-nachos: I think definite assignment operators like this are really dicey, and especially for a test, it's not worth silencing the LSP, especially as this file grows over time. I feel like a better option would be to do a runtime check for `getReader`:

```ts
const reader = response.body?.getReader();
if (reader === undefined) {
	throw new Error("Reader is not defined")
}

return { mockStream, reader };
```

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+
+describe("SseConnection", () => {
+	let mockAxios: AxiosInstance;
+	let mockLogger: ReturnType<typeof createMockLogger>;
+
+	beforeEach(() => {
+		vi.resetAllMocks();
+		mockAxios = axios.create();
+		mockLogger = createMockLogger();
+	});
```
buenos-nachos: Similar concern here. I feel like it'd be better to move the values into the test cases themselves, and maximize isolation

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+
+	describe("URL Building", () => {
+		it.each([
+			[undefined, `${TEST_URL}${API_ROUTE}`],
+			[
+				{ follow: "true", after: "123" },
+				`${TEST_URL}${API_ROUTE}?follow=true&after=123`,
+			],
+			[new URLSearchParams({ foo: "bar" }), `${TEST_URL}${API_ROUTE}?foo=bar`],
+		])("constructs URL with %s search params", (searchParams, expectedUrl) => {
```
buenos-nachos: Just because this test case is so polymorphic, I feel like it'd be good to add a specific type to make the code easier to read during review:

```ts
type TestEntry = readonly [
  params: undefined | URLSearchParams | Record<string, string>,
  url: string,
];
it.each<TestEntry>( /* Rest of the code is the same as before */)
```

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+					if (event === "open") {
+						setImmediate(() => handler(new Event("open")));
+					}
+				}),
+			});
+			setupEventSourceMock(mockES);
+
+			const connection = createConnection();
+			const events1: unknown[] = [];
+			const events2: unknown[] = [];
```
buenos-nachos: Do these need to be typed as `unknown`?

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+		});
+
+		it("fires close event when connection closes on error", async () => {
+			const mockES = createMockEventSource({
+				addEventListener: vi.fn((event, handler) => {
+					if (event === "error") {
+						setImmediate(() => {
+							// Simulate EventSource behavior: state transitions to CLOSED when error occurs
+							(mockES as { readyState: number }).readyState =
+								EventSource.CLOSED;
```
buenos-nachos: I get why we have this (`readyState` is readonly), but it feels a little funky to me. Could we split this up across multiple lines, and have a comment explaining why we need the type assertion?

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+				undefined,
+				undefined,
+				{ code: 1000, reason: "Normal closure", wasClean: true },
+			],
+			[
+				4000,
+				"Custom close",
+				{ code: 4000, reason: "Custom close", wasClean: true },
+			],
+		])(
```
buenos-nachos: Similar feedback here:
```ts
type TestEntry = readonly [
	code: number | undefined,
	reason: string | undefined,
	event: Partial<CloseEvent>
];
it.each<TestEntry>()
```

test/unit/websocket/sseConnection.test.ts
buenos-nachos: Type aliases can be placed anywhere, so I think it'd make the most sense to put it right above the `it.each` call

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+		it.each([
+			["open", new Event("open")],
+			["message", new MessageEvent("data", { data: '{"test": true}' })],
+			["error", new Event("error")],
+		] as const)(
+			"logs error and continues when %s callback throws",
+			async (sseEvent, eventData) => {
+				const mockES = createMockEventSource({
+					addEventListener: vi.fn((event, handler) => {
+						const esEvent = sseEvent === "message" ? "data" : sseEvent;
```
buenos-nachos, line 274: Can this get a comment explaining why we're sometimes translating the event type?

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+			},
+		);
+	});
+
+	describe("Callback Error Handling", () => {
+		it.each([
+			["open", new Event("open")],
+			["message", new MessageEvent("data", { data: '{"test": true}' })],
+			["error", new Event("error")],
+		] as const)(
```
buenos-nachos: Same thing here: I feel like the `as const` in the test case already feels like a bit of a code smell, because it doesn't actually give us much type safety, even though the values should be structured as known types

I tried to figure out a way to get the type more specific via the `EventPayloadMap` type, but it got really clunky, really fast. I think this is fine for now:
```ts
type TestEntry = readonly [
	messageType: WebSocketEventType,
	event: Event | MessageEvent
]
```

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+
+				connection.addEventListener(sseEvent, () => {
+					throw new Error("Handler error");
+				});
+				connection.addEventListener(sseEvent, (event: unknown) =>
+					events.push(event),
+				);
+
+				await waitForNextTick();
+				expect(events).toHaveLength(1);
```
buenos-nachos, line 295: Something about the test setup feels a little bit off to me (for this one and a few other tests, too), because we've given `events` the ability to hold any arbitrary data, but we're never checking the contents after the induction step

Is there any way we can verify that we got back specific data?

test/unit/websocket/sseConnection.test.ts
```diff
@@ -0,0 +1,318 @@
...
+					wasClean: false,
+				},
+			]);
+		});
+	});
+
+	describe("Event Listener Management", () => {
+		it("removes event listener without affecting others", async () => {
+			const mockES = createMockEventSource({
+				addEventListener: vi.fn((event, handler) => {
```
buenos-nachos, line 183: It looks like TypeScript isn't smart enough to infer the types for the `vi.fn` callback, so they're getting downgraded into `any` types

In practice, I think this is fine, since the loose code is strictly in the test files, but I'm wondering if there's a way we can tighten things up

Parkreiner (commented):
I tried to review as much as I could, but I feel like I need a bit more context on what we're trying to assert, and the specific concerns we have for the tests

Happy to pair-program on this next week. I'm located in the EST time zone

src/websocket/sseConnection.ts
EhabY: > I think your changes are probably the safest way to address a mismatch that we can't control, but I can't help but wonder if there's a "deeper" fix we can make

As if you've said, it's normal that the node environment has no `ErrorEvent` but the VS Code one has (it uses Electron so it's a browser).

> we broadened it to accept generic events as well?

It is technically `Event` by default but I think the types are extended to be `ErrorEvent | Event`. See https://developer.mozilla.org/en-US/docs/Web/API/EventSource/error_event here it's only `Event`. If I try to make the type only `ErrorEvent` TS will complain since it does not cover `Event` as well.

test/unit/api/streamingFetchAdapter.test.ts
EhabY: Oooh you are right, this provides much better test isolation and concurrency. I've inlined the mocking to each test

test/unit/api/streamingFetchAdapter.test.ts
EhabY: Done, changed to `overrides`

test/unit/api/streamingFetchAdapter.test.ts
EhabY: This is making similar to RESTful requests and WebSocket connections, see the `coderApi.test.ts` where:

> command overrides config overrides axios default

So here the config is passed first, but the thing passed later is the "init" value as far as I understood :thinking:

test/unit/api/streamingFetchAdapter.test.ts
EhabY: I added a comment explaining that headers are lowercased when retrieved.

test/unit/api/streamingFetchAdapter.test.ts
EhabY: It's `AxiosResponse<IncomingMessage>` but yes good point!

test/unit/api/streamingFetchAdapter.test.ts
EhabY: Yes true, done!

test/unit/websocket/sseConnection.test.ts
EhabY: Done!

test/unit/websocket/sseConnection.test.ts
EhabY: Oooh that is much cleaner and clearer, done!

test/unit/websocket/sseConnection.test.ts
EhabY: I changed them `object` instead. Open events do not have any data but it has to emulate the WS interface

test/unit/websocket/sseConnection.test.ts
EhabY: Changed all `it.each`!

test/unit/websocket/sseConnection.test.ts
EhabY: Done

test/unit/websocket/sseConnection.test.ts
EhabY: Added comment `All SSE events are streaming data and attach a listener on the "data" type in the EventSource`

test/unit/websocket/sseConnection.test.ts
EhabY: Made it:

```ts
		type CallbackErrorTestCase = [
			sseEvent: WebSocketEventType,
			eventData: Event | MessageEvent,
		];
		it.each<CallbackErrorTestCase>([
			["open", new Event("open")],
			["message", new MessageEvent("data", { data: '{"test": true}' })],
			["error", new Event("error")],
		])
```

test/unit/websocket/sseConnection.test.ts
EhabY: I suppose we can pass a 3rd argument here to assert the exact object type but this means we aren't testing the correct thing. There are tests already for specific data, this one here just makes sure that the 2nd listener is actually executed. I suppose we can just put a counter there and it would be the same :sweat:

test/unit/websocket/sseConnection.test.ts
EhabY: Nice catch, we can use explicit typing here but could be a bit more verbose syntax wise

src/websocket/sseConnection.ts
buenos-nachos: Oh, yeah, I thought that `EventSource` types the error event as `ErrorEvent` when you specify `"error"`, but it looks like it doesn't

I'm curious if there's anything we can leverage from JS DOM or HappyDOM to bridge the gaps between the server and browser-ish environments, but I'm okay with the current code for now

test/unit/api/streamingFetchAdapter.test.ts
buenos-nachos: Yeah, if there's a convention and the behavior is expected, I'm okay with keeping it

test/unit/websocket/sseConnection.test.ts
buenos-nachos: Yeah, in that case, it's probably not worth it. I'm okay with keeping the code as-is, especially since we have the other test cases

src/websocket/sseConnection.ts
EhabY: Hmmm the issue here is that VS Code is still run in a NodeJS environment, yes it's electron but it's kind of mixed? Running the tests as node matches the production environment as well.

This can be seen in the types even, the VS Code "DOM" types extend the actual DOM types defined in the Mozilla documentation

test/unit/websocket/sseConnection.test.ts
EhabY: The fact that this has overrides is making it very difficult to force type checking except by us explicitly giving them the types which defeats the purpose of it. I'll leave it as is for now..

PR #627, opened 2025-10-17
Show Remote SSH Output panel on workspace start

### LINKED ISSUE #626: Show agent startup script when connecting to a Workspace
## New Approach

In case the template has a blocking script, then we should wait for the script to finish executing and show the user the progress as it's happening. This will make it clear that something is happening and there are no network issues instead of thinking it's just stuck.

## Old Approach

This will show the agent startup script logs and provide much better feedback to the waiting user than what we show by default.

In the VS Code extension, we support 4 Remote SSH extensions:
* ms-vscode-remote.remote-ssh -> `opensshremotes.showLog`
* jeanp413.open-remote-ssh -> `opensshremotes.showLog`
* codeium.windsurf-remote-openssh -> `windsurf-remote-openssh.showLog`
* anysphere.remote-ssh -> `opensshremotes.showLog`~~

Since they are never installed together we can just execute all of them and ignore any errors (best effort approach)

EhabY:
Closes #626

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Changed
+
+- Improved workspace connection progress messages and enhanced the workspace build terminal
+  with better log streaming. The extension now also waits for blocking startup scripts to
+  complete before connecting, providing clear progress indicators during the wait.
+
 ## [v1.11.3](https://github.com/coder/vscode-coder/releases/tag/v1.11.3) 2025-10-22
 
 ### Fixed


src/extension.ts
```diff
@@ -460,9 +462,27 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 			}
 		}
 	}
+
+	if (shouldShowSshOutput) {
+		showSshOutput();
+	}
```
EhabY: I'm aiming for a less intrusive UX by only showing the `Output: Remote SSH` channel when a workspace starts (i.e., when the startup script runs). However, since the channel logs activity during (normal) connection as well, we could always show it to help catch issues sooner. The caveat is that we can't automatically hide the output channel once a connection succeeds.

What would be the best UX approach here?

* Show the output only after a workspace startup (current approach)
* Show it always when connecting to a remote SSH (whether it's a startup or not).

EhabY: I'm now leaning more towards the latter because those logs can be useful still even if this part is faster than the startup script part

code-asher: If we could hide it after connecting I think it might make sense but as a user I feel like I might be annoyed if I have to keep closing the window every time I connect (but I am not a VS Code user so I could be wrong).  Do we have users asking for this?

If we just want users to be aware something is happening, the progress notifications should cover that, but maybe we need to show more detail in the notification?  Although, that is on the remote SSH extension side and not something we can change.  :thinking:

EhabY: > Do we have users asking for this?

This scenario was encountered by @dannykopping where their workspace seemed like it was stuck but it was just executing the startup script and working normally. Seeing the logs just shows that things are moving I suppose while the progress notification doesn't tell us if things are ACTUALLY working (can't tell if it's stuck).

I can keep it as is where we just show the output logs when we start the workspace, that way it's less annoying but might not show the output in some cases.

code-asher: Makes sense.  Actually, I was thinking at this stage we handed off to the remote extension and had no more control, but we do actually still control this part right?  We start the workspace before handing off, I think?  What if we updated the notification with the last line from the log?  Or would that look janky?  Maybe even just adding a "open build log" button would do the trick.

EhabY: > but we do actually still control this part right?

The progress monitor is actually from the Remote SSH extension so we have no control over that one and I'm not sure if we can even hook into those events to show/hide the output or some other notification that could be more descriptive.

We can instead show an info notification when we hand off with a simple message saying we are connecting to the SSH and a button to open the logs, it would go away after a few seconds if not clicked though so it's a bit more subtle than just showing the output.

code-asher: But we start the workspace and wait for it to be running before handing over control no?

https://github.com/coder/vscode-coder/blob/e64350fd448e63ceb982e1894668816bd091b42e/src/remote/remote.ts#L430-L431

code-asher: And we have this progress notification: https://github.com/coder/vscode-coder/blob/e64350fd448e63ceb982e1894668816bd091b42e/src/remote/remote.ts#L159-L164

EhabY: Yes, but I think we are talking about different progress monitors. The one I'm talking about is shown by the Remote SSH extension. See [the 4th point here](https://code.visualstudio.com/docs/remote/ssh#_connect-to-a-remote-host).

> After a moment, VS Code will connect to the SSH server and set itself up. VS Code will keep you up-to-date using a progress notification and you can see a detailed log in the Remote - SSH output channel.

We already focus the terminal during a workspace build by showing a "Build Logs" terminal (and destroying it after). However, the startup script is executed AFTER the build has finished and is printed to the SSH Remote output.

code-asher: Ohhhh right we pop that up already, my bad.  No need to modify the notification there.  :facepalm: 

~So is this like in a reconnect scenario where our plugin never gets involved?~

~I think my confusion is, if we are able to run code to show the remote SSH window, then we can equally as well run code to wait for the workspace to be up and show whatever nice UI we want (maybe we pop up then close the build log like in the other scenario).~ ignore I am dumb

code-asher: Ohhhhhhhhhh we are not waiting for startup scripts though, just that the workspace itself is up.  I wonder if we are able to do that.

This is what I get for somehow not reading your last sentence lol

EhabY: > wait for the workspace to be up

I couldn't find any way to actually hook up to the Remote SSH extension to get these events (connection established). From what I tested, the Remote SSH extension is activated after we finish Coder's activation.

I was very confused as well since I thought the startup script was the same as the build script, but no, we do wait for the build script and show the output in a nice way, but the start up script is executed on the server and through the Remote SSH extension. You can test it by starting a workspace like Coder on Coder.

EhabY: Hahaha I went through the same exact confusion so I understand

code-asher: On the one hand, it feels surprising the build watch endpoint does not also wait for startup scripts, on the other hand it does make sense, I suppose they are distinct steps.

Thinking out loud, maybe we could check if the agent has any blocking startup scripts (`start_blocks_login` I think) and then pop up the remote SSH log window if so.

The best long-term fix I suppose would be to actually follow and print the startup script logs, like with the `/api/v2/workspaceagents/$id/logs?follow=true` endpoint (`watchWorkspaceAgentLogs` in coder/coder's `api.ts`).  But I think this endpoint stays open indefinitely.  We would need some new endpoint specifically for the startup scripts I think.

code-asher: Wait actually, I think the workspace status does tell us if the startup scripts finished?  I vaguely remember doing this in the Jetbrains plugin now that I think about it.

Yeah there are lifecycle states to tell us if the scripts are done: https://github.com/coder/jetbrains-coder/blob/51d1d57bc80e8173b6603a8c4c39b8ffc0e6cbde/src/main/kotlin/com/coder/gateway/models/WorkspaceAndAgentStatus.kt#L31

We could combine that with watching logs

EhabY: Ooooh that could be nice, watching the logs and showing a similar terminal to the `Build Log` one is very clear. But how do we know that the startup script finished executing besides just polling `getWorkspace`, extracting the agents and checking if the lifecycle is not `starting`?

code-asher: Polling is how we did it in Gateway but we could also use `watchWorkspace` I believe (I think we already use it, but would need some refactoring).

src/remote/remote.ts
```diff
@@ -558,39 +542,29 @@ export class Remote {
...
 				this.logger.info(`Waiting for ${workspaceName}/${agent.name}...`);
-				await vscode.window.withProgress(
+				const updatedAgent = await this.waitForAgentWithProgress(
+					monitor,
+					agent,
 					{
-						title: "Waiting for the agent to connect...",
-						location: vscode.ProgressLocation.Notification,
+						progressTitle: "Waiting for the agent to connect...",
+						timeoutMs: 3 * 60 * 1000,
```
EhabY: We used to potentially wait forever, should we set a timeout for this? What's a good value 🤔 ?

src/remote/remote.ts
```diff
@@ -608,6 +582,64 @@ export class Remote {
...
+								agent,
+							);
+						socket = agentSocket;
+
+						const agentStatePromise = this.waitForAgentWithProgress(
+							monitor,
+							agent,
+							{
+								progressTitle: "Waiting for agent startup scripts...",
+								timeoutMs: 5 * 60 * 1000,
```
EhabY: Same here

src/api/workspace.ts
```diff
@@ -36,35 +41,33 @@ export async function startWorkspaceIfStoppedOrFailed(
...
+			const lines = data
 				.toString()
 				.split(/\r*\n/)
-				.forEach((line: string) => {
-					if (line !== "") {
-						writeEmitter.fire(line.toString() + "\r\n");
-					}
-				});
+				.filter((line) => line !== "");
+			for (const line of lines) {
```
code-asher, line 57: I see a few places have been changed from `forEach` to a `for of` loop, was there a specific reason for this or just preference?

src/api/workspace.ts
```diff
@@ -130,3 +135,55 @@ export async function waitForBuild(
...
+	agent: WorkspaceAgent,
+): Promise<{
+	socket: OneWayWebSocket<WorkspaceAgentLog[]>;
+	completion: Promise<void>;
+}> {
+	// This fetches the initial bunch of logs.
+	const logs = await client.getWorkspaceAgentLogs(agent.id);
+	for (const log of logs) {
+		writeEmitter.fire(log.output + "\r\n");
+	}
```
code-asher: The watch endpoint already streams all the logs, I think (including previous ones, since the default for the `after` query param is zero which means all logs), so I believe we would not need the extra `getWorkspaceAgentLogs` call.

src/remote/remote.ts
code-asher: I think there is no good timeout value really, agents can take an arbitrary amount of time to connect.  As long as there is no error or network disconnect, we should keep waiting.

If we did need a timeout, it should probably be something like where we send a ping and if we do not get a pong back in N seconds after M attempts then we error.  I think maybe websockets already do something like that though?

Ultimately though as long as the connection looks good, and we are communicating the status to the user, we should let the user decide whether they want to wait, IMO.

src/remote/remote.ts
```diff
@@ -733,6 +765,95 @@ export class Remote {
...
+	 */
+	private async waitForAgentWithProgress(
+		monitor: WorkspaceMonitor,
+		agent: WorkspaceAgent,
+		options: {
+			progressTitle: string;
+			timeoutMs: number;
+			errorDialogTitle: string;
+		},
+		checker: (agent: WorkspaceAgent) => WorkspaceAgent | undefined,
```
code-asher: This threw me off a little, I think I expected a truthy value to mean the condition has been met (like the callback to `filter` or `some` for example).

Also could this just return a boolean?  It looks like all we ever do is pass back the same agent that was passed in.

src/remote/remote.ts
code-asher: Same thoughts here, I could imagine scripts have even more variable timing than an agent starting.  Granted, five minutes would be a long time, but I think we should not be prescriptive about how long scripts are allowed to run.

src/remote/remote.ts
```diff
@@ -608,6 +582,64 @@ export class Remote {
...
+					this.logger.info(
+						`Waiting for ${workspaceName}/${agent.name} startup...`,
+					);
+
+					let terminalSession: TerminalSession | undefined;
+					let socket: OneWayWebSocket<WorkspaceAgentLog[]> | undefined;
+					try {
+						terminalSession = new TerminalSession("Agent Log");
+						const { socket: agentSocket, completion: logsCompletion } =
+							await streamAgentLogs(
```
code-asher: The `await` here I think means we could possibly miss a state change in between the previous watch and this?  Maybe we could store the current state in `monitor` and check it before registering the `onChange`.

But, I do feel like we have this state machine that is broken up into separate sections and it could make a lot of sense to put it all together.  A single machine that can take a workspace object, explicitly lists out all the possible states, and does the right thing.  

For example what if the workspace is stopped while we wait for the scripts?  We are only checking the lifecycle state so we would miss that change I think.  The web socket would close eventually, although I think we might not be watching for that either (I think this may have already been an issue though, or I am misreading the code haha).

src/remote/remote.ts
```diff
@@ -733,6 +765,95 @@ export class Remote {
...
+				{
+					useCustom: true,
+					modal: true,
+					detail: error instanceof Error ? error.message : String(error),
+				},
+				"Close Remote",
+			);
+			if (result === "Close Remote") {
+				await this.closeRemote();
+			}
```
code-asher: Error dialogs can be closed without pressing the button I think, right?  Should we always close the remote?  I think there is no scenario in the case of an error where we would want to keep the remote open.

Actually, could we let this bubble up?  We close the remote on errors from setting up the remote already.  Kinda nice to have one place where we catch an error and display it.

src/remote/remote.ts
EhabY: Yeah fair enough, timing out for something that can potentially take arbitrary amount of time and would lead to the product being unusable is bad. Removed the timeout!

src/remote/remote.ts
EhabY: Removed

src/api/workspace.ts
EhabY: I installed sonarqube which has some default warnings, see https://next.sonarqube.com/sonarqube/coding_rules?open=javascript%3AS7728&rule_key=javascript%3AS7728

It's not really that significant but it's better I suppose

src/api/workspace.ts
EhabY: Is this also true for the `waitForBuild`? because there we also do:

```typescript
// This fetches the initial bunch of logs.
const logs = await client.getWorkspaceBuildLogs(workspace.latest_build.id);
for (const log of logs) {
	writeEmitter.fire(log.output + "\r\n");
}

const socket = await client.watchBuildLogsByBuildId(
	workspace.latest_build.id,
	logs,
);
```

src/remote/remote.ts
EhabY: Ohh actually fair, I made it `(agent: WorkspaceAgent) => boolean` and we resolve with the new agent info

src/remote/remote.ts
EhabY: This actually brings up a question, should we make the progress indicator cancellable? Currently it's not which might make sense since cancelling means we close the remote connection.

src/remote/remote.ts
EhabY: Oooh yeah that could work, I'll make sure the thrown error has the best message though!

src/remote/remote.ts
EhabY: Do you mean that we have one big state machine that checks the workspace state AND the life cycle state of the agents?

So for example, if we are starting the workspace then handle that as before, if we switch to `started` then we handle the agent life cycle all in one big switch that is executed on every `onChange`

src/remote/remote.ts
EhabY: Done

src/remote/remote.ts
EhabY: Done, added an agent status/lifecycle state machine and made sure to throw an error if at any point the workspace is not `running`

src/api/workspace.ts
EhabY: Seems to be the case, I've removed the initial fetching and the logs seem to be complete for both requests :thinking:

src/remote/remote.ts
code-asher: Oh right good point!  Yeah being able to cancel it does seem ideal.

src/api/workspace.ts
code-asher: Oooo gotcha

src/api/workspace.ts
code-asher: Oh yup good catch!

src/remote/remote.ts
code-asher: Yup exactly, sorry for the delayed reply.

src/remote/remote.ts
```diff
@@ -733,6 +672,193 @@ export class Remote {
...
+
+				case "connected":
+					// Agent connected, now handle lifecycle state
+					break;
+
+				case "disconnected":
+				case "timeout":
+					throw new Error(
+						`${workspaceName}/${currentAgent.name} ${currentAgent.status}`,
+					);
```
code-asher: I think timeout means the agent is taking longer than expected to connect, but we can keep waiting because it might eventually connect.

src/remote/remote.ts
```diff
@@ -733,6 +672,193 @@ export class Remote {
...
+
+				case "off":
+				case "start_error":
+				case "start_timeout":
+				case "shutdown_error":
+				case "shutdown_timeout":
+				case "shutting_down":
+					throw new Error(
+						`${workspaceName}/${currentAgent.name} lifecycle state: ${currentAgent.lifecycle_state}`,
+					);
```
code-asher: We should still connect in the case of `start_error`, it could be that their script has an error or a bug, but we can still connect to the agent.

`start_timeout` means their script is taking a really long time, but we can still wait, it might eventually complete.

With the `shutdown_*` ones, this means the workspace is shutting down, I wonder if we should wait for the workspace to shut down and then ask the user if they want to start it?  Not sure if `off` means a similar thing.

**edit**: if we incorporate workspace build status checks then we might be able to handle shutdown/off for free, I imagine in these cases the workspace goes to `stopping` and `stopped`.

src/remote/remote.ts
```diff
@@ -733,6 +672,193 @@ export class Remote {
...
+	 * Returns the result when the condition is met or throws an error on timeout.
+	 */
+	private async waitForAgentCondition(
+		monitor: WorkspaceMonitor,
+		agent: WorkspaceAgent,
+		checker: (agent: WorkspaceAgent) => boolean,
+	): Promise<WorkspaceAgent> {
+		return new Promise<WorkspaceAgent>((resolve, reject) => {
+			const updateEvent = monitor.onChange.event((workspace) => {
+				if (workspace.latest_build.status !== "running") {
```
code-asher: What do you think about moving the build status checks into `ensureAgentReady` as well and then call it on the change event?  So we would have something like:

```typescript
new Promise((resolve, reject) => {
  monitor.onChange.event((workspace) => {
    try {
      if (ensureAgentReady(workspace) {
        resolve(workspace)
      }
    } catch (error) {
      reject(error)
    }
  })
})
````

And `ensureAgentReady` would merge in that switch we have on `workspace.latest_build.status`.

Partially because, otherwise we have to duplicate checks.  For example, we need to allow connecting with `start_error`, `start_timeout`, or `ready`, so we would have those checks once in the switch, and then once again in `checker`.  We also are duplicating the agent and workspace build status checks, which granted is simpler since they only have a single connectable state right now, but could be nice if they were derived from a single spot.

The socket handling and logging could be a bit tricky though, maybe we share one progress notification and terminal to make it easier.

Not saying we should block on this though, just a suggestion.

**Review by code-asher — COMMENTED** [review]
Looking really good!  I think we just need to tweak how we handle the timeout and error states, and then we are good to go.

File: src/remote/remote.ts (hunk 10 above)
Comment (EhabY) [reply src/remote/remote.ts]: So for `shutdown_*` we can rely on the workspace shutting down - Does that mean the state machine for the agent should be a no-op here? IDK if the workspace can be running but the agent is shutting down for example, in that case we'd be stuck doing nothing...

Comment (code-asher) [reply src/remote/remote.ts]: Definitely agree a no-op would be sketchy.

If the agent is shutting down but the workspace is running, I think it would be reasonable to throw an error.

Or we could wait for one of the final states (`off` and `shutdown_error` I think) and offer to restart the workspace, but idk if the extra work is worth it.

Also just had a thought that we could indicate the current state in the progress notification, that might help the user understand what they are waiting for, in addition to the logs we already have (not sure that the notification lets us dynamically update the title though).

Comment (code-asher) [reply src/remote/remote.ts]: Also I had `start_timeout` wrong, we should not keep waiting, but rather stop waiting and just connect when we get `start_timeout`, I think this is what the cli does.  So `ready`, `start_timeout`, and `start_error` are all "connectable" states.

Comment (EhabY) [reply src/remote/remote.ts]: > not sure that the notification lets us dynamically update the title though

Indeed it does but combining the agent with the workspace seems to be a bit tricky while handling the sockets and terminal etc.. I think I am almost there and it does look a lot better

Comment (EhabY) [reply src/remote/remote.ts]: > Or we could wait for one of the final states (off and shutdown_error I think) and offer to restart the workspace, but idk if the extra work is worth it.

This issue here is that I'm not entirely sure if this is the "agents" state job. Like shouldn't restarting the workspace depend on the workspace state only?

Comment (code-asher) [reply src/remote/remote.ts]: Yeah it would be better if we had an option to just restart the agent, but I think there is no API for that, so the best we could offer is to restart the whole workspace in the hopes the agent will not shut down this time.

But idk if we can even get into this state.

File: src/remote/remote.ts (hunk 9 above)
Comment (EhabY) [reply src/remote/remote.ts]: Done, add this to the state machine

File: src/remote/remote.ts (hunk 11 above)
Comment (EhabY) [reply src/remote/remote.ts]: I have added a workspace status state machine along with an agent one. I shared the terminal so that we log to one place and added handling for all the states as best I can. Seems to work well.

If we get a new workspace while processing the first one then we queue them while keeping the latest one only (so queue of size 1).

File: src/remote/remote.ts [hunk 12]
```diff
@@ -427,36 +283,110 @@ export class Remote {
...
+										w,
+										progress,
+									);
+									if (isReady) {
+										subscription.dispose();
+										resolve();
+										return;
+									}
+
+									if (pendingWorkspace) {
```
code-asher: This would have to be in a loop I think right?  If we get an update while processing a workspace, we set `pendingWorkspace` and return.  Then the workspace processing ends, and we start processing the pending workspace that was set earlier.  Then if another workspace update comes in we would set `pendingWorkspace` again and return, but nothing after this would then run that latest update.

Maybe it would be as simple as something like this after the try/catch/finally:

```
if (pendingWorkspace) {
  processWorkspace(pendingWorkspace)
}
```

src/remote/remote.ts
```diff
@@ -427,36 +283,110 @@ export class Remote {
...
+						cancellable: false,
+						title: "Preparing workspace",
+					},
+					async (progress) => {
+						let inProgress = false;
+						let pendingWorkspace: Workspace | null = null;
+
+						await new Promise<void>((resolve, reject) => {
+							const processWorkspace = async (w: Workspace) => {
+								workspace = w;
```
code-asher: Does the notification let us resolve with any value?  I wonder if we could do `resolve(w)` and then `workspace = await this.vscodeProposed.window.WithProgress`.

src/workspace/workspaceMonitor.ts
```diff
@@ -131,7 +133,10 @@ export class WorkspaceMonitor implements vscode.Disposable {
 		this.maybeNotifyOutdated(workspace);
 		this.maybeNotifyAutostop(workspace);
 		this.maybeNotifyDeletion(workspace);
-		this.maybeNotifyNotRunning(workspace);
+		if (this.isReady) {
+			// This instance might be created before the workspace is running
+			this.maybeNotifyNotRunning(workspace);
+		}
```
code-asher: This notification can be a little bit weird, if I stop my workspace while trying to connect for example it says "reload to reconnect" but actually in the background we are handling it and reconnecting automatically already.  Maybe we only show it after fully connecting rather than when ready?  Not sure.

src/remote/workspaceStateMachine.ts
```diff
@@ -0,0 +1,272 @@
...
+
+			default:
+				this.closeBuildLogSocket();
+				throw new Error(
+					`${workspaceName} unknown status: ${workspace.latest_build.status}`,
+				);
+		}
+
+		const agents = extractAgents(workspace.latest_build.resources);
+		if (this.agentId === undefined) {
```
code-asher: The agent ID can change after every restart, so if we are handling a case where the workspace shuts down in the middle of trying to reconnect, then we eventually get back here but then we can get an error because the original agent ID we had no longer exists.

Maybe we could get or ask for the ID again if there is no match?  Or, maybe we clear `this.agentID` when the workspace becomes un-ready.

code-asher (commented):
Awesome!!  I experimented with it and it has a few quirks when I try to stop my workspace while it waits for scripts, but otherwise looks excellent.

src/remote/remote.ts
EhabY: Smart, done!

src/remote/remote.ts
EhabY: Oooh that's right, doing that after the `try/catch/finally` would just trigger a recursive loop as long as we get new workspaces. Done!

src/workspace/workspaceMonitor.ts
EhabY: Ah that was my first solution but it seemed a bit quirky to set the readiness from outside, though that might be the only way since we should only trigger this AFTER the state machine is done. I've added `markInitialSetupComplete` to the monitor that is triggered after we have a running workspace and agent

src/remote/workspaceStateMachine.ts
EhabY: I suppose we can clear it on the following conditions since those indicate a restart and do not throw:

```
case "pending":
case "starting":
case "stopping":
```

Like if we get any of these states then we'll have to wait for the workspace build so we can also clear the agent ID then.

code-asher: Makes sense to me!

src/workspace/workspaceMonitor.ts
```diff
@@ -133,7 +136,7 @@ export class WorkspaceMonitor implements vscode.Disposable {
 		this.maybeNotifyOutdated(workspace);
 		this.maybeNotifyAutostop(workspace);
 		this.maybeNotifyDeletion(workspace);
```
code-asher: Was thinking about this some more, and we probably want to skip the deletion notification while in the setup phase too, otherwise I think we would get one notification here, and then another from the error thrown during setup.

Autostop and outdated I think are OK to show even during setup.

src/remote/workspaceStateMachine.ts
```diff
@@ -95,6 +95,8 @@ export class WorkspaceStateMachine implements vscode.Disposable {
 			case "pending":
 			case "starting":
 			case "stopping":
+				// Clear the agent ID since it could change after a restart
+				this.agentId = undefined;
 				progress?.report({ message: "Waiting for workspace build..." });
```
code-asher: What do you think about putting the build state in here?  `Waiting for workspace build: stopping` or `Waiting for workspace to queue/start/stop` if we want to be fancy.

src/remote/workspaceStateMachine.ts
```diff
@@ -0,0 +1,274 @@
...
+				);
+				this.logger.info(`${workspaceName} status is now running`);
+				return false;
+			}
+
+			case "pending":
+			case "starting":
+			case "stopping":
+				// Clear the agent ID since it could change after a restart
+				this.agentId = undefined;
```
code-asher: I think we need to clear the agent socket here as well or something, when I stop my workspace and it starts again, I do not see the agent script logs the second time.

Or, maybe we clear the socket when the agent closes, or both.

src/workspace/workspaceMonitor.ts
EhabY: Done!

src/remote/workspaceStateMachine.ts
EhabY: Done!

src/remote/workspaceStateMachine.ts
EhabY: I fixed this here but I do not know how to re-produce? Can you give it a go or give me the steps?

code-asher: I think what I did was:

1. Make sure workspace is off.
2. Connect, watch the logs, see startup script logs.
3. Turn off the workspace before it can finish connecting (I did this via the dashboard).
4. It waits for the stop then starts the workspace again, but this time no startup script logs.

code-asher: Tested and it works!

EhabY: Ahhh tested and yes seems to work well now :)!

PR #628, opened 2025-10-17
Update CLAUDE.md

EhabY:


CLAUDE.md
```diff
@@ -8,20 +43,22 @@
...
-- No semicolons (see `.prettierrc`)
+- Use semicolons
 - Trailing commas for all multi-line lists
 - 120 character line width
 - Use ES6 features (arrow functions, destructuring, etc.)
 - Use `const` by default; `let` only when necessary
+- Never use `any`, and use exact types when you can
 - Prefix unused variables with underscore (e.g., `_unused`)
-- Sort imports alphabetically in groups: external → parent → sibling
+- Sort imports alphabetically in groups (see `import/order` in `.eslintrc.json`): external → parent → sibling
```
jaaydenh: since this can be automatically handled when running linting, there should be no need to use context for this.

EhabY: That's true, I've removed this rule and 3 other prettier rules and replaced them with:

> Use Prettier for code formatting and ESLint for code linting

PR #633, opened 2025-10-27
Implement OAuth client in VS Code

### LINKED ISSUE #586: feat: support OAuth in the VS Code extension
With Coder eventually becoming an OAuth provider, there is an opportunity to replace the existing copy-paste token authentication with a smoother experience. 

# Issues with Token Auth 
- **Longer default token lifetimes** **due to lack of refresh**, which means more risk of abuse - a stolen token remains valid for a longer period of time, while real-time revocation becomes far more difficult (requires a server-side blacklist that must be maintained)
- **If tokens are leaked, potential for widespread damage is higher** - because tokens are stateless and self-contained, the validating server does not check with a database for verification
- Requires **more user input and changes of context** (redirection to a browser page, copy the token to the clipboard, paste it into the original context) which results in a heavier UX

# Suggested Flow 

Image

1. Once the user chooses to Login, they get directed to the following page.  
2. Upon entering the URL, a browser page will open and prompt the user to log in to their Coder account.
    1. Offer an optional link with text to the effect of, “Can’t authenticate? Generate a token” that will reload the page with a generated token if clicked.
    2. If this link is clicked, the remainder of the flow will return to the token authentication experience (i.e. IDE will prompt for the token to be pasted.)
3. In the next prompt, the Coder IDE extension requests access to act on their behalf. The user clicks **Grant app access**.
4. When authentication is successfully completed, a confirmation message appears. The user closes the browser window and returns to the IDE. 
    1. The IDE reads and saves authentication on the user’s local machine.

The permissions required on the API key for IDE clients are: 

- read workspaces
- read workspace agents + apps
- read your own user details
- authentication
- read access to SSH keys
- connect to workspace agents and show their resources
- inbox notifications

EhabY:
Closes #586

### DIFF
--- a/src/api/oauthInterceptors.ts
+++ b/src/api/oauthInterceptors.ts
@@ -24,16 +24,10 @@ export function attachOAuthInterceptors(
 	client.getAxiosInstance().interceptors.response.use(
 		// Success response interceptor: proactive token refresh
 		(response) => {
-			if (oauthSessionManager.shouldRefreshToken()) {
-				logger.debug(
-					"Token approaching expiry, triggering proactive refresh in background",
-				);
-
-				// Fire-and-forget: don't await, don't block response
-				oauthSessionManager.refreshToken().catch((error) => {
-					logger.warn("Background token refresh failed:", error);
-				});
-			}
+			// Fire-and-forget: don't await, don't block response
+			oauthSessionManager.refreshIfAlmostExpired().catch((error) => {
+				logger.warn("Proactive background token refresh failed:", error);
+			});
 
 			return response;
 		},
--- a/src/oauth/sessionManager.ts
+++ b/src/oauth/sessionManager.ts
@@ -34,7 +34,7 @@ const RESPONSE_TYPE = "code" as const;
 const PKCE_CHALLENGE_METHOD = "S256" as const;
 
 /**
- * Token refresh threshold: refresh when token expires in less than this time
+ * Token refresh threshold: refresh when token expires in less than this time.
  */
 const TOKEN_REFRESH_THRESHOLD_MS = 10 * 60 * 1000;
 
@@ -44,9 +44,14 @@ const TOKEN_REFRESH_THRESHOLD_MS = 10 * 60 * 1000;
 const ACCESS_TOKEN_DEFAULT_EXPIRY_MS = 60 * 60 * 1000;
 
 /**
- * Minimum time between refresh attempts to prevent thrashing
+ * Minimum time between refresh attempts to prevent thrashing.
  */
 const REFRESH_THROTTLE_MS = 30 * 1000;
+
+/**
+ * Background token refresh check interval.
+ */
+const BACKGROUND_REFRESH_INTERVAL_MS = 5 * 60 * 1000;
 
 /**
  * Minimal scopes required by the VS Code extension.
@@ -69,6 +74,7 @@ export class OAuthSessionManager implements vscode.Disposable {
 	private storedTokens: StoredOAuthTokens | undefined;
 	private refreshPromise: Promise<TokenResponse> | null = null;
 	private lastRefreshAttempt = 0;
+	private refreshTimer: NodeJS.Timeout | undefined;
 
 	private pendingAuthReject: ((reason: Error) => void) | undefined;
 
@@ -88,6 +94,7 @@ export class OAuthSessionManager implements vscode.Disposable {
 			extensionId,
 		);
 		await manager.loadTokens();
+		manager.scheduleBackgroundRefresh();
 		return manager;
 	}
 
@@ -158,6 +165,25 @@ export class OAuthSessionManager implements vscode.Disposable {
 		this.storedTokens = undefined;
 		this.refreshPromise = null;
 		this.lastRefreshAttempt = 0;
+	}
+
+	/**
+	 * Schedule the next background token refresh check.
+	 * Only schedules the next check after the current one completes.
+	 */
+	private scheduleBackgroundRefresh(): void {
+		if (this.refreshTimer) {
+			clearTimeout(this.refreshTimer);
+		}
+
+		this.refreshTimer = setTimeout(async () => {
+			try {
+				await this.refreshIfAlmostExpired();
+			} catch (error) {
+				this.logger.warn("Background token refresh failed:", error);
+			}
+			this.scheduleBackgroundRefresh();
+		}, BACKGROUND_REFRESH_INTERVAL_MS);
 	}
 
 	/**
@@ -636,13 +662,23 @@ export class OAuthSessionManager implements vscode.Disposable {
 	}
 
 	/**
+	 * Refreshes the token if it is approaching expiry.
+	 */
+	public async refreshIfAlmostExpired(): Promise<void> {
+		if (this.shouldRefreshToken()) {
+			this.logger.debug("Token approaching expiry, triggering refresh");
+			await this.refreshToken();
+		}
+	}
+
+	/**
 	 * Check if token should be refreshed.
 	 * Returns true if:
 	 * 1. Token expires in less than TOKEN_REFRESH_THRESHOLD_MS
 	 * 2. Last refresh attempt was more than REFRESH_THROTTLE_MS ago
 	 * 3. No refresh is currently in progress
 	 */
-	public shouldRefreshToken(): boolean {
+	private shouldRefreshToken(): boolean {
 		if (
 			!this.isLoggedInWithOAuth() ||
 			!this.storedTokens?.refresh_token ||
@@ -757,6 +793,10 @@ export class OAuthSessionManager implements vscode.Disposable {
 	 * Clears all in-memory state.
 	 */
 	public dispose(): void {
+		if (this.refreshTimer) {
+			clearTimeout(this.refreshTimer);
+			this.refreshTimer = undefined;
+		}
 		if (this.pendingAuthReject) {
 			this.pendingAuthReject(new Error("OAuth session manager disposed"));
 		}
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -9,7 +9,6 @@ import { type CoderApi } from "./api/coderApi";
 import { type CliManager } from "./core/cliManager";
 import { type ServiceContainer } from "./core/container";
 import { type ContextManager } from "./core/contextManager";
-import { type Deployment } from "./core/deployment";
 import { type MementoManager } from "./core/mementoManager";
 import { type PathResolver } from "./core/pathResolver";
 import { type SecretsManager } from "./core/secretsManager";
@@ -45,11 +44,11 @@ export class Commands {
 	// if you use multiple deployments).
 	public workspace?: Workspace;
 	public workspaceLogPath?: string;
-	public workspaceRestClient?: CoderApi;
+	public remoteWorkspaceClient?: CoderApi;
 
 	public constructor(
 		serviceContainer: ServiceContainer,
-		private readonly restClient: CoderApi,
+		private readonly extensionClient: CoderApi,
 		private readonly oauthSessionManager: OAuthSessionManager,
 	) {
 		this.vscodeProposed = serviceContainer.getVsCodeProposed();
@@ -65,12 +64,12 @@ export class Commands {
 	/**
 	 * Get the current deployment, throwing if not logged in.
 	 */
-	private async requireDeployment(): Promise<Deployment> {
-		const deployment = await this.secretsManager.getCurrentDeployment();
-		if (!deployment) {
+	private requireExtensionBaseUrl(): string {
+		const url = this.extensionClient.getAxiosInstance().defaults.baseURL;
+		if (!url) {
 			throw new Error("You are not logged in");
 		}
-		return deployment;
+		return url;
 	}
 
 	/**
@@ -117,9 +116,9 @@ export class Commands {
 
 		// Set client immediately so subsequent operations in this function have the correct host/token.
 		// The cross-window listener will also update the client, but that's async.
-		this.restClient.setCredentials(url, result.token);
+		this.extensionClient.setCredentials(url, result.token);
 
-		// Set as current deployment (triggers cross-window sync).
+		// Set as current deployment
 		await this.secretsManager.setCurrentDeployment({ url, label });
 
 		// Update contexts
@@ -173,14 +172,12 @@ export class Commands {
 	 * Log out from the currently logged-in deployment.
 	 */
 	public async logout(): Promise<void> {
-		const deployment = await this.requireDeployment();
-		await this.forceLogout(deployment.label);
-	}
-
-	public async forceLogout(label: string): Promise<void> {
+		const baseUrl = this.requireExtensionBaseUrl();
 		if (!this.contextManager.get("coder.authenticated")) {
 			return;
 		}
+
+		const label = toSafeHost(baseUrl);
 		this.logger.info(`Logging out of deployment: ${label}`);
 
 		// Fire and forget OAuth logout
@@ -190,7 +187,7 @@ export class Commands {
 
 		// Clear from the REST client.  An empty url will indicate to other parts of
 		// the code that we are logged out.
-		this.restClient.setCredentials(undefined, undefined);
+		this.extensionClient.setCredentials(undefined, undefined);
 
 		// Clear current deployment (triggers cross-window sync)
 		await this.secretsManager.setCurrentDeployment(undefined);
@@ -211,8 +208,8 @@ export class Commands {
 	 * Must only be called if currently logged in.
 	 */
 	public async createWorkspace(): Promise<void> {
-		const deployment = await this.requireDeployment();
-		const uri = deployment.url + "/templates";
+		const baseUrl = this.requireExtensionBaseUrl();
+		const uri = baseUrl + "/templates";
 		await vscode.commands.executeCommand("vscode.open", uri);
 	}
 
@@ -226,13 +223,13 @@ export class Commands {
 	 */
 	public async navigateToWorkspace(item: OpenableTreeItem) {
 		if (item) {
-			const deployment = await this.requireDeployment();
+			const baseUrl = this.requireExtensionBaseUrl();
 			const workspaceId = createWorkspaceIdentifier(item.workspace);
-			const uri = deployment.url + `/@${workspaceId}`;
+			const uri = baseUrl + `/@${workspaceId}`;
 			await vscode.commands.executeCommand("vscode.open", uri);
-		} else if (this.workspace && this.workspaceRestClient) {
+		} else if (this.workspace && this.remoteWorkspaceClient) {
 			const baseUrl =
-				this.workspaceRestClient.getAxiosInstance().defaults.baseURL;
+				this.remoteWorkspaceClient.getAxiosInstance().defaults.baseURL;
 			const uri = `${baseUrl}/@${createWorkspaceIdentifier(this.workspace)}`;
 			await vscode.commands.executeCommand("vscode.open", uri);
 		} else {
@@ -250,13 +247,13 @@ export class Commands {
 	 */
 	public async navigateToWorkspaceSettings(item: OpenableTreeItem) {
 		if (item) {
-			const deployment = await this.requireDeployment();
+			const baseUrl = this.requireExtensionBaseUrl();
 			const workspaceId = createWorkspaceIdentifier(item.workspace);
-			const uri = deployment.url + `/@${workspaceId}/settings`;
+			const uri = baseUrl + `/@${workspaceId}/settings`;
 			await vscode.commands.executeCommand("vscode.open", uri);
-		} else if (this.workspace && this.workspaceRestClient) {
+		} else if (this.workspace && this.remoteWorkspaceClient) {
 			const baseUrl =
-				this.workspaceRestClient.getAxiosInstance().defaults.baseURL;
+				this.remoteWorkspaceClient.getAxiosInstance().defaults.baseURL;
 			const uri = `${baseUrl}/@${createWorkspaceIdentifier(this.workspace)}/settings`;
 			await vscode.commands.executeCommand("vscode.open", uri);
 		} else {
@@ -274,7 +271,7 @@ export class Commands {
 	 */
 	public async openFromSidebar(item: OpenableTreeItem) {
 		if (item) {
-			const baseUrl = this.restClient.getAxiosInstance().defaults.baseURL;
+			const baseUrl = this.extensionClient.getAxiosInstance().defaults.baseURL;
 			if (!baseUrl) {
 				throw new Error("You are not logged in");
 			}
@@ -329,15 +326,14 @@ export class Commands {
 					const terminal = vscode.window.createTerminal(app.name);
 
 					// If workspace_name is provided, run coder ssh before the command
-					const deployment = await this.requireDeployment();
+					const baseUrl = this.requireExtensionBaseUrl();
+					const label = toSafeHost(baseUrl);
 					const binary = await this.cliManager.fetchBinary(
-						this.restClient,
-						deployment.label,
+						this.extensionClient,
+						label,
 					);
 
-					const configDir = this.pathResolver.getGlobalConfigDir(
-						deployment.label,
-					);
+					const configDir = this.pathResolver.getGlobalConfigDir(label);
 					const globalFlags = getGlobalFlags(
 						vscode.workspace.getConfiguration(),
 						configDir,
@@ -374,14 +370,14 @@ export class Commands {
 		folderPath?: string,
 		openRecent?: boolean,
 	): Promise<void> {
-		const baseUrl = this.restClient.getAxiosInstance().defaults.baseURL;
+		const baseUrl = this.extensionClient.getAxiosInstance().defaults.baseURL;
 		if (!baseUrl) {
 			throw new Error("You are not logged in");
 		}
 
 		let workspace: Workspace | undefined;
 		if (workspaceOwner && workspaceName) {
-			workspace = await this.restClient.getWorkspaceByOwnerAndName(
+			workspace = await this.extensionClient.getWorkspaceByOwnerAndName(
 				workspaceOwner,
 				workspaceName,
 			);
@@ -417,7 +413,7 @@ export class Commands {
 		localWorkspaceFolder: string = "",
 		localConfigFile: string = "",
 	): Promise<void> {
-		const baseUrl = this.restClient.getAxiosInstance().defaults.baseURL;
+		const baseUrl = this.extensionClient.getAxiosInstance().defaults.baseURL;
 		if (!baseUrl) {
 			throw new Error("You are not logged in");
 		}
@@ -473,7 +469,7 @@ export class Commands {
 	 * this is a no-op.
 	 */
 	public async updateWorkspace(): Promise<void> {
-		if (!this.workspace || !this.workspaceRestClient) {
+		if (!this.workspace || !this.remoteWorkspaceClient) {
 			return;
 		}
 		const action = await this.vscodeProposed.window.showWarningMessage(
@@ -486,7 +482,7 @@ export class Commands {
 			"Update",
 		);
 		if (action === "Update") {
-			await this.workspaceRestClient.updateWorkspaceVersion(this.workspace);
+			await this.remoteWorkspaceClient.updateWorkspaceVersion(this.workspace);
 		}
 	}
 
@@ -501,7 +497,7 @@ export class Commands {
 		let lastWorkspaces: readonly Workspace[];
 		quickPick.onDidChangeValue((value) => {
 			quickPick.busy = true;
-			this.restClient
+			this.extensionClient
 				.getWorkspaces({
 					q: value,
 				})
@@ -564,7 +560,7 @@ export class Commands {
 			// we need to fetch the agents through the resources API, as the
 			// workspaces query does not include agents when off.
 			this.logger.info("Fetching agents from template version");
-			const resources = await this.restClient.getTemplateVersionResources(
+			const resources = await this.extensionClient.getTemplateVersionResources(
 				workspace.latest_build.template_version_id,
 			);
 			return extractAgents(resources);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -151,6 +151,7 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 				client.setCredentials(auth?.url, auth?.token);
 
 				// Update authentication context for current deployment
+				// TODO(ehab) this might never even happen :thinking:
 				contextManager.set("coder.authenticated", auth !== undefined);
 			},
 		);
@@ -159,6 +160,55 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 	// Initialize auth listener for current deployment
 	registerAuthListener(deployment?.label);
 	ctx.subscriptions.push({ dispose: () => authChangeDisposable?.dispose() });
+
+	const changeDeployment = async (
+		deployment: Deployment | null,
+		sessionToken?: string,
+	) => {
+		// Update client
+		if (deployment) {
+			const token =
+				sessionToken ||
+				(await secretsManager.getSessionToken(deployment.label));
+			client.setCredentials(deployment.url, token);
+			await oauthSessionManager.setDeployment(deployment);
+		} else {
+			client.setCredentials(undefined, undefined);
+			oauthSessionManager.clearDeployment();
+		}
+		registerAuthListener(deployment?.label);
+
+		// Update context
+		contextManager.set("coder.authenticated", Boolean(deployment));
+
+		// Refresh workspaces
+		myWorkspacesProvider.fetchAndRefresh();
+		allWorkspacesProvider.fetchAndRefresh();
+	};
+
+	const changeDeploymentAndPersist = async (
+		deployment: Deployment | null,
+		sessionToken?: string,
+	) => {
+		await changeDeployment(deployment, sessionToken);
+		// Persist and sync deployment across windows
+		await secretsManager.setCurrentDeployment(deployment ?? undefined);
+		await mementoManager.addToUrlHistory(deployment?.url ?? "");
+	};
+
+	// Listen for deployment changes from other windows (cross-window sync)
+	ctx.subscriptions.push(
+		secretsManager.onDidChangeCurrentDeployment(async ({ deployment }) => {
+			const isLoggedIn = contextManager.get("coder.authenticated");
+			if (isLoggedIn) {
+				// We keep whatever deployment we have if we're logged in
+				return;
+			}
+
+			output.info("Deployment changed from another window");
+			return changeDeployment(deployment);
+		}),
+	);
 
 	// Handle vscode:// URIs.
 	const uriHandler = vscode.window.registerUriHandler({
@@ -189,7 +239,11 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 					throw new Error("workspace must be specified as a query parameter");
 				}
 
-				await setupDeploymentFromUri(params, client, serviceContainer);
+				await setupDeploymentFromUri(
+					params,
+					serviceContainer,
+					changeDeploymentAndPersist,
+				);
 
 				vscode.commands.executeCommand(
 					"coder.open",
@@ -238,7 +292,11 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 					);
 				}
 
-				await setupDeploymentFromUri(params, client, serviceContainer);
+				await setupDeploymentFromUri(
+					params,
+					serviceContainer,
+					changeDeploymentAndPersist,
+				);
 
 				vscode.commands.executeCommand(
 					"coder.openDevContainer",
@@ -319,31 +377,6 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 
 	const remote = new Remote(serviceContainer, commands, ctx);
 
-	// Listen for deployment changes from other windows (cross-window sync)
-	ctx.subscriptions.push(
-		secretsManager.onDidChangeCurrentDeployment(async ({ deployment }) => {
-			output.info("Deployment changed from another window");
-
-			// Update client
-			if (deployment) {
-				const token = await secretsManager.getSessionToken(deployment.label);
-				client.setCredentials(deployment.url, token);
-				await oauthSessionManager.setDeployment(deployment);
-			} else {
-				client.setCredentials(undefined, undefined);
-				oauthSessionManager.clearDeployment();
-			}
-			registerAuthListener(deployment?.label);
-
-			// Update context
-			contextManager.set("coder.authenticated", Boolean(deployment));
-
-			// Refresh workspaces
-			myWorkspacesProvider.fetchAndRefresh();
-			allWorkspacesProvider.fetchAndRefresh();
-		}),
-	);
-
 	// Since the "onResolveRemoteAuthority:ssh-remote" activation event exists
 	// in package.json we're able to perform actions before the authority is
 	// resolved by the remote SSH extension.
@@ -363,15 +396,10 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 			if (details) {
 				ctx.subscriptions.push(details);
 
-				// Set client host/token immediately for subsequent operations.
-				client.setCredentials(details.url, details.token);
-
-				// Persist and sync deployment across windows
-				await secretsManager.setCurrentDeployment({
-					url: details.url,
-					label: details.label,
-				});
-				await mementoManager.addToUrlHistory(details.url);
+				await changeDeploymentAndPersist(
+					{ label: details.label, url: details.url },
+					details.token,
+				);
 			}
 		} catch (ex) {
 			if (ex instanceof CertificateError) {
@@ -494,15 +522,17 @@ async function showTreeViewSearch(id: string): Promise<void> {
  * Sets up deployment from URI parameters. Handles URL prompting, client setup,
  * and token storage. Throws if user cancels URL input.
  *
- * Sets client host/token immediately for subsequent operations.
- * Other updates (auth listener, OAuth manager, context, workspaces) are handled
- * asynchronously by the onDidChangeCurrentDeployment listener.
+ * Updates the client host/token, auth listener, OAuth manager, context, etc.
+ * through the `changeDeploymentAndPersist` callback.
  */
 async function setupDeploymentFromUri(
 	params: URLSearchParams,
-	client: CoderApi,
 	serviceContainer: ServiceContainer,
-): Promise<Deployment> {
+	changeDeploymentAndPersist: (
+		deployment: Deployment | null,
+		sessionToken?: string,
+	) => Promise<void>,
+): Promise<void> {
 	const secretsManager = serviceContainer.getSecretsManager();
 	const mementoManager = serviceContainer.getMementoManager();
 	const currentDeployment = await secretsManager.getCurrentDeployment();
@@ -529,16 +559,11 @@ async function setupDeploymentFromUri(
 	// command currently always requires a token file. However, if there is
 	// a query parameter for non-token auth go ahead and use it anyway;
 	const token = await getToken(params, label, secretsManager);
-	client.setCredentials(url, token);
 	if (token) {
 		await secretsManager.setSessionAuth(label, { url, token });
 	}
 
-	// Persist and sync deployment across windows
-	await secretsManager.setCurrentDeployment({ url, label });
-	await mementoManager.addToUrlHistory(url);
-
-	return { url, label };
+	await changeDeploymentAndPersist({ label, url }, token);
 }
 
 async function getToken(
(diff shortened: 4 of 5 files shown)


src/oauth/types.ts
```diff
@@ -0,0 +1,163 @@
+// OAuth 2.1 Grant Types
```
EhabY: Potentially we can use the types defined in the Coder API but couldn't access those yet..

src/api/coderApi.ts
```diff
@@ -74,38 +74,63 @@ export class CoderApi extends Api {
...
+		const defaults = this.getAxiosInstance().defaults;
+		const currentHost = defaults.baseURL;
+		const currentToken = defaults.headers.common[coderSessionTokenHeader];
 
-		if (currentToken !== token) {
+		defaults.baseURL = host;
+		defaults.headers.common[coderSessionTokenHeader] = token;
+
+		const hostChanged = currentHost !== host;
+		const tokenChanged = currentToken !== token;
```
code-asher, line 99: Do we need to handle the case of comparing an empty string to undefined?  That could trigger a reconnect or `suspend` but maybe going from falsey to falsey should be considered as no change.

src/api/coderApi.ts
```diff
@@ -198,68 +223,62 @@ export class CoderApi extends Api {
 				throw new Error("No base URL set on REST client");
 			}
 
-			const baseUrl = new URL(baseUrlRaw);
-			const token = this.getAxiosInstance().defaults.headers.common[
-				coderSessionTokenHeader
-			] as string | undefined;
+			return this.createOneWayWebSocket<TData>(socketConfigs);
```
code-asher, line 226: We have a check for the base URL just above this line but `createOneWayWebSocket` does this check so it seems like we no longer need the one above here.

Which would make `socketFactory` little more than a call to `createOneWayWebSocket`, perhaps that means something can be simplified here.

src/api/coderApi.ts
```diff
@@ -74,38 +74,63 @@ export class CoderApi extends Api {
...
+	 * Set both host and token together. Useful for login/logout/switch to
+	 * avoid triggering multiple reconnection events.
+	 */
+	setCredentials = (
+		host: string | undefined,
+		token: string | undefined,
+	): void => {
+		const defaults = this.getAxiosInstance().defaults;
+		const currentHost = defaults.baseURL;
+		const currentToken = defaults.headers.common[coderSessionTokenHeader];
```
code-asher, line 93: We grab these often enough that I wonder if we want some getters for the host and token?

src/api/coderApi.ts
```diff
@@ -74,38 +74,63 @@ export class CoderApi extends Api {
...
+
 	setHost = (host: string | undefined): void => {
-		const defaults = this.getAxiosInstance().defaults;
-		const currentHost = defaults.baseURL;
-		defaults.baseURL = host;
+		const currentToken = this.getAxiosInstance().defaults.headers.common[
+			coderSessionTokenHeader
+		] as string | undefined;
+		this.setCredentials(host, currentToken);
+	};
```
code-asher, line 122: `setHost` appears to be unused except in tests.  Can we remove it?

Part of me wonders if we should remove `setSessionToken` too and only expose `setCredentials` just to avoid the possibility of someone setting the host and then the token or vice versa, causing two reconnections with the first failing since it temporarily has the wrong token or url.

src/api/coderApi.ts
```diff
@@ -198,68 +223,62 @@ export class CoderApi extends Api {
...
+		if (enableRetry) {
+			return this.createReconnectingSocket(socketFactory, configs.apiRoute);
+		}
+		return socketFactory();
+	}
 
-			const httpAgent = await createHttpAgent(
-				vscode.workspace.getConfiguration(),
-			);
+	private async createOneWayWebSocket<TData>(
```
code-asher, line 235: The existence of `createOneWayWebSocket` to me implies that `createWebSocket` is two-way socket, but it actually also creates a one-way web socket, except potentially reconnecting.  Maybe we can have just `createOneWayWebSocket` and it handles `enableRetry`?

Or, we could rename `createWebSocket` to `createReconnectingOneWayWebSocket` or something like that.

src/api/coderApi.ts
```diff
@@ -198,68 +223,62 @@ export class CoderApi extends Api {
...
-			] as string | undefined;
+			return this.createOneWayWebSocket<TData>(socketConfigs);
+		};
 
-			const headersFromCommand = await getHeaders(
-				baseUrlRaw,
-				getHeaderCommand(vscode.workspace.getConfiguration()),
-				this.output,
-			);
+		if (enableRetry) {
```
code-asher, line 229: Nit: we call it a reconnecting socket so possibly this should be `enableReconnect`?

src/api/coderApi.ts
```diff
@@ -288,44 +307,79 @@ export class CoderApi extends Api {
...
-				options: configs.options,
-				enableRetry: configs.enableRetry,
-			});
-		} catch {
-			// Failed to create WebSocket, use SSE fallback
-			return this.createSseFallback<TData>(
-				configs.fallbackApiRoute,
-				configs.searchParams,
-				configs.options?.headers,
+	private async createWebSocketWithFallback(
```
code-asher, line 315: Nit: I know this is existing code, but it may not be clear what fallback means, could name this `createWebSocketWithSSEFalllback` maybe.

Or another pattern that might be cool:

```javascript
private webSocketFactory() {
 // returns a one way web socket after waiting for it to open
}
private withSSEFallback(factory) {
  // wraps factory with sse fallback
}
private withReconnect(factory) {
  // wraps factory with reconnecting
}
public createWebSocket(opts) {
  let factory = this.webSocketFactory
  if (opts.sseFallback) {
    factory = this.withSSEFallback(factory)
  }
  if (opts.reconnect) {
    factory = this.withReconnect(factory)
  }
  return factory()
}
```

However if we end up doing something like this we probably should do it in a separate PR IMO.

src/api/coderApi.ts
```diff
@@ -288,44 +307,79 @@ export class CoderApi extends Api {
...
+			logger: this.output,
+		});
+
+		this.attachStreamLogger(sse);
+		return sse;
 	}
 
-	private waitForConnection<TData>(
+	/**
+	 * Wait for a connection to open. Rejects on error.
```
code-asher, line 379: Oo nice yeah this seems much cleaner than passing in a handler.

src/api/coderApi.ts
```diff
@@ -358,32 +404,36 @@ export class CoderApi extends Api {
...
-		});
+			undefined,
+			() =>
+				this.reconnectingSockets.delete(
+					reconnectingSocket as ReconnectingWebSocket<unknown>,
+				),
+		);
+
+		this.reconnectingSockets.add(
+			reconnectingSocket as ReconnectingWebSocket<unknown>,
```
code-asher, line 433: I have not poked into it but seems weird we have to cast this?

src/api/coderApi.ts
```diff
@@ -457,7 +507,7 @@ function addLoggingInterceptors(client: AxiosInstance, logger: Logger) {
 		},
 		(error: unknown) => {
 			logError(logger, error, getLogLevel());
-			return Promise.reject(error);
+			throw error;
```
code-asher, line 510: Just to double check, the docs I saw do use `Promise.reject()`, are we guaranteed `throw` works as well?

src/api/oauthInterceptors.ts
```diff
@@ -0,0 +1,116 @@
...
+const coderSessionTokenHeader = "Coder-Session-Token";
+
+/**
+ * Attach OAuth token refresh interceptors to a CoderApi instance.
+ * This should be called after creating the CoderApi when OAuth authentication is being used.
+ *
+ * Success interceptor: proactively refreshes token when approaching expiry.
+ * Error interceptor: reactively refreshes token on 401 responses.
+ */
+export function attachOAuthInterceptors(
```
code-asher, line 19: Just an opinion so feel free to ignore, but to me it would feel more composable if we export the interceptor and then the caller attaches it, something like:

```
import { oauthInterceptor } ...
client.getAxiosInstance().interceptors.response.use(oauthInterceptor)
```

src/api/oauthInterceptors.ts
```diff
@@ -0,0 +1,116 @@
...
+	logger: Logger,
+	oauthSessionManager: OAuthSessionManager,
+): void {
+	client.getAxiosInstance().interceptors.response.use(
+		// Success response interceptor: proactive token refresh
+		(response) => {
+			// Fire-and-forget: don't await, don't block response
+			oauthSessionManager.refreshIfAlmostExpired().catch((error) => {
+				logger.warn("Proactive background token refresh failed:", error);
+			});
```
code-asher, line 30: Could you refresh my memory, we were going to switch to a background timer instead of doing this per-request right?  Or were we leaving that for a follow up?

src/api/oauthInterceptors.ts
```diff
@@ -0,0 +1,116 @@
...
+				};
+				if (config._oauthRetryAttempted) {
+					throw error;
+				}
+			}
+
+			const status = error.response?.status;
+
+			// These could indicate permanent auth failures that won't be fixed by token refresh
+			if (status === 400 || status === 403) {
```
code-asher, line 52: 403 I get but 400 has nothing to do with auth does it?

src/api/oauthInterceptors.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+async function handle401Error(
+	error: AxiosError,
+	client: CoderApi,
+	logger: Logger,
+	oauthSessionManager: OAuthSessionManager,
+): Promise<void> {
+	if (!oauthSessionManager.isLoggedInWithOAuth()) {
+		throw error;
```
code-asher, line 88: How do we get here?  Do we not only attach the oauth interceptor if oauth was used?

src/api/oauthInterceptors.ts
```diff
@@ -0,0 +1,116 @@
...
+		client.setSessionToken(newTokens.access_token);
+
+		logger.info("Token refresh successful, retrying request");
+
+		// Retry the original request with the new token
+		if (error.config) {
+			const config = error.config as RequestConfigWithMeta & {
+				_oauthRetryAttempted?: boolean;
+			};
+			config._oauthRetryAttempted = true;
```
code-asher, line 104: Is this config global?  Does that mean if you get a second 401 some time later we will no longer try to refresh?

If so, would it work to make a copy of `config` to send to `request` below?

src/api/oauthInterceptors.ts
```diff
@@ -0,0 +1,116 @@
...
+			config.headers[coderSessionTokenHeader] = newTokens.access_token;
```
code-asher, line 105: Does `client.setSessionToken` not already do this?

src/core/cliManager.ts
```diff
@@ -739,30 +738,27 @@ export class CliManager {
...
-		const tokenPath = this.pathResolver.getSessionTokenPath(label);
-		const [url, token] = await Promise.allSettled([
-			fs.readFile(urlPath, "utf8"),
-			fs.readFile(tokenPath, "utf8"),
-		]);
-		return {
-			url: url.status === "fulfilled" ? url.value.trim() : "",
-			token: token.status === "fulfilled" ? token.value.trim() : "",
-		};
+	private async atomicWriteFile(
```
code-asher, line 749: Seems unrelated to oauth right?   Should we merge it separately?  Just in case we need to revert or bisect in the future.

Second question, why does this need to be atomic?

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
-import type { SecretStorage, Disposable } from "vscode";
+import { toSafeHost } from "../util";
```
code-asher, line 1: It seems like we have two big changes in the PR, first changing the way secrets are managed, and second adding oauth.  How difficult do you think think it would be to split these into two PRs?  Again thinking about the potential revert or bisect in the future, as well as for review purposes.

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
 	 */
-	public async getSessionToken(): Promise<string | undefined> {
+	public async getCurrentDeployment(): Promise<Deployment | undefined> {
 		try {
-			return await this.secrets.get(SESSION_TOKEN_KEY);
+			const data = await this.secrets.get(CURRENT_DEPLOYMENT_KEY);
+			if (!data) {
+				return undefined;
+			}
+			const parsed = JSON.parse(data) as { deployment: Deployment | null };
```
code-asher, line 75: Could use `as CurrentDeploymentState`

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
-		if (!sessionToken) {
-			await this.secrets.delete(SESSION_TOKEN_KEY);
-		} else {
-			await this.secrets.store(SESSION_TOKEN_KEY, sessionToken);
-		}
+	public async setCurrentDeployment(
+		deployment: Deployment | undefined,
+	): Promise<void> {
+		const state = {
+			deployment: deployment ?? null,
```
code-asher, line 60: Is there a reason we store as null but the functions take and return undefined?

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+			if (e.key !== CURRENT_DEPLOYMENT_KEY) {
+				return;
+			}
+
+			try {
+				const data = await this.secrets.get(CURRENT_DEPLOYMENT_KEY);
+				if (data) {
+					const parsed = JSON.parse(data) as {
+						deployment: Deployment | null;
+					};
```
code-asher, line 99: Same here with `as`

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+					await listener({ deployment: parsed.deployment });
+				}
+			} catch {
+				// Ignore parse errors
```
code-asher, line 103: Comment mentions specifically parse errors but `listener` could also error so was it intended not to ignore listener errors?  Or is it intended to ignore both?

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+		listener: (state: CurrentDeploymentState) => void | Promise<void>,
+	): Disposable {
+		return this.secrets.onDidChange(async (e) => {
+			if (e.key !== CURRENT_DEPLOYMENT_KEY) {
+				return;
+			}
+
+			try {
+				const data = await this.secrets.get(CURRENT_DEPLOYMENT_KEY);
+				if (data) {
```
code-asher, line 96: Should we treat no data the same as a null deployment?  Not sure yet when/how it could get deleted, but if it did I imagine we would want to treat that as being logged out.  If we use `getCurrentDeployment` here instead that would do the trick too.

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+			}
+
+			try {
+				const data = await this.secrets.get(OAUTH_CALLBACK_KEY);
+				if (data) {
+					const parsed = JSON.parse(data) as OAuthCallbackData;
+					listener(parsed);
 				}
+			} catch {
+				// Ignore parse errors
```
code-asher, line 135: Both parse and listener errors here as well I imagine

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+		label: string,
+		listener: (auth: SessionAuth | undefined) => void | Promise<void>,
+	): Disposable {
+		const key = `${SESSION_KEY_PREFIX}${label}`;
+		return this.secrets.onDidChange(async (e) => {
+			if (e.key !== key) {
+				return;
+			}
+			const auth = await this.getSessionAuth(label);
+			await listener(auth);
```
code-asher, line 153: This time we are not catching listener errors, is that OK?

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+
+	public async getSessionToken(label: string): Promise<string | undefined> {
+		const auth = await this.getSessionAuth(label);
+		return auth?.token;
+	}
+
+	public async getUrl(label: string): Promise<string | undefined> {
+		const auth = await this.getSessionAuth(label);
+		return auth?.url;
+	}
```
code-asher, line 181: These feel a bit low utility to me, could we just call `getSessionAuth` when necessary and get rid of them?

src/core/deployment.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Represents a Coder deployment with its URL and label.
+ * The label is used as a unique identifier for storing credentials and configuration.
+ * It may be derived from the URL hostname (via toSafeHost) or come from SSH host parsing.
+ */
+export interface Deployment {
+	readonly url: string;
+	readonly label: string;
```
code-asher, line 8: I know we touched on this briefly but it does seem like we could omit `label` and derive it from the URL.

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+				// Ignore parse errors
 			}
 		});
 	}
+
+	/**
+	 * Listen for changes to a specific deployment's session auth.
+	 */
+	public onDidChangeSessionAuth(
+		label: string,
```
code-asher, line 144: I think all of these should take the deployment URL rather than the label, because storing these by label may make it harder to remove the label in the future.

Ultimately, the label is really only relevant for configuring the cli, so internally for the plugin we can store it with urls; we should (I think) only need to derive the label when configuring the cli and the ssh config.

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+			if (e.key !== key) {
+				return;
+			}
+			const auth = await this.getSessionAuth(label);
+			await listener(auth);
+		});
+	}
+
+	public async getSessionAuth(label: string): Promise<SessionAuth | undefined> {
+		if (!label) {
```
code-asher, line 158: Would this be blank if it was using the old format?  But if so always returning `undefined` seems like it would break old connections (well, it would just require re-authenticating I think, so no big deal).

Although, IMO, it is fine to break those at this point.  But then we should remove the concept of a label altogether and just go off the URL.

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
-const LOGIN_STATE_KEY = "loginState";
+import type { TokenResponse, ClientRegistrationResponse } from "../oauth/types";
 
-export enum AuthAction {
-	LOGIN,
-	LOGOUT,
-	INVALID,
+import type { Deployment } from "./deployment";
+
+const SESSION_KEY_PREFIX = "coder.session.";
```
code-asher, line 9: Thinking out loud, right now we have `$key_deployment` for each type of secret but would it make sense to have one secret for all the keys?

So instead of:

```
coder.session.dev.coder.com = "foo"
coder.oauth.tokens.dev.coder.com = "foo"

```

we have something like:

```
coder.deployments.dev.coder.com = JSON.stringify({
  session: "foo",
  oauth: { tokens: "foo" },
})
```

We could maybe have a function like `update(url: string, secrets: Partial<Secrets>)` or something like that, or we could still have the separate update functions like we have now.

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+			JSON.stringify(tokens),
+		);
+		await this.recordDeploymentAccess(label);
+	}
+
+	public async clearOAuthTokens(label: string): Promise<void> {
+		await this.secrets.delete(`${OAUTH_TOKENS_PREFIX}${label}`);
+	}
+
+	public async getOAuthClientRegistration(
```
code-asher, line 224: Definitely a nit, but I feel like we have a lot of these set/get/delete patterns doing the same thing with different keys that could maybe be squashed into something generic.

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+		const legacyUrl = this.memento.get<string>("url");
+		if (!legacyUrl) {
+			return undefined;
+		}
+
+		const label = toSafeHost(legacyUrl);
+
+		const existing = await this.getSessionAuth(label);
+		if (existing) {
+			return undefined;
```
code-asher, line 319: Should we delete the legacy `url` and token if we have existing auth?

src/core/secretsManager.ts
```diff
@@ -1,73 +1,338 @@
...
+		}
+
+		const oldToken = await this.secrets.get(LEGACY_SESSION_TOKEN_KEY);
+		if (!oldToken) {
+			return undefined;
+		}
+
+		await this.setSessionAuth(label, { url: legacyUrl, token: oldToken });
+		await this.secrets.delete(LEGACY_SESSION_TOKEN_KEY);
```
code-asher, line 328: Should we also delete the legacy `url`?

src/api/coderApi.ts
EhabY: I would consider this a change, since we can go from `undefined` to `""` due to mTLS authentication

src/api/coderApi.ts
EhabY: Yeah that could be better actually, currently we only use `setSessionToken` for convenience since there are a few cases where host does not change but the token does

src/api/coderApi.ts
EhabY: Ah I can actually use `ReconnectWebSocket<never>`

src/api/oauthInterceptors.ts
EhabY: There is both actually, the timer on per-request. Though maybe we should only keep the error request handler and skip the proactive refresh now 🤔

src/api/oauthInterceptors.ts
EhabY: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

> The authorization server responds with an HTTP 400 (Bad Request) status code (unless specified otherwise)

src/api/oauthInterceptors.ts
EhabY: We always attach because otherwise we have to track the current state and keep attaching and detaching. So we attach but only intercept if we are logged in with OAuth

src/api/oauthInterceptors.ts
EhabY: This config is per request, we re-use this config after setting the retry attempted so we don't get stuck in a retry loop (we pass it as is to `request(...)`

src/api/oauthInterceptors.ts
EhabY: This is for the old `config` - `client.setSessionToken` changes the client default headers but the config is from the pre-refresh and thus has the old headers.

src/core/cliManager.ts
EhabY: It's related because we support multiple deployments now by writing the CLI auth files on connection and not on login - In `remote.ts` we do `cliManager.configure` so if we have multiple windows they could attempt writing the same `session` and `url` files and thus cause a broken state somewhere

src/core/deployment.ts
EhabY: Can we always do `toSafeHost(url)`? The only "custom" label we have is `parts.label` but if that is always the same as the `toSafeHost(url)` then yes I agree this would simplify a lot of the logic here

src/core/secretsManager.ts
EhabY: Hmmm I can look into splitting this - mainly multiple deployments + secrets for one PR and all the oauth stuff in another

src/core/secretsManager.ts
EhabY: I did this at first but it was a headache to ensure concurrent writes and having reliable `onDidChange*` events. Because there is no atomic way of updating this so we have to do: `read` -> `modify` -> `write` which means if we have two windows that are updating two different settings we might drop some info (also events would need compare old and new values)

src/core/secretsManager.ts
EhabY: Just to make sure the value is set `deployment: null` vs. `<nothing>` but I suppose this is harmless might as well use `undefined`

src/core/secretsManager.ts
EhabY: Agreed, I'll remove all `label, url` pairs and just rely on `url` (we can derive the label when need be) - let's just hope URIs are not weird where sometimes they have a trailing slash (`/`) and sometimes they do not or sometimes they use `http` and sometimes `https`

Actually we might need to keep label for the secrets at least (or we can deduce it from the URL), because otherwise our secrets look something like `coder.session.https://dev.coder.com` compared to `coder.session.dev.coder.com`. I guess technically there's no limitation on the key format (I think) :thinking:

src/core/deployment.ts
code-asher: Hmm yeah so currently we get a remote host like `coder-vscode.dev.coder.com--username--workspace--agent` and we derive `label` from that to be `dev.coder.com`, then we read the URL and session token from the cli config based off the label.

So, I think what we could do is first instead of calling this `parts.label` we can call it `parts.safeHostname` or something like that.  But it is lacking the protocol which means we do still need at least one lookup.  We could grab it from the cli config still, or we could store a map of labels to urls.

But now that I am describing it, that does sound like all it does is introduce an extra lookup step lol.  I think for some reason the `label` name keeps throwing me off even though I wrote it hahaha.  Maybe we could call it `safeHostname`?  And throw a comment saying we have to look up things based on the safe hostname, because that is what we get from the remote authority which is all the context we get when connecting.  What do you think?

src/api/oauthInterceptors.ts
code-asher: Ah gotcha, I must not have gotten that far yet.  Yeah I feel like we only need the background one and the error one.

src/api/oauthInterceptors.ts
code-asher: But this comes from some API endpoint, not the refresh or exchange endpoints of the authorization server?  A 400 here would mean we made a bad API request, but the user is authorized (otherwise we would have returned a 401 instead).

code-asher: Should also mention that a 403 here would be a general forbidden error, nothing specific to oauth, so it may not make sense to try parsing it as an oauth error.

It would make sense to handle oauth errors on the refresh/exchange endpoints though, I have not seen if we are doing that though.

code-asher: Ohh wait this interceptor handles both regular API requests *and* the oauth requests??  lol this is what I get for not reading the whole thing yet 🤦  

That makes sense now, but it does feel odd to me to have the oauth error handling in the interceptor.  Would we not handle it where we make the request?  `try { exchangeToken() } catch () { handleOauthError() }` etc

src/api/oauthInterceptors.ts
code-asher: We have to set a flag for being logged in right?  Could we add/remove the interceptor whenever we do that?  Not a big deal of course, just conceptually to me it fits nicely.

src/api/oauthInterceptors.ts
code-asher: Ah perfect, yup I saw that it was used to prevent a loop, but was worried it was the global config.  Sounds good!

src/api/oauthInterceptors.ts
code-asher: Ahh gotcha and plus this is not the global config.  Makes sense.

src/core/cliManager.ts
code-asher: Ooo I see yeah that makes sense.  So related to the login changes, but not oauth.

Well, I suppose even before it could be an issue if one instance was connecting while another was configuring the cli.  So could still consider it as a separate fix.

src/core/secretsManager.ts
code-asher: Ooh yeah good point, if two instances are updating different parts of the settings.  Yeah what you have makes sense.  We might want to leave a comment explaining the architecture decision here.

src/core/secretsManager.ts
code-asher: nbd either way, was just curious.  Your reasoning makes sense to me.

src/core/secretsManager.ts
code-asher: Depending on the discussion above, maybe we do want to keep the label, but perhaps we can rename it to `safeHostname` or similar.

src/core/secretsManager.ts
code-asher: I forgot that the remote authority does not include the protocol so I guess we will need to look it up based on the "label" after all anyway... 😢

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
...
+	RefreshTokenRequestParams,
+	TokenRequestParams,
+	TokenResponse,
+	TokenRevocationRequest,
+} from "./types";
+
+const AUTH_GRANT_TYPE = "authorization_code" as const;
+const REFRESH_GRANT_TYPE = "refresh_token" as const;
+const RESPONSE_TYPE = "code" as const;
+const PKCE_CHALLENGE_METHOD = "S256" as const;
```
code-asher, line 34: Why do we need the `const` cast?  TypeScript being dumb in some way I imagine?

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
...
+	 * Only schedules the next check after the current one completes.
+	 */
+	private scheduleBackgroundRefresh(): void {
+		if (this.refreshTimer) {
+			clearTimeout(this.refreshTimer);
+		}
+
+		this.refreshTimer = setTimeout(async () => {
+			try {
+				await this.refreshIfAlmostExpired();
```
code-asher, line 181: If we only refresh when almost expired, why not set a timer directly for that time rather than checking in a loop?

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
...
+		}, BACKGROUND_REFRESH_INTERVAL_MS);
+	}
+
+	/**
+	 * Check if granted scopes cover all required scopes.
+	 * Supports wildcard scopes like "workspace:*".
+	 */
+	private hasRequiredScopes(grantedScope: string | undefined): boolean {
+		if (!grantedScope) {
+			// TODO server always returns empty scopes
```
code-asher, line 195: Yeah I think this was never implemented or at least not merged.

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
...
+ */
+const DEFAULT_OAUTH_SCOPES = [
+	"workspace:read",
+	"workspace:update",
+	"workspace:start",
+	"workspace:ssh",
+	"workspace:application_connect",
+	"template:read",
+	"user:read_personal",
+].join(" ");
```
code-asher, line 67: Did you have a way to test if these scopes were sufficient?  Not sure if there is a way right now, maybe if we try to generate a key manually with these scopes (I think this can be done on the account page) and paste it in to make sure everything works.

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
...
+	 * Prepare common OAuth operation setup: client, metadata, and registration.
+	 * Used by refresh and revoke operations to reduce duplication.
+	 */
+	private async prepareOAuthOperation(token?: string): Promise<{
+		axiosInstance: AxiosInstance;
+		metadata: OAuthServerMetadata;
+		registration: ClientRegistrationResponse;
+	}> {
+		const deployment = this.requireDeployment();
+		const client = CoderApi.create(deployment.url, token, this.logger);
```
code-asher, line 240: Oh hm so some oauth requests are not using the oauth interceptor that handles oauth errors?  Looks like exchange uses the plugin client which has the interceptor, and refresh and revoke use this one without the interceptor.  But it looks neither of those have their own error handling.

src/oauth/types.ts
```diff
@@ -0,0 +1,163 @@
...
+export type GrantType =
```
code-asher, line 2: Is pulling these types from `typesGenerated.ts` an option?

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
+import { type AxiosInstance } from "axios";
```
code-asher, line 1: Have we considered using an oauth client library?  I could see needing to add more parts of the spec in the future, or oidc or something, and a library could possibly handle it all for us.

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
...
+
+		const tokens: StoredOAuthTokens = {
+			...tokenResponse,
+			deployment_url: deployment.url,
+			expiry_timestamp: expiryTimestamp,
+		};
+
+		this.storedTokens = tokens;
+		await this.secretsManager.setOAuthTokens(deployment.label, tokens);
+		await this.secretsManager.setSessionAuth(deployment.label, {
```
code-asher, line 653: From the outside looking in, it feels weird to me that we set oauth tokens and session auth separately.  I think I would expect a single session auth object, and that object would have oauth information as well if it came from oauth.

Or to approach this another way, is there a use case for setting only oauth tokens but not the session auth?

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,808 @@
...
+
+		const axiosInstance = client.getAxiosInstance();
+		const metadataClient = new OAuthMetadataClient(axiosInstance, this.logger);
+		const metadata = await metadataClient.getMetadata();
+
+		// Only register the client on login
+		progress.report({ message: "registering client...", increment: 10 });
+		const registration = await this.registerClient(axiosInstance, metadata);
+
+		progress.report({ message: "waiting for authorization...", increment: 30 });
```
code-asher, line 374: Is it possible to add a cancel button?  If I close the browser window without authorizing it looks like I am stuck without a way to change my authorization method.  

Also if I click login and enter the URL again I wonder if it should kick off the authorization flow again, right now it does nothing so I have no way to reopen the window I closed.

And, if I close the window then try to log into a different deployment, the notification for the previous one still shows but we should probably cancel it when that happens (or disable the login button while auth is ongoing).

code-asher (commented):
I have not tested if this was an existing issue, but when I log in with a token my workspaces are not populated until I manually refresh.  Are you seeing the same thing?  

I have not tested logging in with oauth yet because xdg-open opens the wrong thing, but that is an issue with my system configuration.  😛   But, could be cool if there was an option in the flow to manually paste in the callback URI as a fallback.

src/api/coderApi.ts
code-asher: Could you expand on that use case?  If we are `undefined` then we call `socket.suspend()`, if we then go to `""` we call `socket.suspend()` a second time which I assume is a no-op?

code-asher: For the host, that is.

For the token, this would add a blank token to the headers but this has no effect in the mTLS case I think?  Even in the non-mTLS case the outcome would be the same (the server makes no distinction between empty or missing, both would be a 401).

src/api/coderApi.ts
EhabY: So I actually used `setCredentials` throughout but `setHost`/`setSessionToken` are defined in the super class actually. So they can still be called :/

src/api/coderApi.ts
EhabY: Ah yes, changing the host from `undefined` to `""` or vice-versa won't change anything. However, if we authenticate without a token (mTLS) then we might begin with the following: `{ host: XYZ, token: undefined }`, but when we authenticate we change the token to `""` and thus trigger a reconnect.

The input is the same since `undefined` might be treated the same as `""` except that we might have started with invalid configuration (401s) and then changed the setting and prompted a login again that worked so we trigger this reconnect with the new info.

src/api/coderApi.ts
EhabY: Right, done!

src/api/coderApi.ts
EhabY: Done!

src/api/coderApi.ts
EhabY: Still getting used to writing very `async` code in TS lol

src/api/coderApi.ts
EhabY: Done

src/api/coderApi.ts
EhabY: Oh true that is confusing, simplified this to `createOneWayWebSocket` and `createStreamWithSseFallback`, no need for reconnecting methods when we can just wrap it!

src/api/coderApi.ts
EhabY: Did the first option for now, the wrapping bit is smart :eye:

src/api/coderApi.ts
EhabY: I think they are equivalent but the rethrowing causes less modifications. Looking more into it I found [this](https://brandoncc.dev/blog/how-to-handle-unhandledrejection-errors-using-axios/) which suggests rethrowing since `Promise.reject` would create a new error

src/core/cliManager.ts
EhabY: I've separated this from the OAuth stuff and kept it in the multi-deployment PR because this has a much higher chance of happening when the new changes!

src/core/deployment.ts
EhabY: Yeah I do agree that we need it still because `url` -> `safeHostname` is not a reversible operation (we lose the protocol) so we always need to have some kind of lookup. Perhaps renaming this to `safeHostname` could be the way forward so it's less confusing and matches the function `toSafeHost(url)`

src/core/secretsManager.ts
EhabY: Split into 2, the first is this one: https://github.com/coder/vscode-coder/pull/677

src/core/secretsManager.ts
EhabY: Fair enough, I'll leave a comment explaining this. Also if we are afraid of losing track of those "dynamic" keys then newer VS Code API has a `keys()` that we can use to delete old secrets: https://code.visualstudio.com/api/references/vscode-api#SecretStorage

src/core/secretsManager.ts
EhabY: Right, done!

src/core/secretsManager.ts
EhabY: Ah you are right, we should report an event even if we have no data but somehow this is triggered

src/core/secretsManager.ts
EhabY: Removed this code

src/core/secretsManager.ts
EhabY: The `try/catch` is now for the listener only but I am wondering if we should log here since I do not think this is propagated elsewhere :thinking:

src/core/secretsManager.ts
EhabY: I think I'll rename this to `safeHostname` (maybe `safeHost`?) - As you've said above we still need this label somewhere to be able to access configurations using `parts.label` or `parts.safeHostname`. Otherwise we still need another lookup table :/ (Also the label/safeHostname is normalized which is a huge plus over URLs)

src/core/secretsManager.ts
EhabY: Added logging instead

src/core/secretsManager.ts
EhabY: > Although, IMO, it is fine to break those at this point. But then we should remove the concept of a label altogether and just go off the URL.

We have to at some point and as you've said they just have to login once to fix this. I should make this part of a minor instead of patch release (`0.12.0`)

src/core/secretsManager.ts
EhabY: Added logging instead

src/core/secretsManager.ts
EhabY: Removed both of those

src/core/secretsManager.ts
EhabY: Deleted!

src/core/secretsManager.ts
EhabY: We do this now, only if there's both

src/api/coderApi.ts
EhabY: Done!

src/core/secretsManager.ts
EhabY: Fixed

src/core/secretsManager.ts
EhabY: I'll do this when adding the OAuth back

src/api/coderApi.ts
code-asher: Oooooo that makes a lot of sense.  One of those promise gotchas.

src/api/coderApi.ts
EhabY: Resolved in https://github.com/coder/vscode-coder/pull/677

src/api/oauthInterceptors.ts
EhabY: Removed the success interceptor

src/core/secretsManager.ts
EhabY: Fixed in https://github.com/coder/vscode-coder/pull/677

src/core/secretsManager.ts
EhabY: Done

src/oauth/sessionManager.ts
EhabY: They aren't needed here so I removed them but sometimes they can enforce that types are exact strings instead of `string` type

src/oauth/sessionManager.ts
EhabY: Yup seems so, I'll keep this here for when it is added though we'll always return `true` for backward compatibility with older servers

src/oauth/sessionManager.ts
EhabY: > I think this can be done on the account page

I just checked and no such way unless I am missing an extended accounts page :/

But yes that's a smart idea, we need a way to test these scopes

src/oauth/types.ts
EhabY: I upgraded the `coder` API package and it doesn't include all the types :/, it does include some though. So I'll try to replace those that are available

src/oauth/sessionManager.ts
EhabY: Yes, I wanted the bare minimum implementation first, but I suppose I can something like https://www.npmjs.com/package/openid-client

src/oauth/sessionManager.ts
EhabY: The session auth is simply URL + current token, while the OAuth tokens contains more information like `expiry_timestamp` and `refresh_token` (Also potentially `scope`). Let me try to consolidate this into an optional OAuth field to distinguish OAuth from non-OAuth

src/oauth/sessionManager.ts
EhabY: Done, clicking cancel now would dispose everything. Also I've synchronized the logging in so you cannot initiate a new login without cancelling the current one (also removed the `inProgress` map since it's no longer needed)

src/api/oauthInterceptors.ts
EhabY: Yes you are right, I removed the handling of `400` and `403` from the interceptor here and only handled OAuth requests specifically

src/oauth/sessionManager.ts
EhabY: Added proper error handling

src/api/oauthInterceptors.ts
EhabY: Done!

src/api/oauthInterceptors.ts
EhabY: I've added `OAuthInterceptor` class that automatically attaches and detaches

src/oauth/sessionManager.ts
EhabY: So I've tried this approach just now but I'm thinking of reverting since it adds much more complexity, like we need to update on deployment change, update on token change, keep track of extra state as well..

EhabY: Here is the change BTW: https://github.com/coder/vscode-coder/pull/693/commits/76f089fdcf9b16967ba052875930be7179fd5e71

src/oauth/sessionManager.ts
EhabY: I'm thinking of doing this one in a follow up so we can land something that works first (with tests) then use this library for more reliability

src/oauth/sessionManager.ts
EhabY: Combined them in the following and you can see that we have certain racy calls that essentially do `getSessionAuth` then `setSessionAuth` which means we can overwrite with outdated information between multiple windows: https://github.com/coder/vscode-coder/pull/693/commits/3d1440b590c394ab746bed8716b433a4c54221fc

This is very unlikely to happen but it's possible and there's no way to guard against it unless we separate the fields or use file locks (inter-process communication). This actually makes me want to also separate `hostname` -> `{ url, token }` into `hostname` -> `url` (memento since it's not sensitive) and `hostname` -> `token` because there are cases where we want to set the URL but not the token (currently we just pass an `""` token)

src/oauth/sessionManager.ts
code-asher: Ah bugger yeah I thought maybe there was a page where scopes could be selected but I must be thinking of something else.

Maybe we can do it via the cli.  I will look into it.

PR #648, opened 2025-11-14
Remove node forge dependency

EhabY:
Replaced the `node-forge` dependency with `@peculiar/x509`, which is more modern, lightweight, and still widely adopted (600K+ downloads). I initially attempted to use Node.js's built-in `crypto` module, but encountered an issue where `keyUsage` consistently returned `undefined` in Electron environments.

**Testing in Electron Environments**

Added comprehensive testing for Electron environments. In the latest commit, all `vitest` tests now run in Electron, which mirrors the environment VS Code operates in. This introduces a few seconds of additional overhead compared to running them in pure NodeJS env. An alternative approach would be to create a separate command (e.g., `yarn test:electron`) that runs only the relevant tests in Electron, but this is less robust and could result in missed test coverage as the codebase evolves.

I also explored running these as VS Code integration tests, but this approach has significant limitations. It would require converting the test suite to Mocha and rewriting tests to use VS Code API calls, which would make it difficult to write focused unit tests.

**Electron Dependency**

Added `electron` as a dev dependency to ensure proper BoringSSL usage (rather than OpenSSL, which Node.js uses). Note that the current version is the latest release, while VS Code may be running an older version. We should consider pinning to an older Electron version that more closely matches VS Code's environment.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -80,30 +80,20 @@ export class CertificateError extends Error {
 				const url = new URL(address);
 				const socket = tls.connect(
 					{
-						port: parseInt(url.port, 10) || 443,
+						port: Number.parseInt(url.port, 10) || 443,
 						host: url.hostname,
 						rejectUnauthorized: false,
 					},
-					async () => {
+					() => {
 						const x509 = socket.getPeerX509Certificate();
 						socket.destroy();
 						if (!x509) {
 							throw new Error("no peer certificate");
 						}
 
-						/**
-						 * We use "@peculiar/x509" for two reasons:
-						 * 1. Node's x509 returns an undefined `keyUsage` in Electron environments.
-						 * 2. Electron's checkIssued() will fail because it suffers from same
-						 *    the key usage bug that we are trying to work around here in the first place.
-						 */
+						// We use "@peculiar/x509" because Node's x509 returns an undefined `keyUsage`.
 						const cert = new X509Certificate(x509.toString());
-						let isSelfIssued: boolean;
-						try {
-							isSelfIssued = await cert.isSelfSigned();
-						} catch (err) {
-							return reject(err);
-						}
+						const isSelfIssued = cert.subject === cert.issuer;
 						if (!isSelfIssued) {
 							return resolve(X509_ERR.PARTIAL_CHAIN);
 						}
--- a/test/unit/error.test.ts
+++ b/test/unit/error.test.ts
@@ -1,4 +1,9 @@
+import {
+	KeyUsagesExtension,
+	X509Certificate as X509CertificatePeculiar,
+} from "@peculiar/x509";
 import axios from "axios";
+import { X509Certificate as X509CertificateNode } from "node:crypto";
 import * as fs from "node:fs/promises";
 import https from "node:https";
 import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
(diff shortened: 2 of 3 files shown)


src/error.ts
```diff
@@ -84,36 +84,44 @@ export class CertificateError extends Error {
...
+						 * 2. Electron's checkIssued() will fail because it suffers from same
+						 *    the key usage bug that we are trying to work around here in the first place.
+						 */
+						const cert = new X509Certificate(x509.toString());
+						let isSelfIssued: boolean;
+						try {
+							isSelfIssued = await cert.isSelfSigned();
+						} catch (err) {
+							return reject(err);
+						}
```
EhabY: This is async because the package I used here does verification of the certificate as well:

```ts
public async isSelfSigned(crypto = cryptoProvider.get()): Promise<boolean> {
    return this.subject === this.issuer && await this.verify({ signatureOnly: true }, crypto);
  }
```

Do we need the signature verification?

EhabY: `node-forge` had this to say:

> Note that no signature check is performed.

So if we want to be consistent we can just do `cert.subject === cert.issuer`

src/error.ts
```diff
@@ -84,36 +84,44 @@ export class CertificateError extends Error {
...
-						// We use node-forge for two reasons:
-						// 1. Node/Electron only provide extended key usage.
-						// 2. Electron's checkIssued() will fail because it suffers from same
-						//    the key usage bug that we are trying to work around here in the
-						//    first place.
-						const cert = forge.pki.certificateFromPem(x509.toString());
-						if (!cert.issued(cert)) {
+						/**
+						 * We use "@peculiar/x509" for two reasons:
+						 * 1. Node's x509 returns an undefined `keyUsage` in Electron environments.
```
code-asher: Oh wait does Node return undefined `keyUsage` only inside Electron?  Outside Electron it does actually exist?

src/error.ts
code-asher: Hm yeah I think we do not need the signature verification probably.  We just want to know if it looks like a chain or not.

test/unit/error.test.ts
```diff
@@ -12,14 +12,13 @@ describe("Certificate errors", () => {
...
-	const isElectron =
-		(process.versions.electron || process.env.ELECTRON_RUN_AS_NODE) &&
-		!process.env.VSCODE_PID; // Running from the test explorer in VS Code
+	// These tests run in Electron (BoringSSL) for accurate certificate validation testing.
+
+	it("should run in Electron environment", () => {
+		const isElectron =
+			process.versions.electron || process.env.ELECTRON_RUN_AS_NODE;
+		expect(isElectron).toBeTruthy();
+	});
```
code-asher, line 24: nice! 👍

code-asher (approved):
awesome

src/error.ts
EhabY: Removed the signature verification, now we just do `this.subject === this.issuer`

src/error.ts
EhabY: Hmmm I tested in Electron only (through the test), let me actually check and maybe add a test case

EhabY: Apparently node as well, so I've updated the comment and added a test that parses a no-signing cert to verify this behavior (we can remove this dependency in the future if this is fixed)

code-asher: Smart!

PR #654, opened 2025-11-17
Add reconnecting websocket class and use it in CoderApi

### LINKED ISSUE #595: Add automatic WebSocket reconnection with exponential backoff
WebSocket streams occasionally drop due to client network loss or server-side interruptions. Today, many parts of the app rely on one-way WebSockets for push updates, and a closed socket isn’t re-established unless the user refreshes the window. This leads to stale UIs (e.g., the agent metadata status bar showing an unknown error) until a manual reload.

```
2025-09-25 17:27:02.604 [warning] Failed to query metadata: WebSocket closed unexpectedly: undefined undefined
```

**Proposal:**
Introduce an automatic reconnection policy for all one-way WebSocket clients:

* Reconnect with **exponential backoff + jitter**, capped (e.g., 250ms → 1s → 2s … up to 30s).
* **Stop** retrying on non-recoverable close codes (auth errors, 4xx) and surface a clear message.
* **Resume** prior subscriptions/state on reconnect; reset backoff after stable activity.
* Log reconnection attempts.

**Expected outcome:**
Sockets recover transparently from transient failures; users don’t need to refresh to restore live updates.

EhabY:
Closes #595

This PR implements a reconnecting WebSocket with exponential backoff that automatically uses the most up-to-date authentication information.

**Why not use existing packages?**

Third-party reconnecting WebSocket packages (e.g., [reconnecting-websocket](https://www.npmjs.com/package/reconnecting-websocket)) were considered but don't meet our requirements:
- Most packages are stale/unmaintained
- None support custom async authentication methods (we pass different options based on header commands or session tokens)
- Limited flexibility for future authentication expansion

**Additional features:**

- Automatic reconnection when session tokens or host values change, ensuring we always use current authentication info
- Potential future enhancement: Reconnect on proxy/auth/header setting changes (deferred for now)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
   Other windows now wait and display real-time progress instead of attempting concurrent downloads,
   preventing corruption and failures.
 
+### Changed
+
+- WebSocket connections now automatically reconnect on network failures, improving reliability when
+  communicating with Coder deployments.
+
 ## [v1.11.4](https://github.com/coder/vscode-coder/releases/tag/v1.11.4) 2025-11-20
 
 ### Fixed


src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+export type SocketFactory<TData> = () => Promise<UnidirectionalStream<TData>>;
+
+export type ReconnectingWebSocketOptions = {
+	initialBackoffMs?: number;
+	maxBackoffMs?: number;
+	jitterFactor?: number;
+};
+
+// 403 Forbidden, 410 Gone, 426 Upgrade Required, 1002/1003 Protocol errors
+const UNRECOVERABLE_CLOSE_CODES = new Set([403, 410, 426, 1002, 1003]);
```
EhabY: Part of me thinks we should do the opposite, having a white-list instead of a black-list. But eagerly reconnecting the websocket with exponential backoff **should** be harmless anyway 🤔

src/api/agentMetadataHelper.ts
```diff
@@ -53,7 +53,11 @@ export async function createAgentMetadataWatcher(
 				event.parsedMessage.data,
 			);
 
-			// Overwrite metadata if it changed.
+			if (watcher.error !== undefined) {
+				watcher.error = undefined;
+				onChange.fire(null);
+			}
```
DanielleMaywood, line 59: What is the context behind this change?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+		// connect() will close any existing socket
+		this.connect().catch((error) => {
+			if (!this.#isDisposed) {
+				this.#logger.warn(
+					`Manual reconnection failed for ${this.#apiRoute}: ${error instanceof Error ? error.message : String(error)}`,
+				);
+				this.scheduleReconnect();
+			}
+		});
+	}
```
jakehwll: nit: maybe just how i'd write it, but I like to return early..

can see you do it in the next function 😉 


```suggestion
		this.connect().catch((error) => {
			if (this.#isDisposed) return
			this.#logger.warn(
				`Manual reconnection failed for ${this.#apiRoute}: ${error instanceof Error ? error.message : String(error)}`,
			);
			this.scheduleReconnect();
		});
	}
```

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+						`WebSocket connection closed with unrecoverable error code ${event.code}`,
+					);
+					this.dispose();
+					return;
+				}
+
+				// Don't reconnect on normal closure
+				if (event.code === 1000 || event.code === 1001) {
+					return;
+				}
```
jakehwll, line 191: nit: magic numbers are magic.. `const` these?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+		options: ReconnectingWebSocketOptions = {},
+		onDispose?: () => void,
+	) {
+		this.#socketFactory = socketFactory;
+		this.#logger = logger;
+		this.#apiRoute = apiRoute;
+		this.#options = {
+			initialBackoffMs: options.initialBackoffMs ?? 250,
+			maxBackoffMs: options.maxBackoffMs ?? 30000,
+			jitterFactor: options.jitterFactor ?? 0.1,
```
DanielleMaywood, line 62: How did we come up with these default values? This is non-blocking, just curious 😄

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+
+	get url(): string {
+		return this.#currentSocket?.url ?? "";
+	}
+
+	addEventListener<TEvent extends WebSocketEventType>(
+		event: TEvent,
+		callback: EventHandler<TData, TEvent>,
+	): void {
+		(this.#eventHandlers[event] as Set<EventHandler<TData, TEvent>>).add(
```
DanielleMaywood: `as` scares me 👻 

Just want to sure these are required and unavoidable

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+		}
+
+		// connect() will close any existing socket
+		this.connect().catch((error) => {
+			if (!this.#isDisposed) {
+				this.#logger.warn(
+					`Manual reconnection failed for ${this.#apiRoute}: ${error instanceof Error ? error.message : String(error)}`,
+				);
+				this.scheduleReconnect();
+			}
```
DanielleMaywood:
```suggestion
			if (this.#isDisposed) {
				return;
			}

			this.#logger.warn(
				`Manual reconnection failed for ${this.#apiRoute}: ${error instanceof Error ? error.message : String(error)}`,
			);
			this.scheduleReconnect();
```

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+
+		// Fire close handlers synchronously before disposing
+		if (this.#currentSocket) {
+			this.executeHandlers("close", {
+				code: code ?? 1000,
+				reason: reason ?? "",
+				wasClean: true,
+				type: "close",
+				target: this.#currentSocket,
+			} as CloseEvent);
```
DanielleMaywood: Does this need `as CloseEvent`? Could we make use of `satisfies`?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+	private async connect(): Promise<void> {
+		if (this.#isDisposed || this.#isConnecting) {
+			return;
+		}
+
+		this.#isConnecting = true;
+		try {
+			// Close any existing socket before creating a new one
+			if (this.#currentSocket) {
+				this.#currentSocket.close(1000, "Replacing connection");
```
DanielleMaywood: Forgive me, maybe in WebSocket land the status codes are more well known but is it worth making these `const Status<Eh> = 1000`?

src/websocket/reconnectingWebSocket.ts
DanielleMaywood: Similar as above comment

src/websocket/reconnectingWebSocket.ts
```diff
@@ -0,0 +1,280 @@
...
+
+		this.#reconnectTimeoutId = setTimeout(() => {
+			this.#reconnectTimeoutId = null;
+			this.connect().catch((error) => {
+				if (!this.#isDisposed) {
+					this.#logger.warn(
+						`WebSocket connection failed for ${this.#apiRoute}: ${error instanceof Error ? error.message : String(error)}`,
+					);
+					this.scheduleReconnect();
+				}
```
DanielleMaywood:
```suggestion
				if (this.#isDisposed) {
					return;
				}

				this.#logger.warn(
					`WebSocket connection failed for ${this.#apiRoute}: ${error instanceof Error ? error.message : String(error)}`,
				);
				this.scheduleReconnect();
```

src/api/agentMetadataHelper.ts
EhabY: If we reach this place in the code that means we didn't encounter an error while parsing the message and thus we have a valid message so we clear the error (if any). The clients of this method usually display the error (if it exists) before attempting to access the value

src/websocket/reconnectingWebSocket.ts
EhabY: Oh very random and some AI "researched" values hahaha

src/websocket/reconnectingWebSocket.ts
EhabY: Great minds think a like, done!

src/websocket/reconnectingWebSocket.ts
EhabY: True, less nesting is always better :)!

src/websocket/reconnectingWebSocket.ts
EhabY: Thanks for pointing this out, this exposed a bigger issue with types, I made sure to narrow the types in `eventStreamConnection` so it's reflects the information we have and we don't do some nasty casting

src/websocket/reconnectingWebSocket.ts
EhabY: Understandable, I've removed all casting and instead relied on the all-mighty TS static analyzer (it just needed some help by setting the type):

```ts
readonly #eventHandlers: {
		[K in WebSocketEventType]: Set<EventHandler<TData, K>>;
	}
```

(Generics in TS are confusing ngl)

src/websocket/reconnectingWebSocket.ts
EhabY: Done!

src/websocket/reconnectingWebSocket.ts
EhabY: Added `codes.ts` that exposes these codes with some documentation

src/websocket/reconnectingWebSocket.ts
EhabY: Done

EhabY: Done

DanielleMaywood (approved):
I'm happy with this change, thanks for addressing the types issue 🙏

PR #656, opened 2025-11-18
Handle concurrent binary downloads using file locks

### LINKED ISSUE #575: Confusing error message: "Failed to read signature or binary"
## Description

Users are encountering a confusing error dialog when the VS Code extension fails to read a signature or binary file. The error message is unclear about what specifically failed and what the user should do.

Image

## Issues with Current Error
1. Title says "Failed to read signature" but message mentions "signature or binary" - unclear which failed
2. Error suggests file doesn't exist, but unclear if this is the signature file (.sha1) or the binary itself
3. No clear guidance on what caused the failure (download issue, file naming problem, etc.)
4. User is left guessing whether to "Run anyway" or try to fix the underlying issue

## Suggested Improvements
- Clarify whether the signature file or binary is missing
- Provide more context about what caused the failure
- Give clearer guidance on next steps
- Consider auto-retry for download failures

## Environment
- Reported on macOS (darwin-arm64)
- Multiple users have encountered this issue

Reported by: Danny Kopping

EhabY:
Closes #575

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Fixed
+
+- Fixed race condition when multiple VS Code windows download the Coder CLI binary simultaneously.
+  Other windows now wait and display real-time progress instead of attempting concurrent downloads,
+  preventing corruption and failures.
+
 ## [v1.11.4](https://github.com/coder/vscode-coder/releases/tag/v1.11.4) 2025-11-20
 
 ### Fixed


src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+		// Make sure we can execute this new binary.
+		const version = await cliUtils.version(binPath);
+		this.output.info("Downloaded binary version is", version);
+	}
 
+	/**
+	 * Unified handler for any binary-related failure.
+	 * Checks for existing or old binaries and prompts user once.
+	 */
+	private async handleAnyBinaryFailure(
```
EhabY, line 251: Maybe this is too much of a defensive programming, essentially we try to see if there is a binary and try to just execute that, otherwise, we search for an `.old-*` binary and try to use it

src/core/downloadProgress.ts
```diff
@@ -0,0 +1,48 @@
...
+	} catch {
+		return null;
+	}
+}
+
+export async function clearProgress(logPath: string): Promise<void> {
+	try {
+		await fs.rm(logPath, { force: true });
+	} catch (error) {
+		if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
```
mtojek: Should we add a comment when `ENOENT` may happen, and why is it ignored?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
-				const version = await cliUtils.version(binPath);
-				this.output.info("Downloaded binary version is", version);
+			// Step 2: Move temp to final location
+			this.output.info("Moving downloaded file to", path.basename(binPath));
+			await fs.rename(tempFile, binPath);
+		} catch (error) {
+			throw cliUtils.maybeWrapFileLockError(error, binPath);
+		}
+
+		// For debugging, to see if the binary only partially downloaded.
```
mtojek, line 235: Should we check any file checksums?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+		binPath: string,
+		tempFile: string,
+	): Promise<void> {
+		const oldBinPath =
+			binPath + ".old-" + Math.random().toString(36).substring(8);
+
+		try {
+			// Step 1: Move existing binary to backup (if it exists)
+			const stat = await cliUtils.stat(binPath);
+			if (stat) {
```
mtojek, line 220: what happens if `stat` is undefined? will `fs.rename` on L226 work?

src/core/binaryLock.ts
```diff
@@ -0,0 +1,112 @@
...
+							if (release) {
+								clearInterval(interval);
+								this.output.debug("Download completed by another process");
+								return resolve(release);
+							}
+						} catch (error) {
+							clearInterval(interval);
+							reject(error);
+						}
+					}, 500);
```
mtojek: nit: move to const?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+			this.output.info("Downloading binary from", binSource);
+
+			// Ideally we already caught that this was the right version and returned
+			// early, but just in case set the ETag.
+			const stat = await cliUtils.stat(binPath);
+			const etag = stat ? await cliUtils.eTag(binPath) : "";
+			this.output.info("Using ETag", etag || "<N/A>");
+
+			// Download the binary to a temporary file.
+			const writeStream = createWriteStream(tempFile, {
```
mtojek, line 332: What happens is the stream is chopped due to bad network connection? Should cliManager clean up temp files?

test/unit/core/downloadProgress.test.ts
```diff
@@ -0,0 +1,103 @@
...
+		});
+	});
+
+	describe("readProgress", () => {
+		it("should read progress from log file", async () => {
+			const expectedProgress = {
+				bytesDownloaded: 1500,
+				totalBytes: 10000,
+				status: "downloading",
+				timestamp: Date.now(),
```
mtojek:
```
export interface DownloadProgress {
	bytesDownloaded: number;
	totalBytes: number | null;
	status: "downloading" | "verifying";
}
```

shouldn't we add timestamp here?

src/core/downloadProgress.ts
```diff
@@ -0,0 +1,48 @@
+import { promises as fs } from "node:fs";
+import * as path from "node:path";
+
+export const STALE_TIMEOUT_MS = 15000;
```
mtojek: Is 15s an arbitrary value? You may add a comment on why exactly this :)

src/core/cliUtils.ts
```diff
@@ -97,6 +108,52 @@ export async function rmOld(binPath: string): Promise<RemovalResult[]> {
...
+): unknown {
+	const code = (error as NodeJS.ErrnoException).code;
+	if (code === "EBUSY" || code === "EPERM") {
+		return new FileLockError(binPath);
+	}
+	return error;
+}
+export function isFileLockError(error: unknown): boolean {
+	const code = (error as NodeJS.ErrnoException).code;
+	return code === "EBUSY" || code === "EPERM";
```
mtojek: I see you're checking error codes. Isn't there something like `FileLockError` instance or is it not enough?

test/unit/core/cliManager.concurrent.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+		// Verify binary exists and lock/progress files are cleaned up
+		await expect(fsPromises.access(binaryPath)).resolves.toBeUndefined();
+		await expect(fsPromises.access(binaryPath + ".lock")).rejects.toThrow();
+		await expect(
+			fsPromises.access(binaryPath + ".progress.log"),
+		).rejects.toThrow();
+	});
+
+	it("redownloads when version mismatch is detected concurrently", async () => {
```
mtojek, line 122: should we add more test cases?
* disk storage is insufficient
* connection timeouts

src/core/downloadProgress.ts
EhabY: Actually it seems that the `force` flag ignores this error anyway so we don't need it

src/core/cliUtils.ts
```diff
@@ -97,6 +108,52 @@ export async function rmOld(binPath: string): Promise<RemovalResult[]> {
...
+			.filter((f) => f.startsWith(binName) && f.includes(".old-"))
+			.map((f) => path.join(binDir, f));
+
+		// Sort by modification time, most recent first
+		const stats = await Promise.all(
+			oldBinaries.map(async (f) => ({
+				path: f,
+				mtime: (await fs.stat(f)).mtime,
+			})),
+		);
```
jakehwll: Would we want to use `Promise.allSettled()` here instead so we don't accidentally lose the entire dataset if `fs.stat(f)` were to fail?

src/core/binaryLock.ts
```diff
@@ -0,0 +1,112 @@
...
+	): Promise<() => Promise<void>> {
+		return await this.vscodeProposed.window.withProgress(
+			{
+				location: vscode.ProgressLocation.Notification,
+				title: "Another window is downloading the Coder CLI binary",
+				cancellable: false,
+			},
+			async (progress) => {
+				return new Promise<() => Promise<void>>((resolve, reject) => {
+					const interval = setInterval(async () => {
```
jakehwll: I don't love that this is an `setInterval(async () => { ... }` (async setIntervl), I think if we were to fail to read the `fileProgress` in `downloadProgress` due to system load or something we might run the code twice. 

In its current form this appears safe enough but we should probably wait til the code finishes running before scheduling another tick. YMMV

src/core/binaryLock.ts
```diff
@@ -0,0 +1,112 @@
...
+									currentProgress.status === "verifying"
+										? "Verifying signature..."
+										: `${prettyBytes(currentProgress.bytesDownloaded)} / ${totalBytesPretty}`;
+								progress.report({ message });
+							}
+
+							const release = await this.safeAcquireLock(binPath);
+							if (release) {
+								clearInterval(interval);
+								this.output.debug("Download completed by another process");
```
mafredri: If we acquire the lock, could it also mean that the other process failed to download? Do we need to handle that case separately?

src/core/cliUtils.ts
```diff
@@ -97,6 +108,52 @@ export async function rmOld(binPath: string): Promise<RemovalResult[]> {
...
+	error: unknown,
+	binPath: string,
+): unknown {
+	const code = (error as NodeJS.ErrnoException).code;
+	if (code === "EBUSY" || code === "EPERM") {
+		return new FileLockError(binPath);
+	}
+	return error;
+}
+export function isFileLockError(error: unknown): boolean {
```
mafredri: Nit: I don't see this being used anywhere?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+	/**
+	 * Prompt the user to use an existing binary version.
+	 */
+	private async promptUseExistingBinary(
+		version: string,
+		reason: string,
+	): Promise<boolean> {
+		const choice = await this.vscodeProposed.window.showErrorMessage(
+			`${reason}. Use version ${version} anyway?`,
+			"Run",
```
mafredri, line 201:
```suggestion
			"Yes",
```
Feels like a more clear answer to the question, but feel free to dismiss if you disagree. Alt: Use -> Run above.

src/core/cliManager.ts
mafredri: I believe signature verification should fail if checksum doesn't match.

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+				: "Failed to update CLI binary";
+
+		// Try existing binary first
+		const existingCheck = await this.checkBinaryVersion(
+			binPath,
+			expectedVersion,
+		);
+		if (existingCheck.version) {
+			// Perfect match - use without prompting
+			if (existingCheck.matches) {
```
mafredri, line 268: If this is true, why did we ever try to download and fail?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
 				return binPath;
 			}
-			case 304: {
-				this.output.info("Using existing binary since server returned a 304");
+			// Version mismatch - prompt user
+			if (await this.promptUseExistingBinary(existingCheck.version, message)) {
```
mafredri, line 272: Unless I misread, this prompt says Run/Exit. If I select Exit I would not expect old binary to be used. Consider changing terminology?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+			const oldCheck = await this.checkBinaryVersion(
+				oldBinaries[0],
+				expectedVersion,
+			);
+			if (
+				oldCheck.version &&
+				(oldCheck.matches ||
+					(await this.promptUseExistingBinary(oldCheck.version, message)))
+			) {
+				await fs.rename(oldBinaries[0], binPath);
```
mafredri, line 290: Let's assume `binPath` exists already, but the existing check did not match, and the user selected `Exit` in prompt. Then we fallback to old binary and ask the user again. Now the rename may fail and we throw an error.

Why do we have to rename vs using the old binary path as-is? Too many hard-coded references to the non-old path?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+		// Try .old-* binaries as fallback
+		const oldBinaries = await cliUtils.findOldBinaries(binPath);
+		if (oldBinaries.length > 0) {
+			const oldCheck = await this.checkBinaryVersion(
+				oldBinaries[0],
+				expectedVersion,
+			);
+			if (
+				oldCheck.version &&
+				(oldCheck.matches ||
```
mafredri, line 287: Why do we only try _one_ old binary when we are tracking all of them?

I'm trying to think about _when_ this condition actually might happen. I.e. we have an old binary _and_ it's the right version. This means you either downgraded coderd or switched deployment. In either case, _any_ of the old binaries may be the correct one.

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+		parsedVersion: semver.SemVer,
+		binPath: string,
+		progressLogPath: string,
+	): Promise<string> {
+		const cfg = vscode.workspace.getConfiguration("coder");
+		const tempFile =
+			binPath + ".temp-" + Math.random().toString(36).substring(8);
+
+		try {
+			const removed = await cliUtils.rmOld(binPath);
```
mafredri, line 310: Why do we remove the old before our download has completed successfully? Is this to ensure that we don't try to download in case updating the binary would fail (e.g. in use on Windows)?

src/core/cliManager.ts
```diff
@@ -97,143 +103,342 @@ export class CliManager {
...
+					// Replace existing binary (handles both renames + Windows lock)
+					await this.replaceExistingBinary(binPath, tempFile);
+
+					return binPath;
+				}
+				case 304: {
+					this.output.info("Using existing binary since server returned a 304");
+					return binPath;
+				}
+				case 404: {
```
mafredri, line 395: Theoretically, this could also be caused by a poorly configured reverse proxy. I'm just raising a thought, don't think we need to handle it here. There's no easy way to check this either (e.g. no `https://dev.coder.com/bin/.exists`)

src/core/downloadProgress.ts
EhabY: Understandable, added some documentation!

src/core/downloadProgress.ts
EhabY: Removed the `try/catch`!

src/core/cliUtils.ts
EhabY: Oh, I created `maybeWrapFileLockError` but apparently didn't delete this old method... Removed

src/core/cliUtils.ts
EhabY: I removed this, but this is to convert a NodeJS error into a potential `FileLockError` so we can check the instance later on

src/core/cliUtils.ts
EhabY: Oh good idea, we could at least attempt to run one of them, I've replaced this with:

```ts
const stats = await Promise.allSettled(
	oldBinaries.map(async (f) => ({
		path: f,
		mtime: (await fs.stat(f)).mtime,
	})),
).then((result) =>
	result
		.filter((promise) => promise.status === "fulfilled")
		.map((promise) => promise.value),
);
```

We could potentially log here for file that could not be read, similar to the `rmOld` 🤔

src/core/binaryLock.ts
EhabY: Actually you are right yeah, if this was a takeover it could mean that the other process was stuck. The logic that follows is correct though since we essentially recheck if we have the right binary and attempt to download if need be. 

We even log right after "Acquired download lock". So I'll just remove this logging here.

src/core/binaryLock.ts
EhabY: Done!

src/core/binaryLock.ts
EhabY: Nice catch! 

I updated this to only schedule the next tick if everything is successful.

test/unit/core/downloadProgress.test.ts
EhabY: Oh I failed to update the tests, in the first iteration I had a timestamp in there but it wasn't used since now we depend on the lock file staleness (which is handled by `proper-lockfile`)

src/core/cliManager.ts
EhabY: I prefer to keep "Run" since we use that if we want to skip verification. Though I do like you alt suggestion

src/core/cliManager.ts
EhabY: Yes, it's only `undefined` if the file does not exist (e.g. if we start without a binary to begin with) - The idea here is (re)move the binary so we can create on with the same name

src/core/cliManager.ts
mtojek: Agree, I'm curious if we should switch to checksums rather than file length

EhabY: Indeed, but signature verification could be disabled in the settings. Tbh this is redundant since the following `cliUtils.version` would fail if this file was corrupted anyway 🤔

src/core/cliManager.ts
EhabY: This could be just a lot of defensive programming, see https://github.com/coder/vscode-coder/pull/656/files#r2538524344

We could have encountered an error at any stage here and another process download the binary

src/core/cliManager.ts
EhabY: Hmmm, yeah if you click "Cancel" you might get another prompt with an even older version. We should perhaps show this prompt once only for the first match (whether it's `binPath` or an old binary).

I added `throw error;` so that if `binPath` exists we never even attempt to read old binaries.

src/core/cliManager.ts
EhabY: We also clear old binaries when we download new ones so it's unlikely that we have multiple (it could happen if we never remove them because of some errors). Different deployments don't apply here since each deployment has it's own folder, but yes if you downgrade then getting the most recent one might not matter. Should we just get the first match whatever it is?

src/core/cliManager.ts
EhabY: The flow here is a bit sketchy I agree, I've made such that if you reject `binPath` then we do not attempt to read old binaries since to me that makes even less sense... (if you don't want to run the most up-to-date binary, why would you run an older one?)

See https://github.com/coder/vscode-coder/pull/656/files#r2545981797

We rename because a lot of the logic depends on the proper name like removing old binaries but not touching `binPath`

src/core/cliManager.ts
EhabY: Because we remove old binaries only, we still keep the proper binary (`binPath`). Old binaries should have been removed at the end of the previous download but are done here because there's less conflicts (this was the case already from before).

For example this is what might the folder look like:

* `coder-linux`
* `coder-linux.old-123`

So we remove the "old" binaries only and keep the most recent one if the download fails for example..

src/core/cliManager.ts
EhabY: It would be cleaned up on the next run (see `rmOld`). This was already the behavior but I suppose it's so there's less conflicts and file locks etc.. (clean up on next run vs. clean up immediately)

src/core/cliManager.ts
EhabY: That is true, do keep in mind that their `Coder` logs would contain the errored request on the error level. So potentially they can quickly identify this using the logs 🤔

EhabY:
> would IPC communication between VS Code windows have been an option here, like with the login prompt?

@mafredri We could maybe use this for the progress monitoring but we definitely need a lock file since that would make crashes/staleness/stuck handling much simpler. Libraries like `proper-filelock` are already implemented to handle this. IPC communication using secrets might get a bit tricky and cumbersome to work with. Also, we already writing the binary to the file system, it only feels natural to use a filelock in the same folder :wink:

src/core/cliManager.ts
mtojek: Hm.. even if the file has the same name, the same size, but the content is malicious?

EhabY: Yes, if `disableSignatureVerification` is `true` or the user clicks "Run without verification" (if verification fails) then there is no other check that I know of..

test/unit/core/cliManager.concurrent.test.ts
EhabY: Added tests!

mtojek (approved):
Thanks for addressing my comments! As long as other reviewers approve, you're good to go 👍

jakehwll (approved):

PR #663, opened 2025-11-26
feat: use environment variables for CLI auth instead of file flags

ibetitsmike:
## Summary

Pass `CODER_URL` and `CODER_SESSION_TOKEN` via SSH `SetEnv` directive instead of using `--session-token-file` and `--url-file` CLI flags. This simplifies the auth flow by using environment variables that the CLI natively supports.

## Changes

- Remove `--session-token-file` and `--url-file` from `vscodessh` ProxyCommand
- Add `CODER_URL` and `CODER_SESSION_TOKEN` to SSH `SetEnv` directive
- Update doc comments to reflect persistence-only purpose of file storage
- Keep file-based storage for extension's own credential persistence across sessions

## How it works

Previously:
```
ProxyCommand coder vscodessh --session-token-file /path/to/session --url-file /path/to/url %h
```

Now:
```
ProxyCommand coder vscodessh %h
SetEnv CODER_URL=https://example.coder.com CODER_SESSION_TOKEN=xxx CODER_SSH_SESSION_TYPE=vscode
```

The CLI reads credentials from environment variables, which SSH passes via `SetEnv`.

### DIFF
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -479,6 +479,8 @@ export class Remote {
 					binaryPath,
 					logDir,
 					featureSet,
+					baseUrlRaw,
+					token,
 				);
 			} catch (error) {
 				this.logger.warn("Failed to configure SSH", error);
@@ -597,6 +599,8 @@ export class Remote {
 		binaryPath: string,
 		logDir: string,
 		featureSet: FeatureSet,
+		url: string,
+		token: string,
 	) {
 		let deploymentSSHConfig = {};
 		try {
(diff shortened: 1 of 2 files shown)


src/remote/remote.ts
```diff
@@ -677,9 +681,7 @@ export class Remote {
 			? `${escapeCommandArg(binaryPath)}${globalConfigs} ssh --stdio --usage-app=vscode --disable-autostart --network-info-dir ${escapeCommandArg(this.pathResolver.getNetworkInfoPath())}${await this.formatLogArg(logDir)} --ssh-host-prefix ${hostPrefix} %h`
 			: `${escapeCommandArg(binaryPath)}${globalConfigs} vscodessh --network-info-dir ${escapeCommandArg(
 					this.pathResolver.getNetworkInfoPath(),
-				)}${await this.formatLogArg(logDir)} --session-token-file ${escapeCommandArg(this.pathResolver.getSessionTokenPath(label))} --url-file ${escapeCommandArg(
-					this.pathResolver.getUrlPath(label),
-				)} %h`;
```
EhabY, line 682: This removes the session token and URL file from being passed to `vscodessh` which is the **old** client that I think is only used by some older servers (`< 2.19.0`)

src/remote/remote.ts
```diff
@@ -690,9 +692,9 @@ export class Remote {
 			LogLevel: "ERROR",
 		};
 		if (sshSupportsSetEnv()) {
```
EhabY, line 694: Wouldn't this just fail if the SSH does not support `SetEnv`? is it possible to have this as a fallback?

src/remote/remote.ts
```diff
@@ -690,9 +692,9 @@ export class Remote {
...
-			// This allows for tracking the number of extension
-			// users connected to workspaces!
-			sshValues.SetEnv = " CODER_SSH_SESSION_TYPE=vscode";
+			// Pass Coder URL, session token, and session type via environment.
+			// The CLI reads CODER_URL and CODER_SESSION_TOKEN from the environment.
+			sshValues.SetEnv = ` CODER_URL=${url} CODER_SESSION_TOKEN=${token} CODER_SSH_SESSION_TYPE=vscode`;
```
EhabY, line 697: Here we still have the token in plaintext but it's now in `<globalDir>/<deployment url>/session` and in the proxy command (somewhere in `~/.ssh/config`?)

src/remote/remote.ts
EhabY, line 744: If we want to rely on environment variables fully then we need to update the `getGlobalFlags` function to not return `"--global-config", escapeCommandArg(configDir)`

PR #665, opened 2025-12-01
Refactor SSH process monitoring to support VS Code forks

### LINKED ISSUE #660: `Coder: View Logs` does not work on non-MS VS Code IDEs
**Description:**

The current implementation for retrieving Remote - SSH channel logs returns `undefined` for non-Microsoft VS Code forks.

https://github.com/coder/vscode-coder/blob/118d50ab4a2ad9f3490a05cc17163838ab3de134/src/remote/remote.ts#L898-L905

**Issue:**

VS Code forks use different extension identifiers for Remote SSH:
- **Cursor:** `anysphere.remote-ssh`
- **Google Antigravity:** `google.antigravity-remote-openssh`
- **Other forks:** Likely affected as well

These forks store `Remote - SSH` logs in folders matching their extension names rather than the standard Microsoft folder structure. The log files still contain "Remote - SSH" in their filenames.

**Solution:**

The log retrieval logic should check for fork-specific extension folders when searching for Remote - SSH logs.

EhabY:
Extract SSH process discovery and network status display into a
dedicated `SshProcessMonitor` class. Add centralized Remote SSH extension
detection to support Cursor, Windsurf, and other VS Code forks.

Key changes:
- Extract SSH monitoring logic from `remote.ts` into `sshProcess.ts`
- Add `sshExtension.ts` to detect installed Remote SSH extension
- Use `createRequire` instead of private `module._load` API to load the Remote SSH extension
- Fix port detection to find most recent port (handles reconnects)
- Add Cursor's "Socks port:" log format to port regex

Closes #660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 - WebSocket connections now automatically reconnect on network failures, improving reliability when
   communicating with Coder deployments.
+- Improved SSH process and log file discovery with better reconnect handling and support for
+  VS Code forks (Cursor, Windsurf, Antigravity).
 
 ## [v1.11.4](https://github.com/coder/vscode-coder/releases/tag/v1.11.4) 2025-11-20
 


src/remote/remote.ts
```diff
@@ -741,172 +736,27 @@ export class Remote {
...
+				const newLogDir = this.getLogDir(featureSet);
+				if (newLogDir !== currentLogDir) {
+					vscode.window
+						.showInformationMessage(
+							"Log directory configuration changed. Reload window to apply.",
+							"Reload",
+						)
+						.then((action) => {
+							if (action === "Reload") {
+								vscode.commands.executeCommand("workbench.action.reloadWindow");
```
EhabY: If the user changes the log file location AFTER we connect, we allow them to reload if they want to use this new location

src/util.ts
```diff
@@ -13,27 +13,29 @@ export interface AuthorityParts {
...
+	if (allMatches.length === 0) {
 		return null;
 	}
-	if (matches.length < 2) {
-		return null;
-	}
-	const portStr = matches[1] || matches[2] || matches[3];
+
+	const lastMatch = allMatches.at(-1)!;
+	const portStr = lastMatch[1] || lastMatch[2] || lastMatch[3] || lastMatch[4];
```
EhabY, line 41: We now get the last match in case of reconnects and old logs - I do not love log file parsing to get the port but that's the only **reliable** way since we might have different VS Code windows running with different deployments at the same time

src/remote/sshProcess.ts
```diff
@@ -0,0 +1,421 @@
...
+/**
+ * Options for creating an SshProcessMonitor.
+ */
+export interface SshProcessMonitorOptions {
+	sshHost: string;
+	networkInfoPath: string;
+	proxyLogDir?: string;
+	logger: Logger;
+	// Poll interval for SSH process and file discovery
+	pollInterval?: number;
```
EhabY: Should this one be exponential? Right now it will keep retrying every second forever if it cannot find the SSH process :/

src/remote/sshProcess.ts
```diff
@@ -0,0 +1,421 @@
...
+	 * When found, starts monitoring.
+	 */
+	private async searchForProcess(): Promise<void> {
+		const { pollInterval, logger, sshHost } = this.options;
+		let attempt = 0;
+
+		while (!this.disposed) {
+			attempt++;
+
+			if (attempt % 10 === 0) {
```
EhabY:
```suggestion
			if (attempt === 1 || attempt % 10 === 0) {
```

src/remote/sshProcess.ts
```diff
@@ -0,0 +1,421 @@
...
+				const stats = await fs.stat(networkInfoFile);
+				const ageMs = Date.now() - stats.mtime.getTime();
+
+				if (ageMs > staleThreshold) {
+					// Prevent tight loop: if we just searched due to stale, wait before searching again
+					const timeSinceLastSearch = Date.now() - this.lastStaleSearchTime;
+					if (timeSinceLastSearch < staleThreshold) {
+						await this.delay(staleThreshold - timeSinceLastSearch);
+						continue;
+					}
```
EhabY, line 308: This is the reconnection detection logic, the issue here is that we do not get events for disconnecting or connecting and thus we have no way to subscribe to them. So what I did here, was assume a connection is stale if the network stats file (generated by the CLI) does not change in `poll * 5` seconds - If a network is stale and we find the same process then nothing would break so it's harmless, just means we have to search again.

src/remote/sshExtension.ts
```diff
@@ -0,0 +1,25 @@
+import * as vscode from "vscode";
+
+export const REMOTE_SSH_EXTENSION_IDS = [
+	"jeanp413.open-remote-ssh",
+	"codeium.windsurf-remote-openssh",
+	"anysphere.remote-ssh",
+	"ms-vscode-remote.remote-ssh",
+	"google.antigravity-remote-openssh",
```
mafredri, line 8: Sanity-check: Is this the order we want to prioritize? Can multiple be active at the same time?

src/util.ts
mafredri: A comment explaining what match group each of these represent may be helpful for future eyes.

src/remote/sshProcess.ts
```diff
@@ -0,0 +1,421 @@
...
+		});
+	}
+
+	/**
+	 * Searches for the SSH process indefinitely until found or disposed.
+	 * Tries port-based discovery first (more accurate), falls back to hostname-based.
+	 * When found, starts monitoring.
+	 */
+	private async searchForProcess(): Promise<void> {
+		const { pollInterval, logger, sshHost } = this.options;
```
mafredri: `sshHost` only seems to be used for logging, don't we need to filter anything based on it?

src/remote/sshProcess.ts
```diff
@@ -0,0 +1,421 @@
...
+			this.pendingTimeout = setTimeout(() => {
+				this.pendingTimeout = undefined;
+				resolve();
+			}, ms);
+		});
+	}
+
+	/**
+	 * Searches for the SSH process indefinitely until found or disposed.
+	 * Tries port-based discovery first (more accurate), falls back to hostname-based.
```
mafredri: Is this true, since `sshHost` isn't used?

src/util.ts
```diff
@@ -13,27 +13,29 @@ export interface AuthorityParts {
...
 /**
- * Given the contents of a Remote - SSH log file, find a port number used by the
- * SSH process. This is typically the socks port, but the local port works too.
+ * Given the contents of a Remote - SSH log file, find the most recent port
+ * number used by the SSH process. This is typically the socks port, but the
+ * local port works too.
  *
  * Returns null if no port is found.
  */
 export function findPort(text: string): number | null {
```
mafredri, line 30: Suggestion: Consider if we want to add tests for this? Seeing true log output would also help a perspective viewer/author to understand what's going on.

src/remote/sshProcess.ts
```diff
@@ -0,0 +1,421 @@
...
+			const remoteSSHLog = files.find((f) => f.includes("Remote - SSH"));
+			if (remoteSSHLog) {
+				return path.join(outputPath, remoteSSHLog);
+			}
+		}
+	} catch {
+		// output_logging folder doesn't exist
+	}
+
+	return undefined;
```
mafredri, line 446: Minor concern, if our `try`s always fail, we'll end up in a polling interval, never informing the user of anything (`searchForProcess`)?

src/remote/sshProcess.ts
mafredri: If made exponential, do put a ceiling in, like 30s.

src/remote/sshProcess.ts
```diff
@@ -0,0 +1,421 @@
...
+		});
+		return monitor;
+	}
+
+	/**
+	 * Returns the path to the log file for this connection, or undefined if not found.
+	 */
+	getLogFilePath(): string | undefined {
+		return this.logFilePath;
+	}
```
mafredri, line 105: Is there any reason to use a getter here vs accessing the property directly?

src/remote/remote.ts
```diff
@@ -741,172 +736,27 @@ export class Remote {
...
-			// Then we search the remote SSH log until we find the port.
-			const result = await search(filePath);
-			if (!result) {
-				return new Promise((resolve) => setTimeout(() => resolve(loop()), 500));
+	private watchLogDirSetting(
+		currentLogDir: string,
+		featureSet: FeatureSet,
+	): vscode.Disposable {
+		return vscode.workspace.onDidChangeConfiguration((e) => {
+			if (e.affectsConfiguration("coder.proxyLogDirectory")) {
```
mafredri: Suggestion: Consider early return over nesting, makes the code easier to read.

EhabY: True, done!

src/remote/sshExtension.ts
EhabY: We cannot have multiple at the same time (I think they would conflict with each other and break both extensions)

src/remote/sshProcess.ts
EhabY: Makes it immutable from outside I guess

src/remote/sshProcess.ts
EhabY: Oh you are right, this was part of another experiment that I did. I'll fix the documentation here

src/remote/sshProcess.ts
EhabY: We do not no, the port is sufficient (this is based on the old logic that was refactored here). Using the hostname to search was an approach I tested out but it's worse because we can have the same host connected on multiple windows so there's no way to tell them apart...

src/remote/sshProcess.ts
EhabY: They would get `SSH process search attempt ${attempt} for host: ${sshHost}`, do you mean that they won't know it failed at that process exactly? 🤔

src/util.ts
EhabY: Rightt, I added some tests!

src/util.ts
EhabY: Add some documentation!

src/util.ts
```diff
@@ -34,7 +33,11 @@ export function findPort(text: string): number | null {
 		return null;
 	}
 
+	// Get the last match, which is the most recent port.
 	const lastMatch = allMatches.at(-1)!;
+	// Each capture group corresponds to a different Remote SSH extension log format:
+	// [0] full match, [1] and [2] ms-vscode-remote.remote-ssh,
+	// [3] windsurf/open-remote-ssh/antigravity, [4] anysphere.remote-ssh
```
mafredri, line 40: ❤️

test/unit/util.test.ts
```diff
@@ -124,3 +129,46 @@ describe("countSubstring", () => {
...
+
+	it("returns most recent port when multiple matches exist", () => {
+		const log = `
+[10:30:00] Starting connection -> socksPort 1111 -> initialized
+[10:30:05] Reconnecting => 2222(socks) => retry
+[10:30:10] Final connection Socks port: 3333 established
+		`;
+		expect(findPort(log)).toBe(3333);
+	});
+});
```
mafredri, line 240: This is awesome, thanks for adding it!

src/remote/sshProcess.ts
mafredri: Should we remove it from the options to simplify or still valuable as a log? Either way, up to you 👍🏻.

src/remote/sshProcess.ts
mafredri: I suppose it's more like a potential danger that we never stop trying even if it's _never_ going to match (maybe some expectation changes in the future, etc). And the user has to realize that even though it keeps logging that, it will never resolve.

EhabY: I'll go through the different pollings and try to make sure we have sufficient logging (but not too much). So at least if there's a failure here the users know why it might have happened (in this case, it could be that the `Remote - SSH` log location was changed or somehow never written)

src/remote/sshProcess.ts
EhabY: I kept it for now so it's easier to look at the logs if there are issues there

src/remote/sshProcess.ts
EhabY: Added logging

src/remote/sshProcess.ts
EhabY: Made it exponential with a default ceiling of 30s

PR #667, opened 2025-12-04
Expand tilde in Coder settings paths

### LINKED ISSUE #417: Proxy log directory does not expand tilde
If you set it to `~/logs`  for example, it tries to literally create `~`.

EhabY:
Closes #417

### DIFF
--- a/test/unit/util.test.ts
+++ b/test/unit/util.test.ts
@@ -1,6 +1,13 @@
+import os from "node:os";
 import { describe, it, expect } from "vitest";
 
-import { countSubstring, parseRemoteAuthority, toSafeHost } from "@/util";
+import {
+	countSubstring,
+	escapeCommandArg,
+	expandPath,
+	parseRemoteAuthority,
+	toSafeHost,
+} from "@/util";
 
 it("ignore unrelated authorities", () => {
 	const tests = [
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -145,5 +146,6 @@ export function countSubstring(needle: string, haystack: string): number {
 }
 
 export function escapeCommandArg(arg: string): string {
-	return `"${arg.replace(/"/g, '\\"')}"`;
+	const escapedString = arg.replaceAll('"', String.raw`\"`);
```
mtojek, line 149: nit: what was wrong with the previous form?

src/util.ts
```diff
@@ -119,13 +119,14 @@ export function toSafeHost(rawUrl: string): string {
...
- * Expand a path with ${userHome} in the input string
- * @param input string
- * @returns string
+ * Expand a path if it starts with tilde (~) or contains ${userHome}.
  */
 export function expandPath(input: string): string {
 	const userHome = os.homedir();
-	return input.replace(/\${userHome}/g, userHome);
+	if (input.startsWith("~")) {
+		input = userHome + input.substring("~".length);
```
mtojek, line 127: curious: will it work on Windows or this is on a case? e.g. `New fold~1`

src/util.ts
EhabY: Sonarqube linting, nothing is technically wrong there, just harder to read

src/util.ts
EhabY: Any reason why this wouldn't work on Windows 🤔

I just check the first character because I don't want to replace tildes in actual filenames

mtojek: Oh you're right! just the first character, I missed it 👍

PR #670, opened 2025-12-04
Add configurable autostart behavior for SSH connections

### LINKED ISSUE #666: Add a setting to toggle `--disable-autostart`
In Gateway and Toolbox we have a setting for it (called `disableAutostart`) and we default it to true on macOS so we should do the same here.

EhabY:
- Adds `coder.sshFlags` for optional user configurable SSH flags (defaults to only `--disable-autostart`).
- Renames `globalFlags.ts` → `cliConfig.ts` to consolidate CLI configuration logic

Closes #666

### DIFF
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -501,8 +501,6 @@ export class Remote {
 				sshMonitor.onLogFilePathChange((newPath) => {
 					this.commands.workspaceLogPath = newPath;
 				}),
-				// Watch for logDir configuration changes
-				this.watchLogDirSetting(logDir, featureSet),
 				// Register the label formatter again because SSH overrides it!
 				vscode.extensions.onDidChange(() => {
 					// Dispose previous label formatter
@@ -516,6 +514,18 @@ export class Remote {
 				}),
 				...(await this.createAgentMetadataStatusBar(agent, workspaceClient)),
 			);
+
+			const settingsToWatch = [
+				{ setting: "coder.globalFlags", title: "Global flags" },
+				{ setting: "coder.sshFlags", title: "SSH flags" },
+			];
+			if (featureSet.proxyLogDirectory) {
+				settingsToWatch.push({
+					setting: "coder.proxyLogDirectory",
+					title: "Proxy log directory",
+				});
+			}
+			disposables.push(this.watchSettings(settingsToWatch));
 		} catch (ex) {
 			// Whatever error happens, make sure we clean up the disposables in case of failure
 			disposables.forEach((d) => d.dispose());
@@ -790,29 +800,26 @@ export class Remote {
 		return sshConfig.getRaw();
 	}
 
-	private watchLogDirSetting(
-		currentLogDir: string,
-		featureSet: FeatureSet,
+	private watchSettings(
+		settings: Array<{ setting: string; title: string }>,
 	): vscode.Disposable {
 		return vscode.workspace.onDidChangeConfiguration((e) => {
-			if (!e.affectsConfiguration("coder.proxyLogDirectory")) {
-				return;
+			for (const { setting, title } of settings) {
+				if (!e.affectsConfiguration(setting)) {
+					continue;
+				}
+				vscode.window
+					.showInformationMessage(
+						`${title} setting changed. Reload window to apply.`,
+						"Reload",
+					)
+					.then((action) => {
+						if (action === "Reload") {
+							vscode.commands.executeCommand("workbench.action.reloadWindow");
+						}
+					});
+				break;
 			}
-			const newLogDir = this.getLogDir(featureSet);
-			if (newLogDir === currentLogDir) {
-				return;
-			}
-
-			vscode.window
-				.showInformationMessage(
-					"Log directory configuration changed. Reload window to apply.",
-					"Reload",
-				)
-				.then((action) => {
-					if (action === "Reload") {
-						vscode.commands.executeCommand("workbench.action.reloadWindow");
-					}
-				});
 		});
 	}
 


src/remote/remote.ts
```diff
@@ -674,7 +674,7 @@ export class Remote {
 		const globalConfigs = this.globalConfigs(label);
 
 		const proxyCommand = featureSet.wildcardSSH
-			? `${escapeCommandArg(binaryPath)}${globalConfigs} ssh --stdio --usage-app=vscode --disable-autostart --network-info-dir ${escapeCommandArg(this.pathResolver.getNetworkInfoPath())}${await this.formatLogArg(logDir)} --ssh-host-prefix ${hostPrefix} %h`
+			? `${escapeCommandArg(binaryPath)}${globalConfigs} ssh --stdio --usage-app=vscode${this.disableAutostartConfig()} --network-info-dir ${escapeCommandArg(this.pathResolver.getNetworkInfoPath())}${await this.formatLogArg(logDir)} --ssh-host-prefix ${hostPrefix} %h`
```
code-asher: Not something we need to deal with right now but I wonder if we will want to put our args in an array and do an `args.join(" ")` at some point since I feel like that would be a bit cleaner especially as we continue to add more optional flags.

package.json
```diff
@@ -120,6 +120,21 @@
...
+					"markdownDescription": "Disable starting the workspace automatically when connecting via SSH.",
+					"type": "string",
+					"enum": [
+						"auto",
+						"always",
+						"never"
+					],
+					"markdownEnumDescriptions": [
+						"Disables autostart on macOS only (recommended to avoid sleep/wake issues)",
+						"Disables on all platforms",
```
code-asher: The other two descriptions explicitly mention autostart so it might make sense to call this "Disables autostart on all platforms" or remove "autostart" from the other two, just for consistency.

package.json
```diff
@@ -120,6 +120,21 @@
 					"type": "boolean",
 					"default": false
 				},
+				"coder.disableAutostart": {
+					"markdownDescription": "Disable starting the workspace automatically when connecting via SSH.",
+					"type": "string",
+					"enum": [
+						"auto",
```
code-asher: I do think this enum is a nice solution but I worry about drift from the JetBrains plugins which has this as a boolean.  What do you think?

code-asher (approved):
Thank you for implementing this so quickly!

package.json
EhabY: Done

package.json
EhabY: If we have a boolean then we might have to start with a "false" (which is technically the current behavior anyway) since VS Code does not allow per-OS defaults

src/remote/remote.ts
EhabY: Refactored this bit!

package.json
code-asher: Ah I thought maybe VS Code would give it as undefined if it had not been set yet, then we could apply the OS default, but yeah if it always gives a boolean then nothing we can do :cry:

package.json
```diff
@@ -120,6 +120,28 @@
...
+					],
+					"markdownEnumDescriptions": [
+						"Disables autostart on macOS only (recommended to avoid sleep/wake issues)",
+						"Disables autostart on all platforms",
+						"Keeps autostart enabled on all platforms"
+					],
+					"default": "auto"
+				},
+				"coder.sshFlags": {
+					"markdownDescription": "Additional flags to pass to the `coder ssh` command when establishing SSH connections. Enter each flag as a separate array item; values are passed verbatim and in order. See the [CLI ssh reference](https://coder.com/docs/reference/cli/ssh) for available flags.\n\nNote that the following flags are ignored because they are managed internally by the extension: `--network-info-dir`, `--log-dir`, `--ssh-host-prefix`. To manage the `--disable-autostart` flag, use the `coder.disableAutostart` setting.",
```
code-asher: Oh wait `--usage-app` is also ignored, and maybe we should mention that `--log-dir` is controlled by another setting.

package.json
EhabY: Technically we can make the type `boolean | undefined` with default being `undefined`. But then we do not get the checkbox but instead a button that opens the JSON file.

Alternatively, we can find out if the value is "set" or not by the user but I am against this approach because the value of the checkbox would be a lie (visually it would be set but somehow on macOS it is not). Also users would need to toggle it back and forth to apply the setting that they want in some cases :/

package.json
EhabY: Oh true, I didn't list it here because it's not in the SSH manual but I'll add it anyway.

Actually should we filter `--log-dir` and `--disable-autostart` explicitly? Because technically you can keep `coder.proxyLogDirectory` empty or set the `disable-autostart` to `never` but then input those values in the `coder.sshFlags` which means we can have two sources to set the same thing. This might mess up the network indicator or the `View Logs` command since the log dir was set in a way besides the specific setting

code-asher: Ooo fair yeah usage is a hidden flag at the moment, not sure if that is a permanent thing or only while the feature is in experimental.  But makes sense to me not to mention it for now.

I think there is no need to filter `--disable-autostart` personally, although maybe we do not need the separate setting at all?

The log directory does seem like a bit of a problem, I guess the consequence would be that the command would say "make sure to set coder.proxyLogDirectory" so maybe not a big deal since the user can fix it?  But if we keep it as-is maybe we can document the full behavior in the description to avoid the user misconfiguring it.

Or maybe it would be easy enough for that command to fall back to looking for the value in `sshFlags`?

I think we always set `--network-info-dir` so the network indicator should be OK.

package.json
code-asher: Yeah that all sounds bad, current approach makes the most sense for sure.

code-asher: Well to be fair the second is what JetBrains does, and I would not consider it a lie (it is the practical value of the setting if not the "actual" value).  You could make the same argument for any boolean checkbox where the default is "false": it visually is unchecked but the actual value is neither true nor false, just undefined, or "the default".

Why would they have to toggle it?  If it is checked and they uncheck it, the value is now false.  If it is unchecked, and they check it, the value is now true.  This lines up with the visual display.

EhabY: Because if the default is `true` but actually on macOS it's false by default. The box will be checked in the VS Code setting and thus be a lie.

The user would then need to either add an explicit setting to the JSON file or uncheck then check to make sure the value is actually written. If they want the "default" then they have to remove the entry from the setting file.

package.json
EhabY: > although maybe we do not need the separate setting at all

I do agree tbh, a separate setting is more clear and we can enable auto-start for Linux and Windows by default but it does seem a bit unnecessary.

> Or maybe it would be easy enough for that command to fall back to looking for the value in sshFlags?

I can try that yes, but tbh this is one case where I think simple is better since it causes less confusion.

EhabY: > Or maybe it would be easy enough for that command to fall back to looking for the value in sshFlags?

That is easy actually but it won't work 100% of the time unless we interpret the results. Because those flags are passed to the ProxyCommand which is then executed in the terminal. So they can do something like `["--log-dir", "${SOME_ENV: /default/path}"]` or even simpler like `["--log-dir", "'/path/to/file with space'"]`. In one case, we have to expand the env variable and handle the default, while in the other we need to remove the single or double quotes to get the actual path. Again that's why filtering is much less confusing when we have an option already :/

package.json
code-asher: Ohhhhhhhh I see what you mean.  Since VS Code is resolving the checked status on its own, and does not refer to our own default we might derive in the code.

Yeah you are 100% right, the enum is the only way to do this reasonably.

package.json
code-asher: Yeah we definitely would have to expand. I think ignoring is completely fair, and it makes sense to take the extra effort only if we get complaints/feature requests for it.

EhabY: So I've ignored this setting and would just pass it as is to the `coder ssh`. I've also updated the description there saying to use the `coder.proxyLogDirectory` setting for full functionality.

PR #676, opened 2025-12-08
Upgrade Coder site package

EhabY:
We need to update to get the new types for OAuth but also to get access to all the new APIs.

### DIFF
--- a/src/workspace/workspacesProvider.ts
+++ b/src/workspace/workspacesProvider.ts
@@ -1,9 +1,8 @@
 import {
 	type Workspace,
 	type WorkspaceAgent,
-	type WorkspaceApp,
 } from "coder/site/src/api/typesGenerated";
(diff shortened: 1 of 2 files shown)


src/workspace/workspacesProvider.ts
```diff
@@ -165,22 +164,6 @@ export class WorkspaceProvider
 				showMetadata,
 			);
 
-			// Get app status from the workspace agents
```
code-asher, line 168: ~Was this removed due to something in the upgrade?~

nvm, I think maybe it was because `workspaceTreeItem.appStatus` was unused?

EhabY: Indeed, but it was related because the types have changed slightly. I could've fixed the types but realized it was easier to just remove them fully.

PR #677, opened 2025-12-09
Unify deployment tracking and support multiple deployments

EhabY:
This change centralizes deployment state management and enables seamless multi-deployment support. The extension now properly tracks per-deployment credentials, syncs state across VS Code windows, and handles login/logout flows in a unified way.

Key changes:

Architecture:
- Add `DeploymentManager` to centralize deployment state (`url`, `label`, `token`, `user`) and coordinate extension client updates, auth contexts, and workspace refreshes
- Add `LoginCoordinator` to handle login prompts with cross-window detection, preventing duplicate login dialogs when multiple windows need auth
- Move deployment types to `src/deployment/` module with proper type guards

Storage & Auth:
- `SecretsManager` now stores per-deployment credentials using label-based keys (`coder.session.<label>`) instead of flat `sessionToken` storage
- Add LRU tracking for deployments with automatic cleanup of old credentials
- Add migration from legacy flat storage format
- Cross-window sync via `secrets.onDidChange` events
- Debug command (`coder.debug.listDeployments`) for inspecting stored state

Commands & Remote:
- Commands now use `DeploymentManager` instead of directly manipulating client
- Remote connection uses `LoginCoordinator` for auth prompts during workspace connections
- `CliManager.configure()` now called on every remote connection, with secrets storage as the source of truth for credentials

WebSocket improvements:
- `CoderApi` now implements Disposable to clean up WebSocket connections
- Add `setCredentials()` method to update host+token atomically, avoiding unnecessary reconnection cycles
- Add `disconnect()` support to `ReconnectingWebSocket` for clean disconnects
- Simplify WebSocket fallback logic with cleaner SSE fallback handling

Tests:
- Update `secretsManager` tests for new per-deployment API
- Add comprehensive `reconnectingWebSocket` tests for suspend/reconnect
- Extend `coderApi` tests for credential handling

### DIFF
--- a/package.json
+++ b/package.json
@@ -269,12 +269,15 @@
 			{
 				"command": "coder.debug.listDeployments",
 				"title": "List Stored Deployments",
-				"category": "Coder Debug",
-				"when": "coder.devMode"
+				"category": "Coder Debug"
 			}
 		],
 		"menus": {
 			"commandPalette": [
+				{
+					"command": "coder.debug.listDeployments",
+					"when": "coder.devMode"
+				},
 				{
 					"command": "coder.openFromSidebar",
 					"when": "false"


src/api/coderApi.ts
```diff
@@ -74,75 +74,119 @@ export class CoderApi extends Api {
...
+
+		// We cannot use the super.setHost/setSessionToken methods because they are shadowed here
 		const defaults = this.getAxiosInstance().defaults;
-		const currentHost = defaults.baseURL;
 		defaults.baseURL = host;
+		defaults.headers.common[coderSessionTokenHeader] = token;
 
-		if (currentHost !== host) {
+		const hostChanged = (currentHost || "") !== (host || "");
+		const tokenChanged = currentToken !== token;
```
EhabY: Continuation of https://github.com/coder/vscode-coder/pull/633#discussion_r2602335595

src/api/coderApi.ts
```diff
@@ -74,75 +74,119 @@ export class CoderApi extends Api {
...
 		}
 	};
 
+	override setSessionToken = (token: string): void => {
+		this.setCredentials(this.getHost(), token);
+	};
+
+	override setHost = (host: string | undefined): void => {
+		this.setCredentials(host, this.getSessionToken());
+	};
```
EhabY, line 129: See https://github.com/coder/vscode-coder/pull/633#discussion_r2602329618

src/core/secretsManager.ts
```diff
@@ -1,73 +1,236 @@
...
+
+const CURRENT_DEPLOYMENT_KEY = "coder.currentDeployment";
+
+const DEPLOYMENT_USAGE_KEY = "coder.deploymentUsage";
+const DEFAULT_MAX_DEPLOYMENTS = 10;
+
+const LEGACY_SESSION_TOKEN_KEY = "sessionToken";
+
+export interface DeploymentUsage {
+	safeHostname: string;
```
EhabY: Replaced all `label` with `safeHostname` as per https://github.com/coder/vscode-coder/pull/633#discussion_r2602853652

src/deployment/deploymentManager.ts
```diff
@@ -0,0 +1,228 @@
...
+					if (this.currentDeployment?.safeHostname !== safeHost) {
+						return;
+					}
+
+					this.client.setCredentials(this.currentDeployment.url, auth.token);
+
+					// If we don't have a user yet, try to fetch one
+					if (!this.currentDeployment?.user) {
+						await this.tryFetchAndUpgradeUser();
+					}
```
EhabY: I think we should always try to fetch the user here to make sure the new authentication info is valid. If they are not valid it would be like we are "logged out" so looks more consistent. What do you think?

src/core/cliManager.ts
```diff
@@ -693,76 +696,69 @@ export class CliManager {
...
+		content: string,
+	): Promise<void> {
+		await fs.mkdir(path.dirname(filePath), { recursive: true });
+		const tempPath =
+			filePath + ".temp-" + Math.random().toString(36).substring(8);
+		try {
+			await fs.writeFile(tempPath, content);
+			await fs.rename(tempPath, filePath);
+		} catch (err) {
+			await fs.rm(tempPath, { force: true }).catch(() => {});
```
code-asher: Maybe we can log a warning if the rm fails, or return an array of errors?  Just want to make sure we are not hiding errors even if they are non-fatal.

src/core/contextManager.ts
```diff
@@ -12,10 +12,19 @@ type CoderContext = keyof typeof CONTEXT_DEFAULTS;
...
+	public constructor(extensionContext: vscode.ExtensionContext) {
+		for (const key of Object.keys(CONTEXT_DEFAULTS) as CoderContext[]) {
 			this.set(key, CONTEXT_DEFAULTS[key]);
-		});
+		}
+		this.setInternalContexts(extensionContext);
+	}
+
+	private setInternalContexts(extensionContext: vscode.ExtensionContext): void {
+		vscode.commands.executeCommand(
```
code-asher, line 23: Just curious why this uses `vscode.commands.executeCommand` instead of `this.set`.  Seems like the main result is that `this.context` would not get updated, is that OK?

src/core/secretsManager.ts
```diff
@@ -1,73 +1,241 @@
...
 			}
 		});
 	}
+
+	public async getSessionAuth(
+		safeHostname: string,
+	): Promise<SessionAuth | undefined> {
+		if (!safeHostname) {
+			return undefined;
+		}
```
code-asher: I would guess this is for convenience but just from the perspective of the function contracts it feels unexpected to me because:

```
setSessionAuth("", auth) // no error, so i assume my auth is set
getSessionAuth("") // no error, but also no auth that i thought i set earlier
onDidChangeSessionAuth("") // this will still fire, but with no auth
```

Maybe it would be better for the check to be external like:

```
if (deployment?.safeHostname) {
  getSessionAuth(deployment.safeHostname)
}
```

Or, we make these throw errors if we want to formalize that the hostname should be non-empty strings.

src/deployment/deploymentManager.ts
```diff
@@ -0,0 +1,229 @@
...
+	 */
+	public async setDeploymentWithoutAuth(
+		deployment: Deployment & { token?: string },
+	): Promise<void> {
+		this.setDeploymentCore({ ...deployment });
+
+		await this.tryFetchAndUpgradeUser();
+	}
+
+	private setDeploymentCore(deployment: DeploymentWithAuth): void {
```
code-asher: I think this is `Core` because it is called by two other functions but to me it feels kinda weird to name it based on that rather than name it based on what it does if that makes sense.  Could it just be `setDeployment`?

nbd either way though, this is just a private helper anyway.

src/deployment/deploymentManager.ts
code-asher: Whenever we set session auth, I assume we do so after having already validated the user right?  If so it seems like it would result in an extra request for no benefit.

Also the function makes me a bit nervous because it relies on a property `this.currentDeployment` and then has an async function, and then sets that property again but it could have been changed in the meantime.

IMO we should set the deployment only after auth is validated so we set it all at once without any opportunity for race conditions, but is there a use case for having the current deployment set and not any auth details?  To me they always go together.

src/deployment/deploymentManager.ts
```diff
@@ -0,0 +1,229 @@
...
+		this.authListenerDisposable?.dispose();
+
+		this.logger.debug("Registering auth listener for hostname", safeHostname);
+		this.authListenerDisposable = this.secretsManager.onDidChangeSessionAuth(
+			safeHostname,
+			async (auth) => {
+				if (auth) {
+					if (this.currentDeployment?.safeHostname !== safeHostname) {
+						return;
+					}
```
code-asher: Reasoning out loud, we are always listening on the current `safeHostname` right?  So this check would actually always be true?

I think we were adding this to prevent a window logged into say a.coder.com from being automatically switched to b.coder.com if another window logs into b.coder.com but since the first window is only listening to changes to a.coder.com this would not even fire in the first place I think?

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+	safeHostname: string;
+	url: string | undefined;
+	autoLogin?: boolean;
+}
+
+/**
+ * Coordinates login prompts across windows and prevents duplicate dialogs.
+ */
+export class LoginCoordinator {
+	private readonly inProgressLogins = new Map<string, Promise<LoginResult>>();
```
code-asher, line 31: It looks like this represents logins in flight in this window, but when would that happen?  If a new login is initiated would we not just cancel any pending login?

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	private async persistSessionAuth(
+		result: LoginResult,
+		safeHostname: string,
+		url: string,
+	): Promise<void> {
+		// Empty token is valid for mTLS
+		if (result.success && result.token !== undefined) {
+			await this.secretsManager.setSessionAuth(safeHostname, {
```
code-asher, line 129: Is it safe to skip calling this for mTLS?  That would cause onDidChangeSessionAuth to never fire right?  And then that could possibly cause a number of things waiting for auth to never run?

Er wait...I think the comment means a blank/empty string `""` is what we would get for mTLS?  But not undefined, that would be invalid?  I feel like we should only check `result.success`, if it was successful then it would be weird for the token to be invalid (and we can make the types guarantee this rather than having to caveat every `result.success` check with additional checks).

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	/**
+	 * Same-window guard wrapper.
+	 */
+	private async executeWithGuard(
+		safeHostname: string,
+		executeFn: () => Promise<LoginResult>,
+	): Promise<LoginResult> {
+		const existingLogin = this.inProgressLogins.get(safeHostname);
```
code-asher, line 144: If I want to restart the login process this would prevent me from doing so, which could be trouble.  Also if I decide to cancel and log into something else then the other login would be waiting forever.  What if we canceled the previous login and started a new one each time?

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+			);
+			storedToken = auth?.token;
+			if (storedToken) {
+				client.setSessionToken(storedToken);
+			}
+		}
+
+		// Attempt authentication with current credentials (token or mTLS)
+		try {
+			const user = await client.getAuthenticatedUser();
```
code-asher: Maybe we only do this if `!needsToken || auth?.token` otherwise we end up sending and waiting for a request we already know will fail because it needs a token.

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+			// Return the token that was used (empty string for mTLS since
+			// the `vscodessh` command currently always requires a token file)
+			return { success: true, token: storedToken ?? "", user };
+		} catch (err) {
+			if (needsToken) {
+				// For token auth: silently continue to prompt for new credentials
```
code-asher: Only if the error is 401 probably?  A 500 error for example would not indicate the credentials are invalid.

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+					const message = getErrorMessage(err, "no response from the server");
+					return {
+						message: "Failed to authenticate: " + message,
+						severity: vscode.InputBoxValidationSeverity.Error,
+					};
+				}
+			},
+		});
+
+		if (user === undefined || validatedToken === undefined) {
```
code-asher: Could these be null?  Not sure about `getAuthenticatedUser()` but I guess in theory the body could be null.  And the token could be an empty string.  A truthy check would cover all cases.

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+import type { MementoManager } from "../core/mementoManager";
+import type { SecretsManager } from "../core/secretsManager";
+import type { Deployment } from "../deployment/types";
+import type { Logger } from "../logging/logger";
+
+interface LoginResult {
+	success: boolean;
+	user?: User;
+	token?: string;
+}
```
code-asher: Could be nice to formalize that a successful result always has a user and a failed request has nothing, with `LoginResult` being the union of those two result types.  Although, at a glance we are actually not using `user`?  I suppose we could just drop it from the return then (maybe it is used in another file though).

src/api/coderApi.ts
code-asher: > However, if we authenticate without a token (mTLS) then we might begin with the following: { host: XYZ, token: undefined }, but when we authenticate we change the token to "" and thus trigger a reconnect.

> The input is the same since undefined might be treated the same as "" except that we might have started with invalid configuration (401s) and then changed the setting and prompted a login again that worked so we trigger this reconnect with the new info.

But why should this trigger a reconnect?  Nothing changes from the Coder server's perspective, it has no distinction between a blank token and a missing token.  If a missing token was a 401, so will a blank token, so why reconnect when we know it will fail?  Or is that not the case?

src/api/coderApi.ts
code-asher: > So I actually used setCredentials throughout but setHost/setSessionToken are defined in the super class actually. So they can still be called :/

Oooo did not realize these were in the base class.  Unfortunate, but makes sense!

src/core/secretsManager.ts
```diff
@@ -1,73 +1,241 @@
...
+	 * Migrate from legacy flat sessionToken storage to new format.
+	 * Also sets the current deployment if none exists.
+	 */
+	public async migrateFromLegacyStorage(): Promise<string | undefined> {
+		const legacyUrl = this.memento.get<string>("url");
+		if (!legacyUrl) {
+			return undefined;
+		}
+
+		const oldToken = await this.secrets.get(LEGACY_SESSION_TOKEN_KEY);
```
code-asher, line 215: Oh just realized this will fail to migrate mTLS sessions right?  Since they have no token?  Or...did we store an empty string for them?  In which case no problemo.

src/login/loginCoordinator.ts
```diff
@@ -0,0 +1,286 @@
...
+
+						return result;
+					} else {
+						// User cancelled
+						return { success: false };
+					}
+				});
+
+			// Race between user clicking login and cross-window detection
+			return Promise.race([
```
code-asher: If `dialogPromise` finishes before `waitForCrossWindowLogin` then we could have a hanging listener on `onDidChangeSessionAuth`.  Probably not a big deal, but just an observation.  Maybe we want to dispose it after this `Promise.race`.

src/remote/remote.ts
```diff
@@ -118,160 +94,190 @@ export class Remote {
...
+
+			// Listen for token changes for this deployment
+			disposables.push(
+				this.secretsManager.onDidChangeSessionAuth(
+					parts.safeHostname,
+					(auth) => {
+						workspaceClient.setCredentials(auth?.url, auth?.token);
+					},
+				),
 			);
```
code-asher, line 178: We listen for this above too, could/should we combine them?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -121,6 +136,18 @@ export class ReconnectingWebSocket<TData = unknown>
 		this.connect().catch((error) => this.handleConnectionError(error));
 	}
 
+	/**
+	 * Temporarily suspend the socket. Can be resumed via reconnect().
```
code-asher: To me `suspend` sounds more like it stops processing data but this actually closes the socket.  Looking from the outside in maybe the API could be:

```
close() // closes the socket, but can connect again
connect() // connect or reconnect
dispose() // permanently closes the socket
```

src/websocket/reconnectingWebSocket.ts
```diff
@@ -39,23 +38,23 @@ export class ReconnectingWebSocket<TData = unknown>
 	};
 
 	#currentSocket: UnidirectionalStream<TData> | null = null;
+	#lastRoute = "unknown"; // Cached route for logging when socket is closed
 	#backoffMs: number;
 	#reconnectTimeoutId: NodeJS.Timeout | null = null;
-	#isDisposed = false;
+	#isSuspended = false; // Temporary pause, can be resumed via reconnect()
```
code-asher: Since in the code `isSuspended` is always true when the socket is null (I think?), internally we could check `#currentSocket` and skip the extra boolean.  At the same time, a named variable could be valuable just from a self-documenting perspective, so nbd either way, although maybe a function `isSuspended(): boolean` that checks if the socket is null could fulfill that role too.

src/commands.ts
```diff
@@ -66,53 +79,50 @@ export class Commands {
...
-		const url = await maybeAskUrl(this.mementoManager, args?.url);
+		const currentDeployment = await this.secretsManager.getCurrentDeployment();
+		const url = await maybeAskUrl(
+			this.mementoManager,
+			args?.url,
+			currentDeployment?.url,
+		);
 		if (!url) {
-			return; // The user aborted.
+			return;
```
code-asher: Why remove the comment?  May not be obvious what a falsey URL means.

src/commands.ts
```diff
@@ -66,53 +79,50 @@ export class Commands {
...
+			url,
+			autoLogin: args?.autoLogin,
+		});
 
-		// These contexts control various menu items and the sidebar.
-		this.contextManager.set("coder.authenticated", true);
-		if (res.user.roles.find((role) => role.name === "owner")) {
-			this.contextManager.set("coder.isOwner", true);
+		if (!result.success || !result.user || result.token === undefined) {
+			return;
```
code-asher, line 111: We should only have to check result.success right?  And we can formalize the types to guarantee the rest of the conditions for us.

src/commands.ts
```diff
@@ -66,53 +79,50 @@ export class Commands {
...
-		await this.secretsManager.setSessionToken(res.token);
-
-		// Store on disk to be used by the cli.
-		await this.cliManager.configure(label, url, res.token);
+		// case we want to write with the provided blank hostname instead of
+		// generating one.
+		const safeHostname = args?.safeHostname ?? toSafeHost(url);
+		this.logger.info("Using hostname", safeHostname);
+
+		const result = await this.loginCoordinator.promptForLogin({
```
code-asher: The function name implies to me this will always be a prompt, but it is conditional.  I was using the `maybe` prefix to indicate this but there are any number of names we could use I think.

src/commands.ts
```diff
@@ -250,40 +165,21 @@ export class Commands {
...
 		this.logger.info("Logging out");
-		// Clear from the REST client.  An empty url will indicate to other parts of
-		// the code that we are logged out.
-		this.restClient.setHost("");
-		this.restClient.setSessionToken("");
 
-		// Clear from memory.
-		await this.mementoManager.setUrl(undefined);
-		await this.secretsManager.setSessionToken(undefined);
+		await this.deploymentManager.logout();
```
code-asher: I ran into some confusion between how `deploymentManager` has auth-related stuff like logout but then there is also a `loginCoordinator` and that does some auth-related stuff and it feels like they are cross-contaminated.

I think to me conceptually the current deployment and its auth are one singular object that would not make sense to set separately, and conceptually having them split feels like too much, and it could be unclear on what the responsibilities are for each.

Like right now it seems possible to be in this half-state where you have a deployment set but are not logged in, but when we would want that?

But, this is just me musing out loud, possibly I just need more time with it, so...well if anything stands out to you from this then that is great, if not then feel free to ignore lol

src/extension.ts
```diff
@@ -368,10 +311,13 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
 			);
 			if (details) {
 				ctx.subscriptions.push(details);
-				// Authenticate the plugin client which is used in the sidebar to display
-				// workspaces belonging to this deployment.
-				client.setHost(details.url);
-				client.setSessionToken(details.token);
+
+				// Will automatically fetch the user and upgrade the deployment
```
code-asher: What do you mean by upgrade?  I saw it referenced in `deploymentManager.ts` as well but was not too sure.  Should it be `update`?  Or is it like it gets "upgraded" from a not-logged-in deployment to a logged-in deployment?

Also the function says "without auth" but we pass it a token which feels weird.

src/extension.ts
```diff
@@ -411,31 +357,24 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
-				if (user && user.roles) {
+	// Initialize deployment manager with stored deployment (if any).
+	// Skip if already set by remote.setup above.
+	if (deploymentManager.getCurrentDeployment()) {
+		contextManager.set("coder.loaded", true);
+	} else if (deployment) {
+		output.info(`Initializing deployment: ${deployment.url}`);
+		const auth = await secretsManager.getSessionAuth(deployment.safeHostname);
+		deploymentManager
+			.setDeploymentWithoutAuth({ ...deployment, token: auth?.token })
```
code-asher: Similar to previous comment, the name `setDeploymentWithoutAuth` to me does not convey that this performs an auth check, maybe we could call it `validateAndSet` or something like that.

src/extension.ts
```diff
@@ -411,31 +357,24 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
-					if (user.roles.find((role) => role.name === "owner")) {
-						contextManager.set("coder.isOwner", true);
-					}
-
-					// Fetch and monitor workspaces, now that we know the client is good.
-					myWorkspacesProvider.fetchAndRefresh();
-					allWorkspacesProvider.fetchAndRefresh();
 				} else {
-					output.warn("No error, but got unexpected response", user);
+					output.info("Deployment set but not authenticated");
```
code-asher: Ah yeah this is the use case I am curious about.  Initially I felt like we would not want to set the deployment until we have valid credentials so we can set it all at once together and avoid having to synchronize, but thinking about it, this could allow you to try logging in, fail or cancel, close the window, come back later, and be able to try logging in again without having to enter the URL a second time?

But, if I try to log into a new deployment, then change my mind and cancel, I could lose being logged into that previous deployment, which might feel bad.  We could handle the above case by just having the URL in the history I suppose.

code-asher (commented):
I ran out of time to test it but will do so tomorrow!

src/api/coderApi.ts
EhabY: Here's my thought process:

Let's assume that we authenticate without a token (mTLS) but we are pointing to the wrong files or there's an issue with the `tlsCertFile`/`tlsKeyFile`. Let's assume that we then trigger a login event that actually works, so we set the token to `""` and thus trigger a reconnect. It's not necessarily about the empty vs. `undefined` but more about triggering something when we successfully login.

src/core/cliManager.ts
EhabY: Logged a warning!

src/core/contextManager.ts
EhabY: `this.set(CoderContext, Boolean)` would mean that we have to expose this flag to clients of this class. I want to keep this flag hidden and only be set once so that's why it's not of type `CoderContext` and it cannot be changed.

src/core/secretsManager.ts
EhabY: You are right, when we do not set the session token so we do not migrate those, should we still attempt a migration if we simply have a URL? I think that is harmless 🤔 

https://github.com/coder/vscode-coder/blob/677eee4231baac670f2cee3af5fd8fec7ebac99b/src/core/secretsManager.ts#L19-L25

src/core/secretsManager.ts
EhabY: This actually made me realize that if we open a deployment with no label we might not be able to "login" ever. So instead I've added the following lookup:

```ts
private getSessionKey(safeHostname: string): string {
		return `${SESSION_KEY_PREFIX}${safeHostname || "<legacy>"}`;
}
```

The only place that can provide a blank `hostname` is `remote.ts` when connecting to a workspace that uses this old format.

src/deployment/deploymentManager.ts
EhabY: You aren't wrong, renamed for clairty

src/deployment/deploymentManager.ts
EhabY: This is not to prevent cross window logging in from taking over (that part is handled by `onDidChangeCurrentDeployment`). This is here as a sanity check that if we change `this.currentDeployment` without disposing this watcher then it wouldn't cause the token to be set for another deployment.

src/deployment/deploymentManager.ts
EhabY: > but is there a use case for having the current deployment set and not any auth details? To me they always go together.

We could be logged in to a deployment then the token is revoked or expired and thus no "auth" details. This is why I lean towards adding that extra request since there's no guarantee that the info we just read is always valid. Like we can have a "current" deployment that is not signed in but another window signs in and thus we should react to that and fetch the user to update it.

I do get your concern about the deployment changing in the meantime. I'll capture `safeHostname` and check it before trying to modify the `currentDeployment`

src/login/loginCoordinator.ts
EhabY: You are right, I've used union types for more rigid types. But `user` is actually used by another file though it's not always set (if we login from a different window then we do not have a `user` even though it's a valid login)

src/login/loginCoordinator.ts
EhabY: This was part of the OAuth changes I suppose, where we can have two background refreshes happening at the same time in the same window (if the extension and remote client both use OAuth and both are logged in to the same deployment)

src/login/loginCoordinator.ts
EhabY: Ohh rightt, disposed the listener on a `finally`

src/login/loginCoordinator.ts
EhabY: > and we can make the types guarantee this rather than having to caveat every result.success check with additional checks

After applying the previous suggestion then yes I can guarantee based on types which is much safer and more rigid!

src/login/loginCoordinator.ts
EhabY: Yes there is no way to restart the login process, that part is true. But I do not think there's a fear of "being" stuck since the user cancelling would resolve this promise with `{ successful: false }` (that part might be problematic, maybe we try again?)

Also cancelling adds complexity because we have to propagate the cancellation token and make sure it's tied to all the components (assuming we can).

How about we queue these events actually? Like if we are logged in then we continue as is but if not then we get another login process again. But tbh that's confusing since the user just tried cancelling... Actually, why would we want to restart the login process for the same deployment?

src/login/loginCoordinator.ts
EhabY: Done!

src/login/loginCoordinator.ts
EhabY: Made it `needsToken && is401` otherwise we log the error or show an error depending on the `autoLogin` option

src/login/loginCoordinator.ts
EhabY: Both would be `undefined` together if the user cancels, `validated` token can be an empty string if that is a valid token (somehow).

> A truthy check would cover all cases.

I am not sure I understand 🤔

src/remote/remote.ts
EhabY: Yes, combined into the bottom listener and fixed an issue with a dangling listener!

src/websocket/reconnectingWebSocket.ts
EhabY: We cannot really override `close` to be like that because we want clients of this class to be able to `close` without knowing the implementation details. (This is a reconnecting socket if the close event comes from the server or the error handler).

The current public API is the following:

* `close() // also disposes`
* `suspend() // closes websocket without disposing`
* `reconnect() // forces a reconnection (connected or not)`

We could perhaps depend on the `code` in `close()` to check if we should dispose or not but we never set that code when we close and I cannot force people to do so :/

src/websocket/reconnectingWebSocket.ts
EhabY: It's not entirely correct, technically we can call `suspend` (this prevents `connect`/`scheduleReconnect` from executing) while having a `null` socket would not, also we set the socket to `null` in some cases where `#isSuspended` would be `false`. It might also complicate some other logic tbh

src/commands.ts
EhabY: Fair enough, I think I've just seen a lot of these comments where they say "// The user aborted."

src/commands.ts
EhabY: Yes, done! but I fetch the `user` since it could be undefined (see the `LoginCoordinator` comments).

https://github.com/coder/vscode-coder/pull/677/changes#r2609790051

src/commands.ts
EhabY: Renamed to `ensureLoggedIn`

src/commands.ts
EhabY: I think I see the confusion now, let me explain my perspective on these objects lol:

* LoginCoordinator is solely responsible for **logging in** into a deployment and persisting the auth if it was successful. There is no logic about `logout` and technically we can move the `persist` part out of it if it's confusing (I kept it there to keep code DRY but I can see why this might be confusing).
* DeploymentManager is solely responsible for managing the `extensionClient` and it's authentication state. It does not attempt logins or logout (logout will be renamed to `clear` maybe). It's solely a data structure that contains this info and reacts to changes but does not do actual login/logout. (It manages `CurrentDeployment` only). Maybe this should renamed into `CurrentDeploymentManager`?

src/extension.ts
EhabY: > "upgraded" from a not-logged-in deployment to a logged-in deployment

Yes it means that.

> Also the function says "without auth" but we pass it a token which feels weird.

True... maybe something like `setDeploymentAndValidate`?

src/extension.ts
EhabY: Renamed to `setDeploymentAndValidate`

src/extension.ts
EhabY: > try logging in again without having to enter the URL a second time?

The idea here is that you could login, not open VS Code in a while then open it again and now suddenly you are not authenticated (but still you maintain the last deployment that you were connected to). So you can have stuff like the URL prepopulated when you click `login`.

> I could lose being logged into that previous deployment, which might feel bad

I mean currently you cannot do this WITHOUT clicking `logout` anyway (unless the current deployment is not authenticated because the token is gone) but then `setCurrentDeployment` is only set at the very end anyway...

EhabY:
@code-asher  I've add tests in the last commit BTW, if you wish I can make it a separate PR  :pray:

src/api/coderApi.ts
code-asher: Ahhhhh ok I think I understand, so embedded in the token is an assumption that, in the mTLS case, undefined means an unsuccessful (or pending?) login and a blank string means a successful login?

This context does introduces its own problem, I think: if one calls `setCredentials` with defined host and undefined token.  This can cause `socket.reconnect()`.  But if I understand correctly, `undefined` means there is an issue, so `socket.reconnect()` will fail.  If a blank token means success, then I suppose that means we should check explicitly for undefined and not try to reconnect sockets.  But maybe we want to formalize this with a boolean or something.

Although, it feels unexpected to me that we would call `setCredentials` for a failing state.  I would expect it to be called when logging in and we know we have valid credentials, or to log out.

Or...should `setCredentials` be separate from whether sockets are reconnected?

src/deployment/deploymentManager.ts
code-asher: > We could be logged in to a deployment then the token is revoked or expired

Right but this can happen at any moment, not just at login (and that is the least likely place for it to happen, probably near zero?), so we could just let the 401 handler deal with it when the next request is made.

> Like we can have a "current" deployment that is not signed in but another window signs in and thus we should react to that and fetch the user to update it.

That would make sense to me if we needed the user, but we are not actually using it right?  We only fetch it to validate the credentials, and we already know the credentials are valid because otherwise the other window would not have set them.  It must have just made the exact same request we are about to make here, right?

Even if they were somehow wrong, the next request will be a 401 and kick off the login process again.

Also, even if we fetch the user again, there is still no guarantee the info we just read from that request is valid.  A token could be revoked at any time, so we just have to assume what we have is valid and deal with the 401 if not.

src/core/contextManager.ts
code-asher: Gotcha!

src/core/secretsManager.ts
code-asher: Cool, I like that.

src/deployment/deploymentManager.ts
code-asher: We could formalize that, make it such that if one tries to set `currentDeployment` it always updates the listener appropriately.  It sounds like if that were to happen it could be a developer error and maybe we want to prevent it entirely, or see if we can architecture such that it cannot happen.

Fair to have a sanity check though!

src/login/loginCoordinator.ts
code-asher: Same deployment that would mean they use the same promise right?

If the remote client is a.coder.com and the user logs the extension into b.coder.com, and there happens to be a refresh or expiry of a.coder.com at the same time, then we would have two ongoing logins.  Makes sense.

Plus if we add multi-deployment support to the sidebar, we will need it there as well.

src/login/loginCoordinator.ts
code-asher: > why would we want to restart the login process for the same deployment?

I think just to kick off the browser window opens again.  For token, maybe I want it to open `/cli-auth` again and for oauth maybe I want it to reopen the auth window.

But, also partially because if you click a button (login in this case) and nothing happens, it feels kinda bad, like the plugin might be broken.

I think for the token flow it actually works, by accident, because when I click "login" it shows the prompt for the URL, which replaces/closes the token prompt, which cancels the previous flow lol

code-asher: Possibly an alternative is to put the links in the dialogs/notifications so users can re-open them without having to restart the login process.

But I think we do still need some kind of cancellation?  If I mistype `dev.coder.com` and then correct it by trying to log in again, we probably want cancel the other one.  I think this will mostly just be applicable for oauth though since from what I recall there was no cancel?  So maybe we come back to this on that PR.

src/login/loginCoordinator.ts
code-asher: Awesome

src/login/loginCoordinator.ts
code-asher: I meant a check like `!user || !validatedToken` (technically a falsey check, not truthy I suppose).

> validated token can be an empty string if that is a valid token

It can be a blank string if the user just presses "enter" without entering anything, we only validate truthy inputs (we return null if `!value`), but it is not a valid token.

The user would be undefined in that case, so the check still works.  Honestly I guess we really only need to check `user` alone, if we have `user` we must have had a valid token.

`user` could still be `null` though, granted it would be a crazy thing to happen (maybe if someone's proxy was broken and returned null bodies or something wild).

src/login/loginCoordinator.ts
code-asher: Ah right cross-window, that makes sense.

But, why do we need the user?  All we care about is if the login was successful right?  Anywhere we use `result.user` we can use `result.success` instead.

If there is a case where we can set success to true but not have checked the user, that sounds like a bug.

code-asher: Oh we do use it in the login message.  Fair fair.

src/deployment/deploymentManager.ts
code-asher: Coming back to this after realizing we use the username in the login message, maybe it makes sense to fetch the user purely just to have it for use in the UI?  Maybe we will want to put the name in the bottom bar or the sidebar one day, or something like that.

src/core/secretsManager.ts
code-asher: Yeah I would agree with that!

src/websocket/reconnectingWebSocket.ts
code-asher: Yeah with this change we would probably have to make changes to the base socket class so the reconnecting socket would not have different behavior.  The base socket could have the ability to be `close()`d and then later `connect()`ed again, and then the reconnecting socket is just using that but doing it automatically (so it uses `close` and `connect` as well rather than having to recreate `#currentSocket` entirely).

Buuuut idk I mostly just feel like `suspend` is potentially confusing, at least coming from Node sockets (although there it is `pause`).  Maybe we can call it `softClose` or something.

> We could perhaps depend on the code in close() to check if we should dispose or not but we never set that code when we close and I cannot force people to do so :/

Nah that sounds janky to me.

src/websocket/reconnectingWebSocket.ts
code-asher: Not sure I follow, all `suspend` does is set `isSuspended` to true and `currentSocket` to null (after closing it and the timeout) so they seem equivalent.

If `currentSocket `can be null without being suspended (or disposed) then that sounds like a bug, which the boolean might be masking.  What would that case be?

src/commands.ts
code-asher: Yeah they were all commented, there is no guarantee a dev has seen one or the other or recognizes the pattern.

Alternate solution would be to return an object with `aborted: true` or something like that, then it would be obvious without the comment.

src/commands.ts
code-asher: Ahhhh  I see, that does help.   I was seeing `LoginCoordinator` as more a general auth class.

It did also feel a little strange that deployment manager sets `coder.authenticated`, but if I think about it in terms of "authenticated" meaning more like "the current authenticated deployment" it does seem clearer.  And `coder.isOwner` meaning "is owner of the deployment".  Probably the "auth" name throws me off.

> Maybe this should renamed into CurrentDeploymentManager?

Nah, at least for me, this was not a point of confusion.

src/extension.ts
code-asher: That seems reasonable to me.

src/extension.ts
code-asher: > The idea here is that you could login, not open VS Code in a while then open it again and now suddenly you are not authenticated (but still you maintain the last deployment that you were connected to). So you can have stuff like the URL prepopulated when you click login.

But this does not require that we set the deployment without validation and then set it again with validation later.  In this case, of a successful login, we would have a current deployment set.

> I mean currently you cannot do this WITHOUT clicking logout anyway

Mostly I was thinking of it from an isolated point of view, that this prevents such a feature, but it is possible to trigger this with the URI handler, for example if I am logged in `dev.coder.com` and open `vscode://coder.coder-remote/open?url=test.coder.com&owner=my-username&workspace=my-ws`.

Although, maybe that is actually a bug, I forget if we decided to change the current deployment with the URI handler if something else is logged in.

As an aside, it would be nice if it was possible to directly log in a different deployment without having to log out first (fewer button clicks).

src/login/loginCoordinator.ts
code-asher: lol unrelated to your PR since this was already a problem, but I just had an issue where I forgot to make my proxied test deployment public so it was trying to fetch the user and it got html back, then it tried to do `user.roles` but `user` is a string...

Maybe at some point we will want to validate that `user` is an actual user object

src/deployment/deploymentManager.ts
EhabY: > so we could just let the 401 handler deal with it when the next request is made.

Wait actually what handler? We do not have any such handler (yet). Currently if we have an unauthorized request they would all keep failing until this is fixed.

> We only fetch it to validate the credentials

Ahh I see the confusion now, we fetch it to validate the credentials yes but also to set `coder.isOwner` since that is deduced from the `User` object as well - This is something that we always have to do since it plays a crucial role in the UI.

But actually I do agree that maybe we need a 401 handler (not just for OAuth but also for session token auth).

src/core/secretsManager.ts
EhabY: Done!

src/deployment/deploymentManager.ts
EhabY: I've explored other ways like having a custom `set` method with side-effects but that's more confusing IMO + a separate data class but it solves nothing. Instead I've made the methods here all depend on the local deployment so it's more consistent and I'll keep the sanity check for racy processes!

src/login/loginCoordinator.ts
EhabY: > I think for the token flow it actually works, by accident, because when I click "login" it shows the prompt for the URL, which replaces/closes the token prompt, which cancels the previous flow lol

I wouldn't call this an accident haha, since even for the OAuth I've added a progress monitor that you can cancel from. I just do not like having internal cancellations. So basically for session token auth, closing the dialog would "cancel" while for OAuth they can cancel through the progress monitor. In either case, there is no indefinite waiting AFAIK.

Let's revisit this in the OAuth PR actually

src/login/loginCoordinator.ts
EhabY: > if we have user we must have had a valid token

That is true, I'll make this a user check only.

> Maybe at some point we will want to validate that user is an actual user object

This is an issue with RESTful API and WebSocket types, we need to validated objects but we always assume the happy path :/

src/websocket/reconnectingWebSocket.ts
EhabY: How about `disconnect` which is countered by `(re)connect`?

Though `#isSuspended` might be a bit confusing now :/

src/websocket/reconnectingWebSocket.ts
EhabY: If we are suspended then the socket is guaranteed to be `null` yes but not the other way around.

`#currentSocket` can be null during the connection phase, after `create()` is called but before the socket is established, or during reconnection attempts with backoff. In those states, we're actively trying to connect, not suspended.

src/extension.ts
EhabY: I feel like we are talking about different things now lol

>  it would be nice if it was possible to directly log in a different deployment without having to log out first (fewer button clicks).

We currently can do that, that's why we do `setDeploymentAndValidate` when handling `vscode` URLs. What was the original ask actually? Setting all the deployment info (with the `User`) all at once as compared to setting the deployment info partially (everything BUT `user`)?

src/deployment/deploymentManager.ts
code-asher: > Wait actually what handler?

:scream: what in the blazes am I talking about lol, I could have sworn we added an interceptor for 401s but yeah I do not see it anywhere.  Am I thinking about the previous PR with oauth?

> Ahh I see the confusion now, we fetch it to validate the credentials yes but also to set coder.isOwner

Ooooooo right that does make sense!  Yeah so we just straight-up need it regardless of validation.

src/login/loginCoordinator.ts
code-asher: Haha fair, maybe I should have said "by side effect", I just mean there is no explicit cancel, it happens to work because it happens to use prompts which it just so happens we can only have one at a time (if there was no prompt it would not work, the login button would do nothing, like with oauth).

> Let's revisit this in the OAuth PR actually

:+1:

src/login/loginCoordinator.ts
code-asher: Yuuup too true, too true.  But, if we validate the user at least (which we always fetch) then we have at least some confidence that the rest will work.

src/websocket/reconnectingWebSocket.ts
code-asher: Oh snap I like `disconnect`, nice parallel.

> Though #isSuspended might be a bit confusing now :/

I do still argue we could just use `socket === null` for that :stuck_out_tongue:

src/websocket/reconnectingWebSocket.ts
code-asher: I think that is actually a good illustration of what I mean about a bug masked by the boolean, like suppose you call `suspend` while connecting, it clears the socket (still not set so this is a no-op) and then the connect finishes and we are actually connected, not suspended.

If we used the socket itself as the source of truth instead of keeping a separate boolean in sync, that would require us to set it synchronously (which will require edits to the socket factory), but then it could naturally resolve that issue.

src/extension.ts
code-asher: > I feel like we are talking about different things now lol

Could be lol

> it would be nice if it was possible to directly log in a different deployment without having to log out first (fewer button clicks).
> We currently can do that

Here I mean by clicking "login" again (there is no way to do this currently since we have no login button or menu item once you are logged in) or by running the login command (if you do this, it simply does nothing).  Not something we need to add, just thinking out loud it would be nice if I could switch deployments without having to click logout first, and that if we did one day add that, setting the current deployment in this way (as a two-step process) would cause issues.  

While it is not possible from the UI, it is possible to directly log into a different deployment with the URI handler, though I am not sure that is intended?  Should the URI handler change what I am logged into currently?  Or should it only be used for the remote connection?  I vaguely recall talking about this, but idk what conclusion we reached.

> What was the original ask actually? Setting all the deployment info (with the User) all at once as compared to setting the deployment info partially (everything BUT user)?

Yup yup.  Currently we set the deployment, then we log in, then we set it again with credentials.  This means that if you cancel during the login, you are stuck with this new deployment because we already set it as the current deployment.  Or, if your credentials are invalid, you are stuck with that broken connection (until you log into the other again, which admittedly is not that hard since we are not deleting tokens anymore).

For example I am logged into dev.coder.com and if I open a URI from my test deployment but the token is wrong or missing, my current deployment gets nuked so now my sidebar just shows "login".  I would expect dev.coder.com to still be logged in.  If I restart VS Code, I will actually still be logged into dev.coder.com.  We only set it for the current instance, temporarily.

I propose setting it only once we have validated auth, so we set it just the one time, and a failure or cancellation during login will leave the current deployment in place.

src/api/coderApi.ts
EhabY: Actually let me go through the code paths and try to figure out if `setCredentials` can be called on successful login and somehow not trigger a reconnect (in this case we can add a `forceReconnect` or something along those lines!)

src/websocket/reconnectingWebSocket.ts
EhabY: > which will require edits to the socket factory

I do not think we can cleanly convert this code from async to sync. It's async because we need to get the `headers`, read the files, and have an SSE fallback so we have to wait arbitrarily. That's why we need to rely on an outside flag because we need a way to say "keep/make the current state disconnected/suspended".

Your scenario does showcase a bug here but this bug also applies to `dispose` :/, we can fix it by checking after `async` operations:

```ts

await ...

if (#isDisposed || #isDisconnected) { socket.close(...); return; }
```

src/extension.ts
EhabY: > Should the URI handler change what I am logged into currently? Or should it only be used for the remote connection?

Currently it changes the current deployment yes, that was the behavior from before I maintained it here. We set the deployment and then we call `commands.open`/`commands.openDevContainer` both would eventually execute `remote.setup` when they attempt to connect to the deployment in question. 

Actually though, maybe we shouldn't do `setDeploymentAndValidate`, instead, we just set the token (if any) and then we let `remote.setup` handle the login (if we still need to login) and the `details` return from `remote.setup` would be used to set the deployment AFTER successful login. That way, if you click open and there is no token and you cancel the sign in then essentially no permanent state is changed.

--

We currently have 3 usages of `setDeploymentAndValidate`:
* If we login from a different window and this window is not logged in (`currentDeployment` secret is already changed anyway at this point)
* After we successfully establish a remote connection
* At the end of `activate` when we want to set the correct contexts with the current deployment we have

I could just remove `setDeploymentAndValidate` entirely and always do `client.getAuthenticatedUser` before calling `changeDeployment({ ..., user })`. Do keep in mind though, that we can have authenticated and then we aren't if we switch the token or it expires or is revoked. So having the current deployment set does not necessarily mean it's authenticated but we cannot set a new deployment UNLESS we are signed in which actually makes sense yeah.

I've made some changes here so that we only change AFTER we validate (`changeDeploymentIfValid`).

src/api/coderApi.ts
EhabY: Okay seems safe, the only case that we want to force a reconnect could be if we change the `mTLS` settings but those can be a follow up (if any of those settings change we trigger a refresh).

code-asher: Cool cool cool

src/extension.ts
code-asher: > Currently it changes the current deployment yes, that was the behavior from before

Yeah makes sense, I was not sure if we had decided to change that behavior, I vaguely recall talking about it but not what we concluded.  It makes sense to me to keep the current behavior unless we hear complaints or get a feature request!

> Do keep in mind though, that we can have authenticated and then we aren't if we switch the token or it expires or is revoked. 

Yup yup, yeah at any point in time we have no guarantee that the token is valid, all we can do is handle the result in the best way possible.

src/websocket/reconnectingWebSocket.ts
code-asher: Yeah we can also fix by making sure we sync up with the boolean state after each async operation but it feels a bit brittle to me.

I was thinking something like:

```
const socket = socketFactory() // sync
socket.connect() // async, can do whatever async stuff we need here
socket.disconnect() // if this is called, aborts the connect
```

`connect` and `disconnect` would use an abort controller or something like that (you could argue an abort controller is just a fancy boolean though lol)

But to be fair I have not thought it all the way through, just kinda vibes at this point  :stuck_out_tongue:  As long as we sync up after each async operation, we should be good.

EhabY: Yep we do this now, tbh the hard part is making `socketFactory` sync, I do not think we can do that because we need some info to create the socket. We cannot pass these options AFTER the fact :/

src/api/coderApi.ts
EhabY: Created #691

src/websocket/reconnectingWebSocket.ts
code-asher: Yeah the actual web socket itself could never be created sync, but we have a wrapper around it that could be sync (`OneWayWebSocket` I think) and is what we would return from the factory synchronously (then `connect` creates the actual underlying web socket).

But, one could argue all that really does is move the code from the reconnecting socket to the base socket class.  I do think it could end up being slightly simpler, but I admit it is all in my head at the moment lol.

PR #679, opened 2025-12-10
Add log file picker when viewing logs without active connection

### LINKED ISSUE #284: `Coder: View Logs` can show the wrong log file
We look for the SSH PID then store that, but if there is a reconnection we could get a new SSH process and now we are pointing to the wrong log file.

Also, if you are not currently connected the command gives you nothing.  Maybe we should open the directory in that case and show a file picker ordered by recency?  Or just always open the most recent one?  That would also solve the above case.

EhabY:
Also fix fs.readdir sorting - the order is platform-dependent, not guaranteed sorted.

Closes #284

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -280,10 +280,9 @@ export class Commands {
 		}
 	}
 
-	private async openFile(path: string): Promise<void> {
-		const uri = vscode.Uri.file(path);
-		const doc = await vscode.workspace.openTextDocument(uri);
-		await vscode.window.showTextDocument(doc);
+	private async openFile(filePath: string): Promise<void> {
+		const uri = vscode.Uri.file(filePath);
+		await vscode.window.showTextDocument(uri);
 	}
 
 	/**
--- a/test/unit/remote/sshProcess.test.ts
+++ b/test/unit/remote/sshProcess.test.ts
@@ -1,5 +1,6 @@
 import find from "find-process";
 import { vol } from "memfs";
+import * as fsPromises from "node:fs/promises";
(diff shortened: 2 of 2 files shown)


src/commands.ts
```diff
@@ -239,9 +277,11 @@ export class Commands {
 						);
 					}
 				});
-			return;
 		}
-		const uri = vscode.Uri.file(this.workspaceLogPath);
+	}
+
+	private async openFile(path: string): Promise<void> {
```
mtojek: nit: you're shadowing path (import)

src/remote/sshProcess.ts
```diff
@@ -257,7 +257,7 @@ export class SshProcessMonitor implements vscode.Disposable {
 		while (!this.disposed && this.currentPid === targetPid) {
 			try {
 				const logFiles = await fs.readdir(logDir);
-				logFiles.reverse();
+				logFiles.sort().reverse();
```
mtojek, line 260: Can we test the sorting issue in some way?

EhabY: I've added some tests but they aren't perfect because if the filesystem returns a sorted array then it will always succeed anyway :shrug: (we use `memfs` which always returns a sorted array similar to `readdir` on Linux)

src/commands.ts
EhabY: Renamed!

src/remote/sshProcess.ts
EhabY: I've mocked the `readdir` so it's unsorted now to simulate Windows!

PR #682, opened 2025-12-15
Fix false "setting changed" notifications on remote connect

### LINKED ISSUE #681: "Global flags setting changed" notification appears on every connection
## Description

Users are reporting that a notification "Global flags setting changed. Reload window to apply." pops up every time they connect to a workspace, even when the `coder.globalFlags` setting is the default empty array and hasn't been modified.

## Steps to Reproduce

1. Install the VS Code Coder extension v1.11.5
2. Leave `coder.globalFlags` at the default (empty array `[]`)
3. Connect to any Coder workspace
4. Observe the notification appears
5. Disconnect and reconnect - notification appears again

## Expected Behavior

The notification should only appear when the user actually changes the `coder.globalFlags` setting value.

## Actual Behavior

The notification appears on every connection, regardless of whether the setting has changed.

## Root Cause Analysis

The bug was introduced in commit c3943080860c682e4a976afd9fd5990a59838d58 which added the `watchSettings` function in `src/remote/remote.ts`.

The current implementation uses `e.affectsConfiguration(setting)` to detect changes:

```typescript
private watchSettings(
  settings: Array<{ setting: string; title: string }>,
): vscode.Disposable {
  return vscode.workspace.onDidChangeConfiguration((e) => {
    for (const { setting, title } of settings) {
      if (!e.affectsConfiguration(setting)) {
        continue;
      }
      vscode.window
        .showInformationMessage(
          `${title} setting changed. Reload window to apply.`,
          "Reload",
        )
        // ...
    }
  });
}
```

The issue is that `affectsConfiguration()` can return `true` even when the actual value hasn't changed. This is a known VS Code API behavior - it fires when:
- The configuration scope changes (e.g., workspace vs user settings)
- VS Code re-reads settings during Remote SSH connection setup
- Settings are "touched" without the actual value changing

## Suggested Fix

The watcher should compare the actual value before/after, not just check if the configuration "might" have changed. The previous `watchLogDirSetting` function that was replaced actually did this correctly:

```typescript
const newLogDir = this.getLogDir(featureSet);
if (newLogDir === currentLogDir) {
  return;  // Skip if value unchanged
}
```

A similar pattern should be applied to `watchSettings` - store the initial values of watched settings and only show the notification when the value actually differs.

## Environment

- Extension version: v1.11.5
- VS Code version: Various
- OS: Various (reported on multiple platforms)

EhabY:
## Summary
- Fix false "setting changed" notifications appearing when connecting to remote
- Compare actual setting values instead of relying on `affectsConfiguration()`
- Use getter functions (`getGlobalFlagsRaw`, `getSshFlags`, `getLogDir`, `getHeaderCommand`) to properly resolve defaults
- Show single notification when multiple settings change at once
- Add `coder.headerCommand` to watched settings

Closes #681

### DIFF
--- a/test/unit/cliConfig.test.ts
+++ b/test/unit/cliConfig.test.ts
@@ -1,16 +1,14 @@
 import { it, expect, describe } from "vitest";
-import { type WorkspaceConfiguration } from "vscode";
 
 import { getGlobalFlags, getGlobalFlagsRaw, getSshFlags } from "@/cliConfig";
 
+import { MockConfigurationProvider } from "../mocks/testHelpers";
 import { isWindows } from "../utils/platform";
 
 describe("cliConfig", () => {
 	describe("getGlobalFlags", () => {
 		it("should return global-config and header args when no global flags configured", () => {
-			const config = {
-				get: (_key: string, defaultValue: unknown) => defaultValue,
-			} as unknown as WorkspaceConfiguration;
+			const config = new MockConfigurationProvider();
 
 			expect(getGlobalFlags(config, "/config/dir")).toStrictEqual([
 				"--global-config",
@@ -19,12 +17,11 @@ describe("cliConfig", () => {
 		});
 
 		it("should return global flags from config with global-config appended", () => {
-			const config = {
-				get: (key: string) =>
-					key === "coder.globalFlags"
-						? ["--verbose", "--disable-direct-connections"]
-						: undefined,
-			} as unknown as WorkspaceConfiguration;
+			const config = new MockConfigurationProvider();
+			config.set("coder.globalFlags", [
+				"--verbose",
+				"--disable-direct-connections",
+			]);
 
 			expect(getGlobalFlags(config, "/config/dir")).toStrictEqual([
 				"--verbose",
@@ -35,16 +32,12 @@ describe("cliConfig", () => {
 		});
 
 		it("should not filter duplicate global-config flags, last takes precedence", () => {
-			const config = {
-				get: (key: string) =>
-					key === "coder.globalFlags"
-						? [
-								"-v",
-								"--global-config /path/to/ignored",
-								"--disable-direct-connections",
-							]
-						: undefined,
-			} as unknown as WorkspaceConfiguration;
+			const config = new MockConfigurationProvider();
+			config.set("coder.globalFlags", [
+				"-v",
+				"--global-config /path/to/ignored",
+				"--disable-direct-connections",
+			]);
 
 			expect(getGlobalFlags(config, "/config/dir")).toStrictEqual([
 				"-v",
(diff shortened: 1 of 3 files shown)


test/unit/cliConfig.test.ts
```diff
@@ -82,6 +82,30 @@ describe("cliConfig", () => {
 		});
 	});
 
+	describe("getGlobalFlagsRaw", () => {
+		it("returns empty array when no global flags configured", () => {
+			const config = {
+				get: (_key: string, defaultValue: unknown) => defaultValue,
+			} as unknown as WorkspaceConfiguration;
```
DanielleMaywood: I realize you're just copying from prior tests here but either in this PR or a follow-up PR I'd prefer to see `as unknown as` be replaced with something type-safe.

src/remote/remote.ts
```diff
@@ -801,25 +822,48 @@ export class Remote {
...
-						if (action === "Reload") {
-							vscode.commands.executeCommand("workbench.action.reloadWindow");
-						}
-					});
-				break;
+
+				const newValue = getValue();
+				if (
+					JSON.stringify(newValue) !==
+					JSON.stringify(appliedValues.get(setting))
```
DanielleMaywood: Can we rely on this? If the values we are comparing are anything other than a simple primitive type such as a string, list of primitive types, or integers, given two identical objects with different insertion order `JSON.stringify` will produce a different output.

```
> JSON.stringify({"a": 1, "b": 2})
'{"a":1,"b":2}'
> JSON.stringify({"b": 2, "a": 1})
'{"b":2,"a":1}'
> JSON.stringify({"a": 1, "b": 2}) === JSON.stringify({"b": 2, "a": 1})
false
```

test/unit/cliConfig.test.ts
EhabY: This makes the tests much cleaner, if we wanted to provide exact types for all tests we'd need to override the full functionality. Like `ws` or similar, we can have 100s of fields :/.

Creating exact mocks with all the fields seems overkill when the test would fail immediately if we do not have a mock implementation anyway.

src/remote/remote.ts
EhabY: Oooh nice catch, replaced with node's `util.isDeepStrictEqual`. IIRC all the settings are primitives or arrays of primitives, but this is better if we decide to use `type: "object"` in the future!

test/unit/cliConfig.test.ts
DanielleMaywood: I get what you're saying although we could alternatively narrow down the type of the function since we only care about having a `get` function on it.

```ts
/**
 * Returns the raw global flags from user configuration.
 */
export function getGlobalFlagsRaw(
): string[] {
  return configs.get<string[]>("coder.globalFlags", []);
}
```

EhabY: Huh, fair enough actually, this also communicates better about the functionality this method provide. I'll add this and remove the `as unknown` in tests (I'll use the incomplete `MockConfigurationProvider`)

DanielleMaywood (approved):

PR #684, opened 2025-12-15
Fix SetEnv SSH config parsing and accumulation

### LINKED ISSUE #559: SSH config cannot use SetEnv without multiple problems
1. The only way to set an env is to use `SetEnv=MY_VAR=value`. If you use `SetEnv MY_VAR=true`, the env var gets added to the config incorrectly and SSH fails to load because of bad splitting: https://github.com/coder/vscode-coder/blob/2d7dac8a5f5d9ce2cc6746490c73c8873737034f/src/remote.ts#L716-L733

2. Setting any `SetEnv` variables causes the default `SetEnv CODER_SSH_SESSION_TYPE=vscode` to be removed. This is because we use a map to store the key/value pairs.

3. You cannot have multiple `SetEnv` config lines, only one will be processed.

EhabY:
- Add `parseSshConfig` function to correctly parse `SetEnv MY_VAR=value` (previously split on first `=` producing incorrect key/value)
- Modify `mergeSshConfigValues` to concatenate `SetEnv` values instead of replacing, preserving the default `CODER_SSH_SESSION_TYPE`
- Ignore empty `SetEnv` values to prevent accidentally clearing defaults

Closes #559

### DIFF
--- a/src/remote/sshConfig.ts
+++ b/src/remote/sshConfig.ts
@@ -51,7 +51,7 @@ export function parseSshConfig(lines: string[]): Record<string, string> {
 			}
 
 			const key = keyMatch[0];
-			const separator = line[key.length];
+			const separator = line.at(key.length);
 			if (separator !== "=" && separator !== " ") {
 				return acc; // Malformed line
 			}
@@ -63,7 +63,7 @@ export function parseSshConfig(lines: string[]): Record<string, string> {
 				// Ignore empty SetEnv values
 				if (value !== "") {
 					const existing = acc["SetEnv"];
-					acc["SetEnv"] = existing ? `${existing} ${value}` : ` ${value}`;
+					acc["SetEnv"] = existing ? `${existing} ${value}` : value;
 				}
 			} else {
 				acc[key] = value;
@@ -102,7 +102,11 @@ export function mergeSshConfigValues(
 
 			// Special handling for SetEnv - concatenate values instead of replacing.
 			if (lower === "setenv") {
-				merged["SetEnv"] = `${config[key]}${value}`;
+				if (value === "") {
+					merged["SetEnv"] = config[key];
+				} else {
+					merged["SetEnv"] = `${config[key]} ${value}`;
+				}
 				return;
 			}
 
@@ -126,7 +130,7 @@ export function mergeSshConfigValues(
 
 		// Special handling for SetEnv - concatenate if already exists
 		if (lower === "setenv" && merged["SetEnv"]) {
-			merged["SetEnv"] = `${merged["SetEnv"]}${value}`;
+			merged["SetEnv"] = `${merged["SetEnv"]} ${value}`;
 		} else {
 			merged[correctCaseKey] = value;
 		}
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -46,7 +46,7 @@ import {
 	SSHConfig,
 	type SSHValues,
 	mergeSshConfigValues,
-	parseSshConfig as parseSshConfig,
+	parseSshConfig,
 } from "./sshConfig";
 import { SshProcessMonitor } from "./sshProcess";
 import { computeSSHProperties, sshSupportsSetEnv } from "./sshSupport";
@@ -806,7 +806,7 @@ export class Remote {
 		if (sshSupportsSetEnv()) {
 			// This allows for tracking the number of extension
 			// users connected to workspaces!
-			sshValues.SetEnv = " CODER_SSH_SESSION_TYPE=vscode";
+			sshValues.SetEnv = "CODER_SSH_SESSION_TYPE=vscode";
 		}
 
 		await sshConfig.update(safeHostname, sshValues, sshConfigOverrides);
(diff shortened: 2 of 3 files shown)


src/remote/sshConfig.ts
```diff
@@ -36,10 +36,48 @@ const defaultFileSystem: FileSystem = {
...
+	return lines.reduce(
+		(acc, line) => {
+			// Match key pattern (same as VS Code settings: ^[a-zA-Z0-9-]+)
+			const keyMatch = line.match(/^[a-zA-Z0-9-]+/);
+			if (!keyMatch) {
+				return acc; // Malformed line
+			}
+
+			const key = keyMatch[0];
+			const separator = line[key.length];
```
deansheather: This is just undefined if the line is only a key right? e.g. `Set-Env` only

src/remote/sshConfig.ts
```diff
@@ -62,11 +100,17 @@ export function mergeSSHConfigValues(
 			const value = overrides[correctCaseKey];
 			delete caseInsensitiveOverrides[lower];
 
-			// If the value is empty, do not add the key. It is being removed.
-			if (value === "") {
+			// Special handling for SetEnv - concatenate values instead of replacing.
+			if (lower === "setenv") {
+				merged["SetEnv"] = `${config[key]}${value}`;
```
deansheather: missing space?

src/remote/sshConfig.ts
```diff
@@ -78,7 +122,14 @@ export function mergeSSHConfigValues(
 	// Add remaining overrides.
 	Object.keys(caseInsensitiveOverrides).forEach((lower) => {
 		const correctCaseKey = caseInsensitiveOverrides[lower];
-		merged[correctCaseKey] = overrides[correctCaseKey];
+		const value = overrides[correctCaseKey];
+
+		// Special handling for SetEnv - concatenate if already exists
+		if (lower === "setenv" && merged["SetEnv"]) {
+			merged["SetEnv"] = `${merged["SetEnv"]}${value}`;
```
deansheather: missing space?

src/remote/sshConfig.ts
EhabY: Indeed, used `line.at(key.length)` to enforce the `string | undefined` type

src/remote/sshConfig.ts
EhabY: It would technically work since all `value`s have a leading space (in `parseSshConfig`) but this is a bit fragile so I'll make it more explicit!

src/remote/sshConfig.ts
EhabY: Same here

EhabY:
> Are there any other ssh options that can be specified multiple times? Would be nice to support those as well

I had Claude look into the `ssh_config` documentation and here is the full list. We only care about multiple entries so in this case, it would be `IdentityFile`, `CertificateFile`, `SetEnv`, `SendEnv`, `LocalForward`, `RemoteForward`, `DynamicForward` can be accumulating. Alternatively, why not let the SSH client handle it? Like we can just pass the configs as is (`override` configs would be the put first since it's "first match wins"). I do not love that we have to re-implement this override behavior ourselves :thinking:. If we want leaner files then we can still apply the overriding but we can always add entries from the list above as is (we do not manually combine them but pass them as is, such that `override` is earlier than the base config).

<br/>

# Claude's Analysis

<html><head></head><body><h1>SSH Client (<code>ssh_config</code>) Accumulating Options</h1>
<p>Options that <strong>accumulate</strong> across multiple declarations, unlike standard "first match wins" behavior.</p>

Option | Example | Notes
-- | -- | --
IdentityFile | IdentityFile ~/.ssh/id_ed25519<br>IdentityFile ~/.ssh/id_rsa | Tried in sequence
CertificateFile | CertificateFile ~/.ssh/cert1.pub<br>CertificateFile ~/.ssh/cert2.pub | Tried in sequence
SetEnv | SetEnv FOO=bar<br>SetEnv BAZ=qux | Both sent to server
SendEnv | SendEnv LANG LC_*<br>SendEnv MY_VAR | Prefix with - to clear
LocalForward | LocalForward 8080 localhost:80<br>LocalForward 9090 localhost:90 | Multiple tunnels
RemoteForward | RemoteForward 8080 localhost:80 | Multiple tunnels
DynamicForward | DynamicForward 1080<br>DynamicForward 1081 | Multiple SOCKS proxies
GlobalKnownHostsFile | GlobalKnownHostsFile /etc/ssh/known_hosts /etc/ssh/known_hosts2 | Space-separated
UserKnownHostsFile | UserKnownHostsFile ~/.ssh/known_hosts ~/.ssh/known_hosts2 | Space-separated
PermitRemoteOpen | PermitRemoteOpen host1:80 host2:443 | Space-separated
CanonicalDomains | CanonicalDomains example.com corp.local | Space-separated


<h2>Source</h2>
<p><a href="https://man.openbsd.org/ssh_config.5">OpenBSD ssh_config(5)</a> — the canonical reference (OpenSSH is developed by OpenBSD).</p></body></html>

PR #692, opened 2025-12-16
Fix command visibility in the VS Code Command Palette

EhabY:
## Problem

Commands with authentication requirements (like Logout, Create Workspace, Refresh Workspaces) were appearing in the Command Palette even when the user wasn't logged in. This happened because `when` clauses placed directly on command definitions do not affect the Command Palette visibility.

## Solution

- Move all `when` clauses from command definitions to `menus.commandPalette` where they actually control visibility
- Standardize command titles to use `category: "Coder"` instead of prefixing titles with "Coder:"

## Additional Changes (Sonarlint fixes)

- Use `node:fs/promises` import protocol for consistency
- Make string sorting explicit with `localeCompare()` to avoid locale-dependent behavior

### DIFF
--- a/test/unit/remote/sshProcess.test.ts
+++ b/test/unit/remote/sshProcess.test.ts
@@ -127,6 +127,27 @@ describe("SshProcessMonitor", () => {
 			expect(find).toHaveBeenCalledWith("port", 33333);
 		});
 
+		it("sorts output_logging_ directories using localeCompare for consistent ordering", async () => {
+			// localeCompare differs from default sort() for mixed case
+			vol.fromJSON({
+				"/logs/output_logging_a/1-Remote - SSH.log": "-> socksPort 11111 ->",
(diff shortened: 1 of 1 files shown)


src/remote/sshProcess.ts
```diff
@@ -256,8 +256,9 @@ export class SshProcessMonitor implements vscode.Disposable {
 		const targetPid = this.currentPid;
 		while (!this.disposed && this.currentPid === targetPid) {
 			try {
-				const logFiles = await fs.readdir(logDir);
-				logFiles.sort().reverse();
+				const logFiles = (await fs.readdir(logDir))
+					.sort((a, b) => a.localeCompare(b))
```
mtojek, line 260: > localeCompare

smells like a case you may want to cover with units :)

EhabY: Ah nice catch actually, I've added a couple of tests

PR #693, opened 2025-12-16
Add OAuth 2.1 authentication support

### LINKED ISSUE #586: feat: support OAuth in the VS Code extension

EhabY:
Implements OAuth 2.1 with PKCE as an alternative authentication method to session tokens. When connecting to a Coder deployment that supports OAuth, users can choose between OAuth and legacy token authentication.

Key changes:

OAuth Flow:
- Add OAuthSessionManager to handle the complete OAuth lifecycle: dynamic client registration, PKCE authorization flow, token exchange, automatic refresh, and revocation
- Add OAuthMetadataClient to discover and validate OAuth server metadata from the well-known endpoint, ensuring server meets OAuth 2.1 requirements
- Handle OAuth callbacks via vscode:// URI handler with cross-window support for when callback arrives in a different VS Code window

Token Management:
- Store OAuth tokens (access, refresh, expiry) per-deployment in secrets
- Store dynamic client registrations per-deployment in secrets
- Proactive token refresh when approaching expiry (via ~~response interceptor~~ timers)
- Reactive token refresh on 401 responses with automatic request retry
- Handle OAuth errors (invalid_grant, invalid_client) by prompting for re-authentication

Integration:
- Add auth method selection prompt when server supports OAuth
- Attach OAuth interceptors to CoderApi for automatic token refresh (+ detach when no longer using OAuth)
- Clear OAuth state when user explicitly chooses token auth
- DeploymentManager coordinates OAuth session state with deployment changes

Error Handling:
- Typed OAuth error classes (InvalidGrantError, InvalidClientError, etc.)
- Parse OAuth error responses from token endpoint
- Show re-authentication modal for errors requiring user action

Closes #586


src/oauth/types.ts
```diff
@@ -0,0 +1,163 @@
+// OAuth 2.1 Grant Types
```
code-asher: Did we try to pull these types from coder/coder?  Some of these look slightly different, which to me implies these types are wrong (or maybe coder/coder's types are wrong).

For example `ClientRegistrationRequest` has required `redirect_uris`  while Coder's is optional.  Also we have an `application_type` field that does not seem to exist in either coder/coder nor in RFC 7591 nor 7592?  Where does that come from?  There may be other mismatches.

coder/coder does lack all the string enum types though unfortunately, maybe we can look into adding those on the coder/coder side at some point.

src/core/secretsManager.ts
```diff
@@ -7,7 +8,12 @@ import type { Deployment } from "../deployment/types";
 
 // Each deployment has its own key to ensure atomic operations (multiple windows
 // writing to a shared key could drop data) and to receive proper VS Code events.
-const SESSION_KEY_PREFIX = "coder.session.";
+const SESSION_KEY_PREFIX = "coder.session." as const;
```
code-asher: If we need the type to be the string literal rather than `string` I wonder if something like `const SESSION_KEY_PREFIX: "coder:session." = "coder.session.";` would work.  The repetition is unfortunate but the `as const` is just a bit obscure.

Or maybe an enum of prefixes would be more appropriate?

src/core/secretsManager.ts
```diff
@@ -118,39 +181,27 @@ export class SecretsManager {
...
-
-	private async clearSessionAuth(safeHostname: string): Promise<void> {
-		const sessionKey = this.getSessionKey(safeHostname);
-		await this.secrets.delete(sessionKey);
+		// Extract relevant fields before serializing
+		const state: SessionAuth = {
+			url: auth.url,
+			token: auth.token,
+			...(auth.oauth && { oauth: auth.oauth }),
+		};
```
code-asher: Why do we do this?  At first I thought maybe `SessionAuth` was a superset of what we store or a different type, but we are converting it to another `SessionAuth` and it seems to have no extra fields so could we not serialize it directly?

src/core/secretsManager.ts
```diff
@@ -181,7 +232,10 @@ export class SecretsManager {
 	 * Clear all auth data for a deployment and remove it from the usage list.
 	 */
 	public async clearAllAuthData(safeHostname: string): Promise<void> {
-		await this.clearSessionAuth(safeHostname);
+		await Promise.all([
+			this.clearSessionAuth(safeHostname),
+			this.clearOAuthClientRegistration(safeHostname),
```
code-asher, line 257: I have not gone through the whole code yet, but my first impression here is that it feels racy to store the auth/oauth-related details in two separate keys.  For example using the refresh token in the auth key against the wrong client in the client key.  Could/should they be stored together so they can be updated atomically?

src/core/secretsManager.ts
```diff
@@ -234,4 +288,56 @@ export class SecretsManager {
...
+			}
+
+			try {
+				const data = await this.secrets.get(OAUTH_CALLBACK_KEY);
+				if (data) {
+					const parsed = JSON.parse(data) as OAuthCallbackData;
+					listener(parsed);
+				}
+			} catch {
+				// Ignore parse errors
```
code-asher: If we get invalid JSON from the callback URL should we handle that?  Otherwise I suppose we would keep waiting and not show an error.

src/core/secretsManager.ts
```diff
@@ -234,4 +288,56 @@ export class SecretsManager {
 
 		return safeHostname;
 	}
+
+	/**
+	 * Write an OAuth callback result to secrets storage.
+	 * Used for cross-window communication when OAuth callback arrives in a different window.
+	 */
+	public async setOAuthCallback(data: OAuthCallbackData): Promise<void> {
+		await this.secrets.store(OAUTH_CALLBACK_KEY, JSON.stringify(data));
```
code-asher: If I understand correctly, we are not actually needing to store the callback response, we are just using the secret store for the event handling, to pass the data from the uri handler to the waiting authorizer right?  Wondering if we should emit directly to registered handlers and skip the extra step of storing the state, if that is right.

Also could be potentially weird if two instances are trying to log in and both write this key.

Maybe it should even go on some uri handler class, like `this.uriHandler.onDidGetOauthCallback()`, but idk.  Mostly I was just trying to figure out why we need to store this but I think we are just using it for the side effect of having an event.

src/deployment/deploymentManager.ts
```diff
@@ -125,9 +133,14 @@ export class DeploymentManager implements vscode.Disposable {
...
 		}
 
+		// Register auth listener before setDeployment so background token refresh
+		// can update client credentials via the listener
 		this.registerAuthListener();
 		this.updateAuthContexts();
 		this.refreshWorkspaces();
+
+		await this.oauthSessionManager.setDeployment(deployment);
+		await this.oauthInterceptor.setDeployment(deployment.safeHostname);
```
code-asher: Have not looked into these yet but at a glance feels odd these take slightly different things, could the interceptor be made to take a full `deployment`?

src/oauth/authorizer.ts
```diff
@@ -0,0 +1,347 @@
...
+			this.logger.debug("Redirect URI changed, re-registering client");
+		}
+
+		if (!metadata.registration_endpoint) {
+			throw new Error("Server does not support dynamic client registration");
+		}
+
+		const registrationRequest: ClientRegistrationRequest = {
+			redirect_uris: [redirectUri],
+			application_type: "web",
```
code-asher: Mentioned this before but where does `application_type` come from?  I did see it mentioned in a comment in `coder/coder` but for the life of me cannot find details about it.  Is it a bespoke thing maybe?

src/oauth/authorizer.ts
```diff
@@ -0,0 +1,347 @@
...
+				this.logger.warn(
+					`Requested scopes not in server's supported scopes: ${unsupportedScopes.join(", ")}. Server may still accept them.`,
+					{ supported_scopes: metadata.scopes_supported },
+				);
+			}
+		}
+
+		const params = new URLSearchParams({
+			client_id: clientId,
+			response_type: RESPONSE_TYPE,
```
code-asher, line 204: Just a note to ourselves that we are not exposing these params as a type in `coder/coder` but we could consider it so we can import them.

src/oauth/authorizer.ts
```diff
@@ -0,0 +1,347 @@
...
+
+		if (!metadata.registration_endpoint) {
+			throw new Error("Server does not support dynamic client registration");
+		}
+
+		const registrationRequest: ClientRegistrationRequest = {
+			redirect_uris: [redirectUri],
+			application_type: "web",
+			grant_types: ["authorization_code"],
+			response_types: ["code"],
```
code-asher: Would it make sense to use the consts here?  `[AUTH_GRANT_TYPE]` and `[RESPONSE_TYPE]`

src/oauth/authorizer.ts
```diff
@@ -0,0 +1,347 @@
...
+						);
+					},
+					timeoutMins * 60 * 1000,
+				);
+
+				const listener = this.secretsManager.onDidChangeOAuthCallback(
+					({ state: callbackState, code, error }) => {
+						if (callbackState !== state) {
+							return;
+						}
```
code-asher, line 270: Thinking out loud, how do we get a mismatched callback state?  If we start a new login?  In that case should we go ahead and reject the promise?  Since I think at that point there is no way this promise could recover.

src/oauth/authorizer.ts
```diff
@@ -0,0 +1,347 @@
...
+	private async exchangeToken(
+		code: string,
+		verifier: string,
+		axiosInstance: AxiosInstance,
+		metadata: OAuthServerMetadata,
+		registration: ClientRegistrationResponse,
+	): Promise<TokenResponse> {
+		this.logger.debug("Exchanging authorization code for token");
+
+		const params: TokenRequestParams = {
```
code-asher: Similarly here, we could consider exposing this type from `coder/coder` at some point.

src/oauth/authorizer.ts
```diff
@@ -0,0 +1,347 @@
...
+			metadata,
+			registration.client_id,
+			state,
+			challenge,
+		);
+
+		const callbackPromise = new Promise<{ code: string; verifier: string }>(
+			(resolve, reject) => {
+				// Track reject for disposal
+				this.pendingAuthReject = reject;
```
code-asher, line 249: If there is an existing `this.pendingAuthReject` should we go ahead and call it here?  The timeout will get it eventually, but it does mean `dispose` could miss some promises if multiple were started.  I imagine if the user initiates a new login (idk if they actually can right now, but just in terms of the code) we would want to abort any pending logins.

src/oauth/axiosInterceptor.ts
```diff
@@ -0,0 +1,179 @@
...
+	 * Sync interceptor state with OAuth token presence.
+	 * Attaches when tokens exist, detaches when they don't.
+	 */
+	private async syncWithTokenState(): Promise<void> {
+		const isOAuth = await this.oauthSessionManager.isLoggedInWithOAuth();
+		if (isOAuth && this.interceptorId === null) {
+			this.attach();
+		} else if (!isOAuth && this.interceptorId !== null) {
+			this.detach();
+		}
```
code-asher: The attach/detach combined with async makes me nervous we get in a state where we should be attached/detached but are not yet.

It looks like the only thing we are doing here is handling 401s encountered after the remote setup.  Would it make sense to make a generic interceptor always attached, and then upon getting a 401 check if it was done with oauth and do a token refresh?  And if not oauth, or if the refresh fails, then show a prompt for a new token, so we can handle those cases as well (I think that may conflict with the 401 handling we have during remote setup, so maybe we leave that for later).

src/oauth/metadataClient.ts
```diff
@@ -0,0 +1,142 @@
...
+import type {
+	GrantType,
+	OAuthServerMetadata,
+	ResponseType,
+	TokenEndpointAuthMethod,
+} from "./types";
+
+const OAUTH_DISCOVERY_ENDPOINT = "/.well-known/oauth-authorization-server";
+
+const AUTH_GRANT_TYPE = "authorization_code" as const;
```
code-asher: Similar comment here about `as const`.

Also seeing some repeated consts, wonder if they could go in a common file?

src/oauth/metadataClient.ts
```diff
@@ -0,0 +1,142 @@
...
+			);
+		}
+	}
+
+	private validatePKCEMethods(metadata: OAuthServerMetadata): void {
+		// PKCE has no RFC 8414 default - if undefined, server doesn't advertise support
+		const supported = metadata.code_challenge_methods_supported ?? [];
+		if (!includesAllTypes(supported, [PKCE_CHALLENGE_METHOD])) {
+			throw new Error(
+				`Server does not support required PKCE method: ${PKCE_CHALLENGE_METHOD}. Supported: ${supported.length > 0 ? supported.join(", ") : "none"}`,
```
code-asher: Only this one has a length check, should all of them have one?  I imagine the server could return empty arrays for any of these.

src/oauth/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * OAuth callback path for handling authorization responses (RFC 6749).
+ */
+export const CALLBACK_PATH = "/oauth/callback";
+
+/**
+ * Default expiry time for OAuth access tokens when the server doesn't provide one.
+ */
+const ACCESS_TOKEN_DEFAULT_EXPIRY_MS = 60 * 60 * 1000;
```
code-asher, line 16: The default is just something we arbitrarily decide, right, not determined by an RFC or anything?

src/uri/uriHandler.ts
```diff
@@ -177,3 +179,25 @@ async function setupDeployment(
...
+	const secretsManager = serviceContainer.getSecretsManager();
+
+	const code = params.get("code");
+	const state = params.get("state");
+	const error = params.get("error");
+
+	if (!state) {
+		logger.warn("Received OAuth callback with no state parameter");
+		return;
+	}
```
code-asher, line 195: If this happens should we emit so we can abort the auth?  Not sure if this is a recoverable error.  I suppose the user could try opening the auth page again, but if the state was missing presumably it will still be missing.

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,601 @@
...
+			url: auth.url,
+			token: auth.token,
+		});
+	}
+
+	/**
+	 * Clear all refresh-related state: in-flight promise, throttle, timer, and listener.
+	 */
+	private clearRefreshState(): void {
+		this.refreshPromise = null;
```
code-asher, line 159: Although we clear the refresh promise I think I am not seeing anything that would prevent the refresh promise from finishing out.  Could be a risk that we clear the refresh state, but the promise had a request out and it will still try to update the token once that request finishes.  Possibly need a cancellation token or some such.

src/oauth/sessionManager.ts
```diff
@@ -0,0 +1,601 @@
...
+
+		if (!metadata.revocation_endpoint) {
+			this.logger.info("No revocation endpoint available, skipping revocation");
+			return;
+		}
+
+		this.logger.info("Revoking refresh token");
+
+		const params: TokenRevocationRequest = {
+			token: tokenToRevoke,
```
code-asher, line 507: I think we are not exposing this type in `coder/coder` either if we want to do that eventually.

src/core/secretsManager.ts
EhabY: Apparently TS is already smart enough, we do not need the `as const` so I've removed it!

src/core/secretsManager.ts
EhabY: This is because I used to pass a super set before and realized that `JSON.stringify` would serialize extra fields and thus when reading this we get the superset which caused strange hidden issues (I think we should always constrict this to the exact type)

src/core/secretsManager.ts
EhabY: There will always be some racy stuff here because we have two kinds of synchronization: inter-process, and intra-process. The latter we can sync easily but the former is synced due to the secrets API.

If we combine those two keys into the same key, then it would mean if we want to update only one field we have to do the following:

`read` -> `modify` -> `write`, the secrets API does not support an `update` method or something similar so then we'll have some racy writes which is arguably worse.

src/core/secretsManager.ts
EhabY: > we are just using the secret store for the event handling, to pass the data from the uri handler to the waiting authorizer right?

Yes, this is because when you have a `vscode://` link, ONLY the first window handles it. So potentially, you can have a window A, B, C. Window B logs in using OAuth but the callback is sent to Window A which has no context of the auth that is being performed, so we use the secrets for interprocess communication (we could also write to a shared file and have file watchers but that's a hassle)

src/core/secretsManager.ts
EhabY: Added a log on the error level here

src/deployment/deploymentManager.ts
EhabY: Done

src/oauth/authorizer.ts
EhabY: It's part of the OpenID registration (https://openid.net/specs/openid-connect-registration-1_0.html)

I suppose we do not need this now so I'll stick to OAuth only (and use the types in `coder/coder`)

src/oauth/authorizer.ts
EhabY: Done!

src/oauth/authorizer.ts
EhabY: Sure, I'll add this so it's more encapsulated.

Currently we queue login attempts so it should never be possible  but that's a good anyway!

src/oauth/authorizer.ts
EhabY: Good point, but the annoying thing here is cross-window communication. Technically, we can have multiple windows and attempt to login to OAuth to different deployments. In that case, we trigger this code. Maybe I add a `warn` since it's UNLIKEY to happen but still valid?

BTW checking the state here is a security measure to prevent MITM attack.

src/oauth/metadataClient.ts
EhabY: Yes that's fair, done!

src/oauth/metadataClient.ts
EhabY: Ah, right. Fixed!

src/oauth/utils.ts
EhabY: The RFC only recommends that the server provide this, this one hour value is a very conservative and arbitrary. I've added a comment to explain this

src/uri/uriHandler.ts
EhabY: Similar to the comment above, but we could have multiple windows that potentially each login to a different (or the same) deployment so that's why I'm hesitant about this. Maybe we can show something in the UI though and the user can cancel the flow if they wish?

src/oauth/sessionManager.ts
EhabY: Added `AbortSignal` handling

src/oauth/axiosInterceptor.ts
EhabY: > It looks like the only thing we are doing here is handling 401s encountered after the remote setup

Only handling 401 but not just after the remote setup, but also when you are logged in but not in a workspace yet (extension client). Also `attach` and `detach` do check the `interceptId` so racy conditions should be okay 🤔 

I can add more generic 401 handling here or in a follow up, I don't mind either way

src/oauth/authorizer.ts
EhabY: Removed this for now, I'll look into using the types from `coder/coder`

src/core/secretsManager.ts
code-asher: I suppose if we removed it now someone could change this to a superset later and forget they should constrict it.

But, if we are going that route we should probably do it for all the properties including nested ones rather than just spread `auth.oauth` right?

src/core/secretsManager.ts
code-asher: Fair also after reading and thinking a bit more, it does feel like it would be weird to combine the auth details and client registration, they are not really the same thing.

src/core/secretsManager.ts
code-asher: Ooo good point I had not realized that only one window would get the link event.

So, because one window might start the login and a different window might get the event, we need something that can talk to all the instances, thus the secret store.  Makes sense!

src/core/secretsManager.ts
code-asher: Why not let this throw?  It means the login failed right, if Coder redirected back with invalid state.

src/oauth/authorizer.ts
code-asher: Ooooo an OIDC thing gotcha that makes sense.

src/oauth/authorizer.ts
code-asher: Ohhh right multiple logins could be in flight.

I wonder though, it might be even more unlikely that a user is trying to log into multiple things at the same time, so maybe a mismatched state is actually the more likely thing lol

But yeah a warn at least could be nice, if there is a problem with the state, not logging anything could make it tough to debug.

src/oauth/axiosInterceptor.ts
code-asher: Oh true true, I was thinking about how the remote setup already handles 401s so in my mind this addition was mostly for the calls that came after (and before, forgot about that).

For the racy condition, how would checking the intercept ID help?  If the oauth state or deployment changes while `isLoggedInWithOauth` is in flight, we can end up attaching or detaching when we should do the opposite.

1. log out
2. isLoggedInWithOauth yields the thread
3. log in with oauth to new deployment, set credentials and deployment
4. isLoggedInWithOauth called again, yields
5. first returns with the old token, but it no longer matches the deployment so it clears, which yields
6. second returns with the new token, attaches the interceptor 
8. first finishes clearing and returns undefined, detaching the interceptor, but we should actually be attached

Would be insane timing lol.  But I feel like we probably want an interceptor always attached to handle all 401s eventually anyway, and then we can remove the ones in the remote setup so we have it all in one place (imo it could be a follow up).

Actually I guess this has a second problem in that it would clear the token when it should not, but that seems more a race in `getStoredTokens()` than a race here.

src/oauth/utils.ts
code-asher: sweet, sounds good to me

src/uri/uriHandler.ts
code-asher: Oof and there is no way for us to tell which login this callback was for.  Bummer.

My intuition is that initiating multiple logins at a time will probably be uncommon, so if we do get a bad state *most likely* it does refer to whatever login is currently ongoing, and so it might make sense to surface this.

But yeah it does feel really bad that if someone is logging into multiple things at a time and we show some error or warning when really everything is fine.

If there was a way we could only error if the URI was unhandled, so we knew all logins ignored it because of the mismatched state, that could do the trick, because then we could show an error like "got a callback but none of the active logins matched it".  But not sure how we would figure that out.

I feel like we leave this for now with a warning and see how it goes.

code-asher (approved):
Looks good to me!  I think we may want to consider an always-attached 401 interceptor, and I wonder if we should remove the clearing from `getStoredTokens()` to avoid accidentally deleting tokens due to races (we could just return undefined without clearing I think).

src/core/secretsManager.ts
code-asher: Er, not throw, but more like emit an error on the listener.

But, after some of our other conversations, since this could be for any login in any window, maybe log is the best we can do after all.  :cry:

src/oauth/sessionManager.ts
EhabY: Added in https://github.com/coder/coder/pull/21468

src/oauth/types.ts
EhabY: I pulled all the possible types from `coder/coder` but some are missing or have incorrect fields, this should fix that: https://github.com/coder/coder/pull/21468

src/oauth/authorizer.ts
EhabY: Should be exposed in https://github.com/coder/coder/pull/21468

src/oauth/authorizer.ts
EhabY: https://github.com/coder/coder/pull/21468

src/oauth/authorizer.ts
EhabY: > I wonder though, it might be even more unlikely that a user is trying to log into multiple things at the same time, so maybe a mismatched state is actually the more likely thing lol

Valid lol, but I added logging for now, could make the promise reject if needed :pray:

src/uri/uriHandler.ts
EhabY: Agreed, tbh all the cross-window stuff and error states can be hard to deal with. We'll discover some of these issues as we go on but I made sure to always log something so at least it doesn't silently fail..

src/oauth/axiosInterceptor.ts
EhabY: Made it always attached. I made it handle all 401s using a callback, I can fully unify showing the dialog in a follow up since I didn't want to change the behavior too much now since we might need to reiterate on it.

src/core/secretsManager.ts
EhabY: I've used `zod` to strip the runtime and had the type generated out of the schema. This way if we change this then it will still work!

EhabY:
> I think we may want to consider an always-attached 401 interceptor, and I wonder if we should remove the clearing from `getStoredTokens()` to avoid accidentally deleting tokens due to races (we could just return undefined without clearing I think).

I've made the 401 interceptor always attached but re-auths using a callback because I didn't want to change the behavior now (could be a follow up). So now `remote.ts` uses that to show a dialog and login while `extension.ts` would just tell the user to login from the workspaces view.

Also I've removed the clearing of tokens in `getStoredTokens()`!

src/core/secretsManager.ts
code-asher: amazing!

PR #694, opened 2025-12-19
Ensure login prompt before opening workspace from URI

### LINKED ISSUE #680: Missing login prompt if URI handler is missing the token
In version 1.9.2 at least, when the token was missing we would prompt to log in if the token was not included in the URI (or if it was invalid, I think), but now it seems we silently fail in both cases.

EhabY:
When opening a workspace via URI without a token, ensure we prompt for login and set the deployment before attempting to open. Extract URI handler into dedicated module with route-based dispatch.

Closes #680

### DIFF
--- a/src/login/loginCoordinator.ts
+++ b/src/login/loginCoordinator.ts
@@ -217,7 +217,7 @@ export class LoginCoordinator {
 				providedToken,
 				isAutoLogin,
 			);
-			if (result !== "retry") {
+			if (result !== "unauthorized") {
 				return result;
 			}
 		}
@@ -228,7 +228,7 @@ export class LoginCoordinator {
 		);
 		if (auth?.token && auth.token !== providedToken) {
 			const result = await this.tryTokenAuth(client, auth.token, isAutoLogin);
-			if (result !== "retry") {
+			if (result !== "unauthorized") {
 				return result;
 			}
 		}
@@ -251,20 +251,20 @@ export class LoginCoordinator {
 	}
 
 	/**
-	 * Returns 'retry' on 401 to signal trying next token.
+	 * Returns 'unauthorized' on 401 to signal trying next token source.
 	 */
 	private async tryTokenAuth(
 		client: CoderApi,
 		token: string,
 		isAutoLogin: boolean,
-	): Promise<LoginResult | "retry"> {
+	): Promise<LoginResult | "unauthorized"> {
 		client.setSessionToken(token);
 		try {
 			const user = await client.getAuthenticatedUser();
 			return { success: true, token, user };
 		} catch (err) {
 			if (isAxiosError(err) && err.response?.status === 401) {
-				return "retry";
+				return "unauthorized";
 			}
 			this.showAuthError(err, isAutoLogin);
 			return { success: false };


src/uri/uriHandler.ts
```diff
@@ -0,0 +1,182 @@
...
+	if (!url) {
+		throw new Error("url must be provided or specified as a query parameter");
+	}
+
+	const safeHostname = toSafeHost(url);
+
+	const token: string | null = params.get("token");
+	if (token !== null) {
+		await secretsManager.setSessionAuth(safeHostname, { url, token });
+	}
```
code-asher: Should we wait to set the token until after `ensureLoggedIn`?  To avoid potentially storing an invalid or blank token.  

Not a huge deal since usually the token should be set and valid, I just noticed that if I am already logged in and the URI contains an invalid or blank token, I get partially logged out (my workspace list is empty but I still look logged-in) and if I reload VS Code I am completely logged out.  Just seems unfortunate to lose the perfectly good token I had stored previously.

Actually `ensureLoggedIn` already sets auth so I suppose we can skip it here.  Although, it also sets it before validating.  Maybe something we can visit later.

src/uri/uriHandler.ts
```diff
@@ -0,0 +1,182 @@
...
+	const result = await loginCoordinator.ensureLoggedIn({
+		safeHostname,
+		url,
+	});
+
+	if (!result.success) {
+		throw new Error("Failed to login to deployment from URI");
+	}
+
+	await deploymentManager.setDeploymentIfValid({
```
code-asher: Why not call `setDeployment` directly?  The documentation on the function says "Use this when you already have the user from a successful login" which seems to fit here.

src/deployment/deploymentManager.ts
```diff
@@ -83,8 +83,13 @@ export class DeploymentManager implements vscode.Disposable {
 	 * Returns true if deployment was changed, false otherwise.
 	 */
 	public async setDeploymentIfValid(
-		deployment: Deployment & { token?: string },
+		deployment: DeploymentWithAuth,
 	): Promise<boolean> {
+		if (deployment.user) {
+			await this.setDeployment({ ...deployment, user: deployment.user });
+			return true;
+		}
```
code-asher: Continuing with the previous comment, if we already have a user, could we just call `setDeployment` directly?

The comment on `setDeploymentIfValid` says it validates auth so if we do keep this, we might mention that calling it with an already-validated deployment is the same as calling `setDeployment` directly and no additional validation is performed in that case.

src/uri/uriHandler.ts
EhabY: `setDeployment` expects a defined `User`, `ensureLoggedIn` returns `user?: User`, thus why we MIGHT need to get the user in some cases

src/deployment/deploymentManager.ts
EhabY: Fair enough, I'll update the documentation

src/uri/uriHandler.ts
EhabY: Oooh that's a valid point, I've added an optional `token` parameter so we try to login in the following order:

1. mTLS (non-token auth)
2. Provided token
3. Stored token
4. User prompt

We store whatever succeeds in this order

src/uri/uriHandler.ts
EhabY: Made `LoginCoordinator` always return a `User` now (which fits more into the contracts of this class anyway). So no need for the change in `deploymentManager`

src/deployment/deploymentManager.ts
EhabY: No longer relevant!

src/uri/uriHandler.ts
code-asher: Ahh that makes sense.  Nice!

src/login/loginCoordinator.ts
```diff
@@ -191,54 +201,93 @@ export class LoginCoordinator {
...
+		if (providedToken) {
+			const result = await this.tryTokenAuth(
+				client,
+				providedToken,
+				isAutoLogin,
 			);
-			storedToken = auth?.token;
-			if (storedToken) {
-				client.setSessionToken(storedToken);
+			if (result !== "retry") {
```
code-asher: To me "retry" sounds more like "try this token again later" but if I understand correctly it means to try the next token.  Could we return "next" or something like that?  Or add the status code or `reason` to a failed `LoginResult` and do `code !== 401` or something.

src/login/loginCoordinator.ts
```diff
@@ -191,54 +201,93 @@ export class LoginCoordinator {
...
+		// Try stored token (skip if same as provided)
+		const auth = await this.secretsManager.getSessionAuth(
+			deployment.safeHostname,
+		);
+		if (auth?.token && auth.token !== providedToken) {
+			const result = await this.tryTokenAuth(client, auth.token, isAutoLogin);
+			if (result !== "retry") {
+				return result;
 			}
+		}
```
code-asher, line 234: nice 👍

src/login/loginCoordinator.ts
EhabY: Renamed to "unauthorized"

PR #705, opened 2025-12-29
Reconnect WebSockets when configuration settings change

### LINKED ISSUE #691: Trigger a WebSocket reconnect when changing `mTLS` settings
We should listen to changes to the settings `coder.tlsCertFile` and `coder.tlsKeyFile` and trigger a WS re-connection (similar to `CoderApi#setCredentials`). Do take note of `api/utils.ts#needToken`.

EhabY:
## Summary
- WebSockets now automatically reconnect when proxy, TLS, or header settings change
- Previously users had to reload the window for these changes to take effect (or trigger a WS reconnect)
- Extracted reusable `watchConfigurationChanges` utility from `remote.ts`

## Settings that trigger reconnection
- `coder.headerCommand`
- `coder.insecure`
- `coder.tlsCertFile`, `coder.tlsKeyFile`, `coder.tlsCaFile`, `coder.tlsAltHost`
- `http.proxy`, `coder.proxyBypass`

Closes #691

### DIFF
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -16,6 +16,7 @@ import {
 import * as vscode from "vscode";
 import { type ClientOptions } from "ws";
 
+import { watchConfigurationChanges } from "../configWatcher";
 import { CertificateError } from "../error";
 import { getHeaderCommand, getHeaders } from "../headers";
 import { EventStreamLogger } from "../logging/eventStreamLogger";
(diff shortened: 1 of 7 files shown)


EhabY:
> But does seem like if the web socket is working we could leave it. Probably not worth the effort though, the settings probably change infrequently.

My thoughts as well, we could in theory just do `reconnectIfDisconnected` where we check the socket is valid but not sure if it makes sense (also not sure if we can get stuck in a state where the socket is broken without erroring - somehow -)

Issue #377, opened 2024-10-09
coder.coder-remote/log can grow quite large


aaronlehmann:
Noticed this:

```
$ du -sh ~/Library/Application\ Support/Code/User/globalStorage/coder.coder-remote/log
1.7G    /Users/alehmann/Library/Application Support/Code/User/globalStorage/coder.coder-remote/log
```

PR #706, opened 2025-12-30
Add network info cleanup and missing file retry logic

### LINKED ISSUE #51: Old JSON files are left in `net` folder of extension
The folder in question:

```
~/Library/Application Support/Code/User/globalStorage/coder.coder-remote/net/*.json
```

Contains a bunch (10 on a semi recent install) of JSON files.

EhabY:
## Summary
- Clean up old `<pid>.json` files in the network info directory on monitor start (files older than 1 hour)
- Search for a new SSH process after 5 consecutive failures to read the network info file (similar to existing stale file behavior)

## Details
Network info files (`<basePath>/net/<pid>.json`) are created by the Coder CLI and persist after processes terminate. This adds automatic cleanup of files not modified in the last hour.

Previously, if the network info file was missing (e.g., deleted or never created), the monitor would log errors indefinitely without attempting to find a new process. Now it behaves consistently with stale file detection - after 5 failures (with delays), it searches for a new SSH process.

Closes #51

### DIFF
--- a/src/remote/sshProcess.ts
+++ b/src/remote/sshProcess.ts
@@ -39,7 +39,7 @@ export interface SshProcessMonitorOptions {
 	remoteSshExtensionId: string;
 }
 
-// Cleanup threshold for old network info files
+// 1 hour cleanup threshold for old network info files
 const CLEANUP_MAX_AGE_MS = 60 * 60 * 1000;
 
 /**
@@ -100,8 +100,10 @@ export class SshProcessMonitor implements vscode.Disposable {
 						await fs.unlink(filePath);
 						deletedFiles.push(file);
 					}
-				} catch {
-					// File may have been deleted by another process or doesn't exist, ignore
+				} catch (error) {
+					if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
+						logger.debug(`Failed to clean up network info file ${file}`, error);
+					}
 				}
 			}
 


src/remote/sshProcess.ts
```diff
@@ -39,6 +39,9 @@ export interface SshProcessMonitorOptions {
 	remoteSshExtensionId: string;
 }
 
+// Cleanup threshold for old network info files
+const CLEANUP_MAX_AGE_MS = 60 * 60 * 1000;
```
johnstcn:
```suggestion
// 1 hour cleanup threshold for old network info files
const CLEANUP_MAX_AGE_MS = 60 * 60 * 1000;
```

src/remote/sshProcess.ts
```diff
@@ -70,6 +73,48 @@ export class SshProcessMonitor implements vscode.Disposable {
...
+				try {
+					const stats = await fs.stat(filePath);
+					const ageMs = now - stats.mtime.getTime();
+
+					if (ageMs > maxAgeMs) {
+						await fs.unlink(filePath);
+						deletedFiles.push(file);
+					}
+				} catch {
+					// File may have been deleted by another process or doesn't exist, ignore
```
johnstcn: suggest debug log here in case we need to troubleshoot

src/remote/sshProcess.ts
```diff
@@ -70,6 +73,48 @@ export class SshProcessMonitor implements vscode.Disposable {
...
+	): Promise<void> {
+		try {
+			const files = await fs.readdir(networkInfoPath);
+			const now = Date.now();
+
+			const deletedFiles: string[] = [];
+			for (const file of files) {
+				if (!file.endsWith(".json")) {
+					continue;
+				}
```
johnstcn, line 92: If each filename corresponds to a PID, we should also validate that the PID is no longer active.

EhabY: Oooh I like this actually, would mean that we might not need the 1 hour threshold :thinking: ?

I suppose the OS can re-use a PID though so it's not entirely accurate. We can do something like `isPidAlive(pid) || isAnHourOld` so it's more aggressive though not sure if it's super worth it for a PID check

src/remote/sshProcess.ts
EhabY: Added a debug if it's not `ENOENT` (File not found exception)

src/remote/sshProcess.ts
johnstcn: That's a good point. I guess in the worst case scenario the process would still have the file handle (although deleted)

EhabY: If the process has a file lock then we might error as EBUSY (or maybe PERM depending on the OS) but in any case that wouldn't cause any issues + we only delete it if it's been untouched in an hour (this is usually updated every ~3 seconds). IMO I do not think a PID check here is worthwhile

src/remote/sshProcess.ts
EhabY: Done!

src/remote/sshProcess.ts
code-asher: If I understand correctly (have not reviewed the code, just based off the description) this would delete logs from active connections?  Would make debugging any issue that happens after an hour difficult to diagnose.

EhabY: This is deleting the network info JSON, it contains stuff like the latency and is updated very frequently (otherwise the latency is stale)

EhabY: The log file issue can be handled here https://github.com/coder/vscode-coder/issues/377 (rotation with a maximum number of files or maybe date-based etc)

code-asher: Ahaha this is what I get for not fully reading.  But would this break the connection info on the bottom bar for active connections?

code-asher: Ohhh wait are you checking last modified time, not created?

code-asher: Yeah you are lol next time I will just read the code haha.  All good from me!

EhabY: Hahaha understandable, I have another PR you can read as well :see_no_evil:

PR #716, opened 2026-01-07
Modernize build tooling: esbuild + pnpm and add multi-version CI testing

EhabY:
## Summary

This PR modernizes the build infrastructure, resulting in >2x faster CI builds through better caching (pnpm) and faster tooling (esbuild).

## Build System

- Replace Webpack with esbuild for extension bundling - (x5 - x10 improvement)
- Add `esbuild.mjs` configuration with watch mode support and openpgp CJS alias fix
- Remove webpack, webpack-cli, and ts-loader dependencies
- Add --no-dependencies to vsce packaging (dependencies are bundled by esbuild)

## Package Manager Migration

- Migrate from yarn to pnpm - enables better dependency caching in CI
- Add pnpm-workspace.yaml with onlyBuiltDependencies for native modules
- Enable caching for prettier and eslint (`--cache --cache-strategy content`)

## CI/CD Improvements

- Use `pnpm/action-setup@v4` with built-in caching
- Remove global `npm install -g @vscode/vsce` - use local devDependency instead
- Add multi-version test matrix:
  - Unit tests: Electron 32 (VS Code 1.95 minimum) and latest
  - Integration tests: VS Code 1.95.0 and stable
- Add `scripts/test-electron.sh` for version-specific test runs

## Other Changes

- Update minimum VS Code engine: 1.73.0 → 1.95.0
- Update @types/node to v20, add explicit node >= 20 engine requirement


.github/workflows/ci.yaml
```diff
@@ -17,51 +17,88 @@ jobs:
...
+      - run: pnpm install
 
-      - run: yarn test:ci
+      - name: Run tests with Electron ${{ matrix.electron-version }}
+        run: ./scripts/test-electron.sh ${{ matrix.electron-version }}
+        env:
+          CI: true
+
+  test-integration:
+    name: Integration Test (VS Code ${{ matrix.vscode-version }})
```
code-asher, line 62: Very nice indeed!

eslint.config.mjs
```diff
@@ -157,9 +150,9 @@ export default defineConfig(
 		},
 	},
 
-	// Webpack config - CommonJS with Node globals
+	// Build config - ESM with Node globals
```
code-asher, line 153: This could be me being dense but it looks like the `esbuild.mjs` outputs CJS, not ESM like implied in this comment?

Or is this just saying that the build config itself is ESM, not that the output of it is ESM?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,9 @@
+# Native modules allowed to run install scripts (pnpm blocks by default for security)
```
code-asher, line 1: Oh interesting, did not know about this.  Definitely a plus for pnpm

EhabY: Indeed, also cache being smarter/faster and no phantom dependencies!

.github/workflows/ci.yaml
EhabY: We only have one basic integration test but at least we can now just add them and they would run on CI!

eslint.config.mjs
EhabY: The latter, the build config is ESM (the webpack used to be commonJS with the `require` syntax and all)

The file itself is a module JS (`mjs`) which is ESM

PR #730, opened 2026-01-16
Improve session expiry UX with login prompt and automatic recovery

### LINKED ISSUE #723: UX papercuts when an oauth token is revoked
To reproduce:
1. Log in with oauth locally (no remotes opened)
2. With VS Code still open and the Coder sidebar opened, revoke the token from the Coder dashboard

I assume these issues could also happen if the refresh token expired?  But not sure how to easily test that.

- [ ] The sidebar becomes this half-logged-in state where "my workspaces" is blank and "all workspaces" is populated.  To log back in, you must log out first then log in.  Instead, we should revert the sidebar to the logged out state where the "log in" button is shown.

- [ ] A notification saying to log in with the Coder sidebar appears.  We could add a log in button directly to this notification, might be nicer than having to go through the sidebar.

- [ ] Sometimes you can get two notifications, that both more or less say the same thing.  Could we consolidate them?  Personally I like the one in the corner better than the big modal popup one, at least for this scenario.

Image

- [ ] When I do log back in, my "all workspaces" view was empty.  A manual refresh fixes it, but if I leave VS Code open it never gets populated.  We are probably missing an initial call to populate those workspaces after logging back in.

EhabY:
## Summary

- Improves the user experience when sessions expire by showing a "Log In" button
- Keeps deployment context after expiry so users don't have to re-enter the URL
- Adds automatic session recovery when tokens are updated from another window
- Adds OAuth scope validation to detect when required permissions change

Fixes #723

### DIFF
--- a/src/deployment/deploymentManager.ts
+++ b/src/deployment/deploymentManager.ts
@@ -127,6 +127,7 @@ export class DeploymentManager implements vscode.Disposable {
 		// Register auth listener before setDeployment so background token refresh
 		// can update client credentials via the listener
 		this.registerAuthListener();
+		// Contexts must be set before refresh (providers check isAuthenticated)
 		this.updateAuthContexts(deployment.user);
 		this.refreshWorkspaces();
 
--- a/src/workspace/workspacesProvider.ts
+++ b/src/workspace/workspacesProvider.ts
@@ -50,6 +50,7 @@ export class WorkspaceProvider
 		private readonly getWorkspacesQuery: WorkspaceQuery,
 		private readonly client: CoderApi,
 		private readonly logger: Logger,
+		private readonly isAuthenticated: () => boolean,
 		private readonly timerSeconds?: number,
 	) {
 		// No initialization.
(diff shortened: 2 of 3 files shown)


src/deployment/deploymentManager.ts
```diff
@@ -182,7 +194,18 @@ export class DeploymentManager implements vscode.Disposable {
 				}
 
 				if (auth) {
-					this.client.setCredentials(auth.url, auth.token);
+					if (this.contextManager.get("coder.authenticated")) {
```
code-asher: nbd but would calling `this.isAuthenticated()` make sense here since it does the same thing?

src/workspace/workspacesProvider.ts
```diff
@@ -296,12 +296,21 @@ export class WorkspaceProvider
...
 		this.cancelPendingRefresh();
 		for (const watcher of this.agentWatchers.values()) {
 			watcher.dispose();
 		}
 		this.agentWatchers.clear();
+		this.workspaces = [];
+		this.refresh();
+	}
+
+	public dispose() {
```
code-asher, line 325: Feels like if we have `dispose` and `clear` they should do different things, like is `dispose` meant to ensure the provider cannot be used again or something?  In which case maybe we can formalize that with a boolean check or something.

EhabY: > dispose meant to ensure the provider cannot be used again or something?

Indeed, I'll add a boolean check then!

EhabY: Done

src/deployment/deploymentManager.ts
EhabY: Yes and done!

PR #732, opened 2026-01-16
Add automatic TLS client certificate refresh support

### LINKED ISSUE #714: Add certificate refresh command support for expired client certificates
## Summary

When client TLS certificates (e.g., metatron credentials) expire, the extension shows a cryptic error message:

> Failed to check user authentication: 18468360202752:error:10000415:SSL routines:OPENSSL_internal:SSLV3_ALERT_CERTIFICATE_EXPIRED:../../third_party/boringssl/src/ssl/tls_record.cc:486:SSL alert number 45

The extension should detect this condition and either:
1. Show a user-friendly error message explaining that credentials need to be refreshed
2. Automatically run a configurable refresh command (like the JetBrains Toolbox plugin does)

## Current Behavior

The `SSLV3_ALERT_CERTIFICATE_EXPIRED` SSL error propagates as a raw error message. The existing `CertificateError` class in `src/error.ts` only handles server certificate errors (self-signed, untrusted chain, etc.), not client certificate expiration.

## Proposed Solution

Implement a certificate refresh mechanism similar to the [JetBrains Toolbox plugin's `CertificateRefreshInterceptor`](https://github.com/coder/coder-jetbrains-toolbox/blob/main/src/main/kotlin/com/coder/toolbox/sdk/interceptors/CertificateRefreshInterceptor.kt):

1. **New setting**: `coder.tlsCertRefreshCommand` - A command to run when certificates expire (e.g., `metatron refresh`)

2. **Detection**: Catch SSL errors containing `certificate_expired` or `SSLV3_ALERT_CERTIFICATE_EXPIRED`

3. **Refresh flow**:
   - Execute the configured refresh command
   - Reload the TLS context to pick up new certificates
   - Retry the failed request

4. **Fallback**: If no refresh command is configured, show a user-friendly error message suggesting the user refresh their credentials manually

## Implementation Notes

- The error surfaces at extension activation in `src/extension.ts` (~line 296) in the `.catch()` handler for `deploymentManager.setDeploymentIfValid()`
- The `CertificateError.maybeWrap()` method in `src/error.ts` could be extended to detect this error type
- TLS certificates are configured via `coder.tlsCertFile` and `coder.tlsKeyFile` settings
- The axios HTTP client would need an interceptor or error handler to detect and handle these errors

## Related

- JetBrains Toolbox implementation: https://github.com/coder/coder-jetbrains-toolbox/blob/main/src/main/kotlin/com/coder/toolbox/sdk/interceptors/CertificateRefreshInterceptor.kt

EhabY:
## Summary                                                                                                               
- Add automatic refresh and retry for failed TLS client certificate connections                 
- Detect and classify SSL/TLS client certificate errors (expired, revoked, bad certificate, etc.)
- New `coder.tlsCertRefreshCommand` setting to configure a refresh command (e.g., `metatron refresh`)
- Consistently handle websocket errors across HTTP and WebSocket connections

### Certificate Error Handling                                                                                           
- Split `CertificateError` into `ServerCertificateError` (server cert issues) and `ClientCertificateError` (client cert  
  issues)
- Detect 7 SSL/TLS client certificate alert codes: expired, revoked, bad certificate, unknown, unsupported, unknown CA,  
  and access denied. Also classify errors as "refreshable" (expired, revoked, bad, unknown) vs "non-refreshable" (unsupported, unknown CA, access denied)                                                          

### Automatic Refresh & Retry
- Automatically execute configured refresh command when refreshable certificate errors occur
- Retry failed HTTP requests and WebSocket connections after refresh                                  
- Show user-friendly notifications with appropriate guidance based on error type

### Code Quality
- Add `execCommand` utility for shared command execution with proper logging
- Improve websocket error handling consistency between HTTP and WebSocket paths

Fixed #714

### DIFF
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -3,6 +3,7 @@ import {
 	type AxiosInstance,
 	type AxiosHeaders,
 	type AxiosResponseTransformer,
+	isAxiosError,
 } from "axios";
 import { Api } from "coder/site/src/api/api";
 import {
@@ -498,39 +499,13 @@ function setupInterceptors(client: CoderApi, output: Logger): void {
 	client.getAxiosInstance().interceptors.response.use(
 		(r) => r,
 		async (err: unknown) => {
-			const refreshCommand = getRefreshCommand();
-			const certError = ClientCertificateError.fromError(err);
-			if (certError && certError.isRefreshable && refreshCommand) {
-				const config = (
-					err as {
-						config?: RequestConfigWithMeta & {
-							_certRetried?: boolean;
-						};
-					}
-				).config;
-
-				if (config && !config._certRetried) {
-					config._certRetried = true;
-
-					output.info(
-						`Client certificate error (alert ${certError.alertCode}), attempting refresh...`,
-					);
-					const success = await refreshCertificates(refreshCommand, output);
-					if (success) {
-						// Create new agent with refreshed certificates.
-						const agent = await createHttpAgent(
-							vscode.workspace.getConfiguration(),
-						);
-						config.httpsAgent = agent;
-						config.httpAgent = agent;
-
-						// Retry the request.
-						output.info("Retrying request with refreshed certificates...");
-						return client.getAxiosInstance().request(config);
-					}
-				}
-
-				throw certError;
+			const retryResponse = await tryRefreshClientCertificate(
+				err,
+				client.getAxiosInstance(),
+				output,
+			);
+			if (retryResponse) {
+				return retryResponse;
 			}
 
 			// Handle other certificate errors.
@@ -584,6 +559,61 @@ function addLoggingInterceptors(client: AxiosInstance, logger: Logger) {
 			throw error;
 		},
 	);
+}
+
+/**
+ * Attempts to refresh client certificates and retry the request if the error
+ * is a refreshable client certificate error.
+ *
+ * @returns The retry response if refresh succeeds, or undefined if the error
+ *          is not a client certificate error (caller should handle).
+ * @throws {ClientCertificateError} If this is a client certificate error.
+ */
+async function tryRefreshClientCertificate(
+	err: unknown,
+	axiosInstance: AxiosInstance,
+	output: Logger,
+): Promise<unknown> {
+	const certError = ClientCertificateError.fromError(err);
+	if (!certError) {
+		return undefined;
+	}
+
+	const refreshCommand = getRefreshCommand();
+	if (
+		!certError.isRefreshable ||
+		!refreshCommand ||
+		!isAxiosError(err) ||
+		!err.config
+	) {
+		throw certError;
+	}
+
+	// _certRetried is per-request (Axios creates fresh config per request).
+	const config = err.config as RequestConfigWithMeta & {
+		_certRetried?: boolean;
+	};
+	if (config._certRetried) {
+		throw certError;
+	}
+	config._certRetried = true;
+
+	output.info(
+		`Client certificate error (alert ${certError.alertCode}), attempting refresh...`,
+	);
+	const success = await refreshCertificates(refreshCommand, output);
+	if (!success) {
+		throw certError;
+	}
+
+	// Create new agent with refreshed certificates.
+	const agent = await createHttpAgent(vscode.workspace.getConfiguration());
+	config.httpsAgent = agent;
+	config.httpAgent = agent;
+
+	// Retry the request.
+	output.info("Retrying request with refreshed certificates...");
+	return axiosInstance.request(config);
 }
 
 function wrapRequestTransform(
--- a/src/error/clientCertificateError.ts
+++ b/src/error/clientCertificateError.ts
@@ -55,6 +55,18 @@ const ALERT_PATTERNS: ReadonlyArray<[string, CLIENT_CERT_ALERT]> =
 
 /**
  * Alert codes that may be recoverable by refreshing the certificate.
+ *
+ * Refreshable alerts are those where obtaining/reloading a new certificate
+ * could resolve the issue:
+ * - CERTIFICATE_EXPIRED: Certificate validity period has passed
+ * - CERTIFICATE_REVOKED: Certificate was revoked but a new one may be valid
+ * - BAD_CERTIFICATE: Certificate may be corrupted; reloading may help
+ * - CERTIFICATE_UNKNOWN: Ambiguous rejection; refresh worth attempting
+ *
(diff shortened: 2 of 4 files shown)


src/websocket/reconnectingWebSocket.ts
```diff
@@ -275,12 +287,59 @@ export class ReconnectingWebSocket<
...
 		this.#backoffMs = Math.min(this.#backoffMs * 2, this.#options.maxBackoffMs);
 	}
 
+	/**
+	 * Attempt to refresh certificates and return true if refresh succeeded.
+	 * Returns false if no callback configured or refresh failed.
+	 */
+	private async attemptCertificateRefresh(): Promise<boolean> {
+		if (!this.#options.onCertificateExpired) {
+			return false;
```
johnstcn: suggestion: How about we log that the cert is expired and there's nothing we can do about it because we haven't been told how to refresh the client cert?

CHANGELOG.md
```diff
@@ -2,9 +2,15 @@
...
 
+### Added
+
+- Automatic TLS client certificate refresh via new `coder.tlsCertRefreshCommand` setting. Detects
+  certificate errors (expired, revoked, etc.) and automatically refreshes and retries.
+
 ### Fixed
 
 - Fixed `SetEnv` SSH config parsing and accumulation with user-defined values.
+- Improved WebSocket error handling for more consistent behavior across connection failures.
```
johnstcn, line 13: nit: other PR?

test/unit/websocket/reconnectingWebSocket.test.ts
```diff
@@ -90,13 +90,20 @@ describe("ReconnectingWebSocket", () => {
 					);
 				});
 
-				await expect(
-					ReconnectingWebSocket.create(factory, createMockLogger()),
-				).rejects.toThrow(`Unexpected server response: ${statusCode}`);
+				// create() returns a disconnected instance instead of throwing
+				const ws = await ReconnectingWebSocket.create(
+					factory,
```
johnstcn, line 95: nit: related to other PR?

EhabY: Kinda but no, this was needed because I didn't properly propagate the factory errors in the reconnecting websocket, so it's actually needed for the cert handling here.

The other PR is just refactoring with no behavioral change

CHANGELOG.md
EhabY: As mentioned in https://github.com/coder/vscode-coder/pull/732#discussion_r2708131780, this was always in issue but I only noticed it now when catching cert errors

src/websocket/reconnectingWebSocket.ts
EhabY: We actually always pass a handler in production so I'll just make it required (we can pass `() => false` if need be)

src/api/coderApi.ts
```diff
@@ -479,16 +494,53 @@ function setupInterceptors(client: CoderApi, output: Logger): void {
...
+					const config = (
+						err as {
+							config?: RequestConfigWithMeta & {
+								_certRetried?: boolean;
+							};
+						}
+					).config;
+
+					if (config && !config._certRetried) {
+						config._certRetried = true;
```
johnstcn: I'm not sure I'm following the logic here -- we send `_certRetried` in the _outgoing_ request?

EhabY: We intercept failures (in responses/incoming), then we handle the certificate errors. We set this flag because Axios maintains custom flags in the config so if we fail again on the same exact request then we don't attempt to fix it a second time and get stuck in a loop

src/api/coderApi.ts
```diff
@@ -479,16 +494,53 @@ function setupInterceptors(client: CoderApi, output: Logger): void {
...
+							const agent = await createHttpAgent(
+								vscode.workspace.getConfiguration(),
+							);
+							config.httpsAgent = agent;
+							config.httpAgent = agent;
+
+							// Retry the request.
+							output.info("Retrying request with refreshed certificates...");
+							return client.getAxiosInstance().request(config);
+						}
```
jakehwll: nit: maybe `return` early? 🙂 

```suggestion
						const success = await refreshCertificates(refreshCommand, output);
						
						if (!success) return
						
						// Create new agent with refreshed certificates.
						const agent = await createHttpAgent(
							vscode.workspace.getConfiguration(),
						);
						config.httpsAgent = agent;
						config.httpAgent = agent;

						// Retry the request.
						output.info("Retrying request with refreshed certificates...");
						return client.getAxiosInstance().request(config);
```

src/error/clientCertificateError.ts
```diff
@@ -0,0 +1,175 @@
...
+};
+
+/**
+ * Patterns to match SSL alert types in error messages (case-insensitive).
+ */
+const ALERT_PATTERNS: ReadonlyArray<[string, CLIENT_CERT_ALERT]> =
+	Object.entries(CLIENT_CERT_ALERT)
+		.filter(
+			(entry): entry is [string, CLIENT_CERT_ALERT] =>
+				typeof entry[1] === "number",
```
jakehwll: nit: maybe scope these to be more readable

```suggestion
			([name, code]): entry is [string, CLIENT_CERT_ALERT] =>
				typeof code === "number",
```

src/api/coderApi.ts
EhabY: In that case the fallback is `throw certError;` which is why I'd rather not return early since it might be confusing what to throw back (err or the wrapped error)

src/error/clientCertificateError.ts
EhabY: We cannot do this because we need to define `entry` so we can narrow the type but I've changed this to the following:

```ts
/**
 * Type guard to filter out reverse mappings from TypeScript numeric enums.
 */
function isNumericEnumEntry(
	entry: [string, string | CLIENT_CERT_ALERT],
): entry is [string, CLIENT_CERT_ALERT] {
	return typeof entry[1] === "number";
}

/**
 * Patterns to match SSL alert types in error messages (case-insensitive).
 */
const ALERT_PATTERNS: ReadonlyArray<[string, CLIENT_CERT_ALERT]> =
	Object.entries(CLIENT_CERT_ALERT)
		.filter(isNumericEnumEntry)
		.map(([name, code]) => [name.toLowerCase(), code]);
```

src/api/coderApi.ts
EhabY: Actually this still simplifies so I added this after each step!

PR #736, opened 2026-01-20
Add proxy log file cleanup

### LINKED ISSUE #377: coder.coder-remote/log can grow quite large

EhabY:
## Summary                                                                                                               
- Adds cleanup for proxy log files (`coder-ssh-*.log`) in the proxy log directory                           
- Only triggers when file count exceeds 20 AND oldest files are more than 7 days old
- Deletes oldest stale files (by mtime) to bring count back to 20
- Runs on monitor start alongside existing network info file cleanup

Fixed #377

### DIFF
--- a/src/remote/sshProcess.ts
+++ b/src/remote/sshProcess.ts
@@ -78,42 +78,64 @@ export class SshProcessMonitor implements vscode.Disposable {
 	private lastStaleSearchTime = 0;
 
 	/**
-	 * Cleans up network info files older than the specified age.
+	 * Helper to clean up files in a directory.
+	 * Stats files in parallel, applies selection criteria, then deletes in parallel.
 	 */
-	private static async cleanupOldNetworkFiles(
-		networkInfoPath: string,
-		maxAgeMs: number,
+	private static async cleanupFiles(
+		dir: string,
+		fileType: string,
 		logger: Logger,
+		options: {
+			filter: (name: string) => boolean;
+			select: (
+				files: Array<{ name: string; mtime: number }>,
+				now: number,
+			) => Array<{ name: string }>;
+		},
 	): Promise<void> {
 		try {
-			const files = await fs.readdir(networkInfoPath);
 			const now = Date.now();
+			const files = await fs.readdir(dir);
 
-			const deletedFiles: string[] = [];
-			for (const file of files) {
-				if (!file.endsWith(".json")) {
-					continue;
-				}
-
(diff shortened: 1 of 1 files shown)


src/remote/sshProcess.ts
```diff
@@ -117,6 +121,61 @@ export class SshProcessMonitor implements vscode.Disposable {
...
+
+			const withStats = await Promise.all(
+				files
+					.filter((f) => f.startsWith("coder-ssh") && f.endsWith(".log"))
+					.map(async (name) => {
+						try {
+							const stats = await fs.stat(path.join(logDir, name));
+							return { name, mtime: stats.mtime.getTime() };
+						} catch {
+							return null;
```
mtojek: if there is any weird permission denied error, wouldn't this be an issue?

src/remote/sshProcess.ts
```diff
@@ -117,6 +121,61 @@ export class SshProcessMonitor implements vscode.Disposable {
...
+		logDir: string,
+		maxFilesToKeep: number,
+		maxAgeMs: number,
+		logger: Logger,
+	): Promise<void> {
+		try {
+			const now = Date.now();
+			const files = await fs.readdir(logDir);
+
+			const withStats = await Promise.all(
```
mtojek: I believe that in a typical setup, we will have max 20 log files? no way we will need a handle 100+ = 100 x `fs.stat` in parallel

src/remote/sshProcess.ts
EhabY: It would be swallowed yes, I'm not sure I wanna log these :thinking:

src/remote/sshProcess.ts
EhabY: The files themselves have timestamps in the name, would it make sense to delete based on the filename only? They are named based on the creation but not necessarily when they were last modified so that's why I settled for this. Potentially we can just sort, delete until there's 20 left

src/remote/sshProcess.ts
EhabY: Added logging here

src/remote/sshProcess.ts
EhabY: Ran the deletion in parallel and unified some code here

PR #743, opened 2026-01-21
v1.12.0

EhabY:


CHANGELOG.md
```diff
@@ -2,15 +2,37 @@
...
+## [v1.12.0](https://github.com/coder/vscode-coder/releases/tag/v1.12.0) 2026-01-21
+
 ### Added
 
 - Automatic TLS client certificate refresh via new `coder.tlsCertRefreshCommand` setting. Detects
   certificate errors (expired, revoked, etc.) and automatically refreshes and retries.
+- OAuth 2.1 authentication support: When connecting to an OAuth-enabled Coder deployment, users can
+  now choose between OAuth (with automatic token refresh) or legacy session tokens. OAuth tokens
+  refresh automatically in the background, eliminating manual re-authentication.
+- Multi-deployment support: The extension now properly tracks per-deployment credentials, allowing
```
code-asher: The `allowing
  seamless switching` part makes it sound to me like I can be logged into multiple deployments (technically true) and there is a menu to switch between them, but really you would have to log out and then back in to "switch", so I wonder if users may misinterpret this.  Not sure if there is a different way to word it though.  Maybe we could say something like we remember past credentials for multiple deployments even if you log out and automatically use them if they exist and you log into a different deployment.

CHANGELOG.md
```diff
@@ -2,15 +2,37 @@
...
+- Commands now correctly appear/hide in the command palette based on login state and remote connection.
+- Opening a workspace via URI (`vscode://coder.coder-remote/open?...`) now properly prompts for login
+  when credentials are missing.
+- Network info files are now automatically cleaned up, and the SSH process is re-detected after
+  repeated failures to read network info.
+
+### Changed
+
+- **Breaking**: Minimum VS Code version is now 1.95.0.
+- Build tooling modernized: all dependencies updated, faster builds with esbuild (replaces webpack),
```
code-asher: I think we leave out the pnpm change since it is not user-facing.

code-asher (approved):
An exciting release!

CHANGELOG.md
EhabY: True it's not user-facing (well maybe better tree-shaking) - I do personally like more comprehensive notes but I'll remove it since we only focus on user-facing

CHANGELOG.md
EhabY: Changed to:

> Multi-deployment support: The extension now remembers credentials for multiple Coder deployments.
  When you log into a different deployment, existing credentials are preserved and automatically
  restored if you log back in. Credentials sync automatically across VS Code windows.

CHANGELOG.md
EhabY: Removed the whole entry here

CHANGELOG.md
code-asher: beautiful

CHANGELOG.md
code-asher: Yeah tree-shaking could be user-facing, if we find a measurable difference.

Every time I have to update dependencies I feel like I lose years of my life trying to sort through the noise to figure out how the changes actually affect me so I just wanna be super respectful of anyone's time who reads our changelog.  :smile:

PR #744, opened 2026-01-21
Add state machine to ReconnectingWebSocket

### LINKED ISSUE #754: Only trigger WebSocket reconnection if it's not connected
Users are experiencing memory leaks caused by an issue with WebSocket heartbeats (see https://github.com/coder/coder/pull/21676). To mitigate this on the VS Code extension side, we should only recreate the WebSocket when it has been suspended and is awaiting a refresh due to a configuration change, rather than recreating it unconditionally as we do now.

This should reduce the number of WebSocket connections created and, consequently, minimize hanging connections but it doesn't solve the core issue here.

EhabY:
## Summary

- Replace multiple boolean flags with a single `ConnectionState` enum for clearer state management
- Improves code readability and makes state transitions explicit

## Changes

Refactored `ReconnectingWebSocket` to use a state machine pattern with six states:

- `IDLE` - Initial state, ready to connect
- `CONNECTING` - Actively running socket factory
- `CONNECTED` - Socket is open and working
- `AWAITING_RETRY` - Waiting for backoff timer before reconnection
- `DISCONNECTED` - Temporarily paused, user must call `reconnect()` to resume
- `DISPOSED` - Permanently closed, cannot be reused

This makes the websocket lifecycle easier to understand and debug, and prevents invalid state combinations that could
occur with independent boolean flags.

Closes #754

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Changed
+
+- WebSocket connections are now more robust and reconnect less frequently, only when truly
+  necessary, reducing unnecessary disconnections and improving stability.
+
 ## [v1.12.1](https://github.com/coder/vscode-coder/releases/tag/v1.12.1) 2026-01-23
 
 ### Fixed


src/websocket/reconnectingWebSocket.ts
```diff
@@ -94,11 +110,10 @@ export class ReconnectingWebSocket<
 	}
 
 	/**
-	 * Returns true if the socket is temporarily disconnected and not attempting to reconnect.
-	 * Use reconnect() to resume.
+	 * Returns the current connection state.
 	 */
-	get isDisconnected(): boolean {
-		return this.#isDisconnected;
+	get state(): string {
```
mafredri:
```suggestion
	get state(): ConnectionState {
```
Should this be ConnectionState instead?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -187,11 +205,16 @@ export class ReconnectingWebSocket<
 	}
 
 	private async connect(): Promise<void> {
-		if (this.#isDisposed || this.#isDisconnected || this.#isConnecting) {
+		// Only allow connecting from IDLE, CONNECTED (reconnect), or AWAITING_RETRY states
+		if (
+			this.#state === ConnectionState.DISPOSED ||
+			this.#state === ConnectionState.DISCONNECTED ||
+			this.#state === ConnectionState.CONNECTING
+		) {
```
mafredri: For future proofing, I think checking the correct states would be better here rather than "all other states".

src/websocket/reconnectingWebSocket.ts
```diff
@@ -204,18 +227,20 @@ export class ReconnectingWebSocket<
 
 			const socket = await this.#socketFactory();
 
-			// Check if disconnected/disposed while waiting for factory
-			if (this.#isDisposed || this.#isDisconnected) {
+			// Check if state changed while waiting for factory (e.g., disconnect/dispose called)
+			if (this.#state !== ConnectionState.CONNECTING) {
```
mafredri, line 293: Are we expecting something may asynchronously change this state considering we just set connecting?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -233,7 +258,10 @@ export class ReconnectingWebSocket<
 			});
 
 			socket.addEventListener("close", (event) => {
-				if (this.#isDisposed || this.#isDisconnected) {
+				if (
+					this.#state === ConnectionState.DISPOSED ||
+					this.#state === ConnectionState.DISCONNECTED
+				) {
```
mafredri, line 344: This seems like a weak guard, what if we've already started reconnecting before close handler was run?

Tracking "socket" identity may be a good idea. E.g. only affect state if `this.#currentSocket === socket`. Same would apply to "open" above.

src/websocket/reconnectingWebSocket.ts
```diff
@@ -204,18 +227,20 @@ export class ReconnectingWebSocket<
...
-			if (this.#isDisposed || this.#isDisconnected) {
+			// Check if state changed while waiting for factory (e.g., disconnect/dispose called)
+			if (this.#state !== ConnectionState.CONNECTING) {
 				socket.close(WebSocketCloseCode.NORMAL, "Cancelled during connection");
 				return;
 			}
 
 			this.#currentSocket = socket;
 			this.#lastRoute = this.#route;
+			this.#state = ConnectionState.CONNECTED;
```
mafredri: Should this perhaps be set in the open handler since technically we may not connected yet?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -46,10 +64,8 @@ export class ReconnectingWebSocket<
 	#lastRoute = "unknown"; // Cached route for logging when socket is closed
 	#backoffMs: number;
 	#reconnectTimeoutId: NodeJS.Timeout | null = null;
-	#isDisconnected = false; // Temporary pause, can be resumed via reconnect()
-	#isDisposed = false; // Permanent disposal, cannot be resumed
-	#isConnecting = false;
-	#pendingReconnect = false;
+	#state: ConnectionState = ConnectionState.IDLE;
+	#pendingReconnect = false; // Queue reconnect during CONNECTING state
```
mafredri: Should this be a ConnectionState as well?

src/websocket/reconnectingWebSocket.ts
EhabY: Yes, JS is single threaded but it's `async` so there could be a case where we `await socketFactory` and then during that we call `reconnect`/`dispose`... etc

src/websocket/reconnectingWebSocket.ts
EhabY: Yes, fixed!

src/websocket/reconnectingWebSocket.ts
EhabY: I explored this but it would just complicate the state machine unnecessarily. We'd have to duplicate all `CONNECTING` checks essentially

src/websocket/reconnectingWebSocket.ts
EhabY: True, done!

src/websocket/reconnectingWebSocket.ts
EhabY: I've added `this.#currentSocket !== socket` checks in all handlers

src/websocket/reconnectingWebSocket.ts
EhabY: Done!

src/api/coderApi.ts
```diff
@@ -164,21 +165,23 @@ export class CoderApi extends Api implements vscode.Disposable {
...
-				for (const socket of this.reconnectingSockets) {
-					socket.reconnect();
-				}
+			const socketsToReconnect = [...this.reconnectingSockets].filter(
+				(socket) => socket.state === ConnectionState.DISCONNECTED,
+			);
+			this.output.debug(
+				`Configuration changed, ${socketsToReconnect.length}/${this.reconnectingSockets.size} socket(s) in DISCONNECTED state`,
+			);
+			for (const socket of socketsToReconnect) {
```
johnstcn: Would it make sense to drop a log here?

src/websocket/reconnectingWebSocket.ts
```diff
@@ -17,6 +17,91 @@ import type {
...
+	/** Actively running connect() - WS factory in progress */
+	CONNECTING = "CONNECTING",
+	/** Socket is open and working */
+	CONNECTED = "CONNECTED",
+	/** Waiting for backoff timer before attempting reconnection */
+	AWAITING_RETRY = "AWAITING_RETRY",
+	/** Temporarily paused - user must call reconnect() to resume */
+	DISCONNECTED = "DISCONNECTED",
+	/** Permanently closed - cannot be reused */
+	DISPOSED = "DISPOSED",
```
johnstcn, line 35: nit: would be nice to have a state diagram, but not required in this PR.

src/websocket/reconnectingWebSocket.ts
```diff
@@ -133,26 +233,20 @@ export class ReconnectingWebSocket<
...
-			this.#certRefreshAttempted = false; // User-initiated reconnect, allow retry
+		if (this.#state === ConnectionState.DISPOSED) {
+			return;
 		}
 
-		if (this.#isDisposed) {
-			return;
+		if (this.#state === ConnectionState.DISCONNECTED) {
+			this.#backoffMs = this.#options.initialBackoffMs;
+			this.#certRefreshAttempted = false; // User-initiated reconnect, allow retry
```
johnstcn, line 237: How do we know that this was user-initiated?

src/websocket/reconnectingWebSocket.ts
EhabY: I thought about this but it would need to be maintained alongside the code itself, comments don't get updated much

src/api/coderApi.ts
EhabY: Yes actually, would make debugging easier

src/websocket/reconnectingWebSocket.ts
EhabY: All `reconnect` calls are through user actions (Logging in to a different deployment or changing some configurations)

PR #756, opened 2026-01-26
Support VS Code's built-in proxy settings

### LINKED ISSUE #755: Respect VS Code's built-in proxy settings
VS Code provides a few settings for proxy configuration:

* `http.noProxy`
* `http.proxy`
* `http.proxyAuthorization`
* `http.proxyStrictSSL`
* Possible others

We should adhere and respect those settings while falling back on environment variables if it applies. See the following for more info https://github.com/coder/coder/discussions/21666#discussioncomment-15603918

EhabY:
Add support for VS Code's proxy settings that were not previously respected:
- http.noProxy: Falls back when coder.proxyBypass is not set
- http.proxyAuthorization: Passed directly as Proxy-Authorization header
- http.proxyStrictSSL: Combined with coder.insecure for TLS verification

Also fixes a bug in proxy.ts where the scheme condition was inverted, causing schemes to be added when they already existed.

Fixes #755

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## Unreleased
 
+### Added
+
+- Support for VS Code's built-in proxy settings: `http.noProxy` (as fallback when `coder.proxyBypass`
+  is not set), `http.proxyAuthorization`, and `http.proxyStrictSSL`.
+
+### Fixed
+
+- Fixed proxy scheme handling where URLs with schemes got duplicated and URLs without schemes
+  were not normalized.
+
 ### Changed
 
 - WebSocket connections are now more robust and reconnect less frequently, only when truly


src/api/utils.ts
```diff
@@ -40,21 +44,29 @@ export async function createHttpAgent(
...
 		key,
 		ca,
 		servername: altHost === "" ? undefined : altHost,
-		// rejectUnauthorized defaults to true, so we need to explicitly set it to
-		// false if we want to allow self-signed certificates.
-		rejectUnauthorized: !insecure,
+		// TLS verification is disabled if either:
+		// - http.proxyStrictSSL is false (VS Code's proxy SSL setting)
+		// - coder.insecure is true (backward compatible override for Coder server)
+		rejectUnauthorized: proxyStrictSSL && !insecure,
```
mafredri, line 70: Could this be a potential security issue? It seems like it mixes two different concerns (proxy trust vs coder server trust). Say you have a corporate proxy that should be trusted, but the Coder server is a dev-instance and needs insecure?

EhabY: I mentioned this comment in a thread but I'll put it here:

> Side note, but we should probably deprecate coder.proxyBypass and coder.insecure and rely only on the VS Code settings. At least tell users to use the VS Code settings and eventually we can remove them..

They both affect the same setting, even the `coder.insecure` says "This is useful for self-signed certificates." which is what the VS Code settings is for.

In your example here, we would be `rejectUnauthorized: false` since `insecure === true` which is the expected behavior (this util here is for our extensions Axios instance). But actually, maybe we should keep both settings anyway since we might want the Coder extension to access a dev-instance while the rest of VS Code is behind a corporate proxy

code-asher: Just a couple notes from my memory in case they are useful:

1. When I last tested, setting `rejectUnauthorized` had no effect.  I used an untrusted certificate but setting this to `false` would still not accept the certificate.  Even additionally setting `http.proxyStrictSSL` was not working for me, if I recall correctly.  Does it work now?
2. Yeah `coder.proxyBypass` was added before `http.noProxy` existed, makes sense to deprecate IMO.

Actually, if we deprecate `coder.proxyBypass` and `coder.insecure` there may be no reason to have any bespoke proxy support at all because I think VS Code already handles that assuming `http.proxySupport` is set correctly?  Although, I am not sure if VS Code completely override the agent, meaning we also lose the cert settings, etc.  If so then we would still need to handle the proxy ourselves, unless VS Code also has their own settings for configuring certs.

EhabY: 1. It seems to work yes, I just started a Coder server with a self-signed certificate. It fails with "Your Coder deployment's certificate does not appear to be trusted by this system. The certificate must be added to this system's trust store." unless I check `coder.insecure` or `http.proxyStrictSSL`.
2. Perhaps we should deprecate those if they are 100% equivalent, give people some time to migrate them and then eventually remove it. Potentially also automatically migrating or providing a command for the users to do that. 

I think there is no proxy support for the extensions  (see https://github.com/microsoft/vscode/issues/12588). I do not think we can depend on the agent that VS Code provides and we always have to read these settings and apply them.

code-asher: > It seems to work yes

Niiiice.  Maybe it only has no effect if `http.proxySuport` is set to `override`.  :thinking:   Either way, makes sense to me to get rid of `coder.insecure` and rely on `http.proxyStrictSSL` only.

> Perhaps we should deprecate those if they are 100% equivalent

:+1: from me

> I think there is no proxy support for the extensions (see https://github.com/microsoft/vscode/issues/12588)

That issue does also have a Microsoft employee commenting "Extensions automatically get the proxy support we added" and I do vaguely recall back when I tested this that if I set `http.proxySupport` it was overriding my agent, although it could be I just misconfigured something.  The setting itself also does explicitly say it affects extensions.

PR #757, opened 2026-01-27
Normalize "empty" string values in the config watcher

EhabY:


src/configWatcher.ts
```diff
@@ -37,3 +37,17 @@ export function watchConfigurationChanges(
 		}
 	});
 }
+
+function configValuesEqual(a: unknown, b: unknown): boolean {
+	return isDeepStrictEqual(normalizeEmptyValue(a), normalizeEmptyValue(b));
+}
+
+/**
+ * Normalize "empty" string values to a canonical form for comparison.
```
mtojek: not super accurate comment since we're handling "undefined" or "null" too :)

maybe:

>Normalize empty values (undefined, null, "") to a canonical form for comparison.

src/configWatcher.ts
```diff
@@ -37,3 +37,17 @@ export function watchConfigurationChanges(
...
+ */
+function normalizeEmptyValue(value: unknown): unknown {
+	if (value === undefined || value === null || value === "") {
```
mtojek: do we also want to cover `value === []` ?

src/configWatcher.ts
EhabY: Done!

src/configWatcher.ts
EhabY: Probably actually, I'll add it!

PR #759, opened 2026-01-27
Fix proxy settings types when reading the configurations

EhabY:


test/unit/api/utils.test.ts
```diff
@@ -225,5 +225,23 @@ describe("createHttpAgent", () => {
 				await agent.getProxyForUrl("https://fallback.example.com", mockRequest),
 			).toBe(proxy);
 		});
+
+		it("trims and joins multiple http.noProxy entries", async () => {
+			const cfg = new MockConfigurationProvider();
+			cfg.set("http.proxy", proxy);
+			cfg.set("http.noProxy", [" first.example.com ", "second.example.com "]);
```
mtojek, line 232: nit: maybe another test case with empty `[]` ?

src/api/utils.ts
```diff
@@ -28,8 +28,8 @@ export async function createHttpAgent(
 ): Promise<ProxyAgent> {
 	const insecure = cfg.get<boolean>("coder.insecure", false);
 	const proxyStrictSSL = cfg.get<boolean>("http.proxyStrictSSL", true);
-	const proxyAuthorization = cfg.get<string>("http.proxyAuthorization");
-	const httpNoProxy = cfg.get<string>("http.noProxy");
+	const proxyAuthorization = cfg.get<string | null>("http.proxyAuthorization");
```
mtojek, line 31: hmm.. I don't see any test to exercise this change. Do we need one?

EhabY: TypeScript already enforces the type when it knows it so I think it's fine

test/unit/api/utils.test.ts
EhabY: True, done!

PR #761, opened 2026-01-28
Add React webview system with pnpm workspaces

### LINKED ISSUE #771: Add React webview architecture with pnpm workspaces

EhabY:
## Summary

Adds a React-based webview system to the extension using `pnpm` workspaces. This provides the foundation for building rich UI panels (like the Tasks panel) with React while maintaining type safety between the extension and webviews.

## Architecture

```
packages/
├── shared/                      # Shared types and React utilities
│   ├── src/
│   │   ├── index.ts             # WebviewMessage type
│   │   └── react/               # VS Code webview hooks
│   │       ├── api.ts           # postMessage, getState, setState
│   │       └── hooks.ts         # useMessage, useVsCodeState
│   ├── extension.d.ts           # Types exposed to extension (excludes react/)
│   └── vite.config.base.ts      # Shared Vite config factory
└── tasks/                       # Tasks panel webview (placeholder)
    └── src/
        ├── index.tsx            # Entry point
        └── App.tsx              # Root component

src/webviews/
├── util.ts                      # getWebviewHtml() with CSP
└── tasks/
    └── TasksPanel.ts            # WebviewViewProvider
```

## Key Features

- **Type-safe message passing**: `WebviewMessage` type shared between extension and webviews
- **React hooks**: `useMessage()` for receiving messages, `useVsCodeState()` for persisted state
- **Secure CSP**: Nonce-based Content Security Policy for script execution
- **Type isolation**: Extension can only import types from `@coder/shared` (not React code) via `extension.d.ts`
- **Proper lifecycle management**: Disposable pattern for event listeners, visibility change handling

## Changes

- Add pnpm workspace configuration
- Add `@coder/shared` package with types and React hooks
- Add `@coder/tasks-webview` package (placeholder Tasks panel)
- Add `src/webviews/util.ts` with `getWebviewHtml()` helper
- Add `src/webviews/tasks/TasksPanel.ts` WebviewViewProvider
- Register Tasks panel in `package.json` (visible when authenticated + devMode)
- Add ESLint rules for React
- Add webviews documentation to `CONTRIBUTING.md`

Closes #771


packages/webview-shared/src/api.ts
```diff
@@ -0,0 +1,25 @@
...
+	getVsCodeApi().postMessage(message);
+}
+
+export function getState<T>(): T | undefined {
+	return getVsCodeApi().getState() as T | undefined;
+}
+
+export function setState<T>(state: T): void {
+	getVsCodeApi().setState(state);
+}
```
code-asher, line 23: I have not seen how this is used yet but do we need the wrappers?  Seems like callers could grab the API via `getVsCodeApi()` and then call whatever they need, without having to duplicate everything.

Not a big deal either way.

If we keep these wrappers though there is probably no need to export `getVsCodeApi()` (it seems to be unused externally right now).  Maybe the wrappers are better actually because it could be confusing why there is both a `getVsCodeApi()` and an `acquireVsCodeApi()` with such similar names.

packages/shared/src/react/api.ts
```diff
@@ -0,0 +1,25 @@
+import type { WebviewApi } from "vscode-webview";
+
+import type { WebviewMessage } from "../index";
+
+// Singleton - acquireVsCodeApi can only be called once
+let vscodeApi: WebviewApi<unknown> | undefined;
+
+declare function acquireVsCodeApi(): WebviewApi<unknown>;
```
code-asher: Could we throw a comment on here saying VS Code provides this inside webviews?

packages/webview-shared/src/api.ts
```diff
@@ -0,0 +1,25 @@
+import type { WebviewApi } from "vscode-webview";
```
code-asher, line 1: Bit of a nit, but this is in `react` yet has no React-related code.  It feels more webview related than React related to me.

Also instead of calling this `@coder/shared` what do you think about calling it `@coder/webview`?

packages/shared/src/react/api.ts
```diff
@@ -0,0 +1,25 @@
+import type { WebviewApi } from "vscode-webview";
+
+import type { WebviewMessage } from "../index";
+
+// Singleton - acquireVsCodeApi can only be called once
```
code-asher: Could we expand on this a little, say something like trying to acquire the API more than once will throw an error.

packages/shared/vite.config.base.ts
```diff
@@ -0,0 +1,40 @@
...
+			// Skip gzip size calculation for faster builds
+			reportCompressedSize: false,
+			rollupOptions: {
+				output: {
+					entryFileNames: "index.js",
+					assetFileNames: "index.[ext]",
+				},
+			},
+			// No sourcemaps in production (not needed in packaged extension)
+			sourcemap: !production,
```
code-asher: It could be nice to have sourcemaps for stack traces.

packages/webview-shared/extension.d.ts
```diff
@@ -0,0 +1,2 @@
+// Types exposed to the extension (react/ subpath is excluded).
+export type { WebviewMessage } from "./src/index";
```
code-asher, line 2: noice

packages/tasks/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "../shared/tsconfig.webview.json",
+	"compilerOptions": {
+		"paths": {
+			"@coder/shared": ["../shared/src"],
+			"@coder/shared/*": ["../shared/src/*"]
```
code-asher: Do we need the wildcard one?  I tried removing it and it still seems to resolve.

packages/shared/vite.config.base.ts
```diff
@@ -0,0 +1,40 @@
...
+		},
+		resolve: {
+			alias: {
+				"@coder/shared": resolve(dirname, "../shared/src"),
```
code-asher: Is there some common safe scheme people use for internal packages like these?  I worry about the `@coder` namespace since that is a real namespace on NPM that we use.  Something like `@internal` or `@private` or whatever, idk.

Although personally I would like to just see the direct imports in the code instead of aliases lol.

packages/tasks/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
```
code-asher: Is this for testing it standalone or something?  I did not see where this is being used.

src/webviews/util.ts
```diff
@@ -0,0 +1,43 @@
...
+		vscode.Uri.joinPath(baseUri, "index.css"),
+	);
+
+	return `<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'nonce-${nonce}'; style-src ${webview.cspSource} 'unsafe-inline'; font-src ${webview.cspSource}; img-src ${webview.cspSource} data:;">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <link rel="stylesheet" href="${styleUri.toString()}">
```
code-asher, line 34: Should we throw a title on here?

src/webviews/util.ts
```diff
@@ -0,0 +1,43 @@
...
+		webviewName,
+	);
+	const scriptUri = webview.asWebviewUri(
+		vscode.Uri.joinPath(baseUri, "index.js"),
+	);
+	const styleUri = webview.asWebviewUri(
+		vscode.Uri.joinPath(baseUri, "index.css"),
+	);
+
+	return `<!DOCTYPE html>
```
code-asher, line 27: Wondering if we want to read in `index.html` here, possibly adding some placeholders if we need to inject anything.

CONTRIBUTING.md
```diff
@@ -66,6 +66,48 @@ workspaces if the user has the required permissions.
...
+- **Type sharing**: Extension imports types from `@coder/shared` via path mapping
+  to `extension.d.ts`. Webviews import directly from `@coder/shared/react`.
+- **Message passing**: Use `postMessage()`/`useMessage()` hooks for communication.
+- **Lifecycle**: Dispose event listeners properly (see `TasksPanel.ts` for example).
+
+### Development
+
+```bash
+pnpm watch         # Rebuild extension on changes
+pnpm dev:webviews  # Rebuild webviews on changes (run in separate terminal)
```
code-asher: Ah bummer is there no way to watch everything with one command?

eslint.config.mjs
```diff
@@ -160,6 +164,37 @@ export default defineConfig(
...
+			react: reactPlugin,
+			"react-hooks": reactHooksPlugin,
+		},
+		settings: {
+			react: {
+				version: "detect",
+			},
+		},
+		rules: {
+			// Only add React-specific rules, TS rules already applied via **/*.ts config above
```
code-asher: `**/*.ts` would not match `.tsx` would it?  Or wait does it?

package.json
```diff
@@ -18,8 +18,10 @@
 	"type": "commonjs",
 	"main": "./dist/extension.js",
 	"scripts": {
-		"build": "tsc --noEmit && node esbuild.mjs",
-		"build:production": "tsc --noEmit && node esbuild.mjs --production",
+		"build": "pnpm build:webviews && tsc --noEmit && node esbuild.mjs",
+		"build:production": "NODE_ENV=production pnpm build:webviews && tsc --noEmit && node esbuild.mjs --production",
+		"build:webviews": "pnpm -r --filter \"./packages/*\" build",
+		"dev:webviews": "pnpm -r --filter \"./packages/*\" --parallel dev",
```
code-asher: For parity should this be `watch:webviews`?

src/webviews/util.ts
```diff
@@ -0,0 +1,43 @@
...
+		extensionUri,
+		"dist",
+		"webviews",
+		webviewName,
+	);
+	const scriptUri = webview.asWebviewUri(
+		vscode.Uri.joinPath(baseUri, "index.js"),
+	);
+	const styleUri = webview.asWebviewUri(
+		vscode.Uri.joinPath(baseUri, "index.css"),
```
code-asher, line 24: The tasks example has no stylesheet, should we add a blank one for now?

src/webviews/tasks/TasksPanel.ts
```diff
@@ -0,0 +1,77 @@
...
+			}),
+		);
+
+		webviewView.webview.html = getWebviewHtml(
+			webviewView.webview,
+			this.extensionUri,
+			"tasks",
+		);
+
+		// Re-initialize when visibility changes (webview may have been restored)
```
code-asher: When the webview is restored does the HTML not get loaded again?  Which means it would send `ready` and we would send back `init` and would not need this block.  Actually having this block could mean we would send `init` before the plugin said it was ready.

But, if the HTML does stay loaded when the webview is not visible, maybe we should send a `visible` event instead, so the plugin can tell the difference between a fresh load and being toggled.

code-asher (approved):
Nice, looks like a good base for webviews.

packages/shared/src/react/api.ts
code-asher: Actually do we need it?  I removed it and still get completion and docs for `acquireVsCodeApi`, but maybe something is just cached.

packages/webview-shared/src/api.ts
EhabY: True, I've hidden this function!

packages/shared/src/react/api.ts
EhabY: Oh true, I've removed it for now. If we encounter issues later on I'll add it back!

packages/webview-shared/src/api.ts
EhabY: Oh yeah true, I've moved `api.ts` into `src` and renamed this into `@coder/webview` instead of `shared`

packages/shared/src/react/api.ts
EhabY: Done!

packages/shared/vite.config.base.ts
EhabY: We already do not package the sourcemaps for the extension itself. I'm guessing because they would greatly increase the extension package size. If we need to debug a stack trace, we can build sourcemaps locally from the same version/commit and use those to map the minified locations back to the original source.

In any case, we should handle this separately since the extension also suffers from this.

packages/tasks/tsconfig.json
EhabY: Nice catch actually, we do not need this. Removed!

packages/shared/vite.config.base.ts
EhabY: The recommendation here is to use `@repo` since it's not claimed by npm, see https://turborepo.dev/docs/core-concepts/internal-packages

packages/tasks/index.html
EhabY: It was part of a previous experiment, I've removed it!

src/webviews/util.ts
EhabY: I think I'll just inline this here because it'll never grow beyond that, all the logic will actually be in the `index.js` (w React)

src/webviews/util.ts
EhabY: Done!

CONTRIBUTING.md
EhabY: On all platforms no, but we can use `concurrently` package to output both into the same terminal so I'll use that! (dev dependency)

EhabY: Added `pnpm watch:all`

eslint.config.mjs
EhabY: It does not, but we added the rule to apply to `tsx` files as well, let me update the confusing comment!

package.json
EhabY: Done!

src/webviews/util.ts
EhabY: Added a blank sheet for now, but even if it doesn't exist it'll be a harmless local 404

src/webviews/tasks/TasksPanel.ts
EhabY: The current content of this panel are a placeholder just to see something while debugging. I'll implement the real tasks view after this gets approved and merged so it's done more incrementally!

I've simplified this actually so we do not build on something broken

PR #768, opened 2026-01-30
Improve credential management: clear on logout, add switch and manage commands

### LINKED ISSUE #760: Logout doesn't fully clear credentials
**Issue:**
After logging out, the extension still logs in automatically on the next login without prompting for a token.

**Repro steps:**
1. Log in with a token.
1. Click Logout in the UI.
1. Restart VS Code.
1. Attempt to log in to the same deployment

**Expected**: prompted for an API key.
**Actual**: logs in without prompting, suggesting I wasn't fully logged out.

From looking at the code, it looks like logout clears the active session but leaves the session token in VS Code SecretStorage, so login reuses the saved token instead of prompting.

EhabY:
## Summary

- Logout now clears stored credentials for the current deployment
- Add "Switch Deployment" command to change deployments without clearing credentials
- Add "Manage Stored Credentials" command to view and remove stored credentials

## Changes

**Logout behavior:** Previously, logging out only cleared the session but left credentials stored. Now it properly clears all auth data for the deployment.

**Switch Deployment:** New command for users who work with multiple deployments. Allows switching without losing credentials for other deployments.

**Manage Stored Credentials:** Replaces the debug-only `coder.debug.listDeployments` command with a user-facing command that:
- Shows all stored deployments with $(key) icons
- Allows removing credentials for individual deployments
- Includes "Remove All" option (with confirmation) when multiple deployments exist

Fixes #760

### DIFF
--- a/package.json
+++ b/package.json
@@ -289,7 +289,7 @@
 			},
 			{
 				"command": "coder.manageCredentials",
-				"title": "Manage Stored Credentials",
+				"title": "Manage Credentials",
 				"category": "Coder"
 			}
 		],
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -11,12 +11,12 @@ import { type CoderApi } from "./api/coderApi";
 import { getGlobalFlags } from "./cliConfig";
 import { type CliManager } from "./core/cliManager";
 import { type ServiceContainer } from "./core/container";
-import { type ContextManager } from "./core/contextManager";
 import { type MementoManager } from "./core/mementoManager";
 import { type PathResolver } from "./core/pathResolver";
 import { type SecretsManager } from "./core/secretsManager";
 import { type DeploymentManager } from "./deployment/deploymentManager";
 import { CertificateError } from "./error/certificateError";
+import { toError } from "./error/errorUtils";
 import { type Logger } from "./logging/logger";
 import { type LoginCoordinator } from "./login/loginCoordinator";
 import { maybeAskAgent, maybeAskUrl } from "./promptUtils";
@@ -34,7 +34,6 @@ export class Commands {
 	private readonly mementoManager: MementoManager;
 	private readonly secretsManager: SecretsManager;
 	private readonly cliManager: CliManager;
-	private readonly contextManager: ContextManager;
 	private readonly loginCoordinator: LoginCoordinator;
 
 	// These will only be populated when actively connected to a workspace and are
@@ -58,7 +57,6 @@ export class Commands {
 		this.mementoManager = serviceContainer.getMementoManager();
 		this.secretsManager = serviceContainer.getSecretsManager();
 		this.cliManager = serviceContainer.getCliManager();
-		this.contextManager = serviceContainer.getContextManager();
 		this.loginCoordinator = serviceContainer.getLoginCoordinator();
 	}
 
@@ -258,11 +256,11 @@ export class Commands {
 			const items: Array<{
 				label: string;
 				description: string;
-				hostname: string | undefined;
+				hostnames: string[];
 			}> = hostnames.map((hostname) => ({
 				label: `$(key) ${hostname}`,
 				description: "Remove stored credentials",
-				hostname,
+				hostnames: [hostname],
 			}));
 
 			// Only show "Remove All" when there are multiple deployments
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -282,9 +288,9 @@
 				"icon": "$(search)"
 			},
 			{
-				"command": "coder.debug.listDeployments",
-				"title": "List Stored Deployments",
-				"category": "Coder Debug"
+				"command": "coder.manageCredentials",
+				"title": "Manage Stored Credentials",
```
mtojek: nit: hmm.. isn't "stored" redundant? As a user, I would assume they are stored somewhere.

src/commands.ts
```diff
@@ -221,6 +234,87 @@ export class Commands {
...
+				});
+			}
+
+			const selected = await vscode.window.showQuickPick(items, {
+				title: "Manage Stored Credentials",
+				placeHolder: "Select a deployment to remove",
+			});
+
+			if (!selected) {
+				return;
```
mtojek, line 281: nit: perhaps log debug that quick pick was cancelled?

src/commands.ts
```diff
@@ -221,6 +234,87 @@ export class Commands {
...
+						modal: true,
+						detail: `This will remove credentials for: ${hostnames.join(", ")}\n\nYou'll need to log in again to access them.`,
+					},
+					"Remove All",
+				);
+				if (confirm === "Remove All") {
+					await Promise.all(
+						hostnames.map((h) => this.secretsManager.clearAllAuthData(h)),
+					);
+					vscode.window.showInformationMessage(
```
mtojek, line 311: nit: maybe we should log debug message that some credentials were removed. I know there is an info popup, bo logs are easier to copy and send to support@ :)

src/commands.ts
```diff
@@ -221,6 +234,87 @@ export class Commands {
...
+						"Removed credentials for all deployments",
+					);
+				}
+			}
+		} catch (error: unknown) {
+			this.logger.error("Failed to manage stored credentials", error);
+			vscode.window.showErrorMessage(
+				"Failed to manage stored credentials. Storage may be corrupted.",
+			);
```
johnstcn, line 320: Does this automatically show the error, or should we include it in the message?

src/commands.ts
```diff
@@ -221,6 +234,87 @@ export class Commands {
...
+				description: "Remove stored credentials",
+				hostname,
+			}));
+
+			// Only show "Remove All" when there are multiple deployments
+			if (hostnames.length > 1) {
+				items.push({
+					label: "$(trash) Remove All",
+					description: `Remove credentials for all ${hostnames.length} deployments`,
+					hostname: undefined,
```
johnstcn: This `undefined` is pretty load-bearing. What about a placeholder e.g. `*`?

src/commands.ts
```diff
@@ -206,8 +212,15 @@ export class Commands {
 
 		this.logger.debug("Logging out");
 
+		const safeHostname =
+			this.deploymentManager.getCurrentDeployment()?.safeHostname;
```
johnstcn, line 214: Can `getCurrentDeployment()` be nullish here?

EhabY: Potentially yes since all of this can execute asynchronously, so the authentication check can be outdated

src/commands.ts
EhabY: I could make it `string | null` since `null` is more deliberate than `undefined`

I mean I do not think a hostname be "*" but it's doesn't differentiate as much 🤔

src/commands.ts
EhabY: We do not log anywhere when cancelling so I'm a bit hesitant about logging this tbh 🤔

src/commands.ts
EhabY: Ohh true, I'll add some logging

src/commands.ts
EhabY: Good point actually, I'll include it here as well

package.json
EhabY: Removed "stored"

src/commands.ts
EhabY: I made the type `hostnames: string[]` so that's def clearer now!

PR #770, opened 2026-02-02
Add webview testing infrastructure and pnpm catalog

EhabY:
### Summary

- Add pnpm catalog with strict mode for shared dependency versions
- Add webview unit tests for shared utilities
- Upgrade vite 6→7 and @vitejs/plugin-react-swc 3→4
- Add typed message API and ErrorBoundary component

### DIFF
--- a/packages/webview-shared/src/react/hooks.ts
+++ b/packages/webview-shared/src/react/hooks.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useEffectEvent, useState } from "react";
 
 import { getState, setState } from "../api";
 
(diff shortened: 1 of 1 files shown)


packages/webview-shared/src/react/hooks.ts
```diff
@@ -1,24 +1,23 @@
...
+		const listener = (event: MessageEvent<T>): void => {
+			handlerRef.current(event.data);
 		};
 		window.addEventListener("message", listener);
 		return (): void => {
 			window.removeEventListener("message", listener);
 		};
-	}, [handler]);
+	}, []);
 }
```
DanielleMaywood, line 19: Thoughts on using `useEffectEvent` instead of the ref juggling?

packages/webview-shared/src/react/hooks.ts
```diff
@@ -1,24 +1,23 @@
...
 /**
- * Hook to listen for messages from the extension
+ * Listen for messages from the extension. No need to memoize the handler.
  */
-export function useMessage<T = unknown>(
-	handler: (message: WebviewMessage<T>) => void,
-): void {
+export function useMessage<T>(handler: (message: T) => void): void {
+	const handlerRef = useRef(handler);
+	handlerRef.current = handler;
```
DanielleMaywood: My understanding of this pattern is we should have `handlerRef.current` inside of a `useLayoutEffect`

DanielleMaywood (approved):

packages/webview-shared/src/react/hooks.ts
EhabY: Oh yes that's much cleaner actually, changed to the following:

```ts
function useMessage<T>(handler: (message: T) => void): void {
	const onMessage = useEffectEvent((event: MessageEvent<T>): void => {
		handler(event.data);
	});

	useEffect((): (() => void) => {
		window.addEventListener("message", onMessage);
		return (): void => {
			window.removeEventListener("message", onMessage);
		};
	}, []);
}
```

packages/webview-shared/src/react/hooks.ts
EhabY: Removed the reference in favor of the `useEffectEvent`!

PR #774, opened 2026-02-04
Add task detail view with navigation, messaging, and log history

### LINKED ISSUE #777: Add task detail view with navigation and logs
## Summary
Build the task detail view UI with navigation from the task list, real-time logs, and adaptive polling based on task state.

## Goals
- Detail view showing task status, logs, and actions
- Navigation between task list and detail view
- Real-time logs via push messages
- Adaptive polling intervals (faster for active tasks, slower for idle/complete)
- Message input for interacting with tasks

## Resources
- [Figma designs](https://www.figma.com/design/l3PZmjWg7I9Llzd5VepNCV/VS-Code-extension-for-Tasks)
- [Notion spec](https://www.notion.so/coderhq/Enable-Tasks-in-the-VS-Code-Extension-2bdd579be59280ddba04cd5c3cdb354c)

EhabY:
## Summary

Clicking a task in the list now navigates to a detail page with header, log history, and message input.

### Components
- **TaskDetailView**: composes the header, chat history, error banner, and message input
- **TaskDetailHeader**: back button, status dot, task label with loading text (e.g. "Pausing..."), action menu
- **AgentChatHistory**: log entries grouped by role (user/agent) with auto-scroll via `useFollowScroll`
- **TaskMessageInput**: placeholder changes per task status, shows a pause button when the agent is working, Ctrl+Enter to send
- **ErrorBanner**: error message with "View logs" link, shown when task status is error

### Hooks
- **useSelectedTask**: tracks selected task, validates against the task list, fetches details with adaptive polling (faster when active, slower when idle)
- **useFollowScroll**: keeps chat scrolled to bottom as logs grow, backs off when the user scrolls up

### Other changes
- `sendTaskMessage` promoted from command to request (now returns a response)
- `canSendMessage` added to `TaskPermissions`
- `taskDetails()` test factory and Proxy-based `useTasksApi` mock for cleaner tests
- Tests for all new components and utilities

Closes #777

### DIFF
--- a/packages/shared/src/tasks/api.ts
+++ b/packages/shared/src/tasks/api.ts
@@ -15,7 +15,7 @@ import {
 	defineRequest,
 } from "../ipc/protocol";
 
-import type { Task, TaskDetails, TaskLogEntry, TaskTemplate } from "./types";
+import type { Task, TaskDetails, TaskTemplate } from "./types";
 
 export interface InitResponse {
 	tasks: readonly Task[];
@@ -59,7 +59,6 @@ const viewLogs = defineCommand<TaskIdParams>("viewLogs");
 
 const taskUpdated = defineNotification<Task>("taskUpdated");
 const tasksUpdated = defineNotification<Task[]>("tasksUpdated");
-const logsAppend = defineNotification<TaskLogEntry[]>("logsAppend");
 const refresh = defineNotification<void>("refresh");
 const showCreateForm = defineNotification<void>("showCreateForm");
 
@@ -82,7 +81,6 @@ export const TasksApi = {
 	// Notifications
 	taskUpdated,
 	tasksUpdated,
-	logsAppend,
 	refresh,
 	showCreateForm,
 } as const;


EhabY:
Created this as a followup https://github.com/coder/vscode-coder/issues/793

packages/shared/src/tasks/api.ts
```diff
@@ -47,13 +47,13 @@ const deleteTask = defineRequest<TaskActionParams, void>("deleteTask");
 const pauseTask = defineRequest<TaskActionParams, void>("pauseTask");
 const resumeTask = defineRequest<TaskActionParams, void>("resumeTask");
 const downloadLogs = defineRequest<{ taskId: string }, void>("downloadLogs");
+const sendTaskMessage = defineRequest<
+	{ taskId: string; message: string },
```
mtojek: q about consistency: pauseTask and resumeTask take `TaskActionParams`, should we use a similar pattern here as well?

packages/shared/src/tasks/utils.ts
```diff
@@ -26,10 +26,14 @@ const RESUMABLE_STATUSES: readonly TaskStatus[] = [
 export function getTaskPermissions(task: Task): TaskPermissions {
 	const hasWorkspace = task.workspace_id !== null;
 	const status = task.status;
+	const canSendMessage =
+		task.status === "paused" ||
+		(task.status === "active" && task.current_state?.state !== "working");
```
mtojek, line 36: possibly missing test: active status and current state is null

packages/tasks/src/components/AgentChatHistory.tsx
```diff
@@ -0,0 +1,80 @@
...
+							key={log.id}
+							log={log}
+							isGroupStart={index === 0 || log.type !== logs[index - 1].type}
+						/>
+					))
+				)}
+				{isThinking && (
+					<div className="log-entry log-entry-thinking">Thinking...</div>
+				)}
+				<div ref={bottomRef} />
```
mtojek, line 65: missing "key"?

packages/tasks/src/components/AgentChatHistory.tsx
```diff
@@ -0,0 +1,80 @@
...
+			</VscodeScrollable>
+		</div>
+	);
+}
+
+function getEmptyMessage(logsStatus: LogsStatus): string {
+	switch (logsStatus) {
```
mtojek, line 72: hmm.. shouldn't we handle "ok" for clarity?

packages/tasks/src/components/ErrorBanner.tsx
```diff
@@ -0,0 +1,28 @@
...
+	const message = task.current_state?.message || "Task failed";
+
+	return (
+		<div className="error-banner">
+			<VscodeIcon name="warning" />
+			<span>{message}</span>
+			<button
+				type="button"
+				className="text-link"
+				onClick={() => api.viewLogs(task.id)}
```
mtojek, line 22: is it safe if api.viewLogs throws an exception?

packages/tasks/src/components/PromptInput.tsx
```diff
@@ -47,10 +51,10 @@ export function PromptInput({
...
 					<VscodeIcon
 						actionIcon
-						name="send"
-						label="Send"
-						onClick={() => canSubmit && onSubmit()}
-						className={canSubmit ? "" : "disabled"}
+						name={actionIcon}
+						label={actionLabel}
+						onClick={() => !actionDisabled && onSubmit()}
+						className={actionDisabled ? "disabled" : ""}
```
mtojek: nit: it is a good habit to keep positive wording, so "actionEnabled". Phrases like `!actionDisabled` are cognitively harder to digest :)

packages/tasks/src/components/TaskDetailView.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface TaskDetailViewProps {
+	details: TaskDetails;
+	onBack: () => void;
+}
+
+export function TaskDetailView({ details, onBack }: TaskDetailViewProps) {
+	const { task, logs, logsStatus } = details;
+
+	const isThinking =
```
mtojek: nit: should we extract it to `utils`?

packages/tasks/src/components/TaskMessageInput.tsx
```diff
@@ -0,0 +1,71 @@
...
+		mutationFn: (msg: string) => api.sendTaskMessage(task.id, msg),
+		onSuccess: () => setMessage(""),
+	});
+
+	return (
+		<PromptInput
+			placeholder={placeholder}
+			value={message}
+			onChange={setMessage}
+			onSubmit={showPauseButton ? pauseTask : () => sendMessage(message)}
```
mtojek, line 71: q: is it possible to quickly click the button twice or will UI block this behavior?

packages/tasks/src/components/TaskMessageInput.tsx
```diff
@@ -0,0 +1,71 @@
...
+	const canSubmitMessage = canSendMessage && message.trim().length > 0;
+
+	const { mutate: pauseTask, isPending: isPausing } = useMutation({
+		mutationFn: () =>
+			api.pauseTask({ taskId: task.id, taskName: getTaskLabel(task) }),
+	});
+
+	const { mutate: sendMessage, isPending: isSending } = useMutation({
+		mutationFn: (msg: string) => api.sendTaskMessage(task.id, msg),
+		onSuccess: () => setMessage(""),
```
mtojek, line 62: hmm.. isn't onError missing?

packages/tasks/src/hooks/useFollowScroll.ts
```diff
@@ -0,0 +1,68 @@
...
+			}
+		});
+		ro.observe(container);
+
+		return () => {
+			container.removeEventListener("vsc-scrollable-scroll", onScroll);
+			mo.disconnect();
+			ro.disconnect();
+		};
+	}, []);
```
mtojek, line 65: no deps here?

packages/tasks/src/App.tsx
```diff
@@ -95,14 +100,17 @@ export default function App() {
...
 			>
-				<VscodeScrollable ref={historyScrollRef}>
-					<TaskList
-						tasks={tasks}
-						onSelectTask={(_taskId: string) => {
-							// Task detail view will be added in next PR
-						}}
-					/>
-				</VscodeScrollable>
+				<div ref={historyScrollRef} className="collapsible-content">
```
mtojek, line 117: nit: alternative form, perhaps easier to read

```tsx
function renderHistory() {
    if (selectedTask) return <TaskDetailView .../>;
    if (isLoadingDetails) return <div className="loading-container">...</div>;
    return <TaskList .../>;
}
```

src/webviews/tasks/tasksPanel.ts
```diff
@@ -355,10 +356,44 @@ export class TasksPanel
...
+				task.template_version_id,
+			);
+		}
+
+		try {
+			await this.client.sendTaskInput("me", taskId, message);
+		} catch (err) {
+			if (
+				isAxiosError(err) &&
+				(err.response?.status === 409 || err.response?.status === 400)
```
mtojek, line 385: shouldn't we log somewhere what was the real http code and message?

package.json
```diff
@@ -219,7 +219,7 @@
 					"id": "coder.tasksPanel",
 					"name": "Coder Tasks",
 					"icon": "media/tasks-logo.svg",
-					"when": "coder.authenticated && coder.tasksEnabled"
+					"when": "coder.tasksEnabled"
```
mtojek, line 222: curious: why?

mtojek (commented):
Leaving a bunch of comments, hopefully nothing big. 👍 👍 for adding tests!

packages/shared/src/tasks/api.ts
EhabY: We could and done!

packages/shared/src/tasks/utils.ts
EhabY: Added test case

packages/tasks/src/components/AgentChatHistory.tsx
EhabY: This is outside the `logs.map`, also the eslint rules would catch it like this:

> Missing "key" prop for element in iterator

packages/tasks/src/components/AgentChatHistory.tsx
EhabY: Done!

packages/tasks/src/components/ErrorBanner.tsx
EhabY: This here just calls:

```ts
postMessage({
		method: "viewLogs",
		params: task.id,
});
```

Which is VS Code API

All requests go through `useMutation` which has an `onError` (since those are async). `command` are synchronous and called in line, I'll add some error handling around commands and make sure everything goes through the `logger` actually!

packages/tasks/src/components/PromptInput.tsx
EhabY: Ah, good point, changed :)!

packages/tasks/src/components/TaskDetailView.tsx
EhabY: True actually, done!

packages/tasks/src/components/TaskMessageInput.tsx
EhabY: It was yes, I went through these and fixed them 🙏

packages/tasks/src/components/TaskMessageInput.tsx
EhabY: `canSendMessage`/`sendMessage` are both outputs from `useMutation`, `canSendMessage` disabled the button so it cannot be clicked.

It's blocked in the UI and in the event handler as well

packages/tasks/src/hooks/useFollowScroll.ts
EhabY: There are no changing dependencies here (the `ref` objects are stable) + We only want to run this once on mount which is what the `[]` communicates

packages/tasks/src/App.tsx
EhabY: Oh I like this, done!

src/webviews/tasks/tasksPanel.ts
EhabY: We already log in the interceptor the full error, but I can enhance this as follows:

```
throw new Error(
					`Task is not ready to receive messages (${errToStr(err)})`,
				);
```

package.json
EhabY: We already had a "Welcome screen" that shows when we are not authenticated so this was wrong from the beginning!

packages/tasks/src/components/AgentChatHistory.tsx
mtojek: TIL, thanks!

packages/tasks/src/components/AgentChatHistory.tsx
```diff
@@ -0,0 +1,80 @@
...
+		case "not_available":
+			return "Logs not available in current task state";
+		case "error":
+			return "Failed to load logs";
+		case "ok":
```
mtojek, line 77: Thanks for adding this! Isn't it a problem that we don't have the `default` case? In such cases, I used to add a logger warning about unexpected state.

mtojek (approved):
Ship it!

packages/tasks/src/components/AgentChatHistory.tsx
EhabY: It's okay as long as the type is exhaustive - So if we add another type like "unexpected", this part would not compile even so we have to update it

Issue #784, opened 2026-02-05
feat: notification to update workspace to latest template version should warn user of restart


sreya:
Users may not be aware that updating the workspace will restart their workspace, resulting in them losing their current working session. We should re-prompt them if they hit `Update` and warn them that doing so will restart their workspace, interrupting their session.

EhabY:
Would it not make more sense to show this warning in the same dialog? Would re-prompting be more explicit so users don't skip it 🤔?

sreya:
Yeah I think it's better to reprompt, people tend to skim and not read the whole thing

EhabY:
> Yeah I think it's better to reprompt, people tend to skim and not read the whole thing

I'm not so sure about the UX here, current we have 3 places that can trigger the Update modal:
* Through a notification after clicking `Update` on the button there
* Clicking `Update workspace` in the status bar
* Executing the command `Coder: Update Workspace`

So currently, they are 2 steps (trigger as mentioned above -> click "Update" in the modal), do we want to make this a 3 step in all of these scenarios? I can see enhancing the message perhaps, currently it's:

>  Update ${createWorkspaceIdentifier(this.workspace)} to the latest version?
>
> Updating will restart your workspace which stops any running processes and may result in the loss of unsaved work.

Which mentions the restart though it could be made more *loud*

matifali:
@EhabY, what do you think about changing it to?

**Use a two-step flow:**
1. Keep the existing notification to inform the user of the available update
2. When the user clicks "Update," show a **modal confirmation** ([`showWarningMessage`](https://code.visualstudio.com/api/references/vscode-api#:~:text=%3CTextEditor%3E-,showWarningMessage,-%3CT%20extends) with `{ modal: true }`) before proceeding

**Suggested modal copy:**
> "Updating to the latest template version will restart your workspace. Any unsaved work or running processes will be interrupted."
[Update and Restart]

PR #792, opened 2026-02-11
Improve workspace update confirmation to clarify restart consequence

EhabY:
## Summary
- Updated the confirmation dialog when updating a workspace to make it clear that the action will restart the workspace
- Renamed the "Update" button to "Update and Restart" so users understand the consequence before confirming

Before:
image

After:
image

Updates #784

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -596,7 +596,7 @@ export class Commands {
 			{
 				useCustom: true,
 				modal: true,
-				detail: `Updating ${createWorkspaceIdentifier(this.workspace)} to the latest template version will restart your workspace. Running processes will be stopped and unsaved work may be lost.`,
+				detail: `Update ${createWorkspaceIdentifier(this.workspace)} to the latest version?\n\nUpdating will restart your workspace which stops any running processes and may result in the loss of unsaved work.`,
 			},
 			"Update and Restart",
 		);


src/commands.ts
```diff
@@ -596,11 +596,11 @@ export class Commands {
 			{
 				useCustom: true,
 				modal: true,
-				detail: `Update ${createWorkspaceIdentifier(this.workspace)} to the latest version?\n\nUpdating will restart your workspace which stops any running processes and may result in the loss of unsaved work.`,
+				detail: `Updating ${createWorkspaceIdentifier(this.workspace)} to the latest template version will restart your workspace. Running processes will be stopped and unsaved work may be lost.`,
```
matifali: Let's keep the line break as it was, making it look cleaner.

EhabY: I suppose with the same (old) message then? So essentially we just change the action to indicate the restart (not ignorable)

matifali: Yes, I think. That sounds good to me. Is it possible to make the button red, orange, or yellow?

EhabY: Sadly no, we cannot change this. Even they use info for deleting files (with the same 'ol themed button):

image

matifali (dismissed):

src/commands.ts
matifali: > Yeah I think it's better to reprompt, people tend to skim and not read the whole thing

Can we add another prompt?
I don't think renaming the button alone solves the pain point.

EhabY: We could do a follow up prompt yes, but see this comment: https://github.com/coder/vscode-coder/issues/784#issuecomment-3873613127

Essentially it would be a 3-step process instead of 2:
1. Click "Update Workspace" on the statusbar OR Click "Update" on the notification OR execute "Update Workspace" command from the palette
2. Get the modal above (currently we stop here)
3.  Ask for confirmation again

If we want to do the 3-step process, then what would we show differently to the modal on step (2)?

matifali: yes. I think @sreya should help us here on what would be best. IMO, changing the button is sufficient, but it would be best to validate with the end user who reported it.

Let's merge this as it is with a renamed button, and we can consider a 3rd prompt in a follow-up if needed.

matifali (approved):

src/commands.ts
matifali: Lets keep the issue open and update that we updated the button name

sreya: Yeah I don't think a 3rd prompt would really achieve anything. I thought for some reason we weren't prompting the user after they clicked update. The rename is sufficient imo,

matifali: Thanks. @EhabY we can close the issue now I think.

PR #794, opened 2026-02-13
Update dependencies (incl. Coder)

EhabY:
Remove the defined `getTaskLog` from `CoderApi` and rely on the exported `Api` method.


pnpm-workspace.yaml
```diff
@@ -1,30 +1,28 @@
 packages:
   - packages/*
 
-# Shared dependency versions (use "catalog:" in package.json to reference)
```
EhabY, line 4: This is automatically formatted when we run `pnpm update coder` so I'll just let it use the formatting defined here

PR #797, opened 2026-02-16
Stream workspace startup logs in TaskDetailView

### LINKED ISSUE #793: Stream Startup Logs in TaskDetailView
When a task's workspace is building or its agent is initializing, the `TaskDetailView` shows "Logs not available in current task state." Similar to the Coder UI, we should stream the build and agent startup logs and display them in-place, reusing the same CSS as `AgentChatHistory`.

EhabY:
When a task's workspace is building or its agent is initializing, stream build/agent logs via WebSocket and display them in the detail view instead of "Logs not available."

- Add LazyStream class and generalize stream functions to accept onOutput callbacks; refactor WorkspaceStateMachine to use it
- Add streamWorkspaceLogs to TasksPanel with ANSI stripping and phase-aware stream selection (build vs agent logs)
- Extract LogViewer/LogViewerPlaceholder shared components from AgentChatHistory; add WorkspaceLogs with dynamic headers
- Add useWorkspaceLogs hook for log accumulation and socket cleanup
- Add isBuildingWorkspace, isAgentStarting, isWorkspaceStarting helpers
- Add workspace test factory with full type coverage

Closes #793

### DIFF
--- a/packages/tasks/src/hooks/useWorkspaceLogs.ts
+++ b/packages/tasks/src/hooks/useWorkspaceLogs.ts
@@ -2,16 +2,34 @@ import { useEffect, useState } from "react";
 
 import { useTasksApi } from "./useTasksApi";
 
+/**
+ * Subscribes to workspace log lines pushed from the extension.
+ * Batches updates per animation frame to avoid excessive re-renders
+ * when many lines arrive in quick succession.
+ */
 export function useWorkspaceLogs(): string[] {
 	const { onWorkspaceLogsAppend, closeWorkspaceLogs } = useTasksApi();
 	const [lines, setLines] = useState<string[]>([]);
 
 	useEffect(() => {
+		let pending: string[] = [];
+		let frame = 0;
+
 		const unsubscribe = onWorkspaceLogsAppend((newLines) => {
-			setLines((prev) => [...prev, ...newLines]);
+			pending.push(...newLines);
+			if (frame === 0) {
+				frame = requestAnimationFrame(() => {
+					const batch = pending;
+					pending = [];
+					frame = 0;
+					setLines((prev) => prev.concat(batch));
+				});
+			}
 		});
+
 		return () => {
 			unsubscribe();
+			cancelAnimationFrame(frame);
 			closeWorkspaceLogs();
 		};
 	}, [closeWorkspaceLogs, onWorkspaceLogsAppend]);
--- a/test/webview/tasks/useWorkspaceLogs.test.ts
+++ b/test/webview/tasks/useWorkspaceLogs.test.ts
@@ -1,5 +1,5 @@
 import { act, renderHook } from "@testing-library/react";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { TasksApi } from "@repo/shared";
 import { useWorkspaceLogs } from "@repo/tasks/hooks/useWorkspaceLogs";
@@ -33,6 +33,8 @@ function renderLogs() {
 						},
 					}),
 				);
+				// Flush the requestAnimationFrame batch used by useWorkspaceLogs.
+				vi.runAllTimers();
 			});
 		},
 		unmount() {
@@ -43,6 +45,9 @@ function renderLogs() {
 }
 
 describe("useWorkspaceLogs", () => {
+	beforeEach(() => vi.useFakeTimers());
+	afterEach(() => vi.useRealTimers());
+
 	it("returns empty array initially", () => {
 		const h = renderLogs();
 		expect(h.lines).toEqual([]);


packages/shared/src/tasks/api.ts
```diff
@@ -56,9 +56,11 @@ const sendTaskMessage = defineRequest<TaskIdParams & { message: string }, void>(
 
 const viewInCoder = defineCommand<TaskIdParams>("viewInCoder");
 const viewLogs = defineCommand<TaskIdParams>("viewLogs");
+const closeWorkspaceLogs = defineCommand<void>("closeWorkspaceLogs");
```
mtojek, line 59: nit: it is intriguing that `viewLogs` requires params where `closeWorkspaceLogs` does not

packages/shared/src/tasks/types.ts
```diff
@@ -29,16 +29,18 @@ export interface TaskPreset {
 	isDefault: boolean;
 }
 
-/** Status of log fetching */
-export type LogsStatus = "ok" | "not_available" | "error";
+/** Result of fetching task logs: either logs or an error/unavailable state. */
+export type TaskLogs =
```
mtojek, line 33: 👍

packages/shared/src/tasks/utils.ts
```diff
@@ -55,3 +55,23 @@ export function isStableTask(task: Task): boolean {
...
+	const ws = task.workspace_status;
+	return ws === "pending" || ws === "starting";
+}
+
+/** Whether the workspace is running but the agent hasn't reached "ready" yet. */
+export function isAgentStarting(task: Task): boolean {
+	if (task.workspace_status !== "running") {
+		return false;
+	}
+	const lc = task.workspace_agent_lifecycle;
```
mtojek, line 70: is it a problem if `workspace_agent_lifecycle` is null?

packages/tasks/src/components/TaskDetailView.tsx
```diff
@@ -1,29 +1,35 @@
...
 	return (
 		<div className="task-detail-view">
 			<TaskDetailHeader task={task} onBack={onBack} />
 			{task.status === "error" && <ErrorBanner task={task} />}
-			<AgentChatHistory
-				logs={logs}
-				logsStatus={logsStatus}
-				isThinking={isThinking}
-			/>
+			{starting ? (
```
mtojek, line 28: nit: is there a risk that empty logs will appear on the screen for a 0.1s?

packages/tasks/src/hooks/useSelectedTask.ts
```diff
@@ -48,14 +41,14 @@ export function useSelectedTask(tasks: readonly Task[]) {
...
-		return onNotification(TasksApi.taskUpdated, (updatedTask) => {
+		return api.onTaskUpdated((updatedTask) => {
 			if (updatedTask.id !== selectedTaskId) return;
 			queryClient.setQueryData<TaskDetails>(
 				[QUERY_KEY, selectedTaskId],
 				(prev) => (prev ? { ...prev, task: updatedTask } : undefined),
 			);
 		});
-	}, [onNotification, selectedTaskId, queryClient]);
+	}, [api.onTaskUpdated, selectedTaskId, queryClient]);
```
mtojek, line 51: hmm.. how often will `api.onTaskUpdated` be rebinding? every render?

packages/tasks/src/hooks/useWorkspaceLogs.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { useTasksApi } from "./useTasksApi";
+
+export function useWorkspaceLogs(): string[] {
+	const { onWorkspaceLogsAppend, closeWorkspaceLogs } = useTasksApi();
+	const [lines, setLines] = useState<string[]>([]);
+
+	useEffect(() => {
+		const unsubscribe = onWorkspaceLogsAppend((newLines) => {
+			setLines((prev) => [...prev, ...newLines]);
```
mtojek: are there any performance concerns with long builds (many lines)? maybe concat?

src/api/workspace.ts
```diff
@@ -16,6 +15,20 @@ import { type UnidirectionalStream } from "../websocket/eventStreamConnection";
 import { errToStr, createWorkspaceIdentifier } from "./api-helper";
 import { type CoderApi } from "./coderApi";
 
+/** Opens a stream once; subsequent open() calls are no-ops until closed. */
+export class LazyStream<T> {
+	private stream: UnidirectionalStream<T> | null = null;
+
+	async open(factory: () => Promise<UnidirectionalStream<T>>): Promise<void> {
```
mtojek, line 23: this is async, so I suspect that there is a risk of calling `open` twice.. do we need `opening` flag?

src/webviews/tasks/tasksPanel.ts
```diff
@@ -447,6 +471,44 @@ export class TasksPanel
...
+	private async streamWorkspaceLogs(task: Task): Promise<void> {
+		if (task.id !== this.streamingTaskId) {
+			this.buildLogStream.close();
+			this.agentLogStream.close();
+			this.streamingTaskId = task.id;
+		}
+
+		const onOutput = (line: string) => {
+			const clean = stripAnsi(line);
+			if (clean.length === 0) return;
```
mtojek: what if logs have empty lines on purpose?

src/webviews/tasks/tasksPanel.ts
```diff
@@ -584,6 +642,8 @@ export class TasksPanel
 	}
 
 	dispose(): void {
+		this.buildLogStream.close();
```
mtojek, line 646: should we add

`this.streamingTaskId = null`

packages/shared/src/tasks/api.ts
EhabY: `closeWorkspaceLogs` just closes whatever open logs without caring about the ID

packages/tasks/src/components/TaskDetailView.tsx
EhabY: If it's empty we will get `Waiting for logs...` which is expected

packages/shared/src/tasks/utils.ts
EhabY: That should be fine, the UI would show the AgentChatHistory which is still correct since we are not waiting for builds anymore

packages/tasks/src/hooks/useSelectedTask.ts
EhabY: It should never, it would rebind on every render but because we are using the React Compiler here then these values are all memoized

packages/tasks/src/hooks/useWorkspaceLogs.ts
EhabY: Concat and spread both copy, since this is a React immutable state we have to copy it. Potentially, we can cap the number of lines and drop older ones :thinking:

src/api/workspace.ts
EhabY: Oh, nice catch. I added an `opening` promise that is assigned without `await`

src/webviews/tasks/tasksPanel.ts
EhabY: Changed it to `if (line.length > 0 && clean.length === 0) ` so that we only skip pure ANSI stripped lines

src/webviews/tasks/tasksPanel.ts
EhabY: Done!

mtojek (approved):
Ship it!

PR #804, opened 2026-02-19
Split init into separate queries and simplify tasks panel

EhabY:
- Replace InitResponse with separate getTasks and getTemplates queries
  with independent poll intervals (10s tasks, 5min templates)
- Remove extension-side template cache in favor of React Query staleTime
- Add resume button to TaskMessageInput for paused tasks
- Reject message sends to paused tasks instead of auto-starting workspace
- Extract TasksPanel component and usePersistedState hook from App
- Centralize query keys in config.ts
- Rename TasksPanel to TasksPanelProvider on the extension side
- Add refresh bar indicator for initial load and user-triggered refresh

### DIFF
--- a/packages/tasks/src/components/TasksPanel.tsx
+++ b/packages/tasks/src/components/TasksPanel.tsx
@@ -32,14 +32,18 @@ interface TasksPanelProps {
 	};
 }
 
-export function TasksPanel({ tasks, templates, persisted }: TasksPanelProps) {
+export function TasksPanel({
+	tasks,
+	templates,
+	persisted: { initialCreateExpanded, initialHistoryExpanded, save },
+}: TasksPanelProps) {
 	const { selectedTask, isLoadingDetails, selectTask, deselectTask } =
 		useSelectedTask(tasks);
 
 	const [createRef, createOpen, setCreateOpen] =
-		useCollapsibleToggle<CollapsibleElement>(persisted.initialCreateExpanded);
+		useCollapsibleToggle<CollapsibleElement>(initialCreateExpanded);
 	const [historyRef, historyOpen] = useCollapsibleToggle<CollapsibleElement>(
-		persisted.initialHistoryExpanded,
+		initialHistoryExpanded,
 	);
 
 	const createScrollRef = useRef<ScrollableElement>(null);
@@ -53,13 +57,13 @@ export function TasksPanel({ tasks, templates, persisted }: TasksPanelProps) {
 	}, [onShowCreateForm, setCreateOpen]);
 
 	useEffect(() => {
-		persisted.save({
+		save({
 			tasks,
 			templates,
 			createExpanded: createOpen,
 			historyExpanded: historyOpen,
 		});
-	}, [persisted, tasks, templates, createOpen, historyOpen]);
+	}, [save, tasks, templates, createOpen, historyOpen]);
 
 	function renderHistory() {
 		if (selectedTask) {


packages/tasks/src/components/TasksPanel.tsx
```diff
@@ -0,0 +1,105 @@
...
+	}, [onShowCreateForm, setCreateOpen]);
+
+	useEffect(() => {
+		persisted.save({
+			tasks,
+			templates,
+			createExpanded: createOpen,
+			historyExpanded: historyOpen,
+		});
+	}, [persisted, tasks, templates, createOpen, historyOpen]);
```
mtojek: can you pull `persisted` out and just operate on `save`?

`const { save } = persisted;`

packages/tasks/src/components/TasksPanel.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+	const createScrollRef = useRef<ScrollableElement>(null);
+	const historyScrollRef = useRef<HTMLDivElement>(null);
+	useScrollableHeight(createRef, createScrollRef);
+	useScrollableHeight(historyRef, historyScrollRef);
+
+	const { onShowCreateForm } = useTasksApi();
+	useEffect(() => {
+		return onShowCreateForm(() => setCreateOpen(true));
+	}, [onShowCreateForm, setCreateOpen]);
```
mtojek, line 57: alternative way without deps?

```
const api = useTasksApi();
useEffect(() => {
    return api.onShowCreateForm(() => setCreateOpen(true));
}, []);
```

packages/tasks/src/hooks/useTasksQuery.ts
```diff
@@ -1,55 +1,81 @@
...
 			}),
 		];
-
 		return () => unsubs.forEach((fn) => fn());
-	}, [api.onTasksUpdated, api.onTaskUpdated, api.onRefresh, queryClient]);
+	}, [api, queryClient]);
+
+	useEffect(() => {
+		if (hasInitialData) void refreshAll();
+	}, []);
```
mtojek, line 71: hmm ... aren't deps missing?

packages/tasks/src/hooks/useTasksQuery.ts
```diff
@@ -1,55 +1,81 @@
...
 			api.onRefresh(() => {
-				void queryClient.invalidateQueries({ queryKey: QUERY_KEY });
+				setRefreshing(true);
+				void refreshAll();
 			}),
 		];
-
 		return () => unsubs.forEach((fn) => fn());
-	}, [api.onTasksUpdated, api.onTaskUpdated, api.onRefresh, queryClient]);
+	}, [api, queryClient]);
```
mtojek, line 67: Does it mean we recreate api on every render?

EhabY: If we do not use the React compiler yes, we'd need to use `useMemo` - But because of the React compiler this is memoized automatically so no need!

packages/tasks/src/hooks/useTasksQuery.ts
EhabY: This is meant to be executed once (one-shot on mount) - We do not want to subscribe like in the previous example

packages/tasks/src/components/TasksPanel.tsx
EhabY: With the react compiler those are memoized so there's no re-rendering happening here, it's just technically correct to include those if they change (they won't ever)

packages/tasks/src/components/TasksPanel.tsx
EhabY: Done

PR #805, opened 2026-02-20
Reduce duplication in Tasks API layer

### LINKED ISSUE #798: Reduce duplication across TasksApi / useTasksApi / TasksPanel handler maps
## Problem

Every time a new request, command, or notification is added to the Tasks IPC layer, it must be manually defined in three places:

1. **`packages/shared/src/tasks/api.ts`** — canonical `defineRequest`/`defineCommand`/`defineNotification` + `TasksApi` object
2. **`packages/tasks/src/hooks/useTasksApi.ts`** — webview-side wrapper that re-lists every entry as a convenience method
3. **`src/webviews/tasks/tasksPanel.ts`** — extension-side `requestHandlers` / `commandHandlers` maps that re-list every entry by method name

Nothing enforces that these three stay in sync. Forgetting to wire up a new method in one of them is a silent bug — the TypeScript compiler won't catch it.

## Desired outcome

- A single source of truth (`TasksApi`) should _drive_ the handler/hook shape so the compiler errors when a method is missing.
- Ideally, `useTasksApi` and the `TasksPanel` handler maps could be partially or fully generated/inferred from the `TasksApi` definition (e.g. mapped types, helper that iterates over `TasksApi` entries, or codegen).
- Adding a new request/command/notification should require touching as few files as possible — ideally just the definition and the implementation.

## Ideas

- Use mapped types over `typeof TasksApi` to derive the required handler signatures, making `requestHandlers` / `commandHandlers` type-incomplete if a handler is missing.
- Generate `useTasksApi` from `TasksApi` entries automatically (the hook is already a thin `request()`/`command()` pass-through).
- Explore a registry pattern or codegen script that reads the `TasksApi` object and emits boilerplate.

---

Created on behalf of @EhabY

EhabY:
## Summary

- Replace per-definition `requestHandler`/`commandHandler` wrappers with `buildRequestHandlers` and `buildCommandHandlers` mapped-type builders that enforce compile-time completeness over the full API surface
- Add `buildApiHook` and `ApiHook<Api>` so that `useTasksApi` is auto-generated from the `TasksApi` definition object at runtime, eliminating ~60 lines of manual method listings that had to stay in sync
- Add `kind` discriminant to `RequestDef`, `CommandDef`, and `NotificationDef` so builders can dispatch by kind at runtime
- Inline intermediate const declarations in `api.ts` directly into the `TasksApi` object and remove unused `ParamsOf`/`ResponseOf` utility types
- Replace raw `sendNotification` calls in `tasksPanelProvider` with a typed `notify` helper

Closes #798

### DIFF
--- a/packages/shared/src/ipc/protocol.ts
+++ b/packages/shared/src/ipc/protocol.ts
@@ -182,8 +182,8 @@ export function buildApiHook<
 export function buildApiHook(
 	api: Record<string, { kind: string; method: string }>,
 	ipc: {
-		request: (def: { method: string }, params: unknown) => Promise<unknown>;
-		command: (def: { method: string }, params: unknown) => void;
+		request: (def: { method: string }, params?: unknown) => Promise<unknown>;
+		command: (def: { method: string }, params?: unknown) => void;
 		onNotification: (
 			def: { method: string },
 			cb: (data: unknown) => void,
--- a/src/webviews/tasks/tasksPanelProvider.ts
+++ b/src/webviews/tasks/tasksPanelProvider.ts
@@ -468,7 +468,7 @@ export class TasksPanelProvider
 		try {
 			const tasks = await this.fetchTasks();
 			if (tasks !== null) {
-				this.notify(TasksApi.tasksUpdated, [...tasks]);
+				this.notify(TasksApi.tasksUpdated, tasks);
 			}
 		} catch (err) {
 			this.logger.warn("Failed to refresh tasks after action", err);


src/webviews/tasks/tasksPanelProvider.ts
```diff
@@ -514,10 +468,7 @@ export class TasksPanelProvider
 		try {
 			const tasks = await this.fetchTasks();
 			if (tasks !== null) {
-				this.sendNotification({
-					type: TasksApi.tasksUpdated.method,
-					data: tasks,
-				});
+				this.notify(TasksApi.tasksUpdated, [...tasks]);
```
DanielleMaywood: How come we're performing `[...tasks]` here?

If it is because `tasks` is `readonly Task[]` and `tasksUpdated` only accepts `Task[]`, could we investigate having `tasksUpdated` accept `readonly Task[]`?

packages/shared/src/ipc/protocol.ts
```diff
@@ -73,28 +76,135 @@ export interface IpcNotification<D = unknown> {
...
+			def: { method: string; _types?: { params: P } },
+			...args: P extends void ? [] : [params: P]
+		) => void;
+		onNotification: <D>(
+			def: { method: string; _types?: { data: D } },
+			cb: (data: D) => void,
+		) => () => void;
+	},
+): ApiHook<Api>;
+export function buildApiHook(
```
DanielleMaywood, line 182: non-blocking: thoughts on a unit test for this function? If it is sufficiently tested indirectly then no worries.

DanielleMaywood (approved):

packages/shared/src/ipc/protocol.ts
```diff
@@ -73,28 +76,135 @@ export interface IpcNotification<D = unknown> {
...
+	api: Record<string, { kind: string; method: string }>,
+	ipc: {
+		request: (def: { method: string }, params: unknown) => Promise<unknown>;
```
mtojek: shouldn't be `params?`

packages/shared/src/ipc/protocol.ts
```diff
@@ -9,20 +9,23 @@
 
 /** Request definition: params P, response R */
 export interface RequestDef<P = void, R = void> {
+	readonly kind: "request";
 	readonly method: string;
 	/** @internal Phantom types for inference - not present at runtime */
 	readonly _types?: { params: P; response: R };
```
mtojek, line 15: with never?

`readonly _types?: never & { params: P; response: R };`

src/webviews/tasks/tasksPanelProvider.ts
mtojek: I think it isn't necessary?

src/webviews/tasks/tasksPanelProvider.ts
```diff
@@ -100,71 +100,28 @@ export class TasksPanelProvider
...
-				this.streamingTaskId = null;
-				this.buildLogStream.close();
-				this.agentLogStream.close();
-			},
-		),
-	};
+	private readonly requestHandlers = buildRequestHandlers(TasksApi, {
+		getTasks: () => this.fetchTasks(),
+		getTemplates: () => this.fetchTemplates(),
+		getTask: (p) => this.client.getTask("me", p.taskId),
```
mtojek, line 106: is there a risk that client will not be initialized at the moment?

EhabY: These initializers are run after the constructor is invoked, and because they are callback, the `this` is only invoked when we actually invoke the method (and thus only when the webview itself is ready!)

packages/shared/src/ipc/protocol.ts
EhabY: `never &` breaks inference, and TS simplifies it to `never` only so we cannot really use it

packages/shared/src/ipc/protocol.ts
EhabY: Ah true, we would have had to pass `undefined` instead of omitting but this is cleaner!

packages/shared/src/ipc/protocol.ts
EhabY: It is tested through the tasks API yes, but this is all TS and would fail on runtime if we use it wrong somehow (`useTasksApi` has all the correct types)

src/webviews/tasks/tasksPanelProvider.ts
EhabY: Yep, it's not needed. That was left from a previous change!

PR #808, opened 2026-02-25
Store session tokens in the OS keyring

EhabY:
## Summary
- On macOS and Windows with CLI >= 2.29.0, store session tokens in the OS keyring (Keychain / Credential Manager) instead of plaintext files, matching the Coder CLI's native keyring support
- Pass `--url` instead of `--global-config` so the CLI reads tokens from the keyring
- Fall back to file storage (with `0o600` permissions) on Linux, older CLIs, or if the keyring write fails
- Read tokens from the keyring at login time so credentials written by `coder login` in the terminal are picked up automatically

## How it works
- `KeyringStore` wraps `@napi-rs/keyring` using the exact credential format the CLI expects (JSON map keyed by host, base64 on macOS, raw UTF-8 bytes on Windows)
- `CliAuth` discriminated union (`"global-config"` | `"url"`) is threaded through proxy command building and the workspace state machine
- `shouldUseKeyring()` gates on CLI version, platform, and `coder.useKeyring` setting
- `remote.ts` `setup()` is restructured so `configure()` runs after the CLI feature set is known
- `vendor-keyring.mjs` copies native binaries into `dist/node_modules/` at build time for VSIX packaging (vsce can't follow pnpm symlinks)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Added
+
+- Session tokens are now stored in the OS keyring (Keychain on macOS, Credential Manager on
+  Windows) instead of plaintext files, when using CLI >= 2.29.0. Falls back to file storage on
+  Linux, older CLIs, or if the keyring write fails. Controlled via the `coder.useKeyring` setting.
+
 ## [v1.13.0](https://github.com/coder/vscode-coder/releases/tag/v1.13.0) 2026-03-03
 
 ### Added


scripts/vendor-keyring.mjs
```diff
@@ -0,0 +1,61 @@
...
+
+// Native binary packages live as siblings of the resolved keyring package in
+// pnpm's content-addressable store (they aren't hoisted to node_modules).
+const siblingsDir = resolve(resolvedPkg, "..");
+const nativePackages = [
+	"keyring-darwin-arm64",
+	"keyring-darwin-x64",
+	"keyring-win32-arm64-msvc",
+	"keyring-win32-x64-msvc",
+];
```
EhabY, line 40: To make a universal VSIX, we have to package all of these (x64 and arm64 for macOS and Windows). Potentially we can split this into separate VSIXs (esp. when we have linux which has even more).

The universal VSIX went from 1.02MB to 1.80MB. Almost doubled but still acceptable IMO

src/core/cliManager.ts
```diff
@@ -708,13 +715,51 @@ export class CliManager {
...
+		featureSet: FeatureSet,
 	) {
+		if (shouldUseKeyring(featureSet) && url && token !== null) {
+			try {
+				this.keyringStore.setToken(url, token);
+				this.output.info("Stored token in OS keyring for", url);
+				return;
+			} catch (error) {
+				this.output.warn(
+					"Keyring write failed, falling back to file storage",
```
zedkipp: I think we should not silently fallback to plain text storage.

I debated doing the fallback in the CLI, but after considering we should try to be secure by default and looking into what the GitHub CLI did I decided against. Here's two examples: https://redirect.github.com/cli/cli/issues/8954 https://redirect.github.com/cli/cli/issues/10108

Having some explicit feedback to the user about the keyring failure and an explicit way for the user to opt-out of keyring use would be best.

src/cliConfig.ts
```diff
@@ -1,39 +1,76 @@
...
 
+/**
+ * Single source of truth: should the extension use the OS keyring for this session?
+ * Requires CLI >= 2.29.0, macOS or Windows, and the coder.useKeyring setting enabled.
+ */
+export function shouldUseKeyring(featureSet: FeatureSet): boolean {
+	return (
+		featureSet.keyringAuth &&
+		isKeyringSupported() &&
+		vscode.workspace.getConfiguration().get<boolean>("coder.useKeyring", true)
```
zedkipp: Similar functions in this file seem to parameterize this configuration. Should we do the same here?

src/login/loginCoordinator.ts
```diff
@@ -231,12 +235,32 @@ export class LoginCoordinator implements vscode.Disposable {
...
 				return result;
 			}
 		}
 
+		// Try keyring token (picks up tokens written by `coder login` in the terminal)
+		let keyringToken: string | undefined;
+		try {
+			keyringToken = this.keyringStore.getToken(deployment.safeHostname);
+		} catch (error) {
+			this.logger.warn("Failed to read token from keyring", error);
```
zedkipp: Will this be noisy on Linux? Should we gate keyring use with `isKeyringSupported()` here?

src/core/cliManager.ts
```diff
@@ -708,13 +715,51 @@ export class CliManager {
...
 		]);
 	}
 
+	/**
+	 * Remove credentials for a deployment from both keyring and file storage.
+	 */
+	public async clearCredentials(safeHostname: string): Promise<void> {
+		if (isKeyringSupported()) {
+			try {
+				this.keyringStore.deleteToken(safeHostname);
```
zedkipp, line 764: It seems like we aren't consistent on the host passed to `deleteToken()` and `setToken()`. Here it seems like `safeHostname` has any port stripped, but `setToken()` in `configure()` gets the raw URL (host+port). I imagine we would want to always use host+port?

EhabY: VS Code internally uses `safeHostname` which is just the hostname without any port info. That's why I tried to use that whenever possible, for deletion and fetching we can do that by searching for the hostname (after stripping the ports). For writing we have to pass the full URL because that's the format that the CLI supports.

Essentially we can go from host+port -> host only but not the other way around.

src/core/cliManager.ts
EhabY: Oooh fair point, thanks for the examples, it seems that's the consensus!

src/cliConfig.ts
EhabY: This is parameterized, I'm confused 🤔

zedkipp: Disclaimer: I'm not an expert here. `vscode.workspace` looked like a global to me.

src/core/cliManager.ts
EhabY: Done!

Showed an error and a button to open the settings + Failed instead of defaulting to plaintext!

src/core/cliManager.ts
EhabY: Added some JSDoc explaining why

src/login/loginCoordinator.ts
EhabY: Added check!

src/cliConfig.ts
EhabY: Ah right, nice catch. I thought you meant generic parameterization :sweat: 

Anyway fixed

src/core/cliManager.ts
```diff
@@ -701,48 +708,95 @@ export class CliManager {
...
+		featureSet: FeatureSet,
 	) {
+		if (!url) {
+			throw new Error("URL is required to configure the CLI");
+		}
+
+		const configs = vscode.workspace.getConfiguration();
+		if (shouldUseKeyring(configs, featureSet)) {
+			try {
+				this.keyringStore.setToken(url, token);
```
ethanndickson, line 730: Because we set the keyring value at workspace connection time, I'm a little worried about the scenario where multiple previously closed VS Code windows open at once and take turns reading and writing from the keyring. 

In saying that, it might not be an issue in practice though, since each window would be writing the same contents?

EhabY: > I'm a little worried about the scenario where multiple previously closed VS Code windows open at once and take turns reading and writing from the keyring.

We do suffer from read-modify-write issues here because we do not sync on anything (is there something we can sync on even? Potentially, the CLI could be writing this as well). But as you said, it should not be an issue in practice. Would perhaps reading the token and checking it's already written THEN only write it be better? That way it's less likely that issues will happen

zedkipp: I think that we would either need to make a keychain entry per-deployment URL, or use a file lock (or similar) to guarantee synchronization.

It seems unlikely to me that there would be multiple CLI invocations modifying the keyring with different values in the same read-modify-write window, but it is possible. I say we run with what the storage format we've got and just make sure there's a way for the user to retry storing the session token in the unlikely even we hit this and auth fails.

src/cliConfig.ts
```diff
@@ -14,21 +21,61 @@ export function getGlobalFlagsRaw(
...
+ * Resolves how the CLI should authenticate: via the keyring (`--url`) or via
+ * the global config directory (`--global-config`).
+ */
+export function resolveCliAuth(
+	configs: Pick<WorkspaceConfiguration, "get">,
+	featureSet: FeatureSet,
+	deploymentUrl: string,
+	configDir: string,
+): CliAuth {
+	if (shouldUseKeyring(configs, featureSet) && deploymentUrl) {
```
zedkipp: Do we still need `&& deploymentUrl` now that `deploymentUrl` is a typed string?

EhabY: Removed!

src/core/cliManager.ts
EhabY: > I think that we would either need to make a keychain entry per-deployment URL

This is similar to what I did in the VS Code secret storage, that seems to be the best way to guarantee synchronization because otherwise we need both the CLI and VS Code to sync on a file

> I say we run with what the storage format we've got and just make sure there's a way for the user to retry storing the session token in the unlikely even we hit this and auth fails.

Indeed, this needs more thought beyond this PR, I'll make sure we handle auth failures gracefully so users can relogin or retry!

PR #811, opened 2026-03-02
Use task pause/resume API instead of workspace stop/start

### LINKED ISSUE #795: Use task pause/resume API instead of workspace stop/start
## Summary

The Coder API now supports dedicated task pause/resume endpoints (`POST /api/v2/tasks/{user}/{task}/pause` and `POST /api/v2/tasks/{user}/{task}/resume`). The VS Code extension should use these instead of the current `stopWorkspace`/`startWorkspace` calls.

## Current Behavior

In `src/webviews/tasks/tasksPanel.ts`, the `handlePauseTask` and `handleResumeTask` methods use generic workspace stop/start:

```typescript
private async handlePauseTask(taskId: string, taskName: string): Promise<void> {
    const task = await this.client.getTask("me", taskId);
    if (!task.workspace_id) {
        throw new Error("Task has no workspace");
    }
    await this.client.stopWorkspace(task.workspace_id);
    // ...
}

private async handleResumeTask(taskId: string, taskName: string): Promise<void> {
    const task = await this.client.getTask("me", taskId);
    if (!task.workspace_id) {
        throw new Error("Task has no workspace");
    }
    await this.client.startWorkspace(task.workspace_id, task.template_version_id);
    // ...
}
```

## Desired Behavior

Use the dedicated task pause/resume API endpoints:
- **Pause**: `POST /api/v2/tasks/{user}/{task}/pause`
- **Resume**: `POST /api/v2/tasks/{user}/{task}/resume`

These endpoints set the correct `build_reason` (`task_manual_pause` / `task_resume`) which enables features like log snapshot capture on pause and proper task lifecycle tracking.

## Backward Compatibility

Older Coder servers won't have these endpoints. The implementation should:

1. First attempt `pause`/`resume` via the task API.
2. If either call returns a `404`, cache that this is an older deployment (e.g. a boolean flag on the client/panel instance).
3. Fall back to the existing `stopWorkspace`/`startWorkspace` approach.
4. On subsequent calls, skip directly to the fallback if the deployment was already identified as older.

This avoids unnecessary 404 round-trips on every pause/resume action after the first attempt.

## Implementation Notes

- The `CoderApi` class (extends `Api` from `coder/site/src/api/api`) would need `pauseTask` and `resumeTask` methods added, or the extension can make direct HTTP calls.
- The Go SDK already has `PauseTask`/`ResumeTask` in `codersdk/aitasks.go` as a reference for the request/response shapes.
- The `PauseTaskResponse` and `ResumeTaskResponse` both return a `workspace_build` field.
- Tests in `test/unit/webviews/tasks/tasksPanel.test.ts` currently mock `stopWorkspace`/`startWorkspace` for pause/resume scenarios — these need updating too.
- The `TasksPanelClient` type in the tests will need to include the new methods.

---

Created on behalf of @EhabY

EhabY:
Use the dedicated pauseTask/resumeTask endpoints which set the correct build_reason for log snapshot capture and task lifecycle tracking. Falls back to stopWorkspace/startWorkspace on 404 for older Coder servers.

Closes #795

### DIFF
--- a/src/webviews/tasks/tasksPanelProvider.ts
+++ b/src/webviews/tasks/tasksPanelProvider.ts
@@ -290,7 +290,7 @@ export class TasksPanelProvider
 		return this.pauseOrResumeTask(
 			taskId,
 			() => this.client.pauseTask("me", taskId),
-			(task) => this.client.stopWorkspace(task.workspace_id!),
+			(workspaceId) => this.client.stopWorkspace(workspaceId),
 		);
 	}
 
@@ -298,38 +298,42 @@ export class TasksPanelProvider
 		return this.pauseOrResumeTask(
 			taskId,
 			() => this.client.resumeTask("me", taskId),
-			(task) =>
-				this.client.startWorkspace(
-					task.workspace_id!,
-					task.template_version_id,
-				),
+			(workspaceId, task) =>
+				this.client.startWorkspace(workspaceId, task.template_version_id),
 		);
 	}
 
 	private async pauseOrResumeTask(
 		taskId: string,
 		taskApiCall: () => Promise<unknown>,
-		legacyCall: (task: Task) => Promise<unknown>,
+		legacyCall: (workspaceId: string, task: Task) => Promise<unknown>,
 	): Promise<void> {
-		if (!this.useLegacyPauseResume) {
-			try {
-				await taskApiCall();
-				await this.refreshAndNotifyTask(taskId);
-				return;
-			} catch (err) {
-				if (isAxiosError(err) && err.response?.status === 404) {
-					this.useLegacyPauseResume = true;
-				} else {
-					throw err;
-				}
-			}
+		if (this.useLegacyPauseResume) {
+			return this.legacyPauseOrResume(taskId, legacyCall);
 		}
 
+		try {
+			await taskApiCall();
+			await this.refreshAndNotifyTask(taskId);
+		} catch (err) {
+			if (isAxiosError(err) && err.response?.status === 404) {
+				this.useLegacyPauseResume = true;
+				return this.legacyPauseOrResume(taskId, legacyCall);
+			}
+			throw err;
+		}
+	}
+
+	private async legacyPauseOrResume(
+		taskId: string,
+		legacyCall: (workspaceId: string, task: Task) => Promise<unknown>,
+	): Promise<void> {
 		const task = await this.client.getTask("me", taskId);
-		if (!task.workspace_id) {
+		const { workspace_id } = task;
+		if (!workspace_id) {
 			throw new Error("Task has no workspace");
 		}
-		await legacyCall(task);
+		await legacyCall(workspace_id, task);
 		await this.refreshAndNotifyTask(taskId);
 	}
 


src/webviews/tasks/tasksPanelProvider.ts
```diff
@@ -285,28 +286,50 @@ export class TasksPanelProvider
...
+	}
 
-		await this.refreshAndNotifyTask(taskId);
+	private handleResumeTask(taskId: string): Promise<void> {
+		return this.pauseOrResumeTask(
+			taskId,
+			() => this.client.resumeTask("me", taskId),
+			(task) =>
+				this.client.startWorkspace(
+					task.workspace_id!,
```
DanielleMaywood: not a _huge_ fan of this assertion `task.workspace_id!`.

src/webviews/tasks/tasksPanelProvider.ts
```diff
@@ -285,28 +286,50 @@ export class TasksPanelProvider
...
 			throw new Error("Task has no workspace");
 		}
-
-		await this.client.startWorkspace(
-			task.workspace_id,
-			task.template_version_id,
-		);
-
+		await legacyCall(task);
 		await this.refreshAndNotifyTask(taskId);
```
DanielleMaywood, line 337: Thoughts on inverting the logic here? It reads (at least to me) a little nicer (also GitHub broke this suggestion because of my earlier one, thanks GitHub)

```suggestion
		if (!this.useLegacyPauseResume) {
			const task = await this.client.getTask("me", taskId);
			if (!task.workspace_id) {
				throw new Error("Task has no workspace");
			}
		
			await legacyCall(task);
			await this.refreshAndNotifyTask(taskId);
			return;
		}

		try {
			await taskApiCall();
			await this.refreshAndNotifyTask(taskId);
		} catch (err) {
			if (isAxiosError(err) && err.response?.status === 404) {
				this.useLegacyPauseResume = true;
			} else {
				throw err;
			}
		}
```

src/webviews/tasks/tasksPanelProvider.ts
```diff
@@ -285,28 +286,50 @@ export class TasksPanelProvider
...
 		if (!task.workspace_id) {
 			throw new Error("Task has no workspace");
 		}
-
-		await this.client.startWorkspace(
-			task.workspace_id,
-			task.template_version_id,
-		);
-
+		await legacyCall(task);
```
DanielleMaywood: As we're enforcing a `task.workspace_id` to exist, can we tighten up the types on `legacyCall: (task: Task) => Promise<unknown>` to always have a `workspace_id`?

src/webviews/tasks/tasksPanelProvider.ts
EhabY: If we invert this then we do not have the fallback logic - I guess that's a bit confusing, let me try to simplify this!

src/webviews/tasks/tasksPanelProvider.ts
EhabY: Oh yes, I'll make it `& { workspace_id: string }`

src/webviews/tasks/tasksPanelProvider.ts
EhabY: Removed!

src/webviews/tasks/tasksPanelProvider.ts
EhabY: Inverted and made it clearer

PR #814, opened 2026-03-02
Upgrade ESLint v9 to v10 and bump dependencies

EhabY:
## Summary
- Bump runtime and dev dependencies (axios, electron, typescript-eslint, etc.)
- Upgrade ESLint from v9 to v10 and `@eslint/js` to v10
- Replace `eslint-plugin-react` + `eslint-plugin-react-hooks` with `@eslint-react/eslint-plugin` (4-7x faster, ESLint v10 compatible, type-checked rules)

## ESLint changes
- Use `globalIgnores()` for idiomatic v10 ignore patterns
- Replace two React config blocks (hooks + JSX) with a single `recommended-type-checked` config
- Fix `preserve-caught-error` violations (new v10 recommended rule) by adding `{ cause }` to re-thrown errors
- Fix `@eslint-react` naming convention warnings: refs use `Ref` suffix, `useState` destructuring matches convention
- Give `LogEntry` objects stable IDs so `WorkspaceLogs` avoids array-index keys; derive `ActionMenu` keys in a pre-processing step
- Scope `package-json` rule overrides to root only so workspace packages are fully linted


src/core/cliUtils.ts
```diff
@@ -55,7 +55,9 @@ export async function version(binPath: string): Promise<string> {
 			if (result.stdout?.startsWith("Coder")) {
 				const v = result.stdout.split(" ")[1]?.trim();
 				if (!v) {
-					throw new Error("No version found in output: ${result.stdout}");
+					throw new Error(`No version found in output: ${result.stdout}`, {
```
mtojek, line 58: 👍

PR #815, opened 2026-03-03
Polish Tasks panel UI/UX and remove experimental feature flag

EhabY:
### Summary

- Polish user-facing messages across the Tasks panel for clarity and consistency (e.g., "logs" to "messages", clearer empty/error states, role labels like `[User]`/`[Agent]`)
- Add snapshot date metadata display in the chat history header with a relative-time tooltip powered by `date-fns`
- Fix a layout bug where collapsible template sections didn't resize when their content changed (e.g., loading spinner replaced by a form) by adding a `MutationObserver` to re-observe new children
- Remove the `coder.experimental.tasks` feature flag so the Tasks panel is always visible when authenticated

### DIFF
--- a/packages/shared/src/tasks/utils.ts
+++ b/packages/shared/src/tasks/utils.ts
@@ -83,10 +83,10 @@ export function isWorkspaceStarting(task: Task): boolean {
 /** Label for the log preview header, matching the Coder dashboard pattern. */
 export function logPreviewLabel(count: number): string {
 	if (count === 0) {
-		return "AI chat logs";
+		return "AI chat messages";
 	}
 	if (count === 1) {
-		return "Last message of AI chat logs";
+		return "Last message of AI chat";
 	}
-	return `Last ${count} messages of AI chat logs`;
+	return `Last ${count} messages of AI chat`;
 }
--- a/test/webview/shared/tasks/utils.test.ts
+++ b/test/webview/shared/tasks/utils.test.ts
@@ -241,10 +241,10 @@ describe("isAgentStarting", () => {
 
 describe("logPreviewLabel", () => {
 	it.each([
-		{ count: 0, expected: "AI chat logs" },
-		{ count: 1, expected: "Last message of AI chat logs" },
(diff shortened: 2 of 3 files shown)


packages/shared/src/tasks/utils.ts
```diff
@@ -79,3 +79,14 @@ export function isAgentStarting(task: Task): boolean {
 export function isWorkspaceStarting(task: Task): boolean {
 	return isBuildingWorkspace(task) || isAgentStarting(task);
 }
+
+/** Label for the log preview header, matching the Coder dashboard pattern. */
+export function logPreviewLabel(count: number): string {
+	if (count === 0) {
+		return "AI chat logs";
```
johnstcn:
```suggestion
		return "AI chat messages";
```

packages/shared/src/tasks/utils.ts
```diff
@@ -79,3 +79,14 @@ export function isAgentStarting(task: Task): boolean {
...
+	}
+	if (count === 1) {
+		return "Last message of AI chat logs";
```
johnstcn:
```suggestion
		return "Last message of AI chat";
```

packages/shared/src/tasks/utils.ts
```diff
@@ -79,3 +79,14 @@ export function isAgentStarting(task: Task): boolean {
...
+	}
+	return `Last ${count} messages of AI chat logs`;
```
johnstcn:
```suggestion
	return `Last ${count} messages of AI chat`;
```

jakehwll (approved):
LGTM 🙂

packages/shared/src/tasks/utils.ts
EhabY: Those are all like the Coder UI, I do agree that your suggestions are better though

johnstcn: I mainly called them out due to the PR description :)

EhabY: I'll change them here at least, IMO makes more sense. But yeah understandable :)

PR #816, opened 2026-03-03
Delegate keyring operations to CLI and simplify deployment identity

### LINKED ISSUE #825: Delegate keyring operations to CLI
Replace `@napi-rs/keyring` (4 native .node binaries) with CLI subprocess calls (`coder login --use-token-as-session` to store, `coder login token --url` to read, `coder logout --url --yes` to delete), keeping the credential format in sync with the CLI automatically and not maintaining the format of the keyring if it ever changes.

EhabY:
## Summary

- Replace `@napi-rs/keyring` (4 native `.node` binaries) with CLI subprocess calls (`coder login --use-token-as-session` to store, `coder login token --url` to read, `coder logout --url --yes` to delete), keeping the credential format in sync with the CLI automatically
- Make `url` the single source of truth for deployment identity — `fetchBinary`, `configure`, and `clearCredentials` now derive `safeHostname` internally via `toSafeHost()`, eliminating redundant parameters
- Download the CLI if it does not exist (or out of date) when reading/deleting the session token on login/logout 
- Remove `KeyringStore`, vendor script, `supportedArchitectures` config, and `@napi-rs/keyring` dependency

Closes #825

### DIFF
--- a/src/cliConfig.ts
+++ b/src/cliConfig.ts
@@ -33,22 +33,27 @@ export function getGlobalFlags(
 			: ["--global-config", escapeCommandArg(auth.configDir)];
 
 	const raw = getGlobalFlagsRaw(configs);
+	const filtered = stripManagedFlags(raw);
+
+	return [...filtered, ...authFlags, ...getHeaderArgs(configs)];
+}
+
+function stripManagedFlags(rawFlags: string[]): string[] {
 	const filtered: string[] = [];
-	for (let i = 0; i < raw.length; i++) {
-		if (isFlag(raw[i], "--use-keyring")) {
+	for (let i = 0; i < rawFlags.length; i++) {
+		if (isFlag(rawFlags[i], "--use-keyring")) {
 			continue;
 		}
-		if (isFlag(raw[i], "--global-config")) {
+		if (isFlag(rawFlags[i], "--global-config")) {
 			// Skip the next item too when the value is a separate entry.
-			if (raw[i] === "--global-config") {
+			if (rawFlags[i] === "--global-config") {
 				i++;
 			}
 			continue;
 		}
-		filtered.push(raw[i]);
+		filtered.push(rawFlags[i]);
 	}
-
-	return [...filtered, ...authFlags, ...getHeaderArgs(configs)];
+	return filtered;
 }
 
 function isFlag(item: string, name: string): boolean {
@@ -67,17 +72,6 @@ export function isKeyringEnabled(
 }
 
 /**
- * Single source of truth: should the extension use the OS keyring for this session?
- * Requires CLI >= 2.29.0, macOS or Windows, and the coder.useKeyring setting enabled.
- */
-export function shouldUseKeyring(
-	configs: Pick<WorkspaceConfiguration, "get">,
-	featureSet: FeatureSet,
-): boolean {
-	return isKeyringEnabled(configs) && featureSet.keyringAuth;
-}
-
-/**
  * Resolves how the CLI should authenticate: via the keyring (`--url`) or via
  * the global config directory (`--global-config`).
  */
@@ -87,7 +81,7 @@ export function resolveCliAuth(
 	deploymentUrl: string,
 	configDir: string,
 ): CliAuth {
-	if (shouldUseKeyring(configs, featureSet)) {
+	if (isKeyringEnabled(configs) && featureSet.keyringAuth) {
 		return { mode: "url", url: deploymentUrl };
 	}
 	return { mode: "global-config", configDir };
--- a/src/core/cliCredentialManager.ts
+++ b/src/core/cliCredentialManager.ts
@@ -1,20 +1,32 @@
 import { execFile } from "node:child_process";
+import fs from "node:fs/promises";
+import path from "node:path";
 import { promisify } from "node:util";
+import * as semver from "semver";
 
 import { isKeyringEnabled } from "../cliConfig";
+import { featureSetForVersion } from "../featureSet";
 import { getHeaderArgs } from "../headers";
+import { toSafeHost } from "../util";
+
+import * as cliUtils from "./cliUtils";
 
 import type { WorkspaceConfiguration } from "vscode";
 
 import type { Logger } from "../logging/logger";
 
+import type { PathResolver } from "./pathResolver";
+
 const execFileAsync = promisify(execFile);
+
+const EXEC_TIMEOUT_MS = 60_000;
+const EXEC_LOG_INTERVAL_MS = 5_000;
 
 /**
  * Resolves a CLI binary path for a given deployment URL, fetching/downloading
  * if needed. Returns the path or throws if unavailable.
  */
-export type BinaryResolver = (url: string) => Promise<string>;
+export type BinaryResolver = (deploymentUrl: string) => Promise<string>;
 
 /**
  * Returns true on platforms where the OS keyring is supported (macOS, Windows).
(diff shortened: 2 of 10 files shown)


src/core/cliCredentialManager.ts
```diff
@@ -0,0 +1,124 @@
...
+	 * Resolves the CLI binary automatically. Returns trimmed stdout,
+	 * or undefined if the binary can't be resolved or the CLI returns no token.
+	 */
+	public async readToken(
+		url: string,
+		configs: Pick<WorkspaceConfiguration, "get">,
+	): Promise<string | undefined> {
+		let binPath: string;
+		try {
+			binPath = await this.resolveBinary(url);
```
EhabY, line 100: This currently blocks to install the binary (on login, with keyring enabled, and CLI >= 2.29)

src/core/cliCredentialManager.ts
```diff
@@ -0,0 +1,124 @@
...
+	 * Resolves the CLI binary automatically. Best-effort: logs warnings
+	 * on failure but never throws.
+	 */
+	async deleteToken(
+		url: string,
+		configs: Pick<WorkspaceConfiguration, "get">,
+	): Promise<void> {
+		let binPath: string;
+		try {
+			binPath = await this.resolveBinary(url);
```
EhabY: This currently blocks to install the binary (on logout, with keyring enabled, and CLI >= 2.29)

src/core/container.ts
```diff
@@ -36,18 +37,30 @@ export class ServiceContainer implements vscode.Disposable {
...
+		// the resolver is only called after construction.
+		this.cliCredentialManager = new CliCredentialManager(
+			this.logger,
+			async (url) => {
+				try {
+					return await this.cliManager.locateBinary(url);
+				} catch {
+					const client = CoderApi.create(url, "", this.logger);
+					return this.cliManager.fetchBinary(client);
+				}
```
EhabY, line 50: This checks if a binary exists BEFORE we try to fetch and download. Potentially, we can just make it throwing and handle this gracefully so we do not read/delete from the keyring if the binary does not exist (writing to the keyring should always work because we call `fetchBinary` when connecting to a remote workspace - unless the user deleted it in the meantime)

EhabY:
We wait only after checking the passed token or the token stored in the secret storage. So essentially it's the last fallback incase there's a token in the OS keyring.

I suppose we could attempt to use `coder login token` globally before attempting the download, this would work for reading only, for writing or deleting (login/logout) we have to use the downloaded CLI since it might be using a different schema/key in the keychain (if we ever change the format, essentially, writing should use the same version)

src/core/cliCredentialManager.ts
```diff
@@ -0,0 +1,131 @@
...
+	constructor(
+		private readonly logger: Logger,
+		private readonly resolveBinary: BinaryResolver,
+	) {}
+
+	/**
+	 * Store a token via `coder login --use-token-as-session`.
+	 * Token is passed via CODER_SESSION_TOKEN env var, never in args.
+	 */
+	public async storeToken(
```
mtojek, line 56: should we check isKeyringEnabled here (consistency)?

src/core/cliCredentialManager.ts
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		const args = [
+			...getHeaderArgs(configs),
+			"login",
+			"--use-token-as-session",
+			url,
+		];
+		try {
+			await execFileAsync(binPath, args, {
```
mtojek: should we define timeout for `execFileAsync` operations, just in case it hangs?

src/core/cliCredentialManager.ts
```diff
@@ -0,0 +1,131 @@
...
+	): Promise<void> {
+		if (!isKeyringEnabled(configs)) {
+			return;
+		}
+
+		let binPath: string;
+		try {
+			binPath = await this.resolveBinary(url);
+		} catch (error) {
+			this.logger.debug(
```
mtojek: why not warn?

src/core/cliCredentialManager.ts
```diff
@@ -0,0 +1,131 @@
...
+
+import type { Logger } from "../logging/logger";
+
+const execFileAsync = promisify(execFile);
+
+/**
+ * Resolves a CLI binary path for a given deployment URL, fetching/downloading
+ * if needed. Returns the path or throws if unavailable.
+ */
+export type BinaryResolver = (url: string) => Promise<string>;
```
mtojek: hmm.. we should somehow indicate tha that this is a deployment URL

src/core/cliManager.ts
```diff
@@ -705,48 +735,87 @@ export class CliManager {
 	/**
 	 * Configure the CLI for the deployment with the provided hostname.
 	 *
-	 * Falsey URLs and null tokens are a no-op; we avoid unconfiguring the CLI to
-	 * avoid breaking existing connections.
+	 * Stores credentials in the OS keyring when available, otherwise falls back
```
mtojek: is it correctly implemented? if `storeToken` throws an error, it will be rethrowed up == no fallback, right?

src/core/cliManager.ts
```diff
@@ -705,48 +735,87 @@ export class CliManager {
...
+		const configs = vscode.workspace.getConfiguration();
+		await this.cliCredentialManager.deleteToken(url, configs);
+
+		const tokenPath = this.pathResolver.getSessionTokenPath(safeHostname);
+		const urlPath = this.pathResolver.getUrlPath(safeHostname);
 		await Promise.all([
-			this.updateUrlForCli(safeHostname, url),
-			this.updateTokenForCli(safeHostname, token),
+			fs.rm(tokenPath, { force: true }).catch((error) => {
+				this.output.warn("Failed to remove token file", tokenPath, error);
```
mtojek: warn or error?

same below

src/core/container.ts
```diff
@@ -34,12 +37,30 @@ export class ServiceContainer implements vscode.Disposable {
...
 			this.logger,
 		);
-		this.cliManager = new CliManager(this.logger, this.pathResolver);
+		// Circular ref: cliCredentialManager ↔ cliManager. Safe because
+		// the resolver is only called after construction.
+		this.cliCredentialManager = new CliCredentialManager(
+			this.logger,
+			async (url) => {
+				try {
+					return await this.cliManager.locateBinary(url);
```
mtojek, line 46: should we check if cliManager is initliazed?

src/login/loginCoordinator.ts
```diff
@@ -231,12 +235,30 @@ export class LoginCoordinator implements vscode.Disposable {
...
 			const result = await this.tryTokenAuth(client, auth.token, isAutoLogin);
 			if (result !== "unauthorized") {
 				return result;
 			}
 		}
 
+		// Try keyring token (picks up tokens written by `coder login` in the terminal)
+		const keyringToken = await this.cliCredentialManager.readToken(
+			deployment.url,
+			vscode.workspace.getConfiguration(),
```
mtojek, line 248: I see you're pulling conf twice in this function. Should we pull it once at the top?

src/cliConfig.ts
```diff
@@ -14,19 +21,76 @@ export function getGlobalFlagsRaw(
...
+		}
+		if (isFlag(raw[i], "--global-config")) {
+			// Skip the next item too when the value is a separate entry.
+			if (raw[i] === "--global-config") {
+				i++;
+			}
+			continue;
+		}
+		filtered.push(raw[i]);
+	}
```
mtojek, line 55: Franky speaking I would add a comment describing what is happening below (easier to skim through the code).

src/login/loginCoordinator.ts
EhabY: We could but this would mean that we if the settings change between the two calls we wouldn't be able to pick it up (it's a snapshot essentially)

mtojek: As long as we accept that discrepancy, let's leave it 👍

src/core/cliCredentialManager.ts
EhabY: Changed name to `deploymentUrl`

src/core/cliCredentialManager.ts
EhabY: Done, restructured this whole class so it's self-contained

src/core/cliCredentialManager.ts
EhabY: Add a timeout of 60s with logging every 5 seconds (also added a progress monitor)

src/core/cliCredentialManager.ts
EhabY: Changed all errors into `warn`

src/core/cliManager.ts
EhabY: Oh right, it used to fallback but for security reasons we should just fail. I'll update the comments!

src/core/cliManager.ts
EhabY: This is best-effort approach, I think we can log it on the `warn` level since this is not breaking

src/cliConfig.ts
EhabY: Added a comment

src/core/container.ts
EhabY: The constructor is executed in one shot, never during construction so this is not needed

PR #826, opened 2026-03-05
feat: set remote.SSH.reconnectionGraceTime and refactor settings manipulation

### LINKED ISSUE #823: Overwrite ReconnectionGraceTime setting on remote connection
VS Code's default connection timeouts create a poor developer experience for Coder users.

**Current Workflow Pain Points:**
- Developers expect to return to their work seamlessly without losing state
- Instead, they must manually "Reload Window" every morning
- All unsaved changes and runtime state are lost during the forced reload

## Technical Root Cause

VS Code used to have a hardcoded timeouts (3-hour ReconnectionGraceTime) that terminate server sessions when connections are lost for extended periods. This conflicts with Coder's cost-optimization model, where workspaces may be offline for 6+ hours overnight.

With the introduction the `ReconnectionGraceTime` setting, we can overwrite that value when establishing a connection to remote. If the value is already modified then we can possibly skip that. 

See https://github.com/microsoft/vscode/pull/274910

EhabY:
- Extract SSH settings manipulation from `remote.ts` into a reusable `userSettings.ts` module with `buildSshOverrides()` and `applySettingOverrides()` functions
- Automatically set `reconnectionGraceTime` (8h), `serverShutdownTimeout` (8h), and `maxReconnectionAttempts` (unlimited) when the user hasn't configured them, preventing disconnects during overnight workspace sleep
- Add a **Coder: Apply Recommended SSH Settings** command that writes all four recommended settings globally via the command palette

Closes #823

### DIFF
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -465,11 +465,16 @@ export class Remote {
 				parts.sshHost,
 				agent.operating_system,
 			);
-			await applySettingOverrides(
-				this.pathResolver.getUserSettingsPath(),
-				overrides,
-				this.logger,
-			);
+			if (overrides.length > 0) {
+				const ok = await applySettingOverrides(
+					this.pathResolver.getUserSettingsPath(),
+					overrides,
+					this.logger,
+				);
+				if (ok) {
+					this.logger.info("Settings modified successfully");
+				}
+			}
 
 			const logDir = this.getLogDir(featureSet);
 
--- a/src/remote/userSettings.ts
+++ b/src/remote/userSettings.ts
@@ -29,7 +29,11 @@ function recommended(
 	return { value, label: `${shortName}: ${humanized}` };
 }
 
-/** Applied by the "Apply Recommended SSH Settings" command. */
+/**
+ * Applied by the "Apply Recommended SSH Settings" command.
+ * These are more aggressive (24h) than AUTO_SETUP_DEFAULTS (8h) because the
+ * user is explicitly opting in via the command palette.
+ */
 export const RECOMMENDED_SSH_SETTINGS = {
 	"remote.SSH.connectTimeout": recommended("Connect Timeout", 1800),
 	"remote.SSH.reconnectionGraceTime": recommended(
@@ -109,7 +113,7 @@ export async function applySettingOverrides(
 	logger: Logger,
 ): Promise<boolean> {
 	if (overrides.length === 0) {
-		return false;
+		return true;
 	}
 
 	let settingsContent = "{}";
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Added
 
+- Automatically set `reconnectionGraceTime`, `serverShutdownTimeout`, and `maxReconnectionAttempts`
+  on first connection to prevent disconnects during overnight workspace sleep.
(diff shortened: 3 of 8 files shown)


src/remote/remote.ts
```diff
@@ -459,85 +459,17 @@ export class Remote {
...
-					mungedPlatforms = mungedConnTimeout = false;
-					this.logger.warn("Failed to configure settings", ex);
-				}
-			}
+			const overrides = buildSshOverrides(
+				vscodeProposed.workspace.getConfiguration(),
+				parts.sshHost,
+				agent.operating_system,
+			);
+			await applySettingOverrides(
```
mtojek: so this function returns bool when the write fails, shouldn't we take an action on this? maybe logging?

src/remote/userSettings.ts
```diff
@@ -0,0 +1,124 @@
...
+	let settingsContent = "{}";
+	try {
+		settingsContent = await fs.readFile(settingsFilePath, "utf8");
+	} catch {
+		// File probably doesn't exist yet.
+	}
+
+	for (const { key, value } of overrides) {
+		settingsContent = jsonc.applyEdits(
+			settingsContent,
```
mtojek, line 128: should we check if settingsContent is empty?

src/remote/userSettings.ts
```diff
@@ -0,0 +1,124 @@
...
+
+	// Set the remote platform for this host to bypass the platform prompt.
+	const remotePlatforms = config.get<Record<string, string>>(
+		"remote.SSH.remotePlatform",
+		{},
+	);
+	if (remotePlatforms[sshHost] !== agentOS) {
+		overrides.push({
+			key: "remote.SSH.remotePlatform",
+			value: { ...remotePlatforms, [sshHost]: agentOS },
```
mtojek, line 81: should we make a deep copy of remotePlatforms? just in case

src/remote/userSettings.ts
```diff
@@ -0,0 +1,124 @@
...
+	"remote.SSH.reconnectionGraceTime": {
+		value: 28800,
+		label: "Reconnection Grace Time: 28800s (8 hours)",
+	},
+	"remote.SSH.serverShutdownTimeout": {
+		value: 28800,
+		label: "Server Shutdown Timeout: 28800s (8 hours)",
+	},
+	"remote.SSH.maxReconnectionAttempts": {
+		value: null,
```
mtojek: hmm... null?

src/remote/remote.ts
EhabY: It does yes, I'll add some logging where it makes sense - If it's best-effort then nothing should happen

src/remote/userSettings.ts
EhabY: No need for this check since it's never empty (or if it is then it's still a valid JSONC)

src/remote/userSettings.ts
EhabY: All the values are `string` -> `string` so this shallow copy should be fine as string is immutable!

src/remote/userSettings.ts
EhabY: As per the setting:

> The maximum number of times to attempt reconnection. Use 0 to disallow reconnection, and `null` to use the maximum of 8.

PR #827, opened 2026-03-05
feat: fall back to extension storage for proxy log directory

### LINKED ISSUE #824: Use the VS Code extension storage as a fallback when the proxyLogDirectory is not set
Currently, if the user does not pass the `coder.proxyLogDirectory`, then they cannot view the logs easily or investigate any SSH issues when connected to a remote environment.  We should instead investigate using the extension storage as a fallback so logs are always available and accessible to VS Code Coder extension users.

EhabY:
- Fall back to extension global storage for the proxy log directory (`getProxyLogPath`) so logs are always written even when `coder.proxyLogDirectory` is not explicitly configured, removing the "No logs available" dead-end in the UI
- Add `CODER_SSH_LOG_DIR` environment variable support as an intermediate fallback (setting > env var > extension storage)
- Extract a shared `resolveOverride` helper in `PathResolver` that both `getBinaryCachePath` and `getProxyLogPath` use, ensuring consistent trim, tilde/`${userHome}` expansion, and normalization for all path overrides
- Simplify `commands.ts` log viewer to always use the resolved log directory instead of branching on whether the setting is configured

Closes #824

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -148,12 +148,12 @@ export class Commands {
 	public async viewLogs(): Promise<void> {
 		if (this.workspaceLogPath) {
 			// Return the connected deployment's log file.
-			return this.openFile(this.workspaceLogPath);
+			return openFile(this.workspaceLogPath);
 		}
 
 		const logDir = this.pathResolver.getProxyLogPath();
 		try {
-			const files = await fs.readdir(logDir);
+			const files = await readdirOrEmpty(logDir);
 			// Sort explicitly since fs.readdir order is platform-dependent
 			const logFiles = files
 				.filter((f) => f.endsWith(".log"))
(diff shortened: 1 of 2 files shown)


src/commands.ts
```diff
@@ -150,51 +150,33 @@ export class Commands {
...
-
-				if (logFiles.length === 0) {
-					vscode.window.showInformationMessage(
-						"No log files found in the configured log directory.",
-					);
-					return;
-				}
+		const logDir = this.pathResolver.getProxyLogPath();
+		try {
+			const files = await fs.readdir(logDir);
```
mtojek: nit: does this directory always exist?

EhabY: No always I suppose, I've added the following helper so we do not show an error dialog when it does not exist:

```ts

async function readdirOrEmpty(dirPath: string): Promise<string[]> {
	try {
		return await fs.readdir(dirPath);
	} catch (err) {
		if (err instanceof Error && "code" in err && err.code === "ENOENT") {
			return [];
		}
		throw err;
	}
}
```

PR #831, opened 2026-03-09
refactor: improve keyring support reliability and reduce duplication

EhabY:
- Separate keyringTokenRead (>= 2.31) from keyringAuth (>= 2.29) feature flags
- Combine duplicate resolveKeyringBinary/resolveReadTokenBinary with explicit feature param
- Extract tempFilePath() utility with crypto.randomUUID, replacing 4 inline patterns
- Add withProgress() wrapper, adopt in commands.ts and promptUtils.ts
- Type-safe circular dependency guard in ServiceContainer via ref object
- Fix stubExecFileAbortable to handle non-pre-aborted signals
- Replace regex test matchers with structural assertions

### DIFF
--- a/src/core/cliCredentialManager.ts
+++ b/src/core/cliCredentialManager.ts
@@ -5,7 +5,7 @@ import { promisify } from "node:util";
 import * as semver from "semver";
 
 import { isKeyringEnabled } from "../cliConfig";
-import { type FeatureSet, featureSetForVersion } from "../featureSet";
+import { featureSetForVersion } from "../featureSet";
 import { getHeaderArgs } from "../headers";
 import { tempFilePath, toSafeHost } from "../util";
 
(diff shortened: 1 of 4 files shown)


src/core/cliCredentialManager.ts
```diff
@@ -131,22 +138,23 @@ export class CliCredentialManager {
...
+	 * when keyring is enabled in settings and the CLI version supports the
+	 * requested feature, or undefined to fall back to file-based storage.
 	 *
 	 * Throws on binary resolution or version-check failure (caller decides
 	 * whether to catch or propagate).
 	 */
 	private async resolveKeyringBinary(
 		url: string,
 		configs: Pick<WorkspaceConfiguration, "get">,
+		feature: keyof Pick<FeatureSet, "keyringAuth" | "keyringTokenRead">,
```
mtojek: maybe: `type KeyringFeature = "keyringAuth" | "keyringTokenRead` ?

EhabY: Done

mtojek (approved):

PR #833, opened 2026-03-10
feat: apply coder config-ssh --ssh-option to VS Code connections

### LINKED ISSUE #832: bug: VSCODE take not the ssh-config settings - like X11
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

i try to develop a Qt app in the coder environment, and will use the X11 forwarding. In a ssh session runs perfect, but this settings are not taken to the vscode connection.

### Relevant Log Output

```shell
xterm: Xt error: Can't open display: 
xterm: DISPLAY is not set
```

### Expected Behavior

_No response_

### Steps to Reproduce

coder config-ssh --yes  --ssh-option ForwardX11=yes   --ssh-option ForwardX11Trusted=yes
-> ssh connection with X11 still perfect
-> vscode terminal
Error:
xterm: Xt error: Can't open display: 
xterm: DISPLAY is not set

### Environment

- Host OS: Linux
- Coder version: 2.30.1

### Additional Context

_No response_

EhabY:
- SSH options set via `coder config-ssh --ssh-option` (e.g. `ForwardX11=yes`) were not applied to VS Code connections because the extension never read the CLI's config block. This parses `# :ssh-option=` comments from the CLI's START-CODER/END-CODER block and merges them into the SSH config with lowest priority (CLI options < deployment API < user VS Code settings).
- Adds `parseCoderSshOptions()` to extract ssh-option values from the Coder CLI block in `~/.ssh/config`
- Renames `computeSSHProperties` to `computeSshProperties` for naming consistency
- Adds managed-section header comments to the generated SSH config block

Fixes #832

### DIFF
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -832,17 +832,17 @@ export class Remote {
 		const sshConfig = new SSHConfig(sshConfigFile);
 		await sshConfig.load();
 
-		// Merge SSH config from three sources (lowest to highest priority):
-		// 1. coder config-ssh --ssh-option flags from the CLI block
-		// 2. Deployment SSH config from the coderd API
-		// 3. User's VS Code coder.sshConfig setting
+		// Merge SSH config from three sources (highest to lowest priority):
+		// 1. User's VS Code coder.sshConfig setting
+		// 2. coder config-ssh --ssh-option flags from the CLI block
+		// 3. Deployment SSH config from the coderd API
 		const configSshOptions = parseCoderSshOptions(sshConfig.getRaw());
 		const userConfigSsh = vscode.workspace
 			.getConfiguration("coder")
 			.get<string[]>("sshConfig", []);
 		const userConfig = parseSshConfig(userConfigSsh);
 		const sshConfigOverrides = mergeSshConfigValues(
-			mergeSshConfigValues(configSshOptions, deploymentSSHConfig),
+			mergeSshConfigValues(deploymentSSHConfig, configSshOptions),
 			userConfig,
 		);
 
--- a/src/remote/sshSupport.ts
+++ b/src/remote/sshSupport.ts
@@ -1,5 +1,9 @@
 import * as childProcess from "child_process";
 
+// Matches the OpenSSH version number from `ssh -V` output.
+// [^,]* prevents greedy matching across comma-separated components
+const openSSHVersionRegex = /OpenSSH[^,]*_([\d.]+)/;
+
 /** Check if the local SSH installation supports the `SetEnv` directive. */
 export function sshSupportsSetEnv(): boolean {
 	try {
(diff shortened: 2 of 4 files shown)


src/remote/sshConfig.ts
```diff
@@ -36,6 +36,12 @@ const defaultFileSystem: FileSystem = {
 	writeFile,
 };
 
+// Matches an SSH config key at the start of a line (e.g. "ConnectTimeout", "LogLevel").
+const sshKeyRegex = /^[a-zA-Z0-9-]+/;
```
mafredri, line 40: Do any keys actually have numbers or `-` in them?

src/remote/sshSupport.ts
```diff
@@ -11,10 +12,10 @@ export function sshSupportsSetEnv(): boolean {
...
-// sshVersionSupportsSetEnv ensures that the version string from the SSH
-// command line supports the `SetEnv` directive.
-//
-// It was introduced in SSH 7.8 and not all versions support it.
+/**
+ * Check if an SSH version string supports the `SetEnv` directive.
+ * Requires OpenSSH 7.8 or later.
+ */
 export function sshVersionSupportsSetEnv(sshVersionString: string): boolean {
 	const match = /OpenSSH.*_([\d.]+)[^,]*/.exec(sshVersionString);
```
mafredri: This is not changed but caught my eye, why do we need `[^,]*`, it seems to serve no purpose as we're only matching numbers and `.`. If we wanted to make sure we're not matching the version from the next entry. Like: `, LibreSSL 3.3.6`, then this makes more sense to me as a guard against `.*` greedily matching the next entry if it was e.g. `LibreSSL_3.3.6`.

```suggestion
	const match = /OpenSSH[^,]*_([\d.]+)/.exec(sshVersionString);
```

Compare: https://regex101.com/r/jLv50G/1 vs https://regex101.com/r/Sl0cr7/1.

src/remote/remote.ts
```diff
@@ -849,6 +839,20 @@ export class Remote {
 		const sshConfig = new SSHConfig(sshConfigFile);
 		await sshConfig.load();
 
+		// Merge SSH config from three sources (lowest to highest priority):
+		// 1. coder config-ssh --ssh-option flags from the CLI block
+		// 2. Deployment SSH config from the coderd API
+		// 3. User's VS Code coder.sshConfig setting
```
mafredri: This order seems wrong.

Coder configssh has the following priority order (highest to lowest):
1. User defined
2. Deployment defined
3. Defaults

Due to SSH evaluation method of using _first encounter_ of a setting, the user has highest even though it seems like the latter ones override.

Here we're introducing a new user-level setting that comes from VS Code. This should have the _highest_ priority:

0. VS Code (user defined in VS Code environment)
1. User defined via configssh
2. Deployment defined
3. Defaults (if applicable?)

src/remote/sshConfig.ts
EhabY: Some keys like `ForwardX11` might have digits, as for `-`, I'm not sure tbh, it could be removed

src/remote/sshSupport.ts
EhabY: Ooooh nice catch, I've changed this and added a test case for it!

src/remote/remote.ts
EhabY: Ah right, I wasn't sure about the order but this makes sense since `configssh` is user defined so should be higher than deployment defined!

PR #835, opened 2026-03-12
feat: make keyring opt-in and store token at login

### LINKED ISSUE #836: Make `coder.useKeyring` opt-in and store token at login
## Problem

`coder.useKeyring` currently defaults to `true` on macOS/Windows, which has unintended side effects: logging out of the IDE logs out the CLI and vice versa, since they share the same keyring entry. Additionally, tokens are only written to the keyring when connecting to a workspace, not at login time — so the CLI doesn't pick them up until a workspace connection occurs.

## Proposed Changes

- Default `coder.useKeyring` to `false` (opt-in) due to shared logout side effects
- Update the setting description to document sync behavior and version requirements
- Store token to the OS keyring at login time so the CLI picks it up immediately
- Show a cancellable progress notification for CLI keyring invocations during login

EhabY:
- Default `coder.useKeyring` to false (opt-in) due to shared logout side effects between CLI and IDE
- Update setting description to document sync behavior and version requirements
- Store token to OS keyring at login time so the CLI picks it up immediately
- Show cancellable progress notification for all CLI keyring invocations during login
- Migrate `CliCredentialManager` public API to options bag pattern with signal and keyringOnly support

Closes #836

### DIFF
--- a/src/core/cliCredentialManager.ts
+++ b/src/core/cliCredentialManager.ts
@@ -6,6 +6,7 @@ import { promisify } from "node:util";
 import * as semver from "semver";
 
 import { isKeyringEnabled } from "../cliConfig";
+import { isAbortError } from "../error/errorUtils";
 import { featureSetForVersion } from "../featureSet";
 import { getHeaderArgs } from "../headers";
 import { renameWithRetry, tempFilePath, toSafeHost } from "../util";
@@ -100,6 +101,7 @@ export class CliCredentialManager {
 	/**
 	 * Read a token via `coder login token --url`. Returns trimmed stdout,
 	 * or undefined on any failure (resolver, CLI, empty output).
+	 * Throws AbortError when the signal is aborted.
 	 */
 	public async readToken(
 		url: string,
@@ -129,7 +131,7 @@ export class CliCredentialManager {
 			const token = stdout.trim();
 			return token || undefined;
 		} catch (error) {
-			if ((error as Error).name === "AbortError") {
+			if (isAbortError(error)) {
 				throw error;
 			}
 			this.logger.warn("Failed to read token via CLI:", error);
@@ -139,7 +141,8 @@ export class CliCredentialManager {
 
 	/**
 	 * Delete credentials for a deployment. Runs file deletion and keyring
-	 * deletion in parallel, both best-effort (never throws).
+	 * deletion in parallel, both best-effort. Throws AbortError when the
+	 * signal is aborted.
 	 */
 	public async deleteToken(
 		url: string,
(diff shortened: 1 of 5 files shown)


src/core/cliCredentialManager.ts
```diff
@@ -34,7 +35,8 @@ export type BinaryResolver = (deploymentUrl: string) => Promise<string>;
  * Returns true on platforms where the OS keyring is supported (macOS, Windows).
  */
 export function isKeyringSupported(): boolean {
-	return process.platform === "darwin" || process.platform === "win32";
+	const platform = os.platform();
+	return platform === "darwin" || platform === "win32";
```
DanielleMaywood, line 40: is `process.platform` not reliable enough?

src/core/cliCredentialManager.ts
```diff
@@ -114,10 +123,15 @@ export class CliCredentialManager {
...
-			const { stdout } = await this.execWithTimeout(binPath, args);
+			const { stdout } = await this.execWithTimeout(binPath, args, {
+				signal: options?.signal,
+			});
 			const token = stdout.trim();
 			return token || undefined;
 		} catch (error) {
+			if ((error as Error).name === "AbortError") {
+				throw error;
+			}
```
DanielleMaywood: Should we gate this `error` with an `instanceof` (or equivalent) check to ensure we aren't casting?

test/unit/progress.test.ts
```diff
@@ -122,6 +126,77 @@ describe("withCancellableProgress", () => {
...
+
+			await withOptionalProgress((ctx) => {
+				captured = ctx;
+				return Promise.resolve();
+			}, disabledOptions);
+
+			expect(captured).toBeDefined();
+			expect(typeof captured!.progress.report).toBe("function");
+			expect(captured!.signal).toBeInstanceOf(AbortSignal);
+			expect(captured!.signal.aborted).toBe(false);
```
DanielleMaywood, line 184: I wish these `expect` functions were typed better 🥲 Would stop us having to do `captured!`

src/core/cliCredentialManager.ts
EhabY: Both are the same, this is just easier to mock

test/unit/progress.test.ts
EhabY: I know but sadly `toBeDefined` doesn't change the typing at all :/

src/core/cliCredentialManager.ts
EhabY: Added this util:

```ts
/**
 * Check whether an unknown thrown value is an AbortError (signal cancellation).
 */
export function isAbortError(error: unknown): boolean {
	return error instanceof Error && error.name === "AbortError";
}
```

PR #837, opened 2026-03-12
chore: upgrade dependencies and parallelize build pipeline

EhabY:
- Upgrade dependencies: proxy-agent v6→v7, esbuild 0.27.3→0.27.4, Vite 7→8, Vitest 4.0→4.1, TypeScript ESLint 8.56→8.57, Electron 40→41, pnpm 10.30→10.32, and others
- Switch to `moduleResolution: "Bundler"` for proxy-agent v7 compatibility (its `exports` field requires Bundler or NodeNext resolution)
- Parallelize the build pipeline: webview builds and extension bundling now run concurrently (~1s vs 3.3s)
- Separate type-checking from building into a dedicated `typecheck` script — esbuild and Vite handle TypeScript transpilation, so `tsc` is only needed for validation
- Remove unused `composite` project references and declaration emit from workspace packages (all packages export raw `.ts` source; emitted `.d.ts` files were unused)
- Add per-package `typecheck` scripts, auto-discovered via `pnpm -r`
- Unify production detection: both esbuild and Vite now read `NODE_ENV`
- Replace `vscode:prepublish` with explicit build steps in `package` scripts and CI workflows to avoid `vsce` invoking npm (which warns about pnpm config)
- Extension bundle size reduced ~25% (2.7MB → 2.0MB) from proxy-agent v7's smaller dependency tree

### DIFF
--- a/test/unit/api/utils.test.ts
+++ b/test/unit/api/utils.test.ts
@@ -14,7 +14,8 @@ vi.mock("node:fs/promises", async () => {
 	return { default: memfs.fs.promises, ...memfs.fs.promises };
 });
 
-// ProxyAgentOptions extends TLS options but TypeScript doesn't resolve the intersection.
+// ProxyAgentOptions uses '' as the URI type parameter, which makes ConnectOpts
+// resolve to `never` and drops TLS fields. Re-add them for test assertions.
 type AgentOpts = ProxyAgentOptions & ConnectionOptions;
 
 describe("needToken", () => {
--- a/package.json
+++ b/package.json
@@ -489,6 +489,7 @@
 		"@eslint-react/eslint-plugin": "^2.13.0",
 		"@eslint/js": "^10.0.1",
 		"@eslint/markdown": "^7.5.1",
+		"@rolldown/plugin-babel": "catalog:",
 		"@tanstack/react-query": "catalog:",
 		"@testing-library/jest-dom": "^6.9.1",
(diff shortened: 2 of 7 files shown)


test/unit/api/utils.test.ts
```diff
@@ -71,36 +71,38 @@ describe("createHttpAgent", () => {
 			cfg.set("coder.tlsCaFile", "/ca.pem");
 
 			const agent = await createHttpAgent(cfg);
-			const opts = agent.connectOpts as AgentOpts;
+			const opts = agent.connectOpts as AgentOpts | undefined;
```
DanielleMaywood, line 75: are all of these casts required btw? feels unfortunate that we have them

EhabY: Sadly yes, see the comment above:

> ProxyAgentOptions extends TLS options but TypeScript doesn't resolve the intersection.

Basically TS does not recognize these fields we are accessing (but they do exist, hence the tests working)

DanielleMaywood: my bad, git diff did me dirty 💀 sucks that typescript cannot handle this

EhabY: I've updated the comment to be more accurate:

> ProxyAgentOptions uses '' as the URI type parameter, which makes ConnectOpts
> resolve to `never` and drops TLS fields.  Re-add them for test assertions.

PR #844, opened 2026-03-17
feat: embed Coder agent chat in VS Code sidebar panel

ThomasK33:
## Summary

Adds a new **Coder Chat** webview panel that embeds the Coder agent chat UI directly inside VS Code, triggered via a deep link.

### Deep link format

```
vscode://coder.coder-remote/openChat?url=<coderUrl>&token=<sessionToken>&agentId=<chatAgentId>
```

### How it works

1. The `/openChat` URI handler calls `setupDeployment()` to authenticate, then opens the chat panel.
2. `ChatPanelProvider` renders a webview containing an iframe pointing at `/agents/{id}/embed` on the Coder server.
3. The embed page detects the user is signed out and sends `postMessage({ type: "coder:vscode-ready" })` to the parent.
4. The webview relays this to the extension host, which replies with the session token.
5. The webview forwards `{ type: "coder:vscode-auth-bootstrap", payload: { token } }` into the iframe.
6. The embed page calls `API.setSessionToken(token)`, fetches user + permissions, and renders the authenticated chat UI.

### Why not cookies?

Cookie-based auth (`SameSite=None; Secure`) does not work in VS Code webview iframes over HTTP. The `setSessionToken` approach uses header-based auth via axios, bypassing cookies and CSRF entirely.

### Files changed

- **`package.json`** — Added `coderChat` view container (panel, not activity bar) and `coder.chatPanel` webview view.
- **`src/webviews/chat/chatPanelProvider.ts`** *(new)* — `ChatPanelProvider` class: postMessage relay, `openChat(agentId)`, iframe HTML generation.
- **`src/uri/uriHandler.ts`** — Added `/openChat` route and `handleOpenChat()` handler.
- **`src/extension.ts`** — Creates and registers `ChatPanelProvider`, passes it to the URI handler.

### Testing

Verified end-to-end against `dev.coder.com`: deep link → extension activation → postMessage auth → full chat UI rendered with multi-turn conversation and tool invocations.

---

> This PR was authored with [Coder Agents](https://coder.com/agents) 🤖

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,7 +196,7 @@
 			"panel": [
 				{
 					"id": "coderChat",
-					"title": "Coder Chat",
+					"title": "Coder Chat (Experimental)",
 					"icon": "media/shorthand-logo.svg"
 				}
 			]
@@ -236,7 +236,7 @@
 				{
 					"type": "webview",
 					"id": "coder.chatPanel",
-					"name": "Coder Chat",
+					"name": "Coder Chat (Experimental)",
 					"icon": "media/shorthand-logo.svg"
 				}
 			]


src/uri/uriHandler.ts
```diff
@@ -98,6 +121,10 @@ async function handleOpen(ctx: UriRouteContext): Promise<void> {
 		folder ?? undefined,
 		openRecent,
 	);
+
+	if (agentId) {
+		chatPanelProvider.openChat(agentId);
+	}
```
EhabY: Does the `openChat` execute actually? Because `commands.open` might trigger a window refresh and thus this code won't run, or possibly might open in a new window but `openChat` is executed on this one

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -0,0 +1,196 @@
...
+
+	public refresh(): void {
+		if (!this.view) {
+			return;
+		}
+		this.renderView();
+	}
+
+	private renderView(): void {
+		const webview = this.view!.webview;
```
EhabY: Maybe just explicitly throw when `this.view` is undefined instead of asserting

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -0,0 +1,196 @@
...
+					"Chat iframe requested auth but no session token available",
+				);
+				return;
+			}
+			this.logger.info("Chat: forwarding token to iframe");
+			this.view?.webview.postMessage({
+				type: "coder:auth-bootstrap-token",
+				token,
+			});
+		}
```
EhabY, line 109: Should this be something like `coder:request-token`? While unlikely but a token can expire mid-session

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	private getIframeHtml(embedUrl: string, allowedOrigin: string): string {
+		const nonce = randomBytes(16).toString("base64");
+
+		return /* html */ `<!DOCTYPE html>
```
EhabY, line 115: Nit: maybe extract the style and script contents into a variable so it's easier to isolate and read

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -0,0 +1,196 @@
...
+        const data = event.data;
+        if (!data || typeof data !== 'object') return;
+
+        if (event.source === iframe.contentWindow) {
+          if (data.type === 'coder:vscode-ready') {
+            status.textContent = 'Authenticating…';
+            vscode.postMessage({ type: 'coder:vscode-ready' });
+          }
+          return;
+        }
```
EhabY, line 166: Should we just forward all messages as is to the extension?

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -0,0 +1,196 @@
...
+  <script nonce="${nonce}">
+    (function () {
+      const vscode = acquireVsCodeApi();
+      const iframe = document.getElementById('chat-frame');
+      const status = document.getElementById('status');
+
+      iframe.addEventListener('load', () => {
+        iframe.style.display = 'block';
+        status.style.display = 'none';
+      });
```
EhabY, line 154: This could be made more flexible so that even if messages arrive mid-run it wouldn't cause the status to always be displayed and never cleared (we could have some state tracking in this webview possibly). Anyway if this is unlikely or too much work then we can skip

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -0,0 +1,196 @@
...
+	private disposeInternals(): void {
+		for (const d of this.disposables) {
+			d.dispose();
+		}
+		this.disposables = [];
+	}
+
+	dispose(): void {
+		this.disposeInternals();
+	}
```
EhabY: nit: Could just be dispose 🤷

src/uri/uriHandler.ts
```diff
@@ -8,19 +8,22 @@ import { CALLBACK_PATH } from "../oauth/utils";
...
 	commands: Commands;
+	chatPanelProvider: ChatPanelProvider;
 }
 
 type UriRouteHandler = (ctx: UriRouteContext) => Promise<void>;
 
 const routes: Readonly<Record<string, UriRouteHandler>> = {
 	"/open": handleOpen,
 	"/openDevContainer": handleOpenDevContainer,
+	"/openChat": handleOpenChat,
```
EhabY: How can we trigger this one actually?

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -0,0 +1,196 @@
...
+		if (!this.agentId) {
+			webview.html = this.getNoAgentHtml();
+			return;
+		}
+
+		const coderUrl = this.client.getHost();
+		if (!coderUrl) {
+			webview.html = this.getNoAgentHtml();
+			return;
+		}
```
EhabY: I keep getting the no agent HTML even though I have a chat running, am I doing something wrong 🤔 ?

src/extension.ts
```diff
@@ -222,8 +223,18 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
+		chatPanelProvider,
+		vscode.window.registerWebviewViewProvider(
+			ChatPanelProvider.viewType,
+			chatPanelProvider,
+			{ webviewOptions: { retainContextWhenHidden: true } },
+		),
+	);
+
 	ctx.subscriptions.push(
-		registerUriHandler(serviceContainer, deploymentManager, commands),
```
EhabY, line 226: Any reason why this was moved to the bottom?

src/webviews/chat/chatPanelProvider.ts
ThomasK33: I think that's a great idea for a follow-up, but it's not needed for this now.

src/webviews/chat/chatPanelProvider.ts
ThomasK33: What benefit do we gain from doing so?

src/uri/uriHandler.ts
ThomasK33: Via a direct deep link

src/webviews/chat/chatPanelProvider.ts
ThomasK33: I don't think there's a single code path that would cause this function to be called with a null view.

src/webviews/chat/chatPanelProvider.ts
ThomasK33: Maybe to be considered for a follow-up, but don't think it makes sense for this PR.

src/webviews/chat/chatPanelProvider.ts
ThomasK33: I don't expect those styles and scripts to be reused elsewhere.
They serve a single purpose for embedding this iframe and the visual scaffolding in the sidebar.

src/webviews/chat/chatPanelProvider.ts
ThomasK33: Also, this isn't an assertion. It's a type assertion that will still cause a runtime exception if a code path executes it without a set view. So, it already does what you're asking for technically.

src/uri/uriHandler.ts
ThomasK33: Thanks, switched it to use the same memento mechanism as other parts of the extension.

src/uri/uriHandler.ts
EhabY: I meant through the Coder UI, I could do `/open` but not `/openChat`, is it not implemented yet in `coder/coder`?

src/webviews/chat/chatPanelProvider.ts
EhabY: Yeah but if would fail in an ambiguous way, doing `if(!this.view) throw new Error("...")` is a lot more clear

src/webviews/chat/chatPanelProvider.ts
EhabY: Fair enough

src/webviews/chat/chatPanelProvider.ts
EhabY: Understandable

src/webviews/chat/chatPanelProvider.ts
EhabY: This ties with the auth comment above, so maybe they should be tackled at the same time when needed!

src/webviews/chat/chatPanelProvider.ts
EhabY: So we don't have to maintain a switch here or a large `if else`, I suppose this is not an issue now..

src/uri/uriHandler.ts
ThomasK33: The updated deep links are not yet part of coder/coder. I want to get the extension changes merged first, in case the deep link needs to change.
I'm happy to remove the `/openChat` and just leave `/open` for now.

src/webviews/chat/chatPanelProvider.ts
ThomasK33: Sure, though I'd argue that it's less a matter of when it's needed and more a matter of if it's needed.

src/uri/uriHandler.ts
EhabY: Oh no worries, let's keep it here then, as long as the current handling makes sense!

src/uri/uriHandler.ts
```diff
@@ -89,6 +89,15 @@ async function handleOpen(ctx: UriRouteContext): Promise<void> {
 		params.has("openRecent") &&
 		(!params.get("openRecent") || params.get("openRecent") === "true");
 
+	// Persist the chat agent ID before commands.open() triggers
+	// a remote-authority reload that wipes in-memory state.
+	// The extension picks this up after the reload in activate().
+	const agentId = params.get("agentId");
+	if (agentId) {
+		const mementoManager = serviceContainer.getMementoManager();
+		await mementoManager.setPendingChatAgentId(agentId);
```
EhabY: Might not be an issue but this is set here while `commands.open` might not even open a new window. It's not major since it means that the next refresh might open the chat agent. Could be addressed in a follow up

src/webviews/chat/chatPanelProvider.ts
EhabY: Apparently this has not made it into `coder/coder` yet, we'll check after

PR #846, opened 2026-03-17
fix: improve SSH process discovery and sleep/wake recovery

### LINKED ISSUE #845: SSH monitor: improve process discovery and sleep/wake recovery
On Linux systems without `netstat`, `find-process` could not resolve the SSH PID for a port. This meant the extension could not find the proxy log file (breaking View Logs) or read network info (breaking the network status bar).

Additionally, after laptop sleep/wake, VS Code may reconnect SSH on a new port. The extension did not detect this, so WebSockets were left in a TCP half-open state and the label formatter was not re-registered.

### Changes

- Bump `find-process` to 2.1.1: uses `ss` -> `netstat` -> `lsof` on Linux, `netstat` -> `lsof` on macOS.
- Detect SSH port changes in the Remote SSH log to reset discovery backoff.
- Reconnect WebSockets and re-register the label formatter on SSH PID change.

EhabY:
## Summary

- Detect SSH port changes in the Remote SSH log to reset discovery backoff after sleep/wake.
- Re-register the label formatter when the SSH PID changes.
- Bump `find-process` to 2.1.1: uses `ss` -> `netstat` -> `lsof` on Linux, `netstat` -> `lsof` on macOS, fixing systems where `netstat` was unavailable and the SSH PID could not be resolved (broke network info display and log viewing).

Closes #845

### DIFF
--- a/test/unit/remote/sshProcess.test.ts
+++ b/test/unit/remote/sshProcess.test.ts
@@ -36,6 +36,7 @@ describe("SshProcessMonitor", () => {
 	});
 
 	afterEach(() => {
+		vi.useRealTimers();
 		for (const m of activeMonitors) {
 			m.dispose();
 		}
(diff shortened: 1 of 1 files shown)


src/core/cliManager.ts
```diff
@@ -232,6 +232,7 @@ export class CliManager {
 	): Promise<boolean> {
 		const choice = await vscodeProposed.window.showErrorMessage(
 			`${reason}. Run version ${version} anyway?`,
+			{ modal: true, useCustom: true },
```
DanielleMaywood, line 235: Struggling to understand the context behind this change. Is this a drive-by fix or related to the ssh process discovery and sleep/wake recovery?

EhabY: Drive-by fix yes, we should use modals for all the blocking decisions

PR #872, opened 2026-04-02
fix: add ProxyCommand retry wrapper for sleep/wake reconnection

EhabY:
- After sleep/wake, DNS resolution often fails transiently (`systemd-resolved` not ready). When `coder ssh` fails instantly, the Remote SSH extension sees unparsable output and throws a **permanent** `SshResolverError.Create()`, showing "Reload Window" instead of retrying.
- This wraps the SSH ProxyCommand with a POSIX shell retry script that retries up to 10 times with 5s delays on quick failures (<10s runtime). The 5s sleep is long enough for the machine to fully suspend during it, so when it wakes, `sleep` returns immediately and the next retry succeeds with DNS working.
- Skipped on Windows (POSIX shell only). Falls back gracefully to the unwrapped command if the script can't be written.

### DIFF
--- a/src/remote/proxyCommandRetry.ts
+++ b/src/remote/proxyCommandRetry.ts
@@ -1,0 +1,71 @@
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+
+import { renameWithRetry, tempFilePath } from "../util";
+
+import type { Logger } from "../logging/logger";
+
+/**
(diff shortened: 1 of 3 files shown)


EhabY:
Closed in favor of the CLI solution: https://github.com/coder/coder/issues/24005

PR #875, opened 2026-04-07
refactor: extract CLI execution into cliExec module with improved UX

EhabY:
- Split `core/cliUtils` into `cliUtils` (binary file management) and `cliExec` (command execution: `version`, `speedtest`, `ping`, `openAppStatusTerminal`)
- Introduce `CliEnv` interface so callers don't manage shell escaping; each `cliExec` function resolves its own flags
- Add `resolveClientAndWorkspace` for shared workspace resolution (sidebar, connected, or picker)
- Use `locateBinary` (stat-only) in `resolveCliEnv` before falling back to `fetchBinary`
- Surface CLI stderr in error messages instead of leaking full binary paths
- Validate Go duration input in speed test prompt
- Rename "Run Speed Test" → "Speed Test Workspace" in command palette, "Speed Test" in sidebar
- Use separate `:views` command IDs for shorter context menu labels (GitLens pattern)
- Remove "Diagnostics" submenu — inline commands directly in sidebar context menu

### DIFF
--- a/src/core/cliExec.ts
+++ b/src/core/cliExec.ts
@@ -19,15 +19,17 @@ export interface CliEnv {
 
 const execFileAsync = promisify(execFile);
 
-// util.promisify types are dynamic so there is no concrete type we can import
-// and we have to make our own.
-type ExecException = ExecFileException & { stdout?: string; stderr?: string };
+function isExecFileException(error: unknown): error is ExecFileException {
+	return error instanceof Error && "code" in error;
+}
 
 /** Prefer stderr over the default message which includes the full command line. */
 function cliError(error: unknown): Error {
-	const stderr = (error as ExecException)?.stderr?.trim();
-	if (stderr) {
-		return new Error(stderr, { cause: error });
+	if (isExecFileException(error)) {
+		const stderr = error.stderr?.trim();
+		if (stderr) {
+			return new Error(stderr, { cause: error });
+		}
 	}
 	return toError(error);
 }
@@ -55,7 +57,10 @@ export async function version(binPath: string): Promise<string> {
 		stdout = result.stdout;
 	} catch (error) {
 		// It could be an old version without support for --output.
-		if ((error as ExecException)?.stderr?.includes("unknown flag: --output")) {
+		if (
+			isExecFileException(error) &&
+			error.stderr?.includes("unknown flag: --output")
+		) {
 			const result = await execFileAsync(binPath, ["version"]);
 			if (result.stdout?.startsWith("Coder")) {
 				const v = result.stdout.split(" ")[1]?.trim();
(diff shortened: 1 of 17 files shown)


src/core/cliExec.ts
```diff
@@ -0,0 +1,262 @@
...
+	binary: string;
+	auth: CliAuth;
+	configs: Pick<vscode.WorkspaceConfiguration, "get">;
+}
+
+const execFileAsync = promisify(execFile);
+
+// util.promisify types are dynamic so there is no concrete type we can import
+// and we have to make our own.
+type ExecException = ExecFileException & { stdout?: string; stderr?: string };
```
jeremyruppel: is this the case? i found this type in `@types/node`, which we should maybe add to this repo?

src/core/cliExec.ts
```diff
@@ -0,0 +1,262 @@
...
+
+/** Prefer stderr over the default message which includes the full command line. */
+function cliError(error: unknown): Error {
+	const stderr = (error as ExecException)?.stderr?.trim();
```
jeremyruppel: i generally prefer a type guard over `as`


```typescript
import {type ExecFileException, execFile} from "node:child_process";
import {promisify} from "node:util";

const execFileAsync = promisify(execFile);

const isExecFileException = (error: unknown): error is ExecFileException => {
	return (
		error instanceof Error &&
		"code" in error &&
		"stdout" in error &&
		"stderr" in error
	);
};

try {
	await execFileAsync("ls", ["-l"]);
} catch (error: unknown) {
	if (isExecFileException(error)) {
        error.stdout; // string | undefined, strongly typed
	} else {
		// some other Error
	}
}
```

src/core/cliExec.ts
```diff
@@ -0,0 +1,262 @@
...
+						cause: error,
+					});
+				}
+				return v;
+			}
+		}
+		throw cliError(error);
+	}
+
+	const json = JSON.parse(stdout) as { version?: string };
```
jeremyruppel, line 78:
```suggestion
	const json: { version?: string } = JSON.parse(stdout);
```

src/core/cliExec.ts
EhabY: Oh right, this was copied code, it's part of the node types now so I'll use it!

src/core/cliExec.ts
EhabY: This one fails linting because `JSON.parse(...)` returns `any` so we cast it (without checking). This is already the pattern in the codebase, maybe we should use Zod or something similar elsewhere 🤔

src/core/cliExec.ts
EhabY: Made it:

```ts
function isExecFileException(error: unknown): error is ExecFileException {
	return error instanceof Error && "code" in error;
}
```

Because `stdout` and `stderr` are optional fields anyway so we cannot use `x in error` like this

jeremyruppel (approved):
Very nice!

PR #876, opened 2026-04-08
fix: Windows compatibility for tests, scripts, and Vitest Explorer

EhabY:
- **cross-env for npm scripts**: Unix-style `VAR=value` syntax in npm scripts doesn't work on Windows. Added `cross-env` to `build:production`, `test`, `test:extension`, and `test:webview` scripts.
- **Vitest Explorer electron path**: Changed `vitest.nodeExecutable` from `node_modules/.bin/electron` (doesn't exist on Windows, only `electron.cmd` does) to `node_modules/electron/dist/electron` which resolves cross-platform.
- **speedtest tests**: `execFile` cannot spawn `.cmd` files on Windows (`EINVAL`). Replaced platform-specific `.cmd`/shell script creation with a unified approach using `process.execPath` to run `echo-args.js` directly on both platforms.
- **sshProcess tests**: Path assertions used hardcoded forward slashes. Changed to `path.join()` so separators match the OS.
- **Integration test compilation**: Added `compile-tests:integration` script and included it in the `build` step so the Mocha test explorer can discover integration tests after `pnpm build`.


test/unit/remote/sshProcess.test.ts
```diff
@@ -395,7 +398,7 @@ describe("SshProcessMonitor", () => {
 
 			// With localeCompare: ["a", "Z"] -> reversed -> "Z" first
 			// With plain sort(): ["Z", "a"] -> reversed -> "a" first (WRONG)
-			expect(logPath).toBe("/proxy-logs/Z-999.log");
+			expect(logPath).toBe(path.join("/proxy-logs", "Z-999.log"));
```
jeremyruppel, line 401: is it a problem if these also begin with `/`?

EhabY: `path.join` also normalizes so `path.join("///proxy-logs/sss", "/abc/")` would be `\proxy-logs\sss\abc` on Windows but `/proxy-logs/sss/abc` on Linux, the path separators only matter on the start of the first segment (to differentiate absolute from relative paths)

jeremyruppel (commented):
great! 👍

jeremyruppel (approved):

PR #879, opened 2026-04-09
feat: allow binaryDestination to accept a full path to the CLI binary

EhabY:
## Summary

The `coder.binaryDestination` setting (and `CODER_BINARY_DESTINATION` env var) now accepts a file path in addition to a directory path.

**File path** (e.g. `/usr/bin/coder`): the extension checks its version and downloads a replacement if it does not match the server (and downloads are enabled). If downloads are disabled, the existing binary is used with a warning. Downloads always go to the platform-specific name first, then rename to the configured path under the lock.

**Directory path** (existing behavior): looks for the platform-specific name (`coder-linux-amd64`) first, then falls back to the simple name (`coder` / `coder.exe`). This supports package-manager-installed CLIs (e.g. via `apt`, `brew`, `winget`) that use the simple name.

### Changes

- Add `cliUtils.simpleName()` and rename `name()` to `fullName()` for clarity
- Add unified `resolveBinaryPath()` in `CliManager` returning a `ResolvedBinary` discriminated union (`source: "file-path" | "directory" | "not-found"`) used by both `locateBinary()` and `fetchBinary()`
- Scope `rmOld` cleanup to only files prefixed with the binary basename, preventing accidental removal in shared directories like `/usr/bin`
- Always download using the platform-specific name for temp files, backups, signatures, and lock files
- On error recovery, `handleAnyBinaryFailure` tries candidates in order (download path, user configured path, old backups) and renames the fallback to `resolved.binPath` so the binary ends up where the user expects
- Linearize `fetchBinary` decision tree: version check, lock-wait, and rename logic now use sequential guard clauses instead of nested branches, and a `renameToFinalPath` helper replaces duplicated rename logic
- Fix `binaryDestination` setting description to accurately reflect version-check-then-download behavior
- Remove flaky lock/progress file assertions from concurrent tests

### Test coverage

- File destination: locate, version match, download+rename, in-place download (platform-specific name matches), downloads disabled, error fallback to configured path, error fallback with rename
- Simple name fallback: locate, prefer platform-specific, version match, download to platform-specific name, error fallback
- Old backup restore: simulates EBUSY during replace, verifies `.old-*` backup is restored to expected path

Resolves #861

### DIFF
--- a/test/unit/core/cliUtils.test.ts
+++ b/test/unit/core/cliUtils.test.ts
@@ -14,8 +14,8 @@ describe("CliUtils", () => {
 		await fs.mkdir(tmp, { recursive: true });
 	});
 
-	it("name", () => {
-		expect(cliUtils.name().startsWith("coder-")).toBeTruthy();
+	it("fullName", () => {
+		expect(cliUtils.fullName().startsWith("coder-")).toBeTruthy();
 	});
 
 	it("simpleName", () => {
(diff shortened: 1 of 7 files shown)


test/unit/core/cliUtils.test.ts
```diff
@@ -14,8 +14,17 @@ describe("CliUtils", () => {
...
+	it("fullName", () => {
+		expect(cliUtils.fullName().startsWith("coder-")).toBeTruthy();
+	});
+
+	it("simpleName", () => {
+		const simple = cliUtils.simpleName();
+		if (process.platform === "win32") {
+			expect(simple).toBe("coder.exe");
+		} else {
+			expect(simple).toBe("coder");
```
jeremyruppel, line 26: does this mean we have tests that run on windows platforms?

EhabY: Currently the CI does not do that but it should, I'll create a ticket to run the tests on Windows and possibly macOS!

jeremyruppel (approved):

PR #880, opened 2026-04-10
refactor: simplify shimExecFile by removing type casts and promisify

jeremyruppel:
I noticed `promisify` was adding some complexity and making strict typing difficult in this test. This uses `Reflect.apply` and a generic type parameter instead of `as unknown as` chains. Also replaces `promisify` usage in tests with plain Promise callbacks, which eliminates the need for the `promisify` custom symbol wrapping entirely.

> Claude helped 🤖 reviewed by me


EhabY:
Ah actually the reflect usage seems much cleaner than casting `as unknown`, I've extract what I can from this in a commit here: https://github.com/coder/vscode-coder/pull/876/changes/ee9f91244c811ea0d60fc33bc09c56c74facddc0

PR #886, opened 2026-04-14
feat: use startup mode for workspace updates

### LINKED ISSUE #884: "Update workspace" command does not stop running workspace before updating
The "Update workspace" command provided by the Coder VS Code extension does not currently stop the user's running workspace before updating and restarting it; instead a "start" workspace build/transition is executed on the running workspace:

Image

Despite that the confirmation prompt communicates that a stop/restart of the workspace will occur:

> Update `${createWorkspaceIdentifier(this.workspace)}` to the latest version?
> 
> Updating will restart your workspace which stops any running processes and may result in the loss of unsaved work.

https://github.com/coder/vscode-coder/blob/c2aa5ca41c23c949ba165382a980e285b50c907d/src/commands.ts#L769

I understand that ideally Coder workspace templates would be implemented to be robust to multiple "start" workspace builds/transitions, but at my company we use KubeVirt VMs for our Coder workspaces which do not easily support the resulting re-apply of our workspace Terraform template with another Coder workspace "start" build/transition - instead, only a "stop" followed by a "start" is expected to work consistently as it ensures the underlying VM is actually updated with the latest template configuration.

The impact to our devs is that clicking the "Update workspace" command in the Coder VS Code extension can sometimes (depending on the changes in the newer template version) put their workspace in an unhealthy state which confuses them and requires the additional time and manual steps of stopping and then restarting the workspace to recover.

I understand there's been an effort to make this "Update and restart" behavior more consistent throughout the Coder UX, for example https://github.com/coder/coder/issues/22020 was recently addressed to fix that in the web UI (see my similar comment on that issue here: https://github.com/coder/coder/issues/22020#issuecomment-4076938128).

Along those lines, can the "Update workspace" command in the Coder VS Code extension be changed to first stop the user's running workspace before updating and [re]starting it?

EhabY:
- Replace the boolean `firstConnect` flag with a typed `StartupMode` (`"none"` | `"start"` | `"update"`) so the extension knows exactly what to do on reconnect
- The "Update Workspace" command now sets the mode to `"update"` and reloads the window, letting the workspace state machine handle the update via `coder update` (or the REST API fallback for older CLIs)
- When reconnecting to a stopped workspace without an explicit mode, the confirm dialog now offers both "Start" and "Update and Start" (the latter only when the workspace is outdated)
- Extract a shared `runCliCommand` helper and bundle `restClient`/`featureSet` into `CliContext` so `startWorkspace` and `updateWorkspace` each take a single argument
- Add the `cliUpdate` feature flag gated on CLI >= 2.25

Closes #884

### DIFF
--- a/src/api/workspace.ts
+++ b/src/api/workspace.ts
@@ -46,10 +46,12 @@ export class LazyStream<T> {
 }
 
 interface CliContext {
+	restClient: Api;
 	auth: CliAuth;
 	binPath: string;
 	workspace: Workspace;
 	writeEmitter: vscode.EventEmitter<string>;
+	featureSet: FeatureSet;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


src/api/workspace.ts
```diff
@@ -45,66 +45,49 @@ export class LazyStream<T> {
...
-				.toString()
-				.split(/\r*\n/)
-				.filter((line) => line !== "");
-			for (const line of lines) {
-				writeEmitter.fire(line.toString() + "\r\n");
+		const cmd = `${escapeCommandArg(ctx.binPath)} ${fullArgs.join(" ")}`;
+		const proc = spawn(cmd, { shell: true });
+
+		proc.stdout.on("data", (data: Buffer) => {
+			for (const line of splitLines(data)) {
```
code-asher: Just occurred to me (not something introduced in this PR) that if we we ever got a chunk of data that ended in the middle of a line, we would end up emitting half that line on one line and then the other half of the line as a separate line in the next event.

Obviously not something we have to address now, but do we actually have to split into lines?  Could we just write these bytes as they are directly to the terminal?

code-asher (approved):
Nice, looks good to me!

src/featureSet.ts
```diff
@@ -44,6 +45,8 @@ export function featureSetForVersion(
 		wildcardSSH: versionAtLeast(version, "2.19.0"),
 		// --reason flag for `coder start`
 		buildReason: versionAtLeast(version, "2.25.0"),
+		// `coder update` subcommand
+		cliUpdate: versionAtLeast(version, "2.25.0"),
```
code-asher: Oh wait forgot to ask why `2.25.0` specifically; `update` was in there before that I think?  I was thinking maybe it was when a stop transition was added to `update` but it looks like that was 2.24.0.

EhabY: Yes, I think we should gate on 2.24.0 instead. I think I got confused because some bug was fixed in 2.25.0

EhabY: Done

src/api/workspace.ts
EhabY: Oooh actually it's a small fix and that does a lot cleaner. I've fixed this!

code-asher: Sick!!

PR #891, opened 2026-04-15
feat: add network slowness detection with configurable threshold

### LINKED ISSUE #887: Proactive slowness detection with status bar warning
Follow-up to [AIGOV-193](https://linear.app/codercom/issue/AIGOV-193/proactive-slowness-detection-diagnostic-prompts-and-speedtest).

When a user is connected to a workspace and the network is degraded, the extension should detect it and surface it in the status bar.

## Detection

The extension already polls `NetworkInfo` from the CLI proxy every 3 seconds. We add three configurable thresholds:

* `coder.networkThreshold.latencyMs` (default: 200) - latency above this triggers a warning
* `coder.networkThreshold.downloadMbps` (default: 5) - download speed below this triggers a warning
* `coder.networkThreshold.uploadMbps` (default: 0) - upload speed below this triggers a warning (disabled by default)

Setting any threshold to 0 disables that check. Each threshold is evaluated independently. To avoid flicker, the condition must persist across 3 consecutive polls (\~9 seconds) before triggering, and 3 consecutive healthy polls to clear.

## Status bar UX

* The existing network status bar item turns yellow (`statusBarItem.warningBackground`) when any threshold is crossed.
* The hover tooltip shows all live metrics with warning markers after the offending values:

  ```
  Latency: 350ms ⚠ (threshold: 200ms)
  Download: 45.2 Mbps
  Upload: 12.1 Mbps
  Connection: us-east (relay)
  Click for diagnostics
  ```
* Click action depends on what's degraded:
  * Only latency is bad: runs Ping Workspace
  * Only download/upload is bad: runs Speed Test
  * Multiple thresholds crossed: opens a QuickPick to choose between Ping, Speed Test, and Support Bundle
* In normal state, the tooltip shows the same metrics without warning markers.

## Files to change

* `src/remote/sshProcess.ts` - threshold checking, warning state, tooltip, click handler
* `package.json` - new setting declarations

Part of [AIGOV-193](https://linear.app/codercom/issue/AIGOV-193/proactive-slowness-detection-diagnostic-prompts-and-speedtest).

EhabY:
## What this adds

- New setting `coder.networkThreshold.latencyMs` (default **250ms**, set to `0` to disable, minimum `0`) that triggers a yellow status-bar warning when the workspace connection latency crosses the threshold.
- Warning state is debounced over **2 consecutive polls** to avoid flicker on single-sample spikes. Counter drains symmetrically on healthy polls.
- Clicking the warning runs `coder.pingWorkspace` (latency diagnostic); tooltip action links also offer direct shortcuts to the ping command and the settings entry for the threshold.

## Tooltip layout

- **P2P**: `$(zap) Directly connected peer-to-peer.`
- **Relay**: `$(broadcast) Connected via <DERP> relay. Will switch to peer-to-peer when available.` (preserves the helpful fallback note from the old tooltip)
- **Coder Connect**: dedicated message. The CLI only sets `using_coder_connect: true` and leaves latency/throughput at zero, so `update()` short-circuits to a fixed label and tooltip and skips the slowness machinery entirely.
- **Slow connection** (non-Coder-Connect): adds `$(warning) Slow connection detected` header plus action links ("Run latency test", "Configure threshold").
- **Stale readings**: tilde in status bar (`(~50ms)`) plus `$(history)` footer in the tooltip.

Closes #887

### DIFF
--- a/src/remote/networkStatus.ts
+++ b/src/remote/networkStatus.ts
@@ -4,81 +4,52 @@ import * as vscode from "vscode";
 import type { NetworkInfo } from "./sshProcess";
 
 /** Number of consecutive polls required to trigger or clear a warning */
-const WARNING_DEBOUNCE_THRESHOLD = 3;
+const WARNING_DEBOUNCE_THRESHOLD = 2;
 
 const WARNING_BACKGROUND = new vscode.ThemeColor(
 	"statusBarItem.warningBackground",
 );
 
-export interface NetworkThresholds {
+interface NetworkThresholds {
 	latencyMs: number;
 }
 
-function getThresholdConfig(): NetworkThresholds {
-	const cfg = vscode.workspace.getConfiguration("coder");
-	return {
-		latencyMs: cfg.get<number>("networkThreshold.latencyMs", 200),
-	};
-}
-
-export function isLatencySlow(
-	network: NetworkInfo,
-	thresholds: NetworkThresholds,
-): boolean {
-	return thresholds.latencyMs > 0 && network.latency > thresholds.latencyMs;
-}
-
-export function buildNetworkTooltip(
-	network: NetworkInfo,
-	latencySlow: boolean,
-	thresholds: NetworkThresholds,
-): vscode.MarkdownString {
-	const fmt = (bytesPerSec: number) =>
-		prettyBytes(bytesPerSec * 8, { bits: true }) + "/s";
-
-	const sections: string[] = [];
-
-	if (latencySlow) {
-		sections.push("$(warning) **Slow connection detected**");
-	}
-
-	const metrics: string[] = [];
-	metrics.push(
-		latencySlow
-			? `Latency: ${network.latency.toFixed(2)}ms (threshold: ${thresholds.latencyMs}ms)`
-			: `Latency: ${network.latency.toFixed(2)}ms`,
-	);
-	metrics.push(`Download: ${fmt(network.download_bytes_sec)}`);
-	metrics.push(`Upload: ${fmt(network.upload_bytes_sec)}`);
-
+function connectionLabel(network: NetworkInfo): string {
 	if (network.using_coder_connect) {
-		metrics.push("Connection: Coder Connect");
-	} else if (network.p2p) {
-		metrics.push("Connection: Direct (P2P)");
-	} else {
-		metrics.push(`Connection: ${network.preferred_derp} (relay)`);
+		return "Coder Connect";
 	}
-
-	// Two trailing spaces + \n = hard line break (tight rows within a section).
-	sections.push(metrics.join("  \n"));
-
-	if (latencySlow) {
-		sections.push(
-			"[$(pulse) Ping workspace](command:coder.pingWorkspace) · " +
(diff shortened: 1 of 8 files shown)


src/remote/networkStatus.ts
```diff
@@ -0,0 +1,163 @@
...
+	private warningCounter = 0;
+	private isWarningActive = false;
+
+	constructor(private readonly statusBarItem: vscode.StatusBarItem) {}
+
+	update(network: NetworkInfo, isStale: boolean): void {
+		const thresholds: NetworkThresholds = {
+			latencyMs: vscode.workspace
+				.getConfiguration("coder")
+				.get<number>("networkThreshold.latencyMs", 250),
```
mtojek, line 62: nit: shouldn't we cache it in the reporter?

src/remote/networkStatus.ts
```diff
@@ -0,0 +1,163 @@
...
+		};
+		const isSlow =
+			!network.using_coder_connect &&
```
mtojek: > !network.using_coder_connect

for readability purposes, I would split it into 2 conditions, somebody may read it as "using Coder Connect is _always_ fast"

src/remote/networkStatus.ts
```diff
@@ -0,0 +1,163 @@
...
+			prettyBytes(bytesPerSec * 8, { bits: true }) + "/s";
+
+		const sections: string[] = [];
+		if (this.isWarningActive) {
+			sections.push("$(warning) **Slow connection detected**");
+		}
+		sections.push(connectionSummary(network));
+
+		const metrics: string[] = [];
+		if (network.latency > 0) {
```
mtojek: latency sie float, so in theory 0.4ms will be skipped :)

mtojek (approved):

src/remote/networkStatus.ts
EhabY: Technically this could change at any time so it should always be checked, potentially we can set up events for it but tbh it's a fast operation every 3s so not critical

src/remote/networkStatus.ts
EhabY: The check is `network.latency > 0`, so 0.4 would pass through (0.4 > 0 is true). The `> 0` guard is just meant to hide the row when the CLI hasn't populated latency yet (it reports exactly 0 in that case). Did you have a different scenario in mind?

EhabY: 0.4ms is above 0, I suppose we can only skip for Coder Connect instead :thinking: ?

mtojek: Skipping specifically for Coder Connect would be a more explicit approach, rather than relying on latency <= 0. Minor point though, the current version works correctly.

src/remote/networkStatus.ts
EhabY: Yeah I think I'll restore the early return for the Coder Connect path since we have no network info at all

src/remote/networkStatus.ts
EhabY: Indeed, on main it was like this: Skip for Coder Connect and never check the latency, now I'm skipping for Coder connect and always showing the latency otherwise (if it's 0.0 then so be it since that helps with diagnostics anyway)

PR #897, opened 2026-04-15
fix: use OutputChannel for build logs instead of pseudoterminal

EhabY:
## Summary

- Replace `TerminalSession` (pseudoterminal) with `TerminalOutputChannel` (OutputChannel wrapper) for streaming build and agent logs during connection
- Strip ANSI escape sequences via `strip-ansi` since OutputChannels render them as raw text
- Simplify the write interface: `CliContext` now takes a plain `write` callback instead of a `vscode.EventEmitter`

## Why OutputChannel instead of a pseudoterminal?

VS Code never fires the pseudoterminal `open()` callback for terminals created during the remote connection resolve flow. Without `open()`, VS Code hasn't subscribed to `onDidWrite`, so all `writeEmitter.fire()` calls are silently dropped — the "Workspace Build" terminal appears but remains empty.

This affects any pseudoterminal created inside the remote resolver's `resolve()` method. Terminals created during normal operation (e.g. "Coder: Ping Workspace") work fine because VS Code has fully initialized the terminal renderer by then.

`OutputChannel`s don't depend on a rendering lifecycle callback, so they work reliably in any context.

This now looks like this and the output channel is disposed (and removed) after this is done, the only downside is that we also ANSI formatting but the terminal was always readonly so this works:

image


code-asher (approved):
It does seem unfortunate to lose any potential coloring (not sure if we did color anything though) but what can ya do if VS Code is being buggy.

PR #899, opened 2026-04-16
fix: defer outdated notification until state machine completes

EhabY:
- Delay the outdated workspace notification until after the state machine finishes so it does not fire during an in-progress update (e.g. after "Update and Restart" reloads the window).
- Re-evaluate deferred notifications when `markInitialSetupComplete()` is called so the notification still fires for workspaces that are outdated after a normal startup.
- Add staleness checks to `Commands.updateWorkspace()` so it verifies the workspace is still outdated before showing the modal and again after the user confirms.
- Pass `latestWorkspace` instead of the stale closure-captured workspace when the user clicks "Update" in the notification.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -761,6 +761,15 @@ export class Commands {
 		if (!this.workspace || !this.remoteWorkspaceClient) {
 			return;
 		}
+		const showUpToDate = () =>
+			vscode.window.showInformationMessage(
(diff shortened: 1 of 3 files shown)


zenithwolf1000 (approved):
Provisional ✅  as I'm still learning the codebase, but doesn't seem too complicated.

PR #916, opened 2026-04-22
feat: include VS Code-side logs in support bundle

### LINKED ISSUE #889: Include VS Code-side logs in support bundle
Follow-up to [AIGOV-193](https://linear.app/codercom/issue/AIGOV-193/proactive-slowness-detection-diagnostic-prompts-and-speedtest).

The current "Create Support Bundle" command shells out to `coder support bundle`, which only collects server-side diagnostics. It would be useful to also include VS Code-side logs so support has the full picture.

## Logs to include

When the user is connected to the workspace:

* **Remote SSH extension logs** - path discovered by `SshProcessMonitor` (only available while connected)
* **SSH proxy logs** - tracked via `proxyLogDir` in `PathResolver`
* **Coder extension output channel logs** - the extension's own log output

When not connected, the bundle contains only the CLI-collected data (current behavior). The VS Code logs are appended as a best-effort addition.

These would go under a `vscode-logs/` folder inside the existing zip.

## Scrubbing

The extension already redacts sensitive headers (`Coder-Session-Token`, `Proxy-Authorization`) in `formatters.ts` before writing to the output channel, so the logs should be safe to include.

The Coder CLI's own docs note that support bundles may still contain info deemed sensitive by your organization. The same caveat applies here.

## Approach options

**A) Only include SSH proxy logs.** These are generated by the Coder CLI proxy itself, lowest risk. Simple to implement since `proxyLogDir` is already tracked.

**B) Include all three log sources.** More useful for support but requires accessing the Remote SSH log path (only available while connected) and the extension's log file on disk.

**C) Skip for now.** Defer until there's a clear need from support cases.

## Files to change

* `src/commands.ts` - post-process the zip after CLI produces it
* `package.json` - add zip manipulation dependency if needed

Part of [AIGOV-193](https://linear.app/codercom/issue/AIGOV-193/proactive-slowness-detection-diagnostic-prompts-and-speedtest).

EhabY:
After the CLI generates the support bundle zip, this appends logs from three VS Code-side sources under a `vscode-logs/` directory:

- **Remote SSH extension log** — single file from `SshProcessMonitor.getLogFilePath()`
- **SSH proxy logs** — all files from `PathResolver.getProxyLogPath()`, filtered to the last 3 days by mtime
- **Extension output channel logs** — all files from `PathResolver.getCodeLogDir()`, filtered to the last 3 days by mtime

Each source is collected independently. If a source is unavailable or a file cannot be read, it is warned and skipped. If the zip read/merge/write fails, the original bundle is left untouched and any partial sibling is cleaned up. If the atomic rename step fails, the merged zip is left alongside the original as `<name>-vscode.zip` and that path is surfaced in a warning log so the user can still find their bundle.

Uses [fflate](https://github.com/101arrowz/fflate) (zero dependencies, 8kB minified, tree-shakeable) via its async `zip` / `unzip` callbacks wrapped with `node:util.promisify`. Memory is bounded by on-disk rotation: VS Code rotates extension logs per session, and `SshProcessMonitor` caps proxy logs at 20 files / 7 days, so the in-memory peak stays small.

<details>
<summary>Files changed</summary>

| File | Change |
|------|--------|
| `package.json` / `pnpm-lock.yaml` | Add `fflate` dependency |
| `src/core/supportBundleLogs.ts` | Log collection and zip append logic |
| `src/commands.ts` | Call `appendVsCodeLogs` after CLI produces the bundle |
| `test/unit/core/supportBundleLogs.test.ts` | 7 tests covering merge, empty sources, age filtering, unreadable files, rename failure, and corrupt zip |

</details>

Closes #889

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -20,6 +20,7 @@ import { type ServiceContainer } from "./core/container";
 import { type MementoManager } from "./core/mementoManager";
 import { type PathResolver } from "./core/pathResolver";
 import { type SecretsManager } from "./core/secretsManager";
+import { appendVsCodeLogs } from "./core/supportBundleLogs";
 import { type DeploymentManager } from "./deployment/deploymentManager";
 import { CertificateError } from "./error/certificateError";
 import { toError } from "./error/errorUtils";
@@ -256,6 +257,18 @@ export class Commands {
 
 				progress.report({ message: "Collecting diagnostics..." });
 				await cliExec.supportBundle(env, workspaceId, outputUri.fsPath, signal);
+
+				progress.report({ message: "Adding VS Code logs..." });
+				await appendVsCodeLogs(
+					outputUri.fsPath,
+					{
+						remoteSshLogPath: this.workspaceLogPath,
+						proxyLogDir: this.pathResolver.getProxyLogPath(),
+						extensionLogDir: this.pathResolver.getCodeLogDir(),
+					},
+					this.logger,
+				);
+
 				return outputUri;
 			},
 			{
--- a/src/core/supportBundleLogs.ts
+++ b/src/core/supportBundleLogs.ts
@@ -12,35 +12,17 @@ export interface LogSources {
 	extensionLogDir?: string;
 }
 
-// Matches proxy log filenames: coder-ssh-YYYYMMDD-HHMMSS-<nonce>.log
-const PROXY_LOG_DATE_RE =
-	/^coder-ssh-(\d{4})(\d{2})(\d{2})-(\d{2})(\d{2})(\d{2})-/;
 const PROXY_LOG_MAX_AGE_MS = 3 * 24 * 60 * 60 * 1000;
-
-function isProxyLogRecent(filename: string, now: Date): boolean {
-	const match = PROXY_LOG_DATE_RE.exec(filename);
-	if (!match) {
-		return true;
-	}
-	const fileDate = new Date(
-		Number(match[1]),
-		Number(match[2]) - 1,
-		Number(match[3]),
-		Number(match[4]),
-		Number(match[5]),
-		Number(match[6]),
-	);
-	return now.getTime() - fileDate.getTime() <= PROXY_LOG_MAX_AGE_MS;
-}
 
 /** Collect regular files from a directory into zip-ready entries. */
 async function collectDirFiles(
 	dirPath: string,
 	zipPrefix: string,
 	logger: Logger,
-	filter?: (filename: string) => boolean,
+	maxAgeMs?: number,
 ): Promise<Record<string, Uint8Array>> {
 	const files: Record<string, Uint8Array> = {};
+	const now = Date.now();
 
 	let entries: string[];
 	try {
@@ -53,13 +35,13 @@ async function collectDirFiles(
 	}
 
 	for (const entry of entries) {
-		if (filter && !filter(entry)) {
-			continue;
-		}
 		const filePath = path.join(dirPath, entry);
 		try {
 			const stat = await fs.stat(filePath);
 			if (!stat.isFile()) {
+				continue;
+			}
+			if (maxAgeMs !== undefined && now - stat.mtimeMs > maxAgeMs) {
 				continue;
 			}
 			const content = await fs.readFile(filePath);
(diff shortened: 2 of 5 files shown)


src/commands.ts
```diff
@@ -256,6 +257,18 @@ export class Commands {
 
 				progress.report({ message: "Collecting diagnostics..." });
 				await cliExec.supportBundle(env, workspaceId, outputUri.fsPath, signal);
+
+				progress.report({ message: "Adding VS Code logs..." });
+				await appendVsCodeLogs(
```
mtojek, line 262: `appendVsCodeLogs` is best-effort internally (catches errors and returns), but there's no try/catch here. If something unexpected throws (out of memory), it'll propagate up and the user loses the entire bundle, even though the original zip already exists on disk

EDIT:

or make sure `appendVsCodeLogs` truly catches everything

src/core/supportBundleLogs.ts
```diff
@@ -0,0 +1,149 @@
...
+import { type Logger } from "../logging/logger";
+import { renameWithRetry } from "../util";
+
+export interface LogSources {
+	remoteSshLogPath?: string;
+	proxyLogDir?: string;
+	extensionLogDir?: string;
+}
+
+const LOG_MAX_AGE_MS = 3 * 24 * 60 * 60 * 1000;
```
mtojek, line 17: The PR description says `SshProcessMonitor` keeps logs up to 7 days / 20 files. Why is the cutoff 3 days instead of 7? If this is intentional, a comment explaining the choice would help. If it's meant to match rotation, it should be 7 days.

src/core/supportBundleLogs.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+async function collectLogFiles(
+	sources: LogSources,
+	logger: Logger,
+): Promise<Map<string, Uint8Array>> {
+	const files = new Map<string, Uint8Array>();
+
+	if (sources.remoteSshLogPath) {
+		try {
+			files.set(
```
mtojek, line 68: is it intentional to skip `collectDirFiles` & `LOG_MAX_AGE_MS` here?

test/unit/core/supportBundleLogs.test.ts
```diff
@@ -0,0 +1,215 @@
...
+		expect(entries["vscode-logs/proxy/coder-ssh-recent.log"]).toBe("proxy");
+	});
+
+	it("does not touch the zip when no logs are found", async () => {
+		const zipPath = await makeBundle();
+		const before = await fs.stat(zipPath);
+
+		await appendVsCodeLogs(zipPath, {}, logger);
+
+		expect((await fs.stat(zipPath)).mtimeMs).toBe(before.mtimeMs);
```
mtojek: hm... you didn't want to rely on `Buffer.compare(before, after)`?

mtojek (commented):
Is it worth adding another zip test with 50+ files?

src/core/supportBundleLogs.ts
EhabY: I'll add a comment, but the idea is that 7 days is a big amount of logs, 3 days seems more than enough to gain context (especially since the rest of the bundle is has a much lower window)

src/core/supportBundleLogs.ts
EhabY: Yes because this is only set for an active session, so we don't care about the mtime

test/unit/core/supportBundleLogs.test.ts
EhabY: This is to ensure that the file was touched, regardless of the content. I could check both I suppose

src/core/supportBundleLogs.ts
mtojek: gotcha 👍

test/unit/core/supportBundleLogs.test.ts
mtojek: I know there are some controversies around mtime, hence raised the suggestion: [link](https://apenwarr.ca/log/20181113)

src/commands.ts
EhabY: Done

test/unit/core/supportBundleLogs.test.ts
EhabY: Ah, this is in the tests and purely on files so should be good. I've added checks for both tho!

mtojek (approved):
Ship it!

PR #917, opened 2026-04-23
feat: enhance parseRemoteAuthority to handle domains containing "--"

ntoofu:
### Problem

As explained in [CONTRIBUTING.md](https://github.com/coder/vscode-coder/blob/98048f2ba65bed9762b00b47d989e1daf91d37f6/CONTRIBUTING.md), 

> The host name takes the format coder-vscode.<domain>--<username>--<workspace>.

The parser splits on "--" to extract each component, but domain labels may contain "--".
In practice, IDNA/Punycode-encoded domain labels use `xn--` as the ACE prefix.
So the naive split("--") misparses hostnames with such labels, and causes errors like the following image.

Screenshot 2026-04-23 02-50-34
                                                                   
For example, a Coder deployment `coder.xn--eckwd4c7cu47r2wf.jp` would produce an SSH host of `coder-vscode.coder.xn--eckwd4c7cu47r2wf.jp--yourname--your-workspace`.
When split on "--", this yields `["coder-vscode.coder.xn", "eckwd4c7cu47r2wf.jp", "yourname", "your-workspace"]` but the parser expects the first element to be the full hostname, so vscode-coder would try to open a workspace `yourname` whose owner is `eckwd4c7cu47r2wf.jp` and fails.

### Solution

Added a reassembly step after the initial split("--").
The algorithm scans from the front of the split result:

- The first segment is always a part of a hostname.
- If the second segment contains a dot, that must be a fragmented part of a hostname. So merge those two segments back with "--". 
- Because more than one domain labels may contain "--", repeat this merging process until a dot does not appear in the second segment. 

This works because [Coder usernames never contain dots](https://github.com/coder/coder/blob/075face3cbf95e99f77c7081b49c28a436009ad6/codersdk/name.go#L14),
so any dot-bearing segment after the split must be a hostname fragment.

Note: a TLD containing "--" would not be handled correctly because such hostname would fragment into segments the last of which does not contain dot, but no such TLD exists today.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -61,14 +61,30 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
 	// coder-vscode(--|.)<username>--<workspace>(--|.)<agent?>
 	// The agent can be omitted; the user will be prompted for it instead.
 	// Anything else is unrelated to Coder and can be ignored.
-	const parts = authorityParts[1].split("--");
+	const rawParts = authorityParts[1].split("--");
 	if (
-		parts.length <= 1 ||
-		(parts[0] !== AuthorityPrefix &&
-			!parts[0].startsWith(`${AuthorityPrefix}.`))
+		rawParts.length <= 1 ||
+		(rawParts[0] !== AuthorityPrefix &&
+			!rawParts[0].startsWith(`${AuthorityPrefix}.`))
 	) {
 		return null;
 	}
+
+	// DNS labels may legally contain "--" (Punycode labels like "xn--{encoded}"
+	// are a common example), which breaks a naive split.  Coder usernames and
+	// workspace names follow a slug pattern that never contains dots, so a
+	// dot-bearing segment after splitting on "--" must be a hostname fragment
+	// rather than a username.  We reassemble from the front: while the next
+	// segment contains a dot, merge it back into the accumulated prefix.
+	let reassemblingParts = rawParts;
+	while (reassemblingParts.length >= 2) {
+		if (reassemblingParts[1].includes(".")) {
+			reassemblingParts = [reassemblingParts.slice(0, 2).join("--"), ...reassemblingParts.slice(2)];
+		} else {
+			break;
+		}
+	}
+	const parts = reassemblingParts;
 
 	// It has the proper prefix, so this is probably a Coder host name.
 	// Validate the SSH host name.  Including the prefix, we expect at least
--- a/test/unit/util.test.ts
+++ b/test/unit/util.test.ts
@@ -91,6 +91,17 @@ describe("parseRemoteAuthority", () => {
 			username: "foo",
 			workspace: "bar",
 		});
+		expect(
+			parseRemoteAuthority(
+				"vscode://ssh-remote+coder-vscode.dev.coder.xn--eckwd4c7cu47r2wf.jp--foo--bar",
+			),
+		).toStrictEqual({
+			agent: "",
+			sshHost: "coder-vscode.dev.coder.xn--eckwd4c7cu47r2wf.jp--foo--bar",
+			safeHostname: "dev.coder.xn--eckwd4c7cu47r2wf.jp",
+			username: "foo",
+			workspace: "bar",
+		});
 	});
 });
 


EhabY:
Nice fix for the punycode case. Before this lands, one realistic case still misparses, and a slightly simpler shape closes it.

### Where can `--` appear in a hostname?

Only two places:

1. As a schema separator (`coder-vscode.<host>--<owner>--<workspace>--<agent>`).
2. Inside a punycode label (`xn--<encoded>`) somewhere in the domain.

Coder's `UsernameValidRegex` forbids `--` in usernames, workspaces, and agents, and `xn--` is the only IDN ACE prefix IANA has allocated. That's the whole space.

### Apex punycode still misparses

A deployment served at the apex of an internationalized domain (e.g. `https://xn--p1ai/`) produces `coder-vscode.xn--p1ai--owner--ws`. Splitting gives `["coder-vscode.xn", "p1ai", "owner", "ws"]`; `parts[1]` is `p1ai` with no dot, so the merge loop exits immediately and `p1ai` becomes the username. Rare but real.

### Simpler parser

Scan forward, growing the prefix one segment at a time, and accept the first boundary where the prefix is valid and the remaining tail is a valid `<owner>--<workspace>[--<agent>]` or `<owner>--<workspace>[.<agent>]`. A valid prefix is `coder-vscode` or `coder-vscode.<...>` and does not end in `.xn` (which is the only way `--` can legally appear inside a domain, so ending in `.xn` means we cut a punycode label).

```ts
const SLUG = /^[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*$/;

function isValidPrefix(prefix: string): boolean {
  if (prefix === AuthorityPrefix) return true;
  return (
    prefix.startsWith(`${AuthorityPrefix}.`) && !prefix.endsWith(".xn")
  );
}

  const sshHost = authority.split("+")[1] ?? "";
  if (
    sshHost !== AuthorityPrefix &&
    !sshHost.startsWith(`${AuthorityPrefix}--`) &&
    !sshHost.startsWith(`${AuthorityPrefix}.`)
  ) {
    return null;
  }

  const parts = sshHost.split("--");

  for (let i = 1; i <= parts.length - 2; i++) {
    const prefix = parts.slice(0, i).join("--");
    if (!isValidPrefix(prefix)) continue;
    const tail = parts.slice(i);

    // 4-part: <prefix>--<owner>--<workspace>--<agent>
    if (tail.length === 3 && tail.every((p) => SLUG.test(p))) {
      return build(sshHost, prefix, tail[0], tail[1], tail[2]);
    }

    // 3-part: <prefix>--<owner>--<workspace>[.<agent>]
    if (tail.length === 2 && SLUG.test(tail[0])) {
      const [workspace, agent = ""] = tail[1].split(/\.(.+)/);
      if (SLUG.test(workspace) && (agent === "" || SLUG.test(agent))) {
        return build(sshHost, prefix, tail[0], workspace, agent);
      }
    }
  }

  throw new Error("Invalid Coder SSH authority");
}
```

### Every existing case parses identically

| Input | Result |
|---|---|
| `coder-vscode--foo--bar` | 3-part, prefix `coder-vscode` |
| `coder-vscode--foo--bar--baz` | 4-part, agent `baz` |
| `coder-vscode.dev.coder.com--foo--bar` | 3-part |
| `coder-vscode.dev.coder.com--foo--bar--baz` | 4-part wins (prefix valid, no `.xn`) |
| `coder-vscode.dev.coder.com--foo--bar.baz` | 3-part with dot-agent |
| `coder-vscode.coder.xn--eckwd4c7cu47r2wf.jp--foo--bar` (this PR's case) | i=1 prefix invalid (ends `.xn`); i=2 accepts |
| `coder-vscode.xn--p1ai--owner--ws` (apex punycode, currently broken) | i=1 prefix invalid (ends `.xn`); i=2 accepts |
| `coder-vscode.xn--abc.xn--def.com--owner--ws` (multiple punycode labels) | i=1 and i=2 invalid; i=3 accepts |

### False negatives

The only input this can reject incorrectly is a deployment whose domain ends in a label literally named `xn` (e.g. `something.xn`). IANA reserves the `xn--` space, `url.domainToASCII` does not produce bare `xn` followed by separator content, and no real TLD or registrable label of `xn` alone exists. So the failure mode is a label that essentially cannot legally be reached.

### Side benefit

The slug regex also rejects empty names and names ending in `-` that the current parser passes through, so input validation tightens up at the same time.

Happy to push this as a follow-up commit on this branch.

ntoofu:
@EhabY 

Thanks for the detailed analysis and the proposed fix! I agree on your proposal.

I wasn't aware that Punycode could also appear in TLDs (e.g. `.xn--p1ai`). You're absolutely right that the apex-punycode case misparses with my implementation. Sorry for missing that.                                                

There are many edge cases. The RFC5891 explicitly restricts only positions 3-4 being `--` for non-ACE labels, which means `--` in other positions within a label is technically valid. So a label like `foo--bar` can legally exist outside of Punycode. Moreover, single label may contain multiple `--` (e.g. `foo--bar--baz`). 

I don't think there's a general parsing strategy that handles all possible uses of `--` in DNS labels. So limiting the fix to Punycode labels which covers the most realistic cases seems like the right scope.

Please go ahead and push if you're still willing to!

EhabY:
Closing this in favor of https://github.com/coder/vscode-coder/pull/930

Issue #548, opened 2025-07-08
Support multiple windows to the same workspace


matifali:
Hey Team!

VS Code plugin - bug / feature req

Replication Steps:
• VS Code window open connected to workspace X
• Open a new VS Code window
• Navigate to Coder plugin and attempt to launch the same workspace into the new VS Code window
• It just brings the other window into focus
Impact is mostly user experience, as there is a workaround to use the command pallet to remote connect, but feels like there should be an option to "Start connection in new window" or whatever

[Slack Message](https://codercom.slack.com/archives/C057CGXNX6G/p1751962123138539?thread_ts=1751962123.138539&cid=C057CGXNX6G)

EhabY:
Already fixed by #918

PR #918, opened 2026-04-27
feat: detect when multiple windows open the same workspace

### LINKED ISSUE #783: Option to open workspace in new window
Right now when you use the coder remote extension if you have a window open with a session to a workspace say for project A, and you want to open another window to the same workspace for Project B, if you click the connect button it just brings the Project A window forward.

Something like the official VSCode remote extension would be better, it has an option to open the remote in the current window or launch a new window. Importantly launching a new window works even if it's a new window to the same remote.

EhabY:
## Summary

Adds `WindowBroadcast<T>`, a generic typed pub/sub channel built on top of `SecretStorage.onDidChange`, and uses it to detect when more than one VS Code window is connected to the same Coder workspace.

Closes #783, #548

## How it works

When a user clicks **Connect**, the extension fires `openFolder()` immediately and sends a PING through `SecretStorage` in parallel. If another window responds with a PONG, the clicking window shows a non-blocking notification with two choices:

- **Duplicate Window**: tells the responding window to run `duplicateWorkspaceInNewWindow`, preserving its tabs, panels, and remote connection.
- **Open Without Folder**: opens a new window connected to the remote with no folder.

The whole chain is fire-and-forget. `openWorkspace()` never blocks on it, so if no other window is listening the behavior is identical to before.

## Layout

| File | Role |
|------|------|
| `src/ipc/windowBroadcast.ts` | Generic typed pub/sub primitive over a single SecretStorage key |
| `src/workspace/duplicateWorkspaceIpc.ts` | PING / PONG / DUPLICATE protocol used by the open-workspace flow |
| `src/oauth/oauthCallback.ts` | Cross-window OAuth redirect signal, also built on `WindowBroadcast` |
| `src/commands.ts` | Sends PING in parallel with `openFolder()` and shows the notification |
| `src/extension.ts` | Responds to PINGs and runs `duplicateWorkspaceInNewWindow` when targeted |

Each consumer of `WindowBroadcast` defines its message shape with a Zod schema, so invalid or stale data on the storage key is dropped before reaching handlers.

## Notes

- `WindowBroadcast` is the only generic piece. Specializations live with their feature (OAuth signal in `oauth/`, workspace duplicate detection in `workspace/`) rather than in a shared IPC bucket.
- The notification uses non-modal `showInformationMessage`, which VS Code may dismiss without resolving, so the chain must not be awaited from the open path.
- The notification is shown in the window where the user clicked, so it always appears in a predictable place regardless of which window VS Code routes focus to.


src/ipc/windowBroadcast.ts
```diff
@@ -0,0 +1,43 @@
...
+export class WindowBroadcast<T> {
+	constructor(
+		private readonly secrets: SecretStorage,
+		private readonly key: string,
+		private readonly schema: ZodType<T>,
+		private readonly logger: Logger,
+	) {}
+
+	async send(msg: T): Promise<void> {
+		await this.secrets.store(this.key, JSON.stringify(msg));
```
mtojek, line 20: what happens if function `send()` is called twice quickly, is there any need to implement mutual exclusion?

src/ipc/windowBroadcast.ts
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	onReceive(handler: (msg: T) => void | Promise<void>): Disposable {
+		return this.secrets.onDidChange(async (e) => {
```
mtojek, line 24: hm.. can window receive its own messages?

src/workspace/duplicateWorkspaceIpc.ts
```diff
@@ -0,0 +1,145 @@
...
+export type PingMessage = z.infer<typeof PingMessageSchema>;
+export type PongMessage = z.infer<typeof PongMessageSchema>;
+export type DuplicateMessage = z.infer<typeof DuplicateMessageSchema>;
+export type RequestMessage = z.infer<typeof RequestMessageSchema>;
+
+/**
+ * Cross-window protocol for the open-workspace flow:
+ *   PING      "anyone connected to this authority?"
+ *   PONG      "yes, with this sessionId"
+ *   DUPLICATE "sessionId, please duplicate yourself"
```
mtojek, line 47: very helpful 👍

src/workspace/duplicateWorkspaceIpc.ts
```diff
@@ -0,0 +1,145 @@
...
+				clearTimeout(timer);
+				listener.dispose();
+				resolve(result);
+			};
+
+			const listener = this.responses.onReceive((msg) => {
+				if (msg.id === id) {
+					settle(msg);
+				}
+			});
```
mtojek, line 98: we just need one PONG, right? no need to collect responses from other windows

src/workspace/duplicateWorkspaceIpc.ts
```diff
@@ -0,0 +1,145 @@
+import crypto from "node:crypto";
+import { z } from "zod";
+
+import { WindowBroadcast } from "../ipc/windowBroadcast";
+
+import type { Disposable, SecretStorage } from "vscode";
+
+import type { Logger } from "../logging/logger";
+
+const MESSAGE_MAX_AGE_MS = 5000;
```
mtojek: hmm... why exactly is max age 5s and ping timeout 1s?

src/workspace/duplicateWorkspaceIpc.ts
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Cross-window protocol for the open-workspace flow:
+ *   PING      "anyone connected to this authority?"
+ *   PONG      "yes, with this sessionId"
+ *   DUPLICATE "sessionId, please duplicate yourself"
+ *
+ * The sender shows the prompt locally on first PONG. If the user picks
+ * Duplicate, it sends DUPLICATE targeted at the responder's sessionId.
+ */
+export class DuplicateWorkspaceIpc {
```
mtojek, line 52: Should it implement `Disposable` ?

src/ipc/windowBroadcast.ts
EhabY: No mutex is needed here because this is all best-effort and the secret storage should be synchronized internally (maybe if we send two messages at the same exact time they collapse into one for the same key?). In either case, this is handled well

src/ipc/windowBroadcast.ts
EhabY: Yes and this works fine, like if I have one window open, and I click "play" on the same workspace then it'll PING -> PONG itself which would show the prompt!

src/workspace/duplicateWorkspaceIpc.ts
EhabY: Honestly it's just an arbitrary wait time so we don't keep waiting forever. PING <-> PONG should be fast and that's why it has a 1s timeout.

The 5s timeout here actually seems useless since this is only checked AFTER `onDidChange` is triggered (so it's guaranteed to be fresh). I'll remove the 5s timeout here

src/workspace/duplicateWorkspaceIpc.ts
EhabY: ❤️

src/workspace/duplicateWorkspaceIpc.ts
EhabY: There is nothing to dispose here, no open resources

src/workspace/duplicateWorkspaceIpc.ts
EhabY: Indeed, we just acknowledge the first one!

src/workspace/duplicateWorkspaceIpc.ts
mtojek: Fair enough for now. Though requests and responses are WindowBroadcast instances that hold SecretStorage references - if those ever need cleanup, this class would be the natural place. Not blocking, just something to keep in mind.

mtojek (approved):

src/workspace/duplicateWorkspaceIpc.ts
EhabY: Oh yeah for sure, I'll keep this in mind!

Issue #901, opened 2026-04-17
Telemetry: LocalJsonlSink with buffering, rotation, and cleanup


EhabY:
Part of the Telemetry Phase A rollout. See the RFC in Linear: [AIGOV-154](https://linear.app/codercom/issue/AIGOV-154/telemetry-audit-and-observability-instrumentation).

## Scope

* Implement `LocalJsonlSink` in `src/telemetry/sinks/localJsonlSink.ts` implementing the `TelemetrySink` interface from [AIGOV-243](https://linear.app/codercom/issue/AIGOV-243/telemetry-core-service-trace-and-servicecontainer-wiring).
* File layout under `globalStorageUri/telemetry/` with daily files `telemetry-YYYY-MM-DD.jsonl`.
* Buffer events in memory, flush every 15 seconds or when the buffer reaches 100 events. Both thresholds are configurable.
* Hard buffer cap of 500 events (configurable). When exceeded, drop oldest events and log a warning.
* Per-file size cap of 5 MB (configurable). When the current day's file exceeds the cap, create numbered segments (`telemetry-YYYY-MM-DD.1.jsonl`, `.2.jsonl`, and so on). Use `stat` before each flush, cached for the flush interval.
* Serialize events to snake_case JSON so they match the Phase B wire format without transformation.
* Use `fs.appendFile` so no file handles remain open between flushes.
* Write failures are logged via `Logger` but never thrown.
* `dispose()` flushes before returning.
* Cleanup on construction (runs at extension activation): delete files older than `maxAgeDays` (default 30), then remove oldest files if total size exceeds `maxSizeBytes` (default 100 MB). Same pattern as the SSH process log cleanup in `sshProcess.ts`.

## Tests

* Mock `fs/promises` with `memfs`, consistent with existing test patterns.
* Events are readable as valid JSONL after flush.
* Old files are removed on construction.
* Oversized storage is pruned oldest-first.
* Per-file cap creates a numbered segment when exceeded.
* Buffer overflow drops oldest events and logs a warning.
* `dispose()` persists all buffered events.

## Out of scope

* No instrumentation sites.
* No export command (separate ticket).

Depends on [AIGOV-243](https://linear.app/codercom/issue/AIGOV-243/telemetry-core-service-trace-and-servicecontainer-wiring).

PR #919, opened 2026-04-27
feat: add telemetry service infrastructure (Phase A core)

### LINKED ISSUE #900: Telemetry: core service, Trace, and ServiceContainer wiring
Part of the Telemetry Phase A rollout. See the RFC in Linear: AIGOV-154.

## Scope

- Define `TelemetryEvent` and `TelemetrySink` in `src/telemetry/telemetryEvent.ts`. Shape per the RFC: `eventId`, `eventName`, `timestamp`, `eventSequence`, `context`, `properties`, `measurements`, optional `traceId`, optional `error`.
- Implement `TelemetryService` in `src/telemetry/telemetryService.ts` with `log`, `logError`, `time`, `trace`, `setDeploymentUrl`, `dispose`. Auto-injects the `context` block (extension version, machine ID, session ID, deployment URL, `os_type`, `os_version`, `host_arch`, `platform_type`, `platform_version`) and a monotonic `eventSequence`. Echoes every event to `Logger.trace` so developers can inspect it in the output channel.
- Implement `Trace` in the same file. `trace.phase(name, fn, props)` emits a child event carrying the parent's `traceId`.
- Register the service in `ServiceContainer` and inject it into constructors the same way `Logger` is today.
- Add the `coder.telemetry.level` setting in `package.json` with enum values `off` and `local` (default `local`). `deployment` is reserved for Phase B and should not be accepted yet.
- Wire `DeploymentManager` to call `telemetry.setDeploymentUrl(url)` once the deployment is known. Events emitted before that point carry an empty `deploymentUrl`, which is expected.

## Tests

- Unit tests with an in-memory `TestSink` covering `log`, `logError`, `time` on success and on failure, `trace` on success and when a middle phase throws, context auto-injection, and `eventSequence` monotonicity.
- Vitest fake timers for deterministic `durationMs` assertions.

## Out of scope

- No sinks beyond `TestSink`. `LocalJsonlSink` is a separate ticket.
- No instrumentation sites are touched here.

Blocks most of the Telemetry Phase A work.

EhabY:
Lays the spine for Phase A telemetry per the [RFC](https://linear.app/codercom/issue/AIGOV-154). No real sinks ship in this PR. `LocalJsonlSink` and the instrumentation sites land in follow-up tickets.

What this adds:
- `TelemetryEvent`, `TelemetryContext`, `SessionContext`, and `TelemetrySink` interfaces in `src/telemetry/event.ts`, plus `buildSession` and `buildErrorBlock` helpers
- `TelemetryService` with `log`/`logError`/`time`/`trace`/`setDeploymentUrl`/`dispose`, including auto-injected context (extension version, machine/session ID, OS, host, platform, deployment URL) and a monotonic `eventSequence`. `time`/`trace`/`Span.phase` accept caller-supplied `measurements` alongside the framework-set `durationMs`.
- `Span` correlation handle (`src/telemetry/span.ts`): every timed event carries a `traceId`; phase events also carry `parentEventId`. Phase eventNames compose hierarchically (e.g. `remote.setup.workspace_lookup`). `Span.phase` callbacks receive a child `Span`, so grandchildren are just `.phase()` recursion. Maps directly to OTel `trace_id`/`span_id`/`parent_span_id` for OTLP/Traces export.
- `coder.telemetry.level` setting with values `off` and `local` (default `local`), tagged so it surfaces under VS Code's Telemetry settings section. `deployment` is reserved for Phase B.
- `ServiceContainer` registers the service; `DeploymentManager` calls `setDeploymentUrl` from `setDeployment` and resets it from `clearDeployment`.

Design choices worth flagging:
- **Sinks are static and self-gating.** Sinks are constructed once at activation. The service holds a global kill switch on `coder.telemetry.level === "off"`. Each sink decides whether to emit based on its own gating signal, which makes Phase B (`CoderServerSink` gated on `deployment`) and Phase C (external sink gated on VS Code's built-in `telemetry.telemetryLevel`) drop in without service changes.
- **Reactive level changes.** `watchConfigurationChanges` passes a `Map<setting, newValue>` to its callback so consumers don't re-read the config. Flipping to `off` flushes all sinks before the kill switch trips, so in-flight buffered events are not lost. The new value is validated by an `isTelemetryLevel` type guard.
- **Errors never reach callers.** All emission goes through `#safeEmit`, which catches and logs any telemetry-internal failure via the regular logger. Per-sink `try/catch` around `sink.write` isolates write failures further. A throwing sink (or any other telemetry bug) cannot affect the user fn's return value or its error.
- **Phase name validation.** Phase names containing `.` collide with the hierarchical path separator; the producer warns and sanitizes (replaces `.` with `_`), never throws — a typo in instrumentation cannot break a feature.
- **Readonly event shape.** `TelemetryEvent` and its nested records are `readonly` at the type level so a sink can't mutate the event in TypeScript without an explicit `as` cast.
- **`dispose()` returns `Promise<void>`.** VS Code does not await deactivation, so flushes during shutdown are best-effort.

Closes #900


src/telemetry/service.ts
```diff
@@ -0,0 +1,193 @@
...
+	#emit(
+		eventName: string,
+		properties: Record<string, string>,
+		measurements: Record<string, number>,
+		traceId?: string,
+		error?: unknown,
+	): void {
+		if (this.#level === "off") {
+			return;
+		}
```
mtojek: isn't it a double guard in some cases?

src/telemetry/service.ts
```diff
@@ -0,0 +1,193 @@
...
+} from "./event";
+import { Trace } from "./trace";
+
+type TelemetryLevel = "off" | "local";
+
+const TELEMETRY_LEVEL_SETTING = "coder.telemetry.level";
+
+const NOOP_TRACE = new Trace("", "", () => {
+	// Off-mode tracer; no events are emitted.
+});
```
mtojek: Not sure what is the best way to represent it in TS, but would rather see NOOP_TRACE defined inside of Trace? `Trace.NoOp`

src/core/container.ts
```diff
@@ -81,6 +83,7 @@ export class ServiceContainer implements vscode.Disposable {
 			context.secrets,
 			this.logger,
 		);
+		this.telemetryService = new TelemetryService(context, [], this.logger);
```
mtojek, line 93: no sink?

src/core/container.ts
```diff
@@ -123,12 +126,18 @@ export class ServiceContainer implements vscode.Disposable {
...
+
 	/**
 	 * Dispose of all services and clean up resources.
 	 */
 	dispose(): void {
 		this.contextManager.dispose();
 		this.logger.dispose();
 		this.loginCoordinator.dispose();
+		// Fire-and-forget: VS Code does not await deactivation.
+		void this.telemetryService.dispose();
```
mtojek: Hm... what are the implications of not waiting? Will we lose any tracing data by not flushing them in time?

package.json
```diff
@@ -189,6 +189,19 @@
...
+					"type": "string",
+					"enum": [
+						"off",
+						"local"
+					],
+					"markdownEnumDescriptions": [
+						"Disable telemetry collection.",
+						"Record events on this machine only."
+					],
+					"default": "local"
```
mtojek: will user be notified about enabled telemetry by default?

src/telemetry/trace.ts
```diff
@@ -0,0 +1,24 @@
+import { emitTimed, type EmitFn } from "./emit";
+
+/** Correlation handle: parent and all child phase events share one `traceId`. */
+export class Trace {
```
mtojek: Now I'm thinking about aligning names with OTel domain. `Trace` -> `Span`, `phase` -> `span`/`childspan`

unless this in on purpose 👍

src/telemetry/event.ts
```diff
@@ -0,0 +1,91 @@
...
+	const rawCode = (value as { code?: unknown } | null | undefined)?.code;
+	if (typeof rawCode === "string" || typeof rawCode === "number") {
+		block.code = String(rawCode);
+	}
+	return block;
+}
+
+// Node uses "win32" on Windows; OTel's os.type is "windows".
+function detectOsType(): string {
+	return process.platform === "win32" ? "windows" : process.platform;
```
mtojek, line 106: should we align with OTel? `os.type` - `uname`

src/core/container.ts
```diff
@@ -29,6 +30,7 @@ export class ServiceContainer implements vscode.Disposable {
 	private readonly loginCoordinator: LoginCoordinator;
 	private readonly duplicateWorkspaceIpc: DuplicateWorkspaceIpc;
 	private readonly oauthCallback: OAuthCallback;
+	private readonly telemetryService: TelemetryService;
```
mtojek, line 35: if telemetry is off, should we inject it?

src/deployment/deploymentManager.ts
```diff
@@ -155,6 +159,7 @@ export class DeploymentManager implements vscode.Disposable {
 		this.#authListenerDisposable?.dispose();
 		this.#authListenerDisposable = undefined;
 		this.#deployment = null;
+		this.telemetryService.setDeploymentUrl("");
```
mtojek, line 162: consistency: `secretsManager.setCurrentDeployment` - should we align?

src/telemetry/event.ts
```diff
@@ -0,0 +1,91 @@
...
+	write(event: TelemetryEvent): void;
+	flush(): Promise<void>;
+	dispose(): Promise<void>;
+}
+
+/** Build session attributes. `extensionVersion` falls back to `"unknown"`. */
+export function buildSession(ctx: vscode.ExtensionContext): SessionContext {
+	const packageJson = ctx.extension.packageJSON as { version?: unknown };
+	const extensionVersion =
+		typeof packageJson.version === "string" ? packageJson.version : "unknown";
```
mtojek, line 76: Maybe comment on when it is unknown

src/core/container.ts
EhabY: Yes, this was a deliberate decision because the setting can be switched on or off at any time. So I think we should always have the object ready and gate internally

src/core/container.ts
EhabY: Not yet, see https://github.com/coder/vscode-coder/issues/901

src/core/container.ts
EhabY: Yes, any data that is not yet flushed would be lost. There is not much we can do here since VS Code shuts down without waiting anyway :/

src/deployment/deploymentManager.ts
EhabY: There we pass the whole deployment object (URL + safeHostname). While here we only pass the URL

package.json
EhabY: Currently no, but I think we can surface this info somewhere when it's almost complete 🤔

src/telemetry/event.ts
EhabY: Added a comment!

src/telemetry/service.ts
EhabY: Yes that is cleaner actually, done!

src/telemetry/service.ts
EhabY: True... I'll make it guard internally only then!

src/telemetry/trace.ts
EhabY: I just looked it up but a `Span` is created only via a `Tracer`, has its own `spanId`, and links to its parent through `parentSpanId`. Our class is a callback handle whose children share a `traceId` and a `phase` property, so it's really a scope/context object, not a span. Renaming would make it seem like it's the same OTel span.

If we want full alignment then we can add `spanId`/`parentSpanId` to `TelemetryEvent`, generate a span id per emit, rename `phase()` to `startChild()`

See [OTel Tracing API](https://opentelemetry.io/docs/specs/otel/trace/api/), [OTel data model](https://opentelemetry.io/docs/concepts/signals/traces/).

src/telemetry/event.ts
EhabY: Already aligned with [OTel `os.type`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/os/) for everything VS Code supports:

- `linux` -> `linux`
- `darwin` -> `darwin`
- `win32` -> `windows`

Only gap is Node's `sunos` -> OTel `solaris`, but VS Code doesn't run on Solaris so it's unreachable

src/telemetry/service.ts
EhabY: Done

src/telemetry/trace.ts
EhabY: Honestly though I think we can produce OTel-compatible output via an adapter at the sink/export layer. Mapping:

- `eventName` -> span `name`
- `traceId` -> `traceId`
- `eventId` -> `spanId` (truncate to 16 hex chars for strict OTel byte-size)
- `properties` -> `attributes`
- `timestamp` + `measurements.durationMs` -> start/end
- `error` -> status
- `context` (session) -> resource attributes

For `parentSpanId`, the convention is **root events have `eventId === traceId`**. `service.trace` already pre-generates the `traceId`, we just thread it through `EmitFn` as an optional `eventIdOverride` so the root event reuses it instead of getting a fresh UUID. Phase events keep their auto-generated `eventId`. That's a ~5-line producer change (one optional param on `EmitFn` + `emitTimed` forwards it + `service.trace` passes `traceId` as the override + `#emit` uses it when present).

Adapter then becomes two lines:

```ts
const isRoot = e.eventId === e.traceId;
return {
  spanId: e.eventId,
  parentSpanId: isRoot ? undefined : e.traceId,
  // ...rest from the mapping above
};
```

src/telemetry/event.ts
mtojek: definitely skip solaris :)

src/telemetry/trace.ts
mtojek: I like this!

package.json
mtojek: Gotcha 👍 Let's not forget about it.

mtojek (approved):
Thanks Ehab! I'm good, but you can try codex agents against it too :)

src/core/container.ts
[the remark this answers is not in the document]
EhabY: Done. `logger.dispose()` now runs last in container teardown

src/telemetry/service.ts
```diff
@@ -0,0 +1,193 @@
...
+			event.error = buildErrorBlock(error);
+		}
+
+		const currentOrder = LEVEL_ORDER[this.#level];
+		for (const sink of this.sinks) {
+			if (currentOrder < LEVEL_ORDER[sink.minLevel]) {
+				continue;
+			}
+			try {
+				sink.write(event);
```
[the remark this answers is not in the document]
EhabY: Done at the type level. All `TelemetryEvent` fields are `readonly`. A sink author writing TypeScript cannot mutate any field, including nested ones, without an explicit `as` cast (outside the threat model for first-party sinks).

src/telemetry/emit.ts
```diff
@@ -0,0 +1,35 @@
...
+	eventName: string,
+	fn: () => Promise<T>,
+	properties: Record<string, string>,
+	traceId?: string,
+): Promise<T> {
+	const start = performance.now();
+	const send = (result: "success" | "error", error?: unknown): void =>
+		emit(
+			eventName,
+			{ ...properties, result },
```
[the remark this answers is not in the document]
EhabY: `phase` has been removed from the properties, so all is good there. As for `result`, we could add some type checks to prevent it but we would have to thread this type everywhere. Instead, we now overwrite it and have a test for this (also added documentation, same for `durationMs` measurement)

test/unit/telemetry/service.test.ts
```diff
@@ -0,0 +1,311 @@
...
+
+			h.service.log("second");
+			expect(h.sink.events).toHaveLength(1);
+
+			h.config.set("coder.telemetry.level", "local");
+			h.service.log("third");
+			expect(h.sink.events).toHaveLength(2);
+		});
+
+		it("treats unknown values (e.g. future deployment) as local", () => {
```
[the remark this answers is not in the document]
EhabY: Done. Harness now starts at `"off"` so `"deployment"` is a real transition through normalization.

package.json
```diff
@@ -189,6 +189,19 @@
...
 					],
 					"default": "basic"
+				},
+				"coder.telemetry.level": {
+					"markdownDescription": "Controls Coder extension telemetry collection. Used to diagnose extension issues.",
+					"type": "string",
+					"enum": [
+						"off",
+						"local"
+					],
```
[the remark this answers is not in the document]
EhabY: The Phase A sink writes JSONL into the extension's storage directory. Nothing leaves the machine, which is functionally an output channel. VS Code's [custom telemetry guidance](https://code.visualstudio.com/api/extension-guides/telemetry#custom-telemetry-setting) says to respect `isTelemetryEnabled` for telemetry that is sent off-device, so the local sink is not gated on it.

Per the RFC, the later sinks split as:

- Phase B (`deployment`): goes to the user's own Coder server. Gated by `coder.telemetry.level = "deployment"` only, not VS Code's setting. Self-hosted Coder is not a third party in the sense VS Code's docs target, and the level enum is the explicit opt-in.
- Phase C (`external`): third-party. Uses `vscode.env.isTelemetryEnabled` directly via `TelemetryLogger`, not the `coder.telemetry.level` enum.

Tagged the setting with `"tags": ["telemetry"]` so it surfaces under VS Code's Telemetry settings section

test/unit/deployment/deploymentManager.test.ts
```diff
@@ -66,11 +66,16 @@ function createTestContext() {
 		validationMockClient as unknown as CoderApi,
 	);
 
+	const telemetryService = {
```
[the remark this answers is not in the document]
EhabY: Done. Two assertions added: `setDeployment` calls with the URL, `clearDeployment` calls with `""`. `telemetryService` is now exposed from `createTestContext`.

src/telemetry/event.ts
```diff
@@ -0,0 +1,97 @@
+import * as os from "node:os";
+import * as vscode from "vscode";
+
+import { toError } from "../error/errorUtils";
+
+/** Telemetry level, mirrors `coder.telemetry.level`. Ordered: off < local. */
+export type TelemetryLevel = "off" | "local";
+
+/** Session-stable resource attributes. Field names follow OTel conventions. */
```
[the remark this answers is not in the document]
EhabY: Done. Softened to "inspired by OTel resource attributes; camelCase TypeScript and not a 1:1 mapping."

src/telemetry/service.ts
```diff
@@ -0,0 +1,193 @@
...
+		ctx: vscode.ExtensionContext,
+		private readonly sinks: readonly TelemetrySink[],
+		private readonly logger: Logger,
+	) {
+		this.#session = buildSession(ctx);
+		this.#level = readLevel();
+		this.#configWatcher = watchConfigurationChanges(
+			[{ setting: TELEMETRY_LEVEL_SETTING, getValue: readLevel }],
+			(changes) => {
+				const newLevel = changes.get(TELEMETRY_LEVEL_SETTING) as TelemetryLevel;
```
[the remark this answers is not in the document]
EhabY: Done. Replaced the cast with an `isTelemetryLevel` type guard that only accepts `"off"` or `"local"`. Anything else is rejected and the change is ignored.

src/telemetry/emit.ts
```diff
@@ -0,0 +1,35 @@
...
+			{ durationMs: performance.now() - start },
+			traceId,
+			error,
+		);
+	try {
+		const value = await fn();
+		send("success");
+		return value;
+	} catch (err) {
+		send("error", err);
```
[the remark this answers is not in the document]
EhabY: Done structurally. All emission goes through `#safeEmit`, a top-level catch-all that logs any telemetry-internal failure via the regular logger and swallows it

src/telemetry/event.ts
```diff
@@ -0,0 +1,97 @@
...
+	readonly osType: string;
+	readonly osVersion: string;
+	readonly hostArch: string;
+	readonly platformType: string;
+	readonly platformVersion: string;
+}
+
+/** Per-event context: session attributes plus the current deployment URL. */
+export interface TelemetryContext extends SessionContext {
+	deploymentUrl: string;
```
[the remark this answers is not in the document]
EhabY: Done. Added `readonly` on `deploymentUrl`, and on the rest of `TelemetryEvent` while I was there.

src/telemetry/event.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/** Session-stable resource attributes. Field names follow OTel conventions. */
+export interface SessionContext {
+	readonly extensionVersion: string;
+	readonly machineId: string;
+	readonly sessionId: string;
+	readonly osType: string;
+	readonly osVersion: string;
+	readonly hostArch: string;
+	readonly platformType: string;
```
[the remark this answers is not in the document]
EhabY: Done. Renamed to `platformName` (will update the RFC)

src/telemetry/trace.ts
EhabY: Added `parentEventId` since this is the only way to track parent IDs esp when we have retries! (will update the RFC)

src/core/container.ts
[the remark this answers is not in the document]
EhabY: Fixed: `dispose()` is now async and awaits telemetry teardown inside try/finally so the logger always disposes last.

src/telemetry/service.ts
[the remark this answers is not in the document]
EhabY: Skipping since all sinks are internal TypeScript and `TelemetryEvent` is already `readonly` at the type level.

src/telemetry/emit.ts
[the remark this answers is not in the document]
EhabY: Fixed: public methods now take `CallerProperties` and `CallerMeasurements`, which reserve `result` and `durationMs` via `?: never` so passing them is a compile error.

test/unit/telemetry/service.test.ts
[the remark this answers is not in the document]
EhabY: The test now starts at `off` and sets the raw config to `deployment`, forcing a real transition through the watcher's normalization path.

package.json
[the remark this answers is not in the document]
EhabY: Fixed: `coder.telemetry.level` is no longer registered in `package.json`; the service still reads it (defaulting to `local`) but nothing user-visible ships until the first sink lands.

src/telemetry/trace.ts
EhabY: Skipping: `parentEventId` already encodes hierarchy and its absence identifies the root, so an OTel adapter can detect roots without the `eventId === traceId` tweak.

src/telemetry/span.ts
```diff
@@ -0,0 +1,31 @@
...
+	traceId: "",
+	eventId: "",
+	eventName: "",
+	phase<T>(
+		_phaseName: string,
+		fn: (span: Span) => Promise<T>,
+		_properties?: Record<string, string>,
+		_measurements?: Record<string, number>,
+	): Promise<T> {
+		return fn(this);
```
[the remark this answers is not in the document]
EhabY: Fixed: `NOOP_SPAN.phase` now calls `fn(NOOP_SPAN)` so detaching the method cannot strip the receiver.

test/unit/telemetry/service.test.ts
```diff
@@ -0,0 +1,470 @@
...
+	);
+	return { service, sink, config };
+}
+
+function makeService(sinks: TelemetrySink[]): TelemetryService {
+	new MockConfigurationProvider().set("coder.telemetry.level", "local");
+	return new TelemetryService(fakeContext(), sinks, createMockLogger());
+}
+
+describe("TelemetryService", () => {
```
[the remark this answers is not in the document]
EhabY: Skipping in tests: services hold a mock-vscode subscription that frees on process exit, and the production lifecycle owns disposal via `ServiceContainer`; tracking and disposing in `afterEach` would add boilerplate without catching anything that actually leaks.

src/telemetry/service.ts
```diff
@@ -0,0 +1,299 @@
...
+			context: { ...this.#session, deploymentUrl: this.#deploymentUrl },
+			properties: { ...properties },
+			measurements: { ...measurements },
+			...(traceId !== undefined && { traceId }),
+			...(parentEventId !== undefined && { parentEventId }),
+			...(error !== undefined && { error: buildErrorBlock(error) }),
+		};
+
+		const currentOrder = LEVEL_ORDER[this.#level];
+		for (const sink of this.sinks) {
```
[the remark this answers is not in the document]
EhabY: Fixed: `trace()` now captures the level at start and threads it through `SpanOptions` so all events in one trace are filtered against the same level, even if the user toggles mid-trace.

src/telemetry/span.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Parent span handle. Children's `eventName` composes as `${parent.eventName}.${phaseName}`;
+ * phase names must not contain `.`. Recurse via `phase` for grandchildren.
```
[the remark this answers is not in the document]
EhabY: Fixed: docstring now says dots in phase names are replaced with `_` after a warning rather than implying a hard constraint.

src/telemetry/service.ts
```diff
@@ -0,0 +1,299 @@
...
+	): void {
+		if (this.#level === "off") {
+			return;
+		}
+		this.#safeEmit(crypto.randomUUID(), eventName, properties, measurements, {
+			error,
+		});
+	}
+
+	/** Time `fn` and emit one event with `durationMs` and `result`. */
```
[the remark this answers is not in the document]
EhabY: Stale: `time()` was removed earlier in the branch, `trace()` is the only timed entry point now.

PR #920, opened 2026-04-27
fix: defer Workspace Build output channel creation until first write

EhabY:
- `TerminalOutputChannel`'s constructor previously called `createOutputChannel` and `show(true)` immediately, so the Output panel popped open every time we built a `WorkspaceStateMachine` — even on reconnects to an already-running workspace where no build logs would ever be written.
- The empty channel was disposed shortly afterward, but the Output panel itself stayed visible (VS Code's API has no way to programmatically close the Output panel).
- Defer both `createOutputChannel` and `show(true)` to the first `write()`. Reconnects to a running workspace never create the channel and never pop the panel; real builds behave exactly as before.

### DIFF
--- a/src/remote/terminalOutputChannel.ts
+++ b/src/remote/terminalOutputChannel.ts
@@ -1,20 +1,26 @@
 import stripAnsi from "strip-ansi";
 import * as vscode from "vscode";
 
-/** Adapts terminal-style output for a VS Code OutputChannel. Strips ANSI escape sequences and carriage returns. */
+/**
+ * Wraps a VS Code OutputChannel for terminal-style output, stripping ANSI
+ * escapes and carriage returns. The channel is created lazily on first write
+ * to avoid surfacing an empty pane in the Output dropdown when nothing is
+ * ever written.
+ */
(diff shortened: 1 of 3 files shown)


src/remote/terminalOutputChannel.ts
```diff
@@ -1,20 +1,25 @@
 import stripAnsi from "strip-ansi";
 import * as vscode from "vscode";
 
-/** Adapts terminal-style output for a VS Code OutputChannel. Strips ANSI escape sequences and carriage returns. */
+/**
+ * Adapts terminal-style output for a VS Code OutputChannel.
+ * Strips ANSI escape sequences and carriage returns;
+ * creates the underlying channel lazily on first write.
```
mtojek: nit: You can add more context explaining why it is essential or what behavior it is addressing.

mtojek (approved):

src/remote/terminalOutputChannel.ts
EhabY: Done!

PR #921, opened 2026-04-28
chore: bump dev deps and override @vscode-elements/elements to 2.5.1

EhabY:
- Bumps dev dependencies to their latest patches: `@tanstack/react-query`, `vite`, `typescript-eslint`, `vitest`, `@vitest/coverage-v8`, `jsdom`, `eslint-plugin-package-json`, and the `pnpm` package manager pin.
- Adds a pnpm `overrides` entry forcing `@vscode-elements/elements` to `^2.5.1`. The companion `@vscode-elements/react-elements` package has not had a release that bumps its underlying web components dep, so the override is the only way to pick up the 2.5.x bug fixes (Table runtime fix, single-select `onOptionClick` fix, SSR fixes) without forking the wrapper.
- Removes a few now-redundant `as` casts in `protocol.ts` and unit tests, surfaced by stricter inference in the updated `typescript-eslint` / `vitest` types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -628,10 +628,10 @@
 		"@types/ua-parser-js": "0.7.39",
 		"@types/vscode": "1.106.0",
 		"@types/ws": "^8.18.1",
-		"@typescript-eslint/eslint-plugin": "^8.58.2",
-		"@typescript-eslint/parser": "^8.58.2",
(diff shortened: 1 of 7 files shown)


mtojek (approved):
stamp

PR #922, opened 2026-04-29
feat: add custom workspace icons

jakehwll:
Updated workspace tree items to resolve app/workspace icon strings into VS Code URIs and restored the default logo fallback when the icon is missing or invalid.

* Added URL resolution for workspace template icons and app icons.
* Preserved the original default `logo-black.svg` / `logo-white.svg` fallback.
* Kept tree items using the default icon only when no valid custom icon is available.

image

### DIFF
--- a/src/workspace/workspacesProvider.ts
+++ b/src/workspace/workspacesProvider.ts
@@ -165,6 +165,7 @@ export class WorkspaceProvider
 				workspace,
 				this.getWorkspacesQuery === WorkspaceQuery.All,
 				showMetadata,
+				this.client.getAxiosInstance().defaults.baseURL,
 			);
 
 			return workspaceTreeItem;
@@ -258,12 +259,16 @@ export class WorkspaceProvider
 								// We need to do this for now because the reporting isn't super accurate
 								// yet.
 								appStatuses.push(
-									new AppStatusTreeItem({
-										id: status.id,
-										name: status.message,
-										command: app.command,
-										workspace_name: element.workspace.name,
-									}),
+									new AppStatusTreeItem(
+										{
+											id: status.id,
+											name: status.message,
+											command: app.command,
+											workspace_name: element.workspace.name,
+											icon: app.icon,
+										},
+										this.client.getAxiosInstance().defaults.baseURL,
+									),
 								);
 							}
 						}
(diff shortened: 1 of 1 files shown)


src/workspace/workspacesProvider.ts
```diff
@@ -258,12 +259,16 @@ export class WorkspaceProvider
...
-									}),
+									new AppStatusTreeItem(
+										{
+											id: status.id,
+											name: status.message,
+											command: app.command,
+											workspace_name: element.workspace.name,
+											icon: app.icon,
+										},
+										this.client.getAxiosInstance().defaults.baseURL,
```
EhabY, line 270: I think we should pass the resolved icon instead of the baseUrl here (esp. since it should always be defined). That way we do not leak abstraction into the `AppStatusTreeItem` which shouldn't be aware of those details

src/workspace/workspacesProvider.ts
```diff
@@ -358,6 +363,45 @@ class AgentMetadataTreeItem extends vscode.TreeItem {
...
+	}
+
+	const trimmed = icon.trim();
+	if (/^[a-zA-Z][a-zA-Z\d+\-.]*:/.test(trimmed)) {
+		return vscode.Uri.parse(trimmed);
+	}
+
+	if (!baseUrl) {
+		return undefined;
+	}
```
EhabY, line 381: This is not needed, the `URL` implements `WHATWG` resolution semantics, so:

```
new URL("https://x/y.svg", "https://base/").toString() // "https://x/y.svg"
new URL("/icon.svg",      "https://base/").toString() // "https://base/icon.svg"                                                      
new URL("data:image/png;base64,...", "https://base/").toString() // unchanged
```

src/workspace/workspacesProvider.ts
```diff
@@ -358,6 +363,45 @@ class AgentMetadataTreeItem extends vscode.TreeItem {
...
+}
+
+function defaultTreeItemIconPath(): {
+	light: vscode.Uri;
+	dark: vscode.Uri;
+} {
+	return {
+		light: vscode.Uri.file(
+			path.join(__dirname, "..", "..", "media", "logo-black.svg"),
```
EhabY, line 397: This seems wrong now, it used to be `__filename` then traversing up by two levels but now it's `__dirname` but by two levels as well

src/workspace/workspacesProvider.ts
```diff
@@ -366,12 +410,16 @@ class AppStatusTreeItem extends vscode.TreeItem {
...
 			workspace_name?: string;
+			icon?: string;
 		},
+		baseUrl?: string,
 	) {
-		super("", vscode.TreeItemCollapsibleState.None);
+		super(app.name, vscode.TreeItemCollapsibleState.None);
 		this.id = app.id;
-		this.description = app.name;
+		this.description = app.workspace_name;
```
EhabY, line 419: Honestly name here is confusing, this is `status.message` which is why it was rendered in the description (to make it less prominent). How would this currently look like actually?

If we want the status to be more prominent here then maybe we just show `app.display_name ?? app.slug` in the description? Since the workspace name is already the parent (or grandparent) so no need to repeat it

PR #923, opened 2026-04-29
ci: run tests on Windows and macOS

### LINKED ISSUE #882: CI: run tests on Windows and macOS in addition to Linux
## Problem

We've hit multiple test failures that only surface on Windows (and potentially macOS):

- Path separator mismatches (`/` vs `\`) in assertions
- Shell quoting differences (single quotes on Unix vs double quotes on Windows)
- `execFile` unable to spawn `.cmd` shims on Windows
- `cross-env` needed for environment variable syntax in npm scripts

These issues are caught late, only when someone manually runs tests on a non-Linux machine.

## Proposal

Add Windows and macOS runners to the CI test matrix so cross-platform regressions are caught automatically.

## References

- #876 - recent fixes for all of the above

EhabY:
- Add Windows and macOS to the unit and integration test matrices to catch cross-platform regressions automatically (path separators, shell quoting, `.cmd` shim resolution, etc. — see #876 for prior offenders).
- Linux keeps the full matrix (Electron `37` + `latest`, VS Code `1.106.0` + `stable`); Windows and macOS each run a single `latest`/`stable` smoke job to keep the matrix tight (+4 jobs total, all parallel).
- `xvfb-run` is gated to Linux; Windows/macOS run `pnpm test:integration` directly since `@vscode/test-electron` handles platform-specific VS Code downloads.

Closes #882

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,21 +28,19 @@ jobs:
       - run: pnpm build
 
   test-unit:
-    name: Unit Test (${{ matrix.os }}, Electron ${{ matrix.electron-version }})
+    name: Unit Test (${{ matrix.name }}, Electron ${{ matrix.electron-version }})
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


mtojek (approved):
stamp

PR #930, opened 2026-05-04
fix: parse Coder authorities with Punycode (xn--) domain labels

### LINKED ISSUE #929: parseRemoteAuthority misparses Punycode (xn--) domains in SSH host names
## Problem

`parseRemoteAuthority` in `src/util.ts` extracts host, owner, workspace, and agent by splitting the SSH host name on `--`. Punycode labels (`xn--{encoded}`) legally contain `--` at positions 3-4, so the split breaks any deployment on an internationalized domain.

## Reproduction

Mid-domain Punycode (`https://coder.xn--eckwd4c7cu47r2wf.jp/`):

- SSH host: `coder-vscode.coder.xn--eckwd4c7cu47r2wf.jp--owner--workspace`
- Split: `["coder-vscode.coder.xn", "eckwd4c7cu47r2wf.jp", "owner", "workspace"]`
- The parser treats `eckwd4c7cu47r2wf.jp` as the owner; connection fails.

Apex Punycode (`https://xn--p1ai/`):

- SSH host: `coder-vscode.xn--p1ai--owner--ws`
- Split: `["coder-vscode.xn", "p1ai", "owner", "ws"]`
- Same misparse.

## Background

First reported and partially fixed in #917. That fix handles mid-domain Punycode but misparses the apex case, where the merged-back fragment has no dot of its own.

## Scope

Non-Punycode `--` in DNS labels is permitted by RFC 5891 outside positions 3-4 but blocked at the TLD level by all major registries. A deployment on a subdomain like `foo--bar.example.com` is theoretically possible and not handled here.

EhabY:
## Summary

`parseRemoteAuthority` (`src/util.ts`) splits the SSH host name on `--`, which breaks any domain containing a Punycode label (`xn--{encoded}`). After the existing split, the loop merges segment pairs back while the prefix ends in `.xn` (the cut landed inside an `xn--...` label). This covers mid-domain, apex (`xn--p1ai`), and consecutive Punycode labels.

The `.xn` check is structural: Punycode's ACE prefix is IANA-reserved. It does not depend on Coder's username, workspace, or agent character rules.

The `parseRemoteAuthority` describe block is refactored to `it.each`.

## Closes

Closes #929. Builds on #917, extending the same idea to the apex case.

## Scope

A deployment on a non-Punycode `--` label (e.g. `foo--bar.example.com`) is not handled. Registries block `--` outside `xn--` at the TLD level, leaving operator-chosen subdomains as the only path. Supporting that case would require slug-shape validation of the tail, which couples this code to Coder's name rules.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -70,6 +70,12 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
 		return null;
 	}
 
+	// Reassemble Punycode labels (xn--...) the split broke apart: when the
+	// prefix ends in ".xn", the cut landed inside an "xn--..." label.
+	while (parts.length >= 2 && parts[0].endsWith(".xn")) {
+		parts.splice(0, 2, `${parts[0]}--${parts[1]}`);
+	}
+
 	// It has the proper prefix, so this is probably a Coder host name.
 	// Validate the SSH host name.  Including the prefix, we expect at least
 	// three parts, or four if including the agent.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@
      from published versions since it shows up in the VS Code extension changelog
      tab and is confusing to users. Add it back between releases if needed. -->
 
+## Unreleased
+
+### Fixed
+
+- Workspaces hosted on internationalized (IDN) domains can now be opened from
+  recent connections. The SSH authority parser was splitting Punycode (`xn--`)
+  domain labels across the field separator and rejecting the host as invalid.
+
 ## [v1.14.5](https://github.com/coder/vscode-coder/releases/tag/v1.14.5) 2026-04-30
 
 ### Added
--- a/test/unit/util.test.ts
+++ b/test/unit/util.test.ts
@@ -2,6 +2,7 @@ import os from "node:os";
 import { afterEach, beforeEach, describe, it, expect, vi } from "vitest";
 
 import {
+	type AuthorityParts,
 	countSubstring,
 	escapeCommandArg,
 	expandPath,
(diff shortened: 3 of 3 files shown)


src/util.ts
```diff
@@ -70,6 +70,12 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
 		return null;
 	}
 
+	// Reassemble Punycode labels (xn--...) the split broke apart: when the
+	// prefix ends in ".xn", the cut landed inside an "xn--..." label.
+	while (parts.length >= 2 && parts[0].endsWith(".xn")) {
+		parts.splice(0, 2, `${parts[0]}--${parts[1]}`);
+	}
+
```
code-asher, line 78: Apparently double dashes are only disallowed as the third and fourth characters, and even then it can depend on the tld?

I imagine this is exceedingly rare, but it means a domain like `https://test--ドメイン名例.jp` (`xn--test---8o4epknhtgo724awkl.jp`, becomes a triple dash apparently) is actually valid.  And also things like `test--domain.com` are valid.

So I think to fully fix any domain issues we might have to take more of a parser approach.  Something like scan through until the last dot, then scan to the first `--` that is not part of a `---`, then split on everything after that index.  Something like that lol

The current change still seems like an improvement though.

code-asher (approved):

src/util.ts
EhabY: Yes this was considered but it would tie us to Coderd semantics. Like this only works because usernames cannot have dots and thus we can keep appending the segments until there is no dot (this would work on all domains even if rare). Do you prefer this approach?

EhabY: I'll merge this as is but I've created this followup, would like your input on it: https://github.com/coder/vscode-coder/issues/937

code-asher: The host names already can only be used with coderd; they are invalid domains and only work on account of the proxy command so I see no issues.  Is there a scenario you are thinking of where these host names are used outside Coder?

But I think it makes sense for us to obey domain semantics anyway and always treat dots as separators.  My bet is that any other reasonable tool would do the same.  If we were to support dots in usernames, I think the only way for us or any other tool to do it correctly would be to escape them, which means the domain parsing part would still be the same (split on dots).

So that would mean basically what we are doing is extracting the tld and our special info is there appended to the tld, and the trick is just figuring out which part of that is the tld and which part is the special info.

Probably what we should have done is something like `coder-vscode.agent--workspace--user.domain.tld` or `agent--workspace--user.domain.tld.coder-vscode`...not sure if switching now would be overly complicated.  :cry: 

I will check out #937!

code-asher: I forgot that we current write `workspace.agent` so the ship has already sailed on obeying domain semantics lol

PR #932, opened 2026-05-04
chore: implement storybook

jakehwll:
> This PR was modified by Coder Agents on behalf of Jake Howell.

Adds Storybook support for developing and previewing webview components in isolation outside of VS Code. Includes stories for all `tasks` package components with a dark/light theme switcher that mirrors VS Code CSS custom properties.

- Add `.storybook/` config with Vite integration, workspace alias resolution, codicon injection, and `acquireVsCodeApi` mock
- Add dark-v2 and light-v2 themes (sourced from `vscode-elements/webview-playground`) with a toolbar switcher in `globalTypes`
- Add 17 component stories under `packages/tasks/src/components/` covering all major UI states
- Add `test/webview/decorators.tsx` with a shared `withQueryClient` decorator for stories using React Query
- Add `storybook` and `build-storybook` scripts to root `package.json`
- Add Storybook dependencies to workspace catalog


package.json
```diff
@@ -18,26 +18,28 @@
...
-		"test:extension": "cross-env ELECTRON_RUN_AS_NODE=1 electron node_modules/vitest/vitest.mjs --project extension",
-		"test:integration": "pnpm compile-tests:integration && node esbuild.mjs && vscode-test",
-		"test:webview": "cross-env ELECTRON_RUN_AS_NODE=1 electron node_modules/vitest/vitest.mjs --project webview",
-		"typecheck": "concurrently -g -n extension,tests,packages \"tsc --noEmit\" \"tsc --noEmit -p test\" \"pnpm typecheck:packages\"",
-		"typecheck:packages": "pnpm -r --filter \"./packages/*\" --parallel typecheck",
-		"watch": "concurrently -g -n extension,webviews \"pnpm watch:extension\" \"pnpm watch:webviews\"",
-		"watch:extension": "node esbuild.mjs --watch",
-		"watch:webviews": "pnpm -r --filter \"./packages/*\" --parallel dev"
-	},
+    "build": "concurrently -g -n webviews,extension,compile-tests:integration \"pnpm build:webviews\" \"node esbuild.mjs\" \"pnpm compile-tests:integration\"",
```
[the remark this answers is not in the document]
jakehwll: Fair enuf. Done.

pnpm-workspace.yaml
```diff
@@ -3,6 +3,11 @@ packages:
 
 catalog:
   "@rolldown/plugin-babel": ^0.2.3
+  "@storybook/addon-a11y": ^8.6.18
+  "@storybook/addon-essentials": ^8.6.18
+  "@storybook/react": ^8.6.18
+  "@storybook/react-vite": ^8.6.18
+  '@storybook/test': ^8.6.15
```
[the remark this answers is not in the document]
jakehwll: Fair enuf. Done.

package.json
```diff
@@ -594,6 +596,7 @@
 		"@abraham/reflection": "^0.13.0",
 		"@peculiar/x509": "^2.0.0",
 		"@repo/shared": "workspace:*",
+		"@vscode/codicons": "catalog:",
```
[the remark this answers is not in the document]
jakehwll: Fair enuf. Done.

pnpm-workspace.yaml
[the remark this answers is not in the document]
jakehwll: Fair enuf. Done.

packages/tasks/src/components/TaskDetailHeader.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { task } from "../../../../test/mocks/tasks";
+import { withQueryClient } from "../../../../test/webview/decorators";
+
+import { TaskDetailHeader } from "./TaskDetailHeader";
+
+import type { Meta, StoryObj } from "@storybook/react";
+
+
+const meta: Meta<typeof TaskDetailHeader> = {
+	title: "Tasks/TaskDetailHeader",
```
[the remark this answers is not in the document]
jakehwll: Fair enuf. Done.

.storybook/preview.ts
```diff
@@ -0,0 +1,122 @@
...
+
+			useEffect(() => {
+				if (context.tags.includes("tasks")) {
+					// Dynamically import tasks CSS
+					void import("../packages/tasks/src/index.css");
+				}
+			}, [context.tags]);
+
+			return createElement(
+				"div",
```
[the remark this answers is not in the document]
jakehwll: This was from the legacy approach. Fair enuf. Done.

packages/tasks/src/components/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export default meta;
+type Story = StoryObj<typeof ActionMenu>;
+
+export const Default: Story = {};
+
+export const Opened: Story = {
+	play: async ({ canvasElement }) => {
+		// The vscode-icon renders a button in its shadow DOM
```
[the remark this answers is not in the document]
jakehwll: Fair enuf. Done.

packages/tasks/src/components/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+		// We need to find the vscode-icon element and click it
+		const icon = canvasElement.querySelector("vscode-icon[action-icon]");
+		expect(icon).toBeTruthy();
+		await userEvent.click(icon!);
```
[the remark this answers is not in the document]
jakehwll: Ignoring this intentionally as we aren't mocking the backend API. Im satisfied enough to check for the loader. YOMV.

package.json
```diff
@@ -21,13 +21,15 @@
 		"build": "concurrently -g -n webviews,extension,compile-tests:integration \"pnpm build:webviews\" \"node esbuild.mjs\" \"pnpm compile-tests:integration\"",
 		"build:production": "cross-env NODE_ENV=production pnpm build",
 		"build:webviews": "pnpm -r --filter \"./packages/*\" --parallel build",
+		"build-storybook": "storybook build --config-dir .storybook",
```
EhabY: Why not `build:storybook` for consistency?

package.json
```diff
@@ -615,6 +617,11 @@
 		"@eslint/js": "^10.0.1",
 		"@eslint/markdown": "^8.0.1",
 		"@rolldown/plugin-babel": "catalog:",
+		"@storybook/addon-a11y": "catalog:",
+		"@storybook/addon-essentials": "catalog:",
+		"@storybook/react": "catalog:",
+		"@storybook/react-vite": "catalog:",
+		"@storybook/test": "catalog:",
```
EhabY: If these and `storybook` deps are not used anywhere then I would rather inline this. I'd keep the catalog for unifying versions between packages

.storybook/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
EhabY, line 1: The `.storybook` directory should be added  to `.vscodeignore`, otherwise we'd package it

```
# Storybook
.storybook/**
storybook-static/**
**/*.stories.*
```

Also we seem to be missing the the `.gitignore` for the build artifacts

.storybook/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+	"extends": "../tsconfig.json",
```
EhabY: I think `tsconfig.packages.json` is a better fit than the extension tsconfig, or just extend `@tsconfig/node22/tsconfig.json` and inline what we need

.storybook/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+		"types": ["node"],
+		"jsx": "react-jsx",
+		"noEmit": true,
+		"rootDir": ".."
+	},
+	"include": [
+		"*.ts",
+		"themes/**/*.ts",
+		"../test/webview/**/*",
+		"../test/mocks/**/*"
```
EhabY: Aren't those already checked by the `test/tsconfig.ts`? I'd rather keep this tsconfig to only the files here

packages/tasks/src/components/CreateTaskSection.stories.tsx
```diff
@@ -0,0 +1,22 @@
+import { taskTemplate } from "../../../../test/mocks/tasks";
+import { withQueryClient } from "../../../../test/webview/decorators";
```
EhabY: `../../../../test/webview/decorators` is rough. Two cleaner options: move them into `.storybook/` if they're storybook-only, or pull them into a workspace package like `@repo/storybook-utils` and import as `@repo/storybook-utils`. The Vite aliases already set up in `main.ts` would resolve it cleanly. Same applies to `test/mocks/tasks` if it's shared between vitest and storybook.

.storybook/main.ts
```diff
@@ -0,0 +1,60 @@
...
+			if (manifest.name?.startsWith("@repo/")) {
+				aliases[manifest.name] = srcDir;
+			}
+		} catch {
+			// Skip directories without a valid package.json.
+		}
+	}
+
+	return aliases;
+}
```
EhabY: We should use `vite-tsconfig-paths` or perhaps even handwrite them somewhere and re-use them. It currently re-reads all `package.json` at config time to rebuild this info

.storybook/preview.ts
```diff
@@ -0,0 +1,126 @@
...
+	if (navigator.userAgent.includes("Linux")) {
+		return 'system-ui, "Ubuntu", "Droid Sans", sans-serif';
+	} else if (navigator.userAgent.includes("Mac")) {
+		return "-apple-system, BlinkMacSystemFont, sans-serif";
+	} else if (navigator.userAgent.includes("Windows")) {
+		return '"Segoe WPC", "Segoe UI", sans-serif';
+	} else {
+		return "sans-serif";
+	}
+};
```
EhabY: The theme decorator already applies `--vscode-font-family` from the theme. `global.css` uses it for `#root` already. Reading `navigator.userAgent` from JS to set `font-family` inline duplicates that and can drift from what VS Code itself does. Remove `getDefaultFontStack` and the inline style, the CSS variable should handle it

.storybook/preview.ts
```diff
@@ -0,0 +1,126 @@
...
+					selectedTheme.forEach(([property]) => {
+						document.documentElement.style.removeProperty(property);
+					});
+				};
+			}, [selectedTheme]);
+
+			const hasTasks = useMemo(
+				() => context.tags.includes("tasks"),
+				[context.tags.join(",")],
+			);
```
EhabY: Honestly I'd just inline this without the memo, we should have the react compiler for the tasks package anyway

.storybook/preview.ts
```diff
@@ -0,0 +1,126 @@
...
+
+			useEffect(() => {
+				if (hasTasks) {
+					// Dynamically import tasks CSS
+					void import("../packages/tasks/src/index.css");
```
EhabY: Loading `tasks/src/index.css` from a global preview decorator means every new package needs its CSS wired in here.

Better to ship a per-package decorator (like `withTasksStyles`) from the same package as the components and have those stories opt in via `decorators: [withTasksStyles]`. Keeps `preview.ts` package-agnostic.

.storybook/themes/dark-v2.ts
```diff
@@ -0,0 +1,959 @@
+// Sourced from `vscode-elements/webview-playground`.
+// https://github.com/vscode-elements/webview-playground/blob/f9a6f90413d0cc535839fb92445b7a5eebc540d7/dist/themes/dark-v2.js
```
EhabY, line 2: To avoid manual refreshes, add a `scripts/sync-vscode-themes.mjs` that fetches both files from the pinned upstream sha (already in the header comments) and rewrites them in place

I'm not sure if this would work but can we have the script filter vars referenced in our source plus those used by `@vscode-elements/elements`, which would cut these files by ~85% (though it's only a big change on this branch only)

packages/storybook-utils/src/decorators/withQueryClient.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import type { Decorator } from "@storybook/react";
+import { useRef } from "react";
+
+/**
+ * Decorator that wraps stories with a QueryClientProvider.
+ * Use this for components that use React Query hooks (useQuery, useMutation, etc.)
+ */
+export const withQueryClient: Decorator = (Story) => {
+	const clientRef = useRef<QueryClient | null>(null);
```
EhabY: How about `const [client] = useState(() => new QueryClient(...))` since that is the convention for lazily initializing

.storybook/preview.ts
```diff
@@ -0,0 +1,126 @@
...
+	postMessage: (message: unknown) => void;
+	getState: () => unknown;
+	setState: (state: unknown) => void;
+}
+
+// Mock the acquireVsCodeApi function for Storybook, so that components
+// that rely on it can function without errors.
+if (
+	typeof window !== "undefined" &&
+	!(window as { acquireVsCodeApi?: () => VsCodeApi }).acquireVsCodeApi
```
EhabY: The local `VsCodeApi` duplicates `WebviewApi` from `@types/vscode-webview`, we can put it in the same `@repo/storybook-utils` dev-only package proposed above, which has `@types/vscode-webview` as a dev dep. `preview.ts` then becomes `import { installMockVsCodeApi } from "@repo/storybook-utils"`.

EhabY:
Love the change and the story books :heart: 

Pls do run the linters and make sure the pipeline passes!

.storybook/themes/dark-v2.ts
jakehwll: If it points to the same SHA why do we want to pull it again? Am I missing something? 🙂 

I think the script filter might be a future improvement

.storybook/preview.ts
jakehwll: Oooo, this one was interesting.

We actually never define the `--vscode-font-family` 🙂

package.json
```diff
@@ -616,12 +616,13 @@
...
+		"@repo/mocks": "workspace:*",
 		"@rolldown/plugin-babel": "catalog:",
-		"@storybook/addon-a11y": "catalog:",
-		"@storybook/addon-essentials": "catalog:",
-		"@storybook/react": "catalog:",
-		"@storybook/react-vite": "catalog:",
-		"@storybook/test": "catalog:",
+		"@storybook/addon-a11y": "^8.6.18",
+		"@storybook/addon-essentials": "^8.6.18",
+		"@storybook/react": "^8.6.18",
```
EhabY: This one could be moved back to the catalog now that we import it in an inner package actually

.storybook/themes/dark-v2.ts
EhabY: I meant to say that we can trigger this script and it would pull a new SHA (most recent) and add that comment in the header so we know what upstream SHA we are currently pinned against

packages/tasks/src/utils/storybook.ts
```diff
@@ -0,0 +1,9 @@
+import "../index.css";
+
+import type { Decorator } from "@storybook/react";
+
+/**
+ * Injects the tasks package CSS into the Storybook preview.
+ * Add to `decorators` in any story that renders tasks components.
+ */
+export const withTasksStyles: Decorator = (Story) => Story();
```
EhabY: The entire reason the file exists is to trigger `import "../index.css"` as a side effect when stories import the symbol. I would rather that we do not mix production utils in with test utils.

Replace with `import.meta.glob` in `.storybook/preview.ts`:

```ts
// packages/tasks/storybook.preview.ts
import "./src/index.css";
```

```ts
// .storybook/preview.ts
import.meta.glob("../packages/*/storybook.preview.ts", { eager: true });
```

Each package opts in with a one-line file outside `src/`. Stories stop importing the decorator entirely. Add `/// <reference types="vite/client" />` at the top of `preview.ts` for the `ImportMeta.glob` typing, and add `../packages/*/storybook.preview.ts` to the storybook tsconfig include so ESLint can resolve it.

.storybook/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "../packages/tsconfig.packages.json",
+	"compilerOptions": {
+		"types": ["node"],
+		"jsx": "react-jsx",
+		"noEmit": true,
+		"rootDir": ".."
+	},
+	"include": ["*.ts", "themes/**/*.ts"],
+	"exclude": ["node_modules"]
```
EhabY, line 10: Not needed actually, TS already ignores this folder

Actually I think this all can be like this now:
```
{
	"extends": "../packages/tsconfig.packages.json",
	"include": ["*.ts", "themes/**/*.ts"]
}
```

Verified: `tsc --noEmit -p .storybook` passes and `pnpm storybook:build` builds clean.

.storybook/preview.ts
```diff
@@ -0,0 +1,104 @@
...
+					});
+				};
+			}, [selectedTheme]);
+
+			return createElement(
+				"div",
+				{
+					id: "root",
+					style: {
+						fontFamily: getDefaultFontStack(),
```
EhabY: There is an issue with how this is used actually, `style={{ fontFamily }}` on `#root` is overridden by descendants like `packages/tasks/src/index.css:8` that read `var(--vscode-font-family)` directly.

Set it as a CSS variable instead (extend the theme arrays or call `setProperty` next to the theme loop), then drop the inline style.

package.json
EhabY: Add a fourth concurrent run:

```text
"tsc --noEmit -p .storybook"
```

.storybook/preview.ts
```diff
@@ -0,0 +1,104 @@
...
+if (
+	typeof window !== "undefined" &&
+	!(window as { acquireVsCodeApi?: () => WebviewApi<unknown> }).acquireVsCodeApi
+) {
+	(window as { acquireVsCodeApi: () => WebviewApi<unknown> }).acquireVsCodeApi =
+		() => ({
+			postMessage: () => undefined,
+			getState: () => undefined,
+			setState: (state) => state,
+		});
```
EhabY: After the tsconfig is slimmed (separate comment), `@types/vscode-webview` is ambient. Add a `Window` augmentation and the casts go away:

```ts
declare global {
	interface Window {
		acquireVsCodeApi?: <T = unknown>() => WebviewApi<T>;
	}
}

if (typeof window !== "undefined") {
	window.acquireVsCodeApi ??= () => ({
		postMessage: () => undefined,
		getState: () => undefined,
		setState: (state) => state,
	});
}
```

eslint.config.mjs
EhabY, line 3: Nothing currently prevents a runtime file like `packages/tasks/src/TasksPanel.tsx` from importing `@repo/mocks` or `@repo/storybook-utils`. The `devDependencies` listing is a human hint, not a build guard. Add a `no-restricted-imports` block scoped to `packages/*/src/**` excluding `*.stories.*`, blocking those two package patterns. Catches the mistake at the import site with a specific error message.

package.json
```diff
@@ -655,6 +665,7 @@
 		"prettier": "^3.8.3",
 		"react": "catalog:",
 		"react-dom": "catalog:",
+		"storybook": "^8.6.18",
```
EhabY: Why not use the most recent version for all of these new packages?

EhabY:
Now that we have storybooks, there seems to be some inconsistencies:

`tasks-actionmenu--opened`, the loading item is not aligned with the others:

image
<hr/>

`tasks-errorbanner--default`, should we align the view logs to the right?

image
<hr />

`tasks-statepanel--error`, the retry here is not a button while it is a button in the error state. Was this intentional or a storybook styling issue :thinking:?

image
<hr />

The status indicators are not applying the right styles it seems, like this is just a pure square:

image
<hr />

`tasks-tasklist--default` could mix and match other states and descriptions IMO for better coverage

image
<hr/>

This is the "Task Selection" test but I do not see it selecting a task (or inside a selected task):

image
<hr/>

packages/tasks/src/components/TasksPanel.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+export default meta;
+type Story = StoryObj<typeof TasksPanel>;
+
+export const Default: Story = {
+	args: {
+		tasks: [
+			task({ id: "task-1" }),
+			task({ id: "task-2" }),
+			task({ id: "task-3" }),
+		],
```
EhabY: This is reused 3 times here. Lift to a module-scope const or push into `meta.args`, then per-story `args` only overrides what differs.

packages/tasks/src/components/AgentChatHistory.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+			logs: [
+				logEntry({
+					id: 1,
+					type: "input",
+					content: "What is the weather today?",
+				}),
+				logEntry({
+					id: 2,
+					type: "output",
+					content: "The weather today is sunny with a high of 25°C.",
```
EhabY: This is reused 2 times here. Lift to a module-scope const or push into `meta.args`, then per-story `args` only overrides what differs.

packages/tasks/src/components/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export default meta;
+type Story = StoryObj<typeof ActionMenu>;
+
+export const Default: Story = {};
+
+export const Opened: Story = {
+	play: async ({ canvasElement }) => {
+		const icon = canvasElement.querySelector("vscode-icon[action-icon]");
+		await expect(icon).toBeTruthy();
```
EhabY: Not needed IMO

packages/tasks/src/components/NotSupportedState.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+	component: NotSupportedState,
+	decorators: [withTasksStyles],
+};
+
+export default meta;
+type Story = StoryObj<typeof NotSupportedState>;
+
+export const Default: Story = {
+	args: {},
+};
```
EhabY: `args: {}` is dead weight. Use `Default: Story = {};` like in `NoTemplateState.stories.tsx`

packages/tasks/src/components/ActionMenu.stories.tsx
jakehwll: This was from the Coder review bot 😢 Happy to remove, but would rather keep it happy.

jakehwll:
> `tasks-actionmenu--opened`, the loading item is not aligned with the others:

Not going to fix just yet, will wait for this branch to merge to tightly scope and follow up 🙂 

> `tasks-errorbanner--default`, should we align the view logs to the right?

Not going to fix just yet, will wait for this branch to merge to tightly scope and follow up 🙂 

> `tasks-statepanel--error`, the retry here is not a button while it is a button in the error state. Was this intentional or a storybook styling issue 🤔?

Hmm, fair enough. I've imported both those components to the story 🙂 

> The status indicators are not applying the right styles it seems, like this is just a pure square:

Looks like the `display: inline` was zorping these. I've added a wrapper decorator for it now!

> `tasks-tasklist--default` could mix and match other states and descriptions IMO for better coverage

Good idea!

> This is the "Task Selection" test but I do not see it selecting a task (or inside a selected task):

Oops, I think I nuked this by accident.

scripts/sync-vscode-themes.mjs
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env node
```
EhabY, line 1: This could be simplified and parallized slightly:

```ts
#!/usr/bin/env node

// Fetches dark-v2 and light-v2 theme files from vscode-elements/webview-playground
// and rewrites .storybook/themes/{dark,light}-v2.ts in place.
//
// Usage:
//   node scripts/sync-vscode-themes.mjs          # fetch latest main
//   node scripts/sync-vscode-themes.mjs <sha>    # fetch a specific commit

import { writeFileSync } from "node:fs";
import { resolve, dirname } from "node:path";
import { fileURLToPath } from "node:url";

const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), "..");
const REPO = "vscode-elements/webview-playground";

const themes = [
	{
		remotePath: "dist/themes/dark-v2.js",
		localPath: ".storybook/themes/dark-v2.ts",
		exportName: "darkTheme",
	},
	{
		remotePath: "dist/themes/light-v2.js",
		localPath: ".storybook/themes/light-v2.ts",
		exportName: "lightTheme",
	},
];

async function fetchText(url, headers) {
	const res = await fetch(url, { headers });
	if (!res.ok) {
		throw new Error(`Failed to fetch ${url}: ${res.status} ${res.statusText}`);
	}
	return res.text();
}

const sha =
	process.argv[2] ??
	(
		await fetchText(`https://api.github.com/repos/${REPO}/commits/main`, {
			Accept: "application/vnd.github.sha",
		})
	).trim();

console.log(`Syncing themes from ${REPO}@${sha}`);

await Promise.all(
	themes.map(async ({ remotePath, localPath, exportName }) => {
		const js = await fetchText(
			`https://raw.githubusercontent.com/${REPO}/${sha}/${remotePath}`,
		);
		const match = js.match(/export\s+const\s+theme\s*=\s*(\[[\s\S]*\])/);
		if (!match) {
			throw new Error(`Could not find theme array in ${remotePath}`);
		}
		// Upstream escapes dots in CSS property names (e.g. `disabled\.background`).
		// These are unnecessary in JS strings and break setProperty lookups.
		const arrayLiteral = match[1].replaceAll("\\.", ".");

		const ts = `// Sourced from \`vscode-elements/webview-playground\`.
// https://github.com/${REPO}/blob/${sha}/${remotePath}

export const ${exportName}: Array<[string, string]> = ${arrayLiteral};
`;
		writeFileSync(resolve(ROOT, localPath), ts);
		console.log(`  ${localPath}`);
	}),
);

console.log("Done. Run `pnpm format` to normalize the output.");
```

EhabY (approved):
There are a few follow ups in the chromatic one but this one is good!

eslint.config.mjs
```diff
@@ -247,6 +249,17 @@ export default defineConfig(
...
 	},
 
+	// Storybook recommended rules for story files
+	...storybook.configs["flat/recommended"],
+	{
+		files: ["**/*.stories.@(ts|tsx|js|jsx|mjs|cjs)"],
+		rules: {
+			// At storybook v8, @storybook/react-vite doesn't re-export Meta/StoryObj;
+			// importing from @storybook/react is correct for now.
+			"storybook/no-renderer-packages": "off",
```
EhabY: We are using v10 now though right?

.gitignore
```diff
@@ -13,4 +13,4 @@ pnpm-debug.log
 packages/*/node_modules/
 
 # Storybook
-storybook-static/
\ No newline at end of file
+storybook-static/debug-storybook.log
```
EhabY: Shouldn't we ignore the whole folder?

EhabY (approved):
All good again, looks great!

.gitignore
EhabY: Nevermind this was fixed in a later commit

jakehwll: Fixed — `.gitignore` now has `storybook-static/` and `debug-storybook.log` as separate entries (the newline was missing before, causing them to concatenate into one path).

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

eslint.config.mjs
jakehwll: Yep — this was already removed in 80693f0 as part of the v10 upgrade. The PR diff you were looking at was from an older commit.

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

jakehwll: To clarify: the `no-renderer-packages` disable and the v8 comment were present in the commit you reviewed (`d04f325`), but were removed in a subsequent force-push (`32dd1bf`). The current HEAD no longer has this block — just `...storybook.configs["flat/recommended"]` with no overrides.

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

PR #933, opened 2026-05-05
chore: implement chromatic

jakehwll:
> 🤖 This PR was modified by Coder Agents on behalf of Jake Howell.

Follow-up to #932. Adds Chromatic to the CI pipeline so we get visual regression testing and snapshot previews on PRs.

- Added `chromatic` job to `ci.yaml` with separate steps for PR builds (review-gated) and mainline builds (auto-accept)
- Uses TurboSnap (`onlyChanged`) and zip uploads for faster, leaner builds
- Skips Renovate/Dependabot branches to avoid unnecessary snapshot churn
- Added `chromatic` and `storybook:ci` scripts to `package.json`
- Pinned `chromatic` dependency via the pnpm catalog


pnpm-workspace.yaml
```diff
@@ -11,6 +11,7 @@ catalog:
   "@vscode-elements/react-elements": ^2.4.0
   "@vscode/codicons": ^0.0.45
   babel-plugin-react-compiler: ^1.0.0
+  chromatic: ^16.7.0
```
EhabY: Nit: 16.8.0 is already out 😛 (also could be inline in `package.json`)

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,73 @@ jobs:
         run: ${{ runner.os == 'Linux' && 'xvfb-run -a' || '' }} pnpm test:integration --label "VS Code ${{ matrix.vscode-version }}"
         shell: bash
 
+  # Reference guide:
```
EhabY: I like the source but the inner comments feel overly verbose to me, we can remove or adapt this to our use-case and make them a bit slimmer!

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,73 @@ jobs:
...
+  chromatic:
+    name: Chromatic
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+        with:
+          # 👇 Ensures Chromatic can read your full git history
+          fetch-depth: 0
+          # 👇 Tells the checkout which commit hash to reference
```
EhabY: Unnecessary comments IMO

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,73 @@ jobs:
...
+          NODE_OPTIONS: "--max_old_space_size=4096"
+          STORYBOOK: true
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          # Do a fast, testing build for change previews
+          buildScriptName: "storybook:ci"
+          exitOnceUploaded: true
+          # This will prevent CI from failing when Chromatic detects visual changes
+          exitZeroOnChanges: true
+          # Prevent excessive build runs on minor version changes
```
EhabY: Misleading comment (I'd say remove most of these inline ones)

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,73 @@ jobs:
...
+
+      - name: Setup pnpm, Node.js, and dependencies
+        uses: ./.github/actions/setup
+
+      # This step is not meant for mainline because any detected changes to
+      # storybook snapshots will require manual approval/review in order for
+      # the check to pass. This is desired in PRs, but not in mainline.
+      - name: Publish to Chromatic (non-mainline)
+        if: github.ref != 'refs/heads/main' && github.repository_owner == 'coder'
+        uses: chromaui/action@latest
```
EhabY: We should pin this like all other actions for security and reproducibility

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,73 @@ jobs:
...
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          autoAcceptChanges: true
+          # This will prevent CI from failing when Chromatic detects visual changes
+          exitZeroOnChanges: true
+          # Do a full build with documentation for mainline builds
+          buildScriptName: "storybook:build"
+          # Run TurboSnap to trace file dependencies to related stories
+          # and tell chromatic to only take snapshots of relevant stories
+          onlyChanged: true
```
EhabY: We should drop this on mainline because `autoAcceptChanges` is set to true (so we have a full baseline build)

.github/workflows/ci.yaml
```diff
@@ -77,6 +77,73 @@ jobs:
...
+          # Avoid uploading single files, because that's very slow
+          zip: true
+
+      # This is a separate step for mainline only that auto accepts and changes
+      # instead of holding CI up. Since we squash/merge, this is defensive to
+      # avoid the same changeset from requiring review once squashed into
+      # main. Chromatic is supposed to be able to detect that we use squash
+      # commits, but it's good to be defensive in case, otherwise CI remains
+      # infinitely "in progress" in mainline unless we re-review each build.
```
EhabY: Very verbose imo

EhabY (approved):

eslint.config.mjs
```diff
@@ -1,3 +1,6 @@
+// For more info, see https://github.com/storybookjs/eslint-plugin-storybook#configuration-flat-config-format
+import storybook from "eslint-plugin-storybook";
```
EhabY: This seems to be no longer maintained, maybe use the monorepo one? https://github.com/storybookjs/storybook/tree/next/code/lib/eslint-plugin

Also don't we need to extend the rules?

EhabY (commented):
Let's merge this after the storybook one lands and we have no conflicts!

.github/workflows/ci.yaml
jakehwll: Dropped `onlyChanged` from the mainline step.

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

.github/workflows/ci.yaml
jakehwll: Pinned to `chromaui/action@5c6ec06f45a2117a25f07b1bf2b2f3009233fac8 # v16.3.0` (same SHA as `coder/coder`).

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

.github/workflows/ci.yaml
jakehwll: Agreed — stripped all the inline noise. Kept one short comment per step explaining _why_ we have two separate steps (PR vs mainline) since that's the only non-obvious bit.

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

pnpm-workspace.yaml
jakehwll: Bumped to `^16.9.1` (latest) in the catalog. 🫡

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

.github/workflows/ci.yaml
jakehwll: Condensed to a single line: `# Auto-accept on mainline to avoid blocking CI after squash merges.`

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

.github/workflows/ci.yaml
jakehwll: Removed all inline `with:` comments.

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

.github/workflows/ci.yaml
jakehwll: Removed.

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

eslint.config.mjs
jakehwll: The installed `eslint-plugin-storybook@10.3.6` is already the monorepo package (npm publishes from `storybookjs/storybook`). Removed the stale URL comment pointing to the old standalone repo.

The plugin was imported but never wired in -- spread `storybook.configs["flat/recommended"]` into the config so the rules actually apply to `**/*.stories.*` files.

> 🤖 Generated by Coder Agents on behalf of Jake Howell.

EhabY: I think this was fixed in the other branch

EhabY:
Still approved, love the chromatic addition!

PR #934, opened 2026-05-05
feat(telemetry): central command dispatcher (#902)

### LINKED ISSUE #902: Telemetry: central command dispatcher
Part of the Telemetry Phase A rollout. See the RFC in Linear: AIGOV-154.

Introduce a command dispatcher modeled after `contextManager` in `src/core/contextManager.ts` so every `coder.*` command goes through one place. This gives us uniform `command.invoked` instrumentation at registration time (no per-command wrapping), a single choke point for error handling, and a harder path for commands to be registered outside the service container.

## Scope

- Implement `CommandManager` in `src/core/commandManager.ts` exposing `register(id: CoderCommandId, handler: CommandHandler): Disposable`.
- `CoderCommandId` is a string literal union validated against the command list in `package.json`. Attempts to register an unknown id throw at dev time.
- Internally the manager wraps every handler with `TelemetryService.time("command.invoked", handler, { commandId })` so duration and success/error are captured once.
- Register `CommandManager` in `ServiceContainer`. Inject where commands are wired.
- Migrate all existing `vscode.commands.registerCommand("coder.*", ...)` sites to use `CommandManager`. Currently around 25 registrations in `src/extension.ts`.
- Add a guard (lint rule or runtime check) that flags direct `vscode.commands.registerCommand("coder.*", ...)` usage going forward. Exact mechanism is the implementer's call.

## Tests

- Registering a valid id dispatches through the wrapped handler.
- Registering an unknown id throws.
- Handler success emits `command.invoked` with `result: "success"` and a `durationMs`.
- Handler throwing emits `command.invoked` with `result: "error"` and the error re-propagates to the caller.
- Disposing the returned `Disposable` unregisters the command.

## Out of scope

- Other cross-cutting concerns beyond telemetry.

Depends on #900.

EhabY:
Closes #902. Adds `CommandManager` so every `coder.*` command emits `command.invoked` (durationMs + success/error) via `TelemetryService.trace`.

- `src/core/commandManager.ts` — `register(id, handler): Disposable`. `CoderCommandId` is a string-literal union; a unit test pins it to `package.json` so drift breaks CI.
- `src/extension.ts` — migrates all 25 `coder.*` registrations. The manager owns disposal via `ServiceContainer`, disposed before `TelemetryService` so in-flight events still flush.
- `eslint.config.mjs` — blocks direct `registerCommand` of `coder.*` ids (carve-out for `commandManager.ts`).
- `package.json` — declares `coder.openDevContainer` (was registered but never declared).

Builds on #921; merge target is `telemetry/local-jsonl-sink`, not `main`.

### DIFF
--- a/src/core/commandManager.ts
+++ b/src/core/commandManager.ts
@@ -1,0 +1,98 @@
+import * as vscode from "vscode";
+
+import { type TelemetryService } from "../telemetry/service";
+
+/**
+ * Every `coder.*` command id contributed by this extension. Kept in sync with
+ * `contributes.commands` in package.json by a unit test.
+ */
+export const CODER_COMMAND_IDS = [
+	"coder.login",
+	"coder.logout",
+	"coder.switchDeployment",
+	"coder.open",
+	"coder.openDevContainer",
+	"coder.openFromSidebar",
+	"coder.openAppStatus",
+	"coder.workspace.update",
+	"coder.createWorkspace",
+	"coder.navigateToWorkspace",
+	"coder.navigateToWorkspaceSettings",
+	"coder.refreshWorkspaces",
+	"coder.viewLogs",
+	"coder.searchMyWorkspaces",
+	"coder.searchAllWorkspaces",
+	"coder.manageCredentials",
+	"coder.applyRecommendedSettings",
+	"coder.pingWorkspace",
+	"coder.pingWorkspace:views",
+	"coder.speedTest",
+	"coder.speedTest:views",
+	"coder.supportBundle",
+	"coder.supportBundle:views",
+	"coder.tasks.refresh",
+	"coder.chat.refresh",
+] as const;
+
+export type CoderCommandId = (typeof CODER_COMMAND_IDS)[number];
+
+const VALID_IDS: ReadonlySet<string> = new Set(CODER_COMMAND_IDS);
+
+const COMMAND_INVOKED_EVENT = "command.invoked";
+
+// `never[]` parameter type lets any concrete handler shape satisfy the
+// constraint, sidestepping function-parameter contravariance.
+type CommandHandler = (...args: never[]) => unknown;
+
+/**
+ * Single registration point for `coder.*` commands. Wraps every handler in
+ * `TelemetryService.trace("command.invoked", ...)` so duration plus
+ * success/error are captured uniformly.
+ */
+export class CommandManager implements vscode.Disposable {
+	private readonly registrations = new Set<vscode.Disposable>();
+
+	public constructor(private readonly telemetry: TelemetryService) {}
+
+	public register(
+		id: CoderCommandId,
+		handler: CommandHandler,
+	): vscode.Disposable {
+		if (!VALID_IDS.has(id)) {
+			throw new Error(`Unknown coder command id: ${id}`);
+		}
+
+		const invoke = handler as (...args: unknown[]) => unknown;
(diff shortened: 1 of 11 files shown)


src/core/commandManager.ts
```diff
@@ -0,0 +1,98 @@
...
+] as const;
+
+export type CoderCommandId = (typeof CODER_COMMAND_IDS)[number];
+
+const VALID_IDS: ReadonlySet<string> = new Set(CODER_COMMAND_IDS);
+
+const COMMAND_INVOKED_EVENT = "command.invoked";
+
+// `never[]` parameter type lets any concrete handler shape satisfy the
+// constraint, sidestepping function-parameter contravariance.
```
DanielleMaywood: Do we _need_ this comment? Fine with keeping, just want to make sure it adds value

src/extension.ts
```diff
@@ -214,14 +215,14 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
-			tasksPanelProvider.refresh(),
-		),
 		// Refresh tasks panel when deployment changes (login/logout/switch)
 		secretsManager.onDidChangeCurrentDeployment(() =>
 			tasksPanelProvider.refresh(),
 		),
 	);
+	commandManager.register("coder.tasks.refresh", () =>
+		tasksPanelProvider.refresh(),
+	);
```
DanielleMaywood: how come we moved this below the secretsManager?

src/extension.ts
```diff
@@ -248,93 +249,84 @@ export async function activate(ctx: vscode.ExtensionContext): Promise<void> {
...
-			commands.supportBundle.bind(commands),
-		),
-		vscode.commands.registerCommand(
-			"coder.supportBundle:views",
-			commands.supportBundle.bind(commands),
-		),
+	);
+
+	// Coder commands. The CommandManager owns disposal via ServiceContainer,
+	// so registrations do not need to be pushed onto ctx.subscriptions.
```
DanielleMaywood: Do we _need_ this comment?

test/unit/core/commandManager.test.ts
```diff
@@ -0,0 +1,171 @@
...
+	createMockLogger,
+	MockConfigurationProvider,
+} from "../../mocks/testHelpers";
+
+import type * as vscodeTypes from "vscode";
+
+function fakeContext(): vscodeTypes.ExtensionContext {
+	return {
+		extension: { packageJSON: { version: "1.2.3-test" } },
+	} as unknown as vscodeTypes.ExtensionContext;
```
DanielleMaywood: Just want to check that `as unknown as` is absolutely required to do what you want? It would be nice for a solution that doesn't require cheating the type system

DanielleMaywood (approved):

src/core/commandManager.ts
EhabY: Shortened to one line

> `never[]` accepts any concrete handler shape (function-parameter contravariance)

src/extension.ts
EhabY: Removed.

src/extension.ts
EhabY: Fixed

test/unit/core/commandManager.test.ts
EhabY: Fixed this by passing in the version and not the context, removed all new `as unknown as`

PR #935, opened 2026-05-05
refactor(telemetry): OTel-format ids and explicit session id

EhabY:
Follow-up to #934 cleaning up the JSONL schema.

- `session_id` was `vscode.env.sessionId` verbatim (UUID + ms timestamp, looks malformed); `event_id` / `trace_id` were UUIDv4, not OTel hex.
- New `src/telemetry/ids.ts` — `newTraceId` / `newSpanId` / `newSessionId` (32 / 16 / 32 hex, OTel widths).
- `TelemetryService` now takes a built `SessionContext` instead of `ExtensionContext`, removing the `as unknown as` test fixtures. `ServiceContainer` builds it once and threads `sessionId` to both the sink filename and the event payload.

Also folds in #934 review feedback: tightened a couple of comments and moved `coder.{tasks,chat}.refresh` registrations adjacent to their panel providers.

Stacked on #934. Retarget once that lands.

### DIFF
--- a/src/core/container.ts
+++ b/src/core/container.ts
@@ -3,7 +3,8 @@ import * as vscode from "vscode";
 import { CoderApi } from "../api/coderApi";
 import { LoginCoordinator } from "../login/loginCoordinator";
 import { OAuthCallback } from "../oauth/oauthCallback";
-import { extractExtensionVersion } from "../telemetry/event";
+import { buildSession, extractExtensionVersion } from "../telemetry/event";
+import { newSessionId } from "../telemetry/ids";
 import { TelemetryService } from "../telemetry/service";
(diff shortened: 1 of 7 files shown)


src/core/container.ts
```diff
@@ -94,15 +95,21 @@ export class ServiceContainer implements vscode.Disposable {
 			context.extensionUri,
 			this.logger,
 		);
+		// Shared by the sink (filename) and the service (event payload).
```
DanielleMaywood: thoughts on the value add of this comment? could it become outdated and lose all value?

DanielleMaywood (approved):

src/core/container.ts
EhabY: Yeah honestly I gotta add a rule to reduce comments - Like you can know this if you just check the references of this var

PR #941, opened 2026-05-07
chore: pin coder to v2.33.1 and bump deps

EhabY:
- Pin the vendored coder client to `github:coder/coder#v2.33.1` (previously `#main`) so installs are reproducible and not re-fetched on every `pnpm install`.
- Update `CONTRIBUTING.md` with the new tag-pinning convention.
- Bump misc deps to latest minors/patches: `axios`, `zod`, `eslint`, `typescript-eslint`, `jsdom`, `globals`, `@eslint-react/eslint-plugin`, `@tanstack/react-query`, `react`, `react-dom`, `vite`.


src/deployment/deploymentManager.ts
```diff
@@ -271,7 +271,7 @@ export class DeploymentManager implements vscode.Disposable {
 				}
 				this.contextManager.set(
 					"coder.agentsEnabled",
-					experiments.includes("agents"),
+					experiments.includes("agents" as Experiment),
```
matifali, line 274: Agents are no longer an experiment in version 2.33.1 and above.

EhabY: True, but VS Code extension can be used by **any** server version where it might be an experiment

matifali: You are also pinning coder to version 2.33.1 which might return it as invalid as it does not exist.

EhabY: This is only for TypeScript types, `Experiment` is a string enum so by doing `as Experiment` here we bypass that compile time error since "agents" is not part of `Experiment`s

matifali (approved):

PR #944, opened 2026-05-07
fix: parse remote authority from right

### LINKED ISSUE #937: parseRemoteAuthority: parse right-to-left to handle all safeHost shapes uniformly
Follow-up from the review thread on #930 (https://github.com/coder/vscode-coder/pull/930#discussion_r3191316236).

## Problem

`parseRemoteAuthority` (`src/util.ts`) splits the SSH host name on `--`, which conflates the field separator with `--` that legitimately appears inside `<safeHost>` (Punycode `xn--abc`, apex Punycode `xn--p1ai`, triple-dash Punycode `xn--test---8o4...`, Latin double-dash `test--domain.com`). The current `parts[0].endsWith(".xn")` merge loop added in #930 only fixes one of these shapes.

## Approach

Don't change the wire format. Parse from the right.

The encoder writes `coder-vscode.<safeHost>--<user>--<ws>(.<agent>)?`. After `split("--")`, the last element is always `<ws>` or `<ws>.<agent>` and the one before it is always `<user>`. Everything earlier rejoins with `--` to give back `coder-vscode.<safeHost>`, no matter how many `--` it contains.

## Algorithm

1. Split on `--`. Require `parts[0]` starts with `"coder-vscode."`, else return null.
2. Require `parts.length >= 3` and no empty parts, else throw `Invalid Coder SSH authority`.
3. `parts[-1]` is `<ws>` or `<ws>.<agent>` (split on `.`). `parts[-2]` is `<user>`.
4. Rejoin `parts.slice(0, -2)` with `--` and strip the `coder-vscode.` prefix to get `safeHostname`.

Single code path, no Unicode awareness.

## What we lose

Two old authority shapes the current parser still accepts:

1. **No safeHost** (`coder-vscode--<user>--<ws>(--|.)<agent?>`): produced before #292 (Jun 2024). About 23 months ago.
2. **SafeHost with `--<agent>`** (`coder-vscode.<safeHost>--<user>--<ws>--<agent>`): produced between #292 (Jun 2024) and #420 (Feb 2025). Window ended about 15 months ago.

Both can still surface in VS Code's Recent menu for users who never reopened the workspace through the Coder panel since. Failure mode is a "not logged in" prompt or a `null` parse, not a crash. Reopening from the panel writes a fresh authority in the current shape and they're back to working.

## Scope

**Parser**
- `src/util.ts`: replace `parseRemoteAuthority` body per the algorithm.

**Tests** (`test/unit/util.test.ts`)
- Move `coder-vscode--foo--bar` and `coder-vscode--foo--bar--baz` (lines 44-54 and 55-65) into `ignores unrelated authority` (now return null).
- Remove `with hostname and -- agent` (lines 77-87).
- Add `test--domain.com`, `xn--test---8o4...`, and a regression case with random `--` in the middle of a safeHost.

**Dead helper cleanup** (empty-hostname branches that were only reachable via the dropped shapes)
- `src/core/pathResolver.ts`: remove empty-hostname branch in `getGlobalConfigDir` and the related doc comments on the `get*Path` helpers.
- `src/remote/sshConfig.ts`: simplify `startBlockComment` / `endBlockComment` to always include the hostname; simplify the malformed-config error messages.
- `src/core/secretsManager.ts`: remove the `|| "<legacy>"` fallback in `buildKey`.
- `src/extension.ts:113`: skip the `getSessionAuth` call when `deployment` is undefined instead of falling back to `""`.
- `test/unit/core/pathResolver.test.ts`: remove `should use base path for empty labels` (lines 22-29).
- `test/unit/remote/sshConfig.test.ts`: remove `creates a new file and adds config with empty label` (lines 48-84) and the `update("")` case at line 447. Keep `does not remove deployment-unaware SSH config` (line 242); it still passes since `getBlock("dev.coder.com")` does not match the legacy block format.

## Why not change the encoding

Escaping `--` in `<safeHost>` at `toRemoteAuthority` time would also work, but it creates the inverse backward-incompat (older extension versions can't parse newer authorities) and adds a custom escape format. The right-to-left parser needs no encoding change.

EhabY:
Closes #937.

## Summary
- Parse Coder remote authorities from the right, rejoining all safeHost segments before username/workspace extraction.
- Drop legacy empty-safeHost and `--agent` parser expectations; add double-dash/triple-dash safeHost regressions.
- Remove empty-hostname fallbacks in path/secrets/SSH config helpers and avoid session lookup when no deployment is selected.

## Tests
- `pnpm test:extension ./test/unit/util.test.ts ./test/unit/core/pathResolver.test.ts ./test/unit/remote/sshConfig.test.ts`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm test:extension`
- `pnpm format:check`

<details>
<summary>Implementation notes (Coder Agents generated)</summary>

- `parseRemoteAuthority` now requires `coder-vscode.<safeHost>` and at least username/workspace fields.
- The parser treats the last `--` segment as workspace/agent, the previous segment as username, and rejoins earlier segments as safeHost.
- Tests cover Latin double-dash safeHost, triple-dash Punycode-like safeHost, and multiple embedded `--` segments.
- Legacy deployment-unaware helpers and tests were removed per issue scope.

</details>

Generated by Coder Agents.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -53,6 +53,10 @@ export function findPort(text: string): number | null {
 /**
  * Given an authority, parse into the expected parts.
  *
+ * The authority looks like `<scheme>://ssh-remote+<ssh host name>`, where the
+ * SSH host names created by this extension match the format:
+ *   coder-vscode.<safeHostname>--<username>--<workspace>(.<agent?>)
+ *
  * If this is not a Coder host, return null.
  *
  * Throw an error if the host is invalid.
@@ -69,18 +73,18 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
 		return null;
 	}
 
-	if (parts.length < 3 || parts.some((p) => !p)) {
+	if (parts.length < 3) {
 		throw new Error(invalidAuthorityMessage);
 	}
 
 	// Parse from the right because safe hostnames can contain "--".
 	const hostPrefix = parts.slice(0, -2).join("--");
 	const safeHostname = hostPrefix.slice(authorityHostPrefix.length);
-	if (!safeHostname) {
-		throw new Error(invalidAuthorityMessage);
-	}
 	const username = parts[parts.length - 2];
 	const workspaceAndAgent = parts[parts.length - 1];
+	if (!safeHostname || !username || !workspaceAndAgent) {
+		throw new Error(invalidAuthorityMessage);
+	}
 
 	let workspace = workspaceAndAgent;
 	let agent = "";
@@ -95,11 +99,11 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
 	}
 
 	return {
-		agent: agent,
-		sshHost: sshHost,
-		safeHostname: safeHostname,
-		username: username,
-		workspace: workspace,
+		agent,
+		sshHost,
+		safeHostname,
+		username,
+		workspace,
 	};
 }
 
--- a/src/util.ts
+++ b/src/util.ts
@@ -56,52 +56,43 @@ export function findPort(text: string): number | null {
 export function parseRemoteAuthority(authority: string): AuthorityParts | null {
 	// The authority looks like: vscode://ssh-remote+<ssh host name>
 	const authorityParts = authority.split("+");
-
-	// We create SSH host names in a format matching:
-	// coder-vscode(--|.)<username>--<workspace>(--|.)<agent?>
-	// The agent can be omitted; the user will be prompted for it instead.
-	// Anything else is unrelated to Coder and can be ignored.
-	const parts = authorityParts[1].split("--");
-	if (
-		parts.length <= 1 ||
-		(parts[0] !== AuthorityPrefix &&
-			!parts[0].startsWith(`${AuthorityPrefix}.`))
-	) {
+	const sshHost = authorityParts[1];
+	if (!sshHost) {
 		return null;
 	}
 
-	// Reassemble Punycode labels (xn--...) the split broke apart: when the
-	// prefix ends in ".xn", the cut landed inside an "xn--..." label.
-	while (parts.length >= 2 && parts[0].endsWith(".xn")) {
-		parts.splice(0, 2, `${parts[0]}--${parts[1]}`);
+	const parts = sshHost.split("--");
+	if (!parts[0].startsWith(`${AuthorityPrefix}.`)) {
+		return null;
 	}
 
-	// It has the proper prefix, so this is probably a Coder host name.
-	// Validate the SSH host name.  Including the prefix, we expect at least
-	// three parts, or four if including the agent.
-	if ((parts.length !== 3 && parts.length !== 4) || parts.some((p) => !p)) {
-		throw new Error(
-			`Invalid Coder SSH authority. Must be: <username>--<workspace>(--|.)<agent?>`,
-		);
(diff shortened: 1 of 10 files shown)


src/util.ts
```diff
@@ -54,54 +58,47 @@ export function findPort(text: string): number | null {
...
+	// Multiple dots are ambiguous because workspace and agent share this separator.
+	if (workspaceParts.length === 2) {
+		workspace = workspaceParts[0];
+		agent = workspaceParts[1];
+		if (!workspace || !agent) {
+			throw new Error(invalidAuthorityMessage);
 		}
 	}
 
 	return {
```
code-asher, line 101: nbd at all but I suppose we could use the short form for these since all the names are identical to the properties

src/util.ts
```diff
@@ -54,54 +58,47 @@ export function findPort(text: string): number | null {
  * Throw an error if the host is invalid.
  */
 export function parseRemoteAuthority(authority: string): AuthorityParts | null {
-	// The authority looks like: vscode://ssh-remote+<ssh host name>
```
code-asher, line 57: I feel like this comment is still useful for anyone new examining this code, it is not necessarily intuitive why we split on a `+` I think.

src/util.ts
```diff
@@ -54,54 +58,47 @@ export function findPort(text: string): number | null {
...
 	const authorityParts = authority.split("+");
+	const sshHost = authorityParts[1];
+	if (!sshHost) {
+		return null;
+	}
 
-	// We create SSH host names in a format matching:
-	// coder-vscode(--|.)<username>--<workspace>(--|.)<agent?>
-	// The agent can be omitted; the user will be prompted for it instead.
-	// Anything else is unrelated to Coder and can be ignored.
```
code-asher, line 63: Same with this comment, maybe we could combine this with the other and put it in the function comment?

src/util.ts
```diff
@@ -54,54 +58,47 @@ export function findPort(text: string): number | null {
...
+	const parts = sshHost.split("--");
+	if (!parts[0].startsWith(authorityHostPrefix)) {
 		return null;
 	}
 
-	// Reassemble Punycode labels (xn--...) the split broke apart: when the
-	// prefix ends in ".xn", the cut landed inside an "xn--..." label.
-	while (parts.length >= 2 && parts[0].endsWith(".xn")) {
-		parts.splice(0, 2, `${parts[0]}--${parts[1]}`);
+	if (parts.length < 3 || parts.some((p) => !p)) {
```
code-asher: I think `foo---------------bar.com` would be a valid domain.  We probably only need to care about the username, workspace/agent, or entire domain being empty.

code-asher (approved):
wonderful

src/util.ts
EhabY: > I think foo---------------bar.com would be a valid domain

Maybe but why :sob:

src/util.ts
EhabY: True true, I'll summerize this!

src/util.ts
EhabY: Fixed!

src/util.ts
EhabY: Added jsdoc

src/util.ts
EhabY: Fixed

PR #947, opened 2026-05-12
feat: add experimental workspaces view

jakehwll:
> 🤖 This PR was modified by Coder Agents on behalf of Jake Howell.

Scaffolds a new experimental Workspaces webview panel behind the `coder.experimental.workspacesPanel` setting, giving us a blank canvas to iterate on a redesigned workspaces experience. The feature is fully gated and hidden from the Settings UI so it won't affect existing users.

- Adds a `packages/workspaces` package with a React + Vite + React Query entrypoint and Storybook support
- Registers a new `coderExperimentalWorkspaces` activity bar view container ("Coder Remote (New)") visible only when the flag is enabled and the user is authenticated
- Introduces `ExperimentalWorkspacesPanelProvider` implementing `WebviewViewProvider` with IPC dispatch scaffolding and deployment-change refresh wiring
- Gates everything behind `coder.experimental.workspacesPanel` (boolean, default `false`) — sets the `coder.experimental.webkitWorkspaces` context key to control view visibility
- Adds a `coder.workspaces.refresh` command for manual panel refresh


packages/shared/src/workspaces/api.ts
```diff
@@ -0,0 +1,28 @@
+/**
```
EhabY: I think we can keep the file but not need for the verbose comments here since there exists other examples of this `api.ts` file!

packages/workspaces/README.md
```diff
@@ -0,0 +1,27 @@
...
+**This setting is hidden from the Settings UI** - it can only be enabled via settings.json:
+
+1. Open your VS Code settings.json (Cmd/Ctrl + Shift + P → "Preferences: Open User Settings (JSON)")
+2. Add the following:
+
+```json
+{
+	"coder.experimental.workspacesPanel": true
+}
+```
```
EhabY, line 18: We can keep these instructions but remove the field from the `package.json` so it's not discoverable by users.

packages/workspaces/README.md
```diff
@@ -0,0 +1,27 @@
+# Coder Experimental Workspaces Webview Panel
+
+This package contains the experimental Workspaces webview panel for the Coder VS Code extension.
+
+This feature is currently in development and hidden from the Settings UI.
```
EhabY: This might need to get updated if we decide to change the visibility of the setting

packages/workspaces/README.md
```diff
@@ -0,0 +1,27 @@
...
+	"coder.experimental.workspacesPanel": true
+}
+```
+
+3. Reload VS Code
+
+A new activity bar icon labeled **"Coder Remote (New)"** will appear in the activity bar when the setting is enabled. This creates a completely separate panel alongside the existing "Coder Remote" and "Coder Tasks" panels, allowing easy side-by-side comparison during development.
+
+> [!NOTE]
+> The new view will only appear after you instantiate the Coder context (i.e clicking Tasks or Workspaces).
```
EhabY: I think I know why this exists, I do a similar hack for the Coder Tasks where we register a dummy empty view that shows the login button if it's not authenticated but show the webview when it is. Maybe we should do something like this?

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
+import * as vscode from "vscode";
+
+import {
+	buildCommandHandlers,
+	buildRequestHandlers,
+	WorkspacesApi,
+} from "@repo/shared";
+
+import { type CoderApi } from "../../api/coderApi";
+import { type Logger } from "../../logging/logger";
```
EhabY: Nit: those should do `import type { ... }` since it's only types

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+	dispatchRequest,
+	isIpcCommand,
+	isIpcRequest,
+} from "../dispatch";
+import { getWebviewHtml } from "../html";
+
+export class ExperimentalWorkspacesPanelProvider
+	implements vscode.WebviewViewProvider, vscode.Disposable
+{
+	public static readonly viewType = "coder.experimental.workspacesPanel";
```
EhabY: I think we can label the setting experimental but do we want the view to be named as such? I see no point tbh (same for the class name)

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	private view?: vscode.WebviewView;
+	private disposables: vscode.Disposable[] = [];
+
+	// Use buildRequestHandlers/buildCommandHandlers for compile-time safety
+	// Empty for now, but will enforce exhaustiveness when WorkspacesApi is defined
```
EhabY: Redundant since if we add any API endpoint this won't compile anyway

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+		if (isIpcRequest(message)) {
+			await dispatchRequest(message, this.requestHandlers, this.view?.webview, {
+				logger: this.logger,
+				showErrorToUser,
+			});
+		} else if (isIpcCommand(message)) {
+			await dispatchCommand(message, this.commandHandlers, {
+				logger: this.logger,
+				showErrorToUser,
+			});
```
EhabY, line 86: We should surface an error (or log it) if it's neither - Let's follow the same pattern as the tasks view or the chat view etc...

src/extension.ts
```diff
@@ -265,6 +266,38 @@ async function doActivate(
...
+	// Register Experimental Workspaces webview panel (behind configuration setting)
+	const workspacesPanelEnabled = vscode.workspace
+		.getConfiguration("coder")
+		.get<boolean>("experimental.workspacesPanel", false);
+
+	// Set context first so the view visibility is correct
+	contextManager.set(
+		"coder.experimental.webkitWorkspaces",
+		workspacesPanelEnabled,
+	);
```
EhabY: I suppose we do not want to watch the setting for now? (that seems reasonable for such a dev feature for now)

Also pls trim/remove comments that are explained by code 🙏

src/core/contextManager.ts
```diff
@@ -7,6 +7,7 @@ const CONTEXT_DEFAULTS = {
 	"coder.agentsEnabled": false,
 	"coder.workspace.connected": false,
 	"coder.workspace.updatable": false,
+	"coder.experimental.webkitWorkspaces": false,
```
EhabY: Maybe something like `coder.workspacesPanelEnabled` or `coder.workspaces.usePanel`?

package.json
```diff
@@ -149,6 +149,17 @@
 						"experimental"
 					]
 				},
+				"coder.experimental.workspacesPanel": {
```
EhabY: Pls remove this from here but we VS Code still reads the `settings.json` even if it's not defined in the `package.json` as a setting!

package.json
```diff
@@ -326,6 +342,15 @@
 					"icon": "media/shorthand-logo.svg",
 					"when": "coder.agentsEnabled"
 				}
+			],
+			"coderExperimentalWorkspaces": [
+				{
+					"type": "webview",
+					"id": "coder.experimental.workspacesPanel",
+					"name": "Workspaces",
+					"icon": "media/logo-white.svg",
```
EhabY: Shouldn't this be `"media/shorthand-logo.svg"` since we have one view within this view container anyway? Like I think this would only render when we move the view OUT of the view container so maybe it's fine

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+		webviewView: vscode.WebviewView,
+		_context: vscode.WebviewViewResolveContext,
+		_token: vscode.CancellationToken,
+	): void {
+		this.view = webviewView;
+
+		webviewView.webview.options = {
+			enableScripts: true,
+			localResourceRoots: [
+				vscode.Uri.joinPath(this.extensionUri, "dist", "webviews"),
```
EhabY: Should be tightened to `dist/webviews/workspaces`

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+		webviewView.onDidDispose(() => {
+			for (const d of this.disposables) {
+				d.dispose();
+			}
+			this.disposables = [];
+		});
+	}
+
+	private async handleMessage(message: unknown): Promise<void> {
+		const showErrorToUser = () => false;
```
EhabY: This is optional anyway so we can just drop it

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+			this.extensionUri,
+			"workspaces",
+			"Coder Workspaces",
+		);
+
+		webviewView.onDidDispose(() => {
+			for (const d of this.disposables) {
+				d.dispose();
+			}
+			this.disposables = [];
```
EhabY: Cleanup logic is duplicated three times (:57-60, 77-82, 100-105), maybe refactor into `disposeView` helper

src/webviews/workspaces/workspacesPanelProvider.ts
```diff
@@ -0,0 +1,106 @@
...
+	private disposables: vscode.Disposable[] = [];
+
+	// Use buildRequestHandlers/buildCommandHandlers for compile-time safety
+	// Empty for now, but will enforce exhaustiveness when WorkspacesApi is defined
+	private readonly requestHandlers = buildRequestHandlers(WorkspacesApi, {});
+	private readonly commandHandlers = buildCommandHandlers(WorkspacesApi, {});
+
+	constructor(
+		private readonly extensionUri: vscode.Uri,
+		private readonly client: CoderApi,
```
EhabY: I say we only add this param when needed, for example, the chat panel picks only the host and token so tests can easily fake this part

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Fixed — `Logger` now uses `import type`. Lint autofix also reordered it to the correct position.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Renamed to `WorkspacesPanelProvider` and `coder.workspacesPanel` for the view type.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Added `else` branch: `this.logger.warn("Unexpected webview message", message);`

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Dropped — matches the chat panel pattern.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Extracted `disposeView()` helper — used by `resolveWebviewView`, `onDidDispose`, and `dispose()`. Follows the `chatPanelProvider` pattern.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

packages/shared/src/workspaces/api.ts
jakehwll: Stripped it down to just `export const WorkspacesApi = {} as const;`

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

package.json
jakehwll: Removed.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Removed `client: CoderApi` from the constructor — will add back when needed.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

package.json
jakehwll: Changed to `media/shorthand-logo.svg`.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

packages/workspaces/README.md
jakehwll: Noted — didn't add the login view pattern in this pass since it's more of a follow-up concern, but it's a good idea. Could be a separate PR once we flesh out the panel.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/core/contextManager.ts
jakehwll: Renamed to `coder.workspacesPanelEnabled` — updated in `contextManager.ts`, `extension.ts`, and the `when` clause in `package.json`.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/extension.ts
jakehwll: Removed the comments — the code is self-explanatory. Agreed not watching the setting is fine for now.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

packages/workspaces/README.md
jakehwll: Removed the `coder.experimental.workspacesPanel` definition from `package.json`. VS Code still reads it from `settings.json` directly.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

packages/workspaces/README.md
jakehwll: Removed the "hidden from the Settings UI" line — it was an implementation detail that would go stale.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Tightened to `vscode.Uri.joinPath(this.extensionUri, "dist", "webviews", "workspaces")`.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

src/webviews/workspaces/workspacesPanelProvider.ts
jakehwll: Removed — the `buildRequestHandlers`/`buildCommandHandlers` calls speak for themselves.

_🤖 Generated with [Coder Agents](https://coder.com/agents)_

EhabY (approved):
Nice! now we can iterate on this view without disrupting users

PR #956, opened 2026-05-15
Add `alternativeWebUrl` param

ibdafna:
This PR adds a new parameter for specifying an alternative URL to use when opening Coder pages in the browser. When set, it replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page). The connection URL is still used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443)."

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,11 +134,11 @@
 					"default": "",
 					"scope": "application"
 				},
-				"coder.alternativeWebUrl": {
-					"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page). The connection URL is still used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
-					"type": "string",
-					"default": ""
-				},
+			"coder.alternativeWebUrl": {
+				"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page, OAuth authorization page). The connection URL is still used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
+				"type": "string",
+				"default": ""
+			},
 				"coder.autologin": {
 					"markdownDescription": "Automatically log into the default URL when the extension is activated. coder.defaultUrl is preferred, otherwise the CODER_URL environment variable will be used. This setting has no effect if neither is set.",
 					"type": "boolean",
--- a/package.json
+++ b/package.json
@@ -137,7 +137,8 @@
 				"coder.alternativeWebUrl": {
 					"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page, OAuth authorization page). When empty, the connection URL is used for the UI as well. The connection URL is always used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
 					"type": "string",
-					"default": ""
+					"default": "",
+					"scope": "application"
 				},
 				"coder.autologin": {
 					"markdownDescription": "Automatically log into the default URL when the extension is activated. coder.defaultUrl is preferred, otherwise the CODER_URL environment variable will be used. This setting has no effect if neither is set.",
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
 					"scope": "application"
 				},
 				"coder.alternativeWebUrl": {
-					"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page, OAuth authorization page). When empty, the connection URL is used for the UI as well. The connection URL is always used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
+					"markdownDescription": "Alternative URL for opening Coder pages in the browser (dashboard, workspace, and login pages). API, SSH, and CLI always use the connection URL; when this is empty, so does the browser. Useful when the API runs on a browser-restricted port (e.g. 7004) but the web UI is served on a standard port (e.g. 443).",
 					"type": "string",
 					"default": "",
 					"scope": "application"
--- a/package.json
+++ b/package.json
@@ -134,6 +134,11 @@
 					"default": "",
 					"scope": "application"
 				},
+				"coder.alternativeWebUrl": {
+					"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page). The connection URL is still used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
+					"type": "string",
+					"default": ""
+				},
 				"coder.autologin": {
 					"markdownDescription": "Automatically log into the default URL when the extension is activated. coder.defaultUrl is preferred, otherwise the CODER_URL environment variable will be used. This setting has no effect if neither is set.",
 					"type": "boolean",
--- a/package.json
+++ b/package.json
@@ -134,11 +134,11 @@
 					"default": "",
 					"scope": "application"
 				},
-			"coder.alternativeWebUrl": {
-				"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page, OAuth authorization page). The connection URL is still used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
-				"type": "string",
-				"default": ""
-			},
+				"coder.alternativeWebUrl": {
+					"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page, OAuth authorization page). When empty, the connection URL is used for the UI as well. The connection URL is always used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
+					"type": "string",
+					"default": ""
+				},
 				"coder.autologin": {
 					"markdownDescription": "Automatically log into the default URL when the extension is activated. coder.defaultUrl is preferred, otherwise the CODER_URL environment variable will be used. This setting has no effect if neither is set.",
 					"type": "boolean",
--- a/src/login/loginCoordinator.ts
+++ b/src/login/loginCoordinator.ts
@@ -10,7 +10,7 @@ import { buildOAuthTokenData } from "../oauth/utils";
 import { withOptionalProgress } from "../progress";
 import { maybeAskAuthMethod, maybeAskUrl } from "../promptUtils";
 import { isKeyringEnabled } from "../settings/cli";
-import { resolveBrowserUrl } from "../util";
+import { openInBrowser } from "../util";
 import { vscodeProposed } from "../vscodeProposed";
 
 import type { User } from "coder/site/src/api/typesGenerated";
@@ -399,9 +399,7 @@ export class LoginCoordinator implements vscode.Disposable {
 		}
 		// This prompt is for convenience; do not error if they close it since
 		// they may already have a token or already have the page opened.
-		await vscode.env.openExternal(
-			vscode.Uri.parse(`${resolveBrowserUrl(url)}/cli-auth`),
-		);
+		await openInBrowser(url, "/cli-auth");
 
 		// For token auth, start with the existing token in the prompt or the last
 		// used token.  Once submitted, if there is a failure we will keep asking
--- a/src/login/loginCoordinator.ts
+++ b/src/login/loginCoordinator.ts
@@ -10,6 +10,7 @@ import { buildOAuthTokenData } from "../oauth/utils";
 import { withOptionalProgress } from "../progress";
 import { maybeAskAuthMethod, maybeAskUrl } from "../promptUtils";
 import { isKeyringEnabled } from "../settings/cli";
+import { resolveBrowserUrl } from "../util";
 import { vscodeProposed } from "../vscodeProposed";
 
 import type { User } from "coder/site/src/api/typesGenerated";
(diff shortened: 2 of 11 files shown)


src/webviews/chat/chatPanelProvider.ts
```diff
@@ -154,7 +155,12 @@ export class ChatPanelProvider
 			const resolved = new URL(url, coderUrl);
 			const expected = new URL(coderUrl);
 			if (resolved.origin === expected.origin) {
-				void vscode.env.openExternal(vscode.Uri.parse(resolved.toString()));
+				const browserBase = resolveBrowserUrl(coderUrl);
+				const browserUrl = new URL(
+					resolved.pathname + resolved.search + resolved.hash,
+					browserBase,
+				);
```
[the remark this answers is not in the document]
ibdafna: Addressed!

src/login/loginCoordinator.ts
```diff
@@ -361,7 +362,9 @@ export class LoginCoordinator implements vscode.Disposable {
 		}
 		// This prompt is for convenience; do not error if they close it since
 		// they may already have a token or already have the page opened.
-		await vscode.env.openExternal(vscode.Uri.parse(`${url}/cli-auth`));
+		await vscode.env.openExternal(
+			vscode.Uri.parse(`${resolveBrowserUrl(url)}/cli-auth`),
+		);
```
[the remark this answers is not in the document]
ibdafna: We didn't touch the OAuth path deliberately, but I think the case highlighted here is worth addressing so we have consistency. Will submit a fix shortly.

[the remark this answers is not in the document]
ibdafna: Addressed!

src/util.ts
```diff
@@ -202,23 +203,25 @@ export async function renameWithRetry(
 	}
 }
 
+/**
+ * Return the URL for opening Coder pages in the browser.  Uses the
+ * `coder.alternativeWebUrl` setting when configured, otherwise returns
+ * the connection URL unchanged.
+ */
+export function resolveBrowserUrl(connectionUrl: string): string {
```
EhabY: `resolveUiUrl` makes more sense to me and fits the setting better `coder.alternativeWebUrl`

src/oauth/authorizer.ts
```diff
@@ -201,7 +202,13 @@ export class OAuthAuthorizer implements vscode.Disposable {
 			code_challenge_method: PKCE_CHALLENGE_METHOD,
 		});
 
-		const url = `${metadata.authorization_endpoint}?${params.toString()}`;
+		// The server-advertised endpoint is authoritative for the path, but the
+		// origin may be unreachable from a browser (e.g. blocked port). When
+		// `coder.alternativeWebUrl` is set, swap the origin so the user lands on
+		// a reachable host while preserving the path the server told us to use.
```
EhabY: Agents tend to over explain, I think this can be trimmed or removed entirely

src/webviews/chat/chatPanelProvider.ts
```diff
@@ -154,7 +155,12 @@ export class ChatPanelProvider
 			const resolved = new URL(url, coderUrl);
 			const expected = new URL(coderUrl);
 			if (resolved.origin === expected.origin) {
-				void vscode.env.openExternal(vscode.Uri.parse(resolved.toString()));
+				const browserBase = resolveBrowserUrl(coderUrl);
+				// Concatenate rather than `new URL(path, base)` so a path prefix on
+				// the alternative URL (e.g. a reverse proxy at https://host/coder)
+				// is preserved.
```
EhabY: // Preserve the server-advertised path; only the origin/prefix may change.

src/webviews/tasks/tasksPanelProvider.ts
```diff
@@ -308,19 +309,21 @@ export class TasksPanelProvider
...
 
 	private async handleViewLogs(taskId: string): Promise<void> {
-		const baseUrl = this.client.getHost();
-		if (!baseUrl) return;
+		const connUrl = this.client.getHost();
+		if (!connUrl) return;
 
+		const baseUrl = resolveBrowserUrl(connUrl);
 		const task = await this.client.getTask("me", taskId);
 		vscode.env.openExternal(vscode.Uri.parse(getTaskBuildUrl(baseUrl, task)));
```
EhabY: Instead of always doing `resolveXUrl` then `vscode.open` or `vscode.env.openExternal`, how about we just add `openInBrowser(connectionUrl, path)` helper

test/unit/webviews/chat/chatPanelProvider.test.ts
```diff
@@ -87,6 +87,13 @@ describe("ChatPanelProvider", () => {
 	beforeEach(() => {
 		vi.resetAllMocks();
 		windowMock.__setActiveColorThemeKind(vscode.ColorThemeKind.Dark);
+
+		vi.mocked(vscode.workspace.getConfiguration).mockReturnValue({
+			get: vi.fn(),
```
EhabY: Do use `MockConfigurationProvider` instead of mocking this directly (here and in `tasksPanelProvider.test.ts`

package.json
```diff
@@ -121,6 +121,11 @@
 					"type": "string",
 					"default": ""
 				},
+			"coder.alternativeWebUrl": {
+				"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page, OAuth authorization page). The connection URL is still used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
```
EhabY: Do add something like "when empty, the connection URL is used for UI". I do think we need to be clear that this only affects calls like `vscode.env.openExternal` (without mentioning this here of course)

EhabY (commented):
Nice improvement. Make sure to run `pnpm format` and `pnpm lint:fix`/`pnpm typecheck`!

src/webviews/tasks/tasksPanelProvider.ts
EhabY: We could even do it in a safer way like `vscode.Uri.joinPath + Uri.with({ query, fragment })` so that we do not have to use some regex magic and concatenation

src/oauth/authorizer.ts
```diff
@@ -201,7 +202,13 @@ export class OAuthAuthorizer implements vscode.Disposable {
...
+		const endpoint = new URL(metadata.authorization_endpoint);
+		const browserBase = resolveBrowserUrl(endpoint.origin);
+		const url = `${browserBase}${endpoint.pathname}?${params.toString()}`;
```
EhabY: The old code was:

```ts
const url = `${metadata.authorization_endpoint}?${params.toString()}`;
```

The new code drops `endpoint.search` and passes `endpoint.origin` (not a full connection URL) into the helper, which is inconsistent with every other call site. Preserve the original shape and just swap the origin:

```ts
const endpoint = new URL(metadata.authorization_endpoint);
const browserOrigin = new URL(resolveUiUrl(coderApi.getHost())).origin;
endpoint.protocol = new URL(browserOrigin).protocol;
endpoint.host = new URL(browserOrigin).host;
for (const [key, value] of Object.entries(params)) {
    endpoint.searchParams.set(key, value);
}
const url = endpoint.toString();
```

Or, if the helper is as per the suggestion above, just call the open helper with the endpoint pathname and the params as the query. Either way, query strings already on `authorization_endpoint` are preserved and the helper receives a real connection URL

src/webviews/chat/chatPanelProvider.ts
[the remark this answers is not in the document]
EhabY: The file has been removed.

src/login/loginCoordinator.ts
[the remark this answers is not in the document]
EhabY: Addressed. OAuth login now uses the alternative URL too.

src/util.ts
EhabY: Renamed to `resolveUiUrl`

src/oauth/authorizer.ts
EhabY: Trimmed

src/webviews/chat/chatPanelProvider.ts
EhabY: The file has been removed.

src/webviews/tasks/tasksPanelProvider.ts
EhabY: Added an `openInBrowser` helper using `Uri.joinPath`

test/unit/webviews/chat/chatPanelProvider.test.ts
EhabY: Switched to `MockConfigurationProvider`

package.json
EhabY: Updated the description

src/oauth/authorizer.ts
EhabY: Query and origin handling look good now. Thanks!

src/oauth/authorizer.ts
```diff
@@ -205,25 +208,36 @@ export class OAuthAuthorizer implements vscode.Disposable {
...
-		const url = `${metadata.authorization_endpoint}?${params.toString()}`;
+		// Server is authoritative for the path; alternativeWebUrl can swap the origin.
+		const endpoint = new URL(metadata.authorization_endpoint);
+		const browserBase = new URL(resolveUiUrl(connectionUrl));
+		endpoint.protocol = browserBase.protocol;
+		endpoint.hostname = browserBase.hostname;
+		endpoint.port = browserBase.port;
+		// Preserve any path prefix on the alternative URL (e.g. reverse proxy).
+		const prefix = browserBase.pathname.replace(/\/$/, "");
+		endpoint.pathname = `${prefix}${endpoint.pathname}`;
```
EhabY: This doubles the path prefix on sub-path deployments. If the connection URL has a path (e.g. `https://example.com/coder`), the server already returns `authorization_endpoint` with that prefix, so prepending `browserBase.pathname` yields `/coder/coder/oauth2/authorize`.

This is a regression: OAuth breaks on sub-path deployments even when `alternativeWebUrl` is unset. Stripping the connection prefix from the endpoint path before applying the UI prefix would fix it.

package.json
```diff
@@ -134,6 +134,11 @@
 					"default": "",
 					"scope": "application"
 				},
+				"coder.alternativeWebUrl": {
+					"markdownDescription": "An alternative URL to use when opening Coder pages in the browser. When set, this replaces the connection URL for browser links only (dashboard, workspace pages, token authentication page, OAuth authorization page). When empty, the connection URL is used for the UI as well. The connection URL is always used for API calls, SSH, and CLI operations. Useful when the Coder API runs on a port that browsers restrict (e.g., 7004) but the web UI is accessible on a standard port (e.g., 443).",
+					"type": "string",
+					"default": ""
```
EhabY: This is missing `scope: "application"` similar to other settings (security issue)

test/unit/util.test.ts
```diff
@@ -397,3 +402,124 @@ describe("findPort", () => {
...
+describe("resolveUiUrl", () => {
+	let configurationProvider: MockConfigurationProvider;
+
+	beforeEach(() => {
+		configurationProvider = new MockConfigurationProvider();
+	});
+
+	afterEach(() => {
+		vi.mocked(vscode.workspace.getConfiguration).mockReset();
+	});
```
EhabY: We do not need to go around the configuration provider, also we can just create a `setup` function that returns a configuration provider and called inside the test itself (this pattern exists in other tests). That way, we do not have to do an `afterEach` nor a `beforeEach`

test/unit/util.test.ts
```diff
@@ -397,3 +402,124 @@ describe("findPort", () => {
...
+	let configurationProvider: MockConfigurationProvider;
+
+	beforeEach(() => {
+		configurationProvider = new MockConfigurationProvider();
+		vi.mocked(vscode.env.openExternal).mockClear();
+	});
+
+	afterEach(() => {
+		vi.mocked(vscode.workspace.getConfiguration).mockReset();
+	});
```
EhabY: Same here

src/oauth/authorizer.ts
```diff
@@ -205,25 +208,48 @@ export class OAuthAuthorizer implements vscode.Disposable {
...
+			connectionPrefix &&
+			(endpointPath === connectionPrefix ||
+				endpointPath.startsWith(`${connectionPrefix}/`))
+		) {
+			endpointPath = endpointPath.slice(connectionPrefix.length);
+		}
+		endpoint.pathname = `${browserPrefix}${endpointPath}`;
+		for (const [key, value] of Object.entries(params)) {
+			endpoint.searchParams.set(key, value);
+		}
```
EhabY, line 227: This could be something like:

```ts
const params = new URLSearchParams({
	client_id: clientId,
	response_type: RESPONSE_TYPE,
	redirect_uri: this.getRedirectUri(),
	scope: DEFAULT_OAUTH_SCOPES,
	state,
	code_challenge: challenge,
	code_challenge_method: PKCE_CHALLENGE_METHOD,
});

const endpoint = new URL(metadata.authorization_endpoint);
const connectionBase = new URL(connectionUrl);
const browserBase = new URL(resolveUiUrl(connectionUrl));
const connectionPrefix = connectionBase.pathname.replace(/\/$/, "");
const browserPrefix = browserBase.pathname.replace(/\/$/, "");
// Only rebase the endpoint when it lives under the connection URL; an
// endpoint advertised on a different host is left untouched.
const underConnection =
	endpoint.origin === connectionBase.origin &&
	(connectionPrefix === "" ||
		endpoint.pathname === connectionPrefix ||
		endpoint.pathname.startsWith(`${connectionPrefix}/`));
if (underConnection) {
	const rest = endpoint.pathname.slice(connectionPrefix.length);
	endpoint.protocol = browserBase.protocol;
	endpoint.hostname = browserBase.hostname;
	endpoint.port = browserBase.port;
	endpoint.pathname = `${browserPrefix}${rest}`;
}
// Merge onto existing query params (URLSearchParams, iterable, no Object.entries).
for (const [key, value] of params) {
	endpoint.searchParams.set(key, value);
}
```

Since the authorization URI can technically use a different origin, also we should trim the comments a bit and use `URLSearchParams` as before if possible

src/util.ts
```diff
@@ -195,3 +196,30 @@ export function escapeShellArg(arg: string): string {
...
+/**
+ * Open a path on the Coder deployment in the user's browser, applying
+ * `coder.alternativeWebUrl` when configured.
+ */
+export function openInBrowser(
+	connectionUrl: string,
+	path: string,
+): Thenable<boolean> {
+	const base = vscode.Uri.parse(resolveUiUrl(connectionUrl));
+	const segment = path.replace(/^\/+/, "");
```
EhabY: We can drop this since `Uri.joinPath` resolves leading/trailing slashes correctly anyway

src/oauth/authorizer.ts
EhabY: Done

package.json
EhabY: Done

test/unit/util.test.ts
EhabY: Done

test/unit/util.test.ts
EhabY: Done

src/oauth/authorizer.ts
EhabY: Done

src/util.ts
EhabY: Done

EhabY (approved):

PR #957, opened 2026-05-15
feat: shared workspaces support

aslilac:
Adds a third tree view, **Shared Workspaces**, that lists workspaces the current user has access to but does not own. The view is visible to any authenticated user, sits between **My Workspaces** and **All Workspaces**, and reuses the existing context-menu actions and refresh/search affordances.

<details>
<summary>Implementation notes</summary>

- New `WorkspaceQuery.Shared = "shared:true"` value; the provider filters out workspaces owned by the current user client-side, because `shared:true` server-side also matches workspaces the current user shared out.
- `DeploymentManager` exposes `getCurrentUserId()` so the new provider can scope the filter to the active session; this is cleared on logout/suspend.
- `coder.refreshWorkspaces` now also refreshes the new view.

</details>

> Generated with Coder Agents on behalf of @aslilac.

### DIFF
--- a/src/deployment/deploymentManager.ts
+++ b/src/deployment/deploymentManager.ts
@@ -37,6 +37,7 @@ export class DeploymentManager implements vscode.Disposable {
 	private readonly telemetryService: TelemetryService;
 
 	#deployment: Deployment | null = null;
+	#currentUserId: string | undefined;
 	#authListenerDisposable: vscode.Disposable | undefined;
 	#crossWindowSyncDisposable: vscode.Disposable | undefined;
 
@@ -84,6 +85,14 @@ export class DeploymentManager implements vscode.Disposable {
 	}
 
 	/**
+	 * Get the id of the currently authenticated user, if any. Used by the
+	 * Shared workspaces view to filter out the current user's workspaces.
+	 */
+	public getCurrentUserId(): string | undefined {
+		return this.#currentUserId;
+	}
+
+	/**
 	 * Attempt to change to a deployment after validating authentication.
 	 * Only changes deployment if authentication succeeds.
 	 * Returns true if deployment was changed, false otherwise.
@@ -127,6 +136,7 @@ export class DeploymentManager implements vscode.Disposable {
 			user: deployment.user.username,
 		});
 		this.#deployment = { ...deployment };
+		this.#currentUserId = deployment.user.id;
 		this.telemetryService.setDeploymentUrl(deployment.url);
 
 		// Updates client credentials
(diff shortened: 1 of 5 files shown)


src/workspace/workspacesProvider.ts
```diff
@@ -23,6 +23,11 @@ import { type Logger } from "../logging/logger";
 export enum WorkspaceQuery {
 	Mine = "owner:me",
 	All = "",
+	// Shared returns workspaces the user has access to via sharing but does not
+	// own. The server-side `shared:true` filter also includes workspaces the
+	// user owns and has shared out, so the provider filters those out
+	// client-side using the current user's id.
```
EhabY, line 29: Agents tend to over explain with what and not why. I'd trim this a bit more (same for all of the new comments here)

src/workspace/workspacesProvider.ts
```diff
@@ -53,6 +58,10 @@ export class WorkspaceProvider
 		private readonly logger: Logger,
 		private readonly isAuthenticated: () => boolean,
 		private readonly timerSeconds?: number,
+		// Returns the id of the currently authenticated user. Used by the Shared
+		// query to filter out workspaces owned by the current user.
+		private readonly getCurrentUserId: () => string | undefined = () =>
+			undefined,
```
EhabY, line 64: I'd rather decouple this, like maybe pass `filterWorkspaces?: (ws: Workspace[]) => Workspace[]` and remove any knowledge of the current user from here

src/deployment/deploymentManager.ts
```diff
@@ -37,6 +37,7 @@ export class DeploymentManager implements vscode.Disposable {
 	private readonly telemetryService: TelemetryService;
 
 	#deployment: Deployment | null = null;
+	#currentUserId: string | undefined;
```
EhabY, line 40: Maybe call it `#authedUser: User | null` so it's more explicit, we can keep the derived `getCurrentUserId`

src/extension.ts
```diff
@@ -174,6 +175,19 @@ async function doActivate(
 	);
 	ctx.subscriptions.push(allWorkspacesProvider);
 
+	const sharedWorkspacesProvider = new WorkspaceProvider(
+		WorkspaceQuery.Shared,
+		client,
+		output,
+		isAuthenticated,
+		undefined,
```
EhabY, line 183: Do we want to poll for refreshes or keep it like `All` and only on manual refreshes?

src/workspace/workspacesProvider.ts
```diff
@@ -112,6 +121,18 @@ export class WorkspaceProvider
...
+		// keep only the ones owned by someone else.
+		let workspaces = resp.workspaces;
+		if (this.getWorkspacesQuery === WorkspaceQuery.Shared) {
+			const currentUserId = this.getCurrentUserId();
+			if (currentUserId) {
+				workspaces = workspaces.filter(
+					(workspace) => workspace.owner_id !== currentUserId,
+				);
+			}
+		}
```
EhabY, line 134: If we make this a filter then we can just make it a `const` and should be cleaner (separation of concerns)

EhabY:
Closing this in favor of https://github.com/coder/vscode-coder/pull/991

Issue #903, opened 2026-04-17
Telemetry: coder.exportTelemetry command (JSON array and OTLP/JSON)


EhabY:
Part of the Telemetry Phase A rollout. See the RFC in Linear: AIGOV-154.

## Scope

- Register the `coder.exportTelemetry` command via the dispatcher from #902.
- Date range selection via `QuickPick` with presets ("Last 24 hours", "Last 7 days", "Last 30 days", "All time", "Custom range…"). Custom range falls back to two `InputBox` prompts with `validateInput` for `YYYY-MM-DD` dates.
- File filtering by filename only, since daily files are date-stamped. No file contents are opened outside the selected range.
- Format selection via a second `QuickPick`:
  - JSON array (default): a single JSON document for human inspection or compliance review.
  - OTLP/JSON: OpenTelemetry Logs data model wrapped in `{ "resourceLogs": [...] }`. Resource attributes are lifted from the stored `context`; each event becomes one `LogRecord` with `body`, `attributes`, and `observedTimeUnixNano`. Can be POSTed unchanged to any OTLP/HTTP endpoint.
- Streaming output via `readline.createInterface` + `createReadStream`. Memory usage stays proportional to a single line, not total file size.
- `vscode.window.showSaveDialog` with suggested filename `coder-telemetry-{range}.{ext}` and a filter matching the chosen format.

## Tests

- Unit tests for date range parsing and the filename-based file filter.
- Unit tests for the JSON-array writer (header, comma-separated lines, footer).
- Unit tests for the OTLP/JSON writer (envelope structure, `LogRecord` shape, resource attribute mapping from `context`).
- Memfs-backed integration test that seeds a few JSONL files, exports a range, and verifies the output.

## Out of scope

- No Prometheus endpoint. Prometheus cannot ingest event files directly. Users who want Prometheus feed the OTLP/JSON output into an OTel Collector and let it forward via the `prometheusremotewrite` exporter.

Depends on #900, #901, #902.

PR #959, opened 2026-05-17
feat(telemetry): add wire-format module and export event reader

EhabY:
## Summary

- Introduce `src/telemetry/wireFormat.ts` as the single source of truth for the on-disk JSONL telemetry shape. `SessionContext`, `TelemetryContext`, and `TelemetryEvent` are now `DeepReadonly<z.infer<...>>` derived from schemas, so the wire format and the TS types cannot drift.
- `serializeTelemetryEvent` and `parseTelemetryEventLine` are shared by the sink (writer) and the exporter (reader); the sink's hand-rolled serializer is gone.
- Add `listTelemetryFilesForRange` (filename-date filtering, chronological order) and `streamTelemetryEvents` (line-by-line `AsyncIterable<TelemetryEvent>` with `TelemetryFileParseError` for parse failures).
- Add `range.ts` with `createPresetDateRange`, `createCustomDateRange`, `validateUtcDateInput`, `isTimestampInRange`, and `fileDateCanContainRangeEvent`. Preset table is the source of truth for `TelemetryRangePresetId` and the UI-facing `TELEMETRY_RANGE_PRESETS`.

Refs #903.

Stack: 1 / 4. Next: #960.

## Notable design choices

- Schema is the source of truth: `SessionContext`/`TelemetryContext`/`TelemetryEvent` come from `z.infer<...>` wrapped in a local `DeepReadonly<T>` so nested fields (including `context`, `properties`, `measurements`, `error`) stay readonly all the way down. `event.ts` re-exports the types for back-compat.
- Wire and in-memory shapes are kept aligned by a single tiny `wireToCamel` helper that renames structural keys (top-level + `context`); `properties` and `measurements` pass through verbatim since they hold caller-supplied keys.
- Parser tags failures with `TelemetryFileParseError` so stream readers can distinguish parse errors from IO errors without sniffing message text.
- Filename pattern is loose on the session slug but anchored on the date (which is all we need for filtering); the comparator orders by (date, session, part).
- Half-open UTC ranges; `fileDateCanContainRangeEvent` is a coarse calendar-day filter so files outside the window are skipped without being opened.

## Tests

- New `test/unit/telemetry/wireFormat.test.ts` covers serializer behavior (snake_case top level and context, optional-field omission, arbitrary properties/measurements keys), round-trip serialize→parse, `TelemetryFileParseError` tagging with `source:line`, and rejection of invalid ISO timestamps and missing structural fields.
- `test/unit/telemetry/export/files.test.ts` (memfs-backed) covers date filtering, missing-directory handling, timestamp-range streaming, and parse-error surfacing with file:line context.
- `test/unit/telemetry/export/range.test.ts` covers preset and custom range construction, UTC validation, and the calendar-day pre-filter.
- `test/unit/telemetry/sinks/localJsonlSink.test.ts` migrated to the shared `createTelemetryEventFactory` and no longer maintains its own event fixture.

### DIFF
--- a/src/telemetry/export/files.ts
+++ b/src/telemetry/export/files.ts
@@ -10,7 +10,7 @@ import {
 	type TelemetryDateRange,
 } from "./range";
 
-import type { ExportTelemetryEvent } from "./types";
+import type { TelemetryEvent } from "../event";
 
 const TELEMETRY_FILE_PATTERN =
 	/^telemetry-(\d{4}-\d{2}-\d{2})-([a-zA-Z0-9]+)(?:\.(\d+))?\.jsonl$/;
@@ -44,8 +44,8 @@ const StoredTelemetryEventSchema = z.object({
 		.optional(),
 });
 
-const ExportTelemetryEventSchema = StoredTelemetryEventSchema.transform(
-	(event): ExportTelemetryEvent => ({
+const TelemetryEventSchema = StoredTelemetryEventSchema.transform(
+	(event): TelemetryEvent => ({
 		eventId: event.event_id,
 		eventName: event.event_name,
 		timestamp: event.timestamp,
@@ -135,7 +135,7 @@ function compareTelemetryFiles(
 export async function* readTelemetryEvents(
 	filePaths: readonly string[],
 	range: TelemetryDateRange,
-): AsyncGenerator<ExportTelemetryEvent> {
+): AsyncGenerator<TelemetryEvent> {
 	for (const filePath of filePaths) {
 		let lineNumber = 0;
 		const lines = readline.createInterface({
@@ -163,9 +163,9 @@ export function parseStoredTelemetryEvent(
 	line: string,
 	filePath = "<telemetry>",
 	lineNumber = 1,
-): ExportTelemetryEvent {
+): TelemetryEvent {
 	try {
-		return ExportTelemetryEventSchema.parse(JSON.parse(line));
+		return TelemetryEventSchema.parse(JSON.parse(line));
 	} catch (err) {
 		throw new Error(
 			`Failed to parse telemetry file ${path.basename(filePath)}:${lineNumber}: ${errorMessage(err)}`,
@@ -175,7 +175,7 @@ export function parseStoredTelemetryEvent(
 }
 
 export function toStoredTelemetryEvent(
-	event: ExportTelemetryEvent,
+	event: TelemetryEvent,
 ): StoredTelemetryEvent {
 	return {
 		event_id: event.eventId,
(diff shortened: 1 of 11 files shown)


src/telemetry/export/files.ts
```diff
@@ -0,0 +1,125 @@
...
+				throw err;
+			}
+			const at = lineNumber > 0 ? `:${lineNumber}` : "";
+			throw new Error(
+				`Failed to read telemetry file ${name}${at}: ${toError(err).message}`,
+				{ cause: err },
+			);
+		} finally {
+			lines.close();
+			stream.destroy();
```
mtojek: is it mandatory since "lines" is closed?

src/telemetry/export/range.ts
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * UTC range that includes the full 24h of `endDate`; `endMs` lands at
+ * exclusive midnight of the day after.
+ */
+export function createCustomDateRange(
+	startDate: string,
+	endDate: string,
+): TelemetryDateRange {
+	const startMs = parseUtcDate(startDate);
+	const endStartMs = parseUtcDate(endDate);
```
mtojek: nit: hmm.. endDateMs?

src/telemetry/sinks/localJsonlSink.ts
```diff
@@ -315,30 +316,5 @@ function toSessionSlug(sessionId: string): string {
...
-			properties: event.properties,
-			measurements: event.measurements,
-			...(event.traceId !== undefined && { trace_id: event.traceId }),
-			...(event.parentEventId !== undefined && {
-				parent_event_id: event.parentEventId,
-			}),
-			...(event.error !== undefined && { error: event.error }),
-		}) + "\n"
-	);
+	return JSON.stringify(serializeTelemetryEvent(event)) + "\n";
```
mtojek: nit: should we move `"\n"` up?

mtojek (approved):
👍

src/telemetry/export/files.ts
EhabY: Kept, and made it unconditional. `readline.Interface#close()` doesn't destroy the underlying `ReadStream`, so the fd would leak (especially on early-exit from the generator). Also wrapped in nested try/finally so `stream.destroy()` still runs if `lines.close()` throws from a `'close'` listener

src/telemetry/export/range.ts
EhabY: Renamed `endStartMs` → `endDateMs`, and `startMs` → `startDateMs` for symmetry

src/telemetry/sinks/localJsonlSink.ts
EhabY: Added `serializeTelemetryEventLine` in `wireFormat.ts` (mirrors `parseTelemetryEventLine`). Sink and the export test call it directly; the local `serializeEvent` and `wireLine` helpers are gone

PR #962, opened 2026-05-18
feat(telemetry): instrument auth recovery

EhabY:
## Summary

Records local telemetry for the OAuth and 401-recovery paths so latency and failure modes show up alongside the rest of the local telemetry stream.

Auth events (`src/instrumentation/auth.ts`):

- `auth.token_refreshed`: OAuth refresh attempt with `trigger` (`background` / `reactive`).
- `auth.token_refresh.deduped`: emitted when a refresh call joins an in-flight refresh so the dropped trigger stays countable.
- `auth.unauthorized_intercepted`: recovery path for a 401, with `recovery` (`refresh_success` / `login_required` / `none`) and `refreshAttempted`.
- `auth.login_prompted`: modal login prompt outcome, with `trigger` and a `reason` on abort/failure.

Alongside the auth instrumentation, the shared telemetry plumbing picks up a `measurements` argument on `TelemetryService.trace`, span outcome helpers (`markFailure` / `markAborted`), typed properties on `Span`, and test helpers (`enableLocalTelemetry`, `TestSink.expectOne`, `createMockServiceContainer`). The SSH and WebSocket instrumentation modules migrate to the typed property API.

## Split

First of two stacked PRs replacing #948 (closing #906 in two pieces). The workspace half is #963 and depends on this PR — its `remote.ts` diff stacks on the auth-side diff here.

Issue #954, opened 2026-05-13
Audit telemetry properties for dot-namespace grouping


EhabY:
## Context

Came up while reviewing #948. We dot-namespace event **names** (`auth.unauthorized_intercepted`, `workspace.state_transitioned`) but property **keys** are flat camelCase. For events whose properties share a real sub-concept, flattening produces compound names (`fromStatus`, `fromLifecycleState`) and forfeits prefix-grep on the group.

## Convention

Derived from OpenTelemetry's [attribute naming spec](https://opentelemetry.io/docs/specs/semconv/general/naming/) and the [span attribute guide](https://opentelemetry.io/blog/2025/how-to-name-your-span-attributes/):

1. **General-to-specific dot-namespacing**: `{dimension}.{aspect}`, mirroring OTel's `{domain}.{component}.{property}`. So `status.from`, not `from.status` (parallel to `http.request.method`, not `request.method.http`).
2. **When to apply**: a property group with 3+ siblings, or 2 siblings expected to grow, all sharing a real sub-concept (a dimension, a metric family, an entity).
3. **Otherwise stay flat**: single-property or atomic-record events keep bare/camelCase names.
4. **Hard constraint** (OTel): a name cannot be both an attribute and a namespace prefix. Never both `status` and `status.from`.

### Rename

**`workspace.agent.state_transitioned`** (two state dimensions, qualified)
- `fromStatus`/`toStatus` → `status.from`/`status.to`
- `fromLifecycleState`/`toLifecycleState` → `lifecycleState.from`/`lifecycleState.to`

**`http.requests`** (two metric families)
- `count_{1xx,2xx,3xx,4xx,5xx,network_error}` → `count.{1xx,2xx,3xx,4xx,5xx,network_error}`
- `p{50,95,99}_duration_ms` → `duration.p{50,95,99}_ms`
- Decide: keep snake_case within keys (status-code classes and percentile names don't camelCase naturally) or normalize.

### Leave as-is

- `workspace.state_transitioned`: single dimension, bare `from`/`to` already correct.
- `connection.state_transitioned`: same.
- All 24 atomic / single-property events (`auth.token_refreshed`, `ssh.process.discovered`, `connection.opened`, etc.).

### Optional clarity tweak

[...]

Issue #906, opened 2026-04-17
Telemetry: instrument auth and workspace lifecycle


EhabY:
Part of the Telemetry Phase A rollout. See the RFC in Linear: AIGOV-154.

## Events

- `auth.token_refresh` with `result`, `trigger`, measurement `durationMs`.
- `auth.intercept_401` with `recovery`.
- `auth.login_prompt` with `trigger`, `result`.
- `workspace.start` with `result`, measurement `durationMs`. Emitted when the user opens a workspace and it transitions to running.
- `workspace.stop` with `trigger` (`user`, `idle`).
- `workspace.update.triggered` with `result`, measurement `durationMs`.

## Sites

- `src/oauth/sessionManager.ts` refresh path.
- `src/api/authInterceptor.ts` 401 handling.
- `src/login/loginCoordinator.ts` login-dialog path.
- Workspace-related command handlers (start, stop, update).

## Tests

- `TestSink` assertions for each event.
- Token refresh emits on both background and reactive triggers.
- 401 path emits with the right `recovery` value for the refresh-success and login-required cases.
- Workspace start emits on successful transition to running, error event on failure.

Depends on #900, #902.

PR #963, opened 2026-05-18
feat(telemetry): instrument workspace lifecycle

EhabY:
## Summary

Records local telemetry for the workspace state machine and watcher so build durations and agent transitions show up alongside the rest of the local telemetry stream.

Workspace events (`src/instrumentation/workspace.ts`, split into three single-purpose classes):

- `workspace.state_transitioned`: deduped on `(status, transition, reason)`; emits `observedDurationMs` between transitions and `observedBuildDurationMs` when a provisioner run resolves.
- `workspace.agent.state_transitioned`: deduped on `(status, lifecycle_state)`; emits `observedDurationMs`.
- `workspace.start.triggered` / `workspace.update.triggered`: traced spans around user-initiated workspace operations.

All workspace events carry `workspaceName`. The workspace event uses bare `from`/`to` (single state dimension); the agent event uses qualified `fromStatus`/`toStatus` plus `fromLifecycleState`/`toLifecycleState` because two dimensions can change in the same emission. Cross-event consistency via dimension-as-prefix property keys is tracked in #954. `WorkspaceMonitor` and `WorkspaceStateMachine` now take the `ServiceContainer` directly so they can pull the telemetry service.

## Split

Stacked on #962. Second of two PRs replacing #948 (closing #906 in two pieces). Review against the auth branch shows only the workspace-specific diff; rebase to `main` once #962 merges.

### DIFF
--- a/src/remote/workspaceStateMachine.ts
+++ b/src/remote/workspaceStateMachine.ts
@@ -13,6 +13,10 @@ import {
 	streamAgentLogs,
 	streamBuildLogs,
 } from "../api/workspace";
+import {
+	WorkspaceAgentTelemetry,
+	WorkspaceOperationTelemetry,
+} from "../instrumentation/workspace";
 import { maybeAskAgent } from "../promptUtils";
 import { vscodeProposed } from "../vscodeProposed";
 
@@ -25,6 +29,7 @@ import type {
 } from "coder/site/src/api/typesGenerated";
 
 import type { CoderApi } from "../api/coderApi";
+import type { ServiceContainer } from "../core/container";
 import type { StartupMode } from "../core/mementoManager";
 import type { FeatureSet } from "../featureSet";
 import type { Logger } from "../logging/logger";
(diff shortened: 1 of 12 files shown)


src/remote/workspaceStateMachine.ts
```diff
@@ -286,7 +306,10 @@ export class WorkspaceStateMachine implements vscode.Disposable {
...
 		});
 		try {
-			this.workspace = await updateWorkspace(this.buildCliContext(workspace));
+			this.workspace = await this.operationTelemetry.traceUpdateTriggered(() =>
+				updateWorkspace(this.buildCliContext(workspace)),
+			);
+			this.logger.info(`${workspaceName} update initiated`);
 			return this.workspace;
 		} catch (error) {
 			if (error instanceof WorkspaceUpdateCancelledError) {
```
ethanndickson, line 321: doesn't this mean if the user says no to the workspace update, we record an error in the telemetry? not sure what a nice fix looks like here 😕

EhabY: Ah nice catch, I've split the parameter collection into a separate step so now we only trace the update!

Also added missing tests and passed the params to the API call which we used it not do..

ethanndickson (approved):

PR #966, opened 2026-05-19
security: release v1.14.6 with SEC-200 scope hardening

EhabY:
## Summary

Cuts **v1.14.6** off the `v1.14.5` tag with the SEC-200 fix for users on the stable channel. The same change is delivered on `main` in a separate PR.

Closes the SEC-200 attack path where a malicious `.vscode/settings.json` could override security-sensitive Coder settings — most notably the two command-execution settings (`coder.headerCommand`, `coder.tlsCertRefreshCommand`) called out in the original report.

The fix is metadata-only: each affected setting gets a `scope` of `application` or `machine`. VS Code drops workspace and folder values for both scopes, so a malicious workspace value never reaches our code. No runtime guard needed.

This PR also relaxes the release workflow so this release branch can publish without first being merged into `main`.

## Why not `window` (the default)

VS Code's default scope is `window`, which lets **workspace and folder `settings.json` override the user setting** — exactly the SEC-200 vector. Any setting that can cause command execution, redirect network traffic, swap credentials, or substitute the CLI binary must not be controllable by a project file.

## `application` vs `machine`

Both block workspace/folder overrides, so both close SEC-200. They differ in Settings Sync behavior:

| Scope | Settable in | Synced via Settings Sync? |
|---|---|---|
| `application` | User settings only | **Yes** |
| `machine` | User OR per-machine remote settings | **No** |

We split by what the setting actually represents:

- **`application`** — user-wide preferences with no OS or filesystem coupling. Safe (and desirable) to sync across machines.
- **`machine`** — paths, shell commands, and network config that are inherently per-machine. A Windows `cmd.exe /c …` `headerCommand` syncing to a Mac would silently break; absolute paths like `tlsCaFile` don't translate.

This follows VS Code's [recommended scope semantics](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration).

## Settings updated

**`application` (synced, user-wide preferences):**

| Setting | Why it's sensitive |
|---|---|
| `coder.insecure` | Disables TLS host verification |
| `coder.disableSignatureVerification` | Toggles CLI signature verification |
| `coder.enableDownloads` | Force-enables downloads in restricted environments |
| `coder.defaultUrl` | Default Coder deployment URL |
| `coder.autologin` | Combined with `defaultUrl` could auto-login elsewhere |
| `coder.useKeyring` | Credential-storage preference |

**`machine` (per-machine, not synced):**

| Setting | Why it's sensitive |
|---|---|
| `coder.headerCommand` | Runs an external shell command (SEC-200) |
| `coder.tlsCertRefreshCommand` | Runs an external shell command (SEC-200) |
| `coder.binarySource` | URL the Coder CLI is downloaded from |
| `coder.binaryDestination` | Filesystem path of the executed CLI binary |
| `coder.sshFlags` | SSH flags accept `-o ProxyCommand=…` (exec) |
| `coder.globalFlags` | Flags applied to every CLI invocation |
| `coder.tlsCertFile` | TLS client certificate path |
| `coder.tlsKeyFile` | TLS client key path |
| `coder.tlsCaFile` | TLS CA path (workspace-controlled CA → MITM) |
| `coder.tlsAltHost` | Hostname override for TLS verification |
| `coder.proxyLogDirectory` | Arbitrary write path |
| `coder.proxyBypass` | Routes traffic around the proxy |

Already `machine`: `coder.sshConfig`.

## CI change

Drops the `Verify tag is on main` step from `.github/workflows/release.yaml` so this `release/v1.14.6` branch can publish a tagged release without first being merged into `main`. The same change lands on `main` in the companion PR.

## Companion PR

The same scope-hardening change is being delivered against `main` separately.

## Release plan

After merge:

1. Tag `v1.14.6` on the merged commit at the tip of `release/v1.14.6`.
2. The release workflow packages the extension and publishes to the Marketplace + OpenVSX.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,13 +21,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Verify tag is on main
-        run: |
-          if ! git merge-base --is-ancestor "$GITHUB_SHA" origin/main; then
-            echo "Error: Release tags must be pushed from the main branch"
-            exit 1
-          fi
-
       - uses: pnpm/action-setup@v6
 
       - uses: actions/setup-node@v6.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,31 @@
      from published versions since it shows up in the VS Code extension changelog
      tab and is confusing to users. Add it back between releases if needed. -->
 
+## [v1.14.6](https://github.com/coder/vscode-coder/releases/tag/v1.14.6) 2026-05-19
+
+### Security
+
+- Hardened the configuration scope of security-sensitive settings so that a
+  malicious `.vscode/settings.json` cannot override them (SEC-200). Workspace
+  and folder values are now ignored by VS Code for these settings. This closes
+  a path where a workspace could redirect command execution
+  (`coder.headerCommand`, `coder.tlsCertRefreshCommand`), substitute the CLI
+  binary or its source (`coder.binarySource`, `coder.binaryDestination`,
+  `coder.disableSignatureVerification`, `coder.enableDownloads`), inject
+  CLI/SSH flags (`coder.globalFlags`, `coder.sshFlags`), swap TLS material or
+  disable TLS verification (`coder.tlsCertFile`, `coder.tlsKeyFile`,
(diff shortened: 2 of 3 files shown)


EhabY:
Closed in favor of https://github.com/coder/vscode-coder/tree/release/v1.14.6

PR #969, opened 2026-05-21
feat(telemetry): add span log correlation

### LINKED ISSUE #925: Telemetry: span.log() / span.logError() for log-in-trace correlation
Part of the Telemetry Phase A rollout.

`service.log()` and `service.logError()` emit `LogRecord`-shaped events with no `traceId` or `parentEventId`, so a log call made inside a `service.trace(...)` callback is uncorrelated with the surrounding span. OTel backends that link logs to traces (Grafana's Trace-to-Logs, Honeycomb's trace events panel, Datadog APM's log-in-trace view) need both `trace_id` and `span_id` on the LogRecord to make the link. We currently can't provide either.

## Scope

Add log methods on the `Span` interface, with the same signature as the top-level `service.log` / `logError`:

```ts
interface Span {
  // ... existing fields
  log(
    eventName: string,
    properties?: Record<string, string>,
    measurements?: Record<string, number>,
  ): void;
  logError(
    eventName: string,
    error: unknown,
    properties?: Record<string, string>,
    measurements?: Record<string, number>,
  ): void;
}
```

Behavior:

* Identical to `service.log` / `logError`, except the emitted event carries the span's `traceId` and `parentEventId = span.eventId`.
* No `durationMs` measurement (these are point-in-time logs, not timed operations).
* Routes through the same `#safeEmit` so a telemetry failure cannot reach the caller.
* `NOOP_SPAN.log` / `logError` are no-ops (consistent with `phase`).

Top-level `service.log` / `logError` are unchanged: they continue to emit uncorrelated LogRecords.

## Why explicit, not ambient

OTel SDKs use `AsyncLocalStorage` to track the "current active span" so that `logger.info(...)` calls anywhere in code automatically pick up the span context. We could do the same, but:

* It's invisible to readers — a `service.log("foo")` call might or might not have trace context depending on the call stack.
* `AsyncLocalStorage` propagation breaks across some async boundaries (worker_threads, certain native callbacks, cross-process IPC), leading to silent loss of correlation.
* The explicit `span.log(...)` form makes the data contract obvious: if you want correlation, take the `Span` parameter and call it.

We can revisit ambient context if it becomes necessary, but explicit is the simpler and clearer default.

## Discriminator at export time

The exporter ([#903](<https://github.com/coder/vscode-coder/issues/903>)) currently uses `traceId` presence to decide whether an event becomes a `LogRecord` or a `Span`. With `span.log()` shipping, that discriminator no longer holds — a logged event in trace context will have a `traceId` but should still export as a `LogRecord`, not a `Span`.

Options to resolve, decided here so coder/vscode-coder#903 can wire the routing correctly:

* **A. Use** `durationMs` presence: Spans always have `durationMs` (framework-set on `time`/`trace`/`phase`); logs never do. Simple, no schema change. Edge case: a caller manually passing `measurements: { durationMs: ... }` to `service.log` would be misclassified.
* **B. Add explicit** `eventKind: "log" | "span"` field: Disambiguates cleanly at the cost of one new field on every event.

Recommend **A**. The edge case is genuinely weird (passing `durationMs` to `log` is misuse) and we can document it as reserved if needed. Add an explicit `eventKind` field if a real bug surfaces.

## Tests

* `span.log` emits an event with the span's `traceId` and `parentEventId`, no `durationMs`.
* `span.logError` emits an event with `traceId`, `parentEventId`, the normalized error block, no `durationMs`.
* Top-level `service.log` / `logError` continue to emit without `traceId` (no regression).
* `NOOP_SPAN.log` / `logError` no-op when telemetry is off.
* Throwing sink does not affect the surrounding `trace`'s return value (existing isolation contract still holds for log events).

## Out of scope

* Ambient context propagation. Tracked here as a future option only.
* Changes to top-level `service.log` / `logError`. They remain trace-context-free.

## Why this is needed

Without it, the [exporter (#903)](<https://github.com/coder/vscode-coder/issues/903>)'s OTLP/Logs output has no `trace_id` on log records, so OTel backends cannot link logs to traces in their UIs. Adding `span.log()` is the producer-side change that completes the loop.

Depends on coder/vscode-coder#900 (the core service this extends).

EhabY:
## Summary

- Adds `Span.log()` and `Span.logError()` for point-in-time telemetry logs correlated to the active span.
- Auto-composes span log names under the owning span and emits `traceId` plus `parentEventId` without `durationMs`.
- Resolves the auth interceptor TODO by emitting `auth.unauthorized_intercepted.received` when 401 recovery starts.

## Tests

- `pnpm test:extension ./test/unit/telemetry/service.test.ts ./test/unit/api/authInterceptor.test.ts`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format:check`

Closes #925

Generated by Coder Agents.

<details>
<summary>Plan and decisions</summary>

- Use `durationMs` presence as the span/log discriminator; do not add `eventKind`.
- Drop and warn for `span.log()` / `span.logError()` after a span has completed.
- Auto-compose span log event names as `${span.eventName}.${logName}`.
- Resolve the auth interceptor TODO with `auth.unauthorized_intercepted.received`.
- Keep parent/correlation docs concise.

</details>

### DIFF
--- a/src/telemetry/service.ts
+++ b/src/telemetry/service.ts
@@ -178,7 +178,7 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 		let mark: "aborted" | "error" | undefined;
 		const warnPostEmit = (op: string, name: string): void => {
 			this.logger.warn(
-				`Telemetry span '${eventName}' ${op}('${name}') called after emit; mutation dropped`,
+				`Telemetry span '${eventName}' ${op}('${name}') called after emit; call ignored`,
 			);
 		};
 		const emitSpanLog = (
--- a/src/telemetry/service.ts
+++ b/src/telemetry/service.ts
@@ -129,7 +129,7 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 	 * Run a timed operation. The emitted event carries `durationMs` and a
 	 * `result` of `success`, `error`, or `aborted` (set via `span.markAborted()`
 	 * for intentional early exits). All events from one call share a `traceId`;
-	 * phase children carry `parentEventId`.
+	 * child phases and span logs carry `parentEventId`.
 	 */
 	public trace<T>(
 		eventName: string,
@@ -181,6 +181,21 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 				`Telemetry span '${eventName}' ${op}('${name}') called after emit; mutation dropped`,
 			);
 		};
+		const emitSpanLog = (
+			logName: string,
+			logProperties: CallerProperties,
+			logMeasurements: CallerMeasurements,
+			error?: unknown,
+		): void => {
+			const safeName = this.#sanitizeChildName(logName, "log");
+			this.#safeEmit(
+				newSpanId(),
+				`${eventName}.${safeName}`,
+				stringifyProps(logProperties),
+				logMeasurements,
+				{ traceId, parentEventId: eventId, traceLevel, error },
+			);
+		};
 		const span: Span = {
 			traceId,
 			eventId,
@@ -189,9 +204,13 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 				phaseName: string,
 				phaseFn: (childSpan: Span) => Promise<U>,
 				phaseProps: CallerProperties = {},
-				phaseMeasurements: Record<string, number> = {},
+				phaseMeasurements: CallerMeasurements = {},
 			): Promise<U> => {
-				const safeName = this.#sanitizePhaseName(phaseName);
+				if (completed) {
+					warnPostEmit("phase", phaseName);
+					return phaseFn(NOOP_SPAN);
+				}
+				const safeName = this.#sanitizeChildName(phaseName, "phase");
 				return this.#startSpan(
 					`${eventName}.${safeName}`,
 					phaseFn,
@@ -199,6 +218,29 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 					phaseMeasurements,
 					{ traceId, parentEventId: eventId, traceLevel },
 				);
+			},
+			log: (
+				logName: string,
+				logProperties: CallerProperties = {},
+				logMeasurements: CallerMeasurements = {},
+			): void => {
+				if (completed) {
+					warnPostEmit("log", logName);
+					return;
+				}
+				emitSpanLog(logName, logProperties, logMeasurements);
+			},
+			logError: (
+				logName: string,
+				error: unknown,
+				logProperties: CallerProperties = {},
+				logMeasurements: CallerMeasurements = {},
+			): void => {
+				if (completed) {
+					warnPostEmit("logError", logName);
+					return;
+				}
+				emitSpanLog(logName, logProperties, logMeasurements, error);
 			},
 			setProperty(name: string, value: CallerPropertyValue): void {
 				if (completed) {
@@ -243,13 +285,13 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 		});
 	}
 
-	#sanitizePhaseName(name: string): string {
+	#sanitizeChildName(name: string, kind: "phase" | "log"): string {
 		if (!name.includes(".")) {
 			return name;
 		}
 		const sanitized = name.replaceAll(".", "_");
 		this.logger.warn(
-			`Telemetry phase name '${name}' contains '.', sanitized to '${sanitized}'`,
+			`Telemetry ${kind} name '${name}' contains '.', sanitized to '${sanitized}'`,
 		);
 		return sanitized;
 	}
--- a/src/telemetry/span.ts
+++ b/src/telemetry/span.ts
@@ -22,13 +22,13 @@ export interface Span {
 		properties?: CallerProperties,
 		measurements?: CallerMeasurements,
 	): Promise<T>;
-	/** Emit a point-in-time log event correlated with this span. */
+	/** Emit a point-in-time log event under this span. No framework-set `result` or `durationMs`. */
 	log(
 		logName: string,
 		properties?: CallerProperties,
 		measurements?: CallerMeasurements,
 	): void;
-	/** Emit a point-in-time error log event correlated with this span. */
+	/** Emit a point-in-time error log event under this span. No framework-set `result` or `durationMs`. */
 	logError(
 		logName: string,
 		error: unknown,
--- a/src/telemetry/span.ts
+++ b/src/telemetry/span.ts
@@ -7,8 +7,8 @@ import type {
 export type SpanResult = "success" | "aborted" | "error";
 
 /**
- * Parent span handle. Children's `eventName` composes as `${parent.eventName}.${phaseName}`.
- * Phase names should not contain `.`; if they do, dots are replaced with `_` and a warning is logged.
+ * Parent span handle. Child phases and logs compose as `${parent.eventName}.${name}`.
+ * Child names should not contain `.`; if they do, dots are replaced with `_` and a warning is logged.
  * Recurse via `phase` for grandchildren.
  */
 export interface Span {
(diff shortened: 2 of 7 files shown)


test/unit/telemetry/service.test.ts
```diff
@@ -323,6 +406,35 @@ describe("TelemetryService", () => {
 			);
 		});
 
+		it("drops span logs called after emit", async () => {
```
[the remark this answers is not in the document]
EhabY: Intentional. We pulled the warn-mock assertion because it tests a side effect rather than the contract. The contract is no event leaks, and that's what we check. The warn stays in production as a runtime aid for the developer who triggers the bug, but tests pinned to logger mock counts and exact wording are fragile to message rewording and don't add behavioral coverage. The sibling `setProperty` test at line 373 still has the old pattern but we left it alone to avoid sweeping changes outside this PR.

test/unit/telemetry/service.test.ts
```diff
@@ -323,6 +406,35 @@ describe("TelemetryService", () => {
...
+				return Promise.resolve();
+			});
+
+			escapedSpan?.log("late");
+			escapedSpan?.logError("late_error", new Error("ignored"));
+
+			expect(h.sink.events).toHaveLength(1);
+		});
+
+		it("runs phase fns called after emit but emits no phase event", async () => {
```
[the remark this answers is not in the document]
EhabY: Same call as the log/logError test above: we deliberately don't assert on `logger.warn` mock calls. The visible contract is fn runs and no phase event emits, which is what the test checks. The warn is a runtime aid only.

src/telemetry/wireFormat.ts
```diff
@@ -32,7 +32,7 @@ const TelemetryEventSchema = z.object({
 	measurements: z.record(z.string(), z.number()),
 	/** Shared by all events in a trace. Maps to OTel `trace_id`. */
 	traceId: z.string().optional(),
-	/** Parent event in the same trace. Maps to OTel `parent_span_id`. */
+	/** Parent or correlated span event in the same trace. */
```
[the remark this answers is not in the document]
EhabY: Good catch. Restored the OTel mapping and made the dual case explicit:

```ts
/**
 * Parent span in the same trace. For phase children this maps to OTel
 * `parent_span_id`. For span logs it maps to OTel `span_id` on the log
 * record (the span the log belongs to).
 */
```

src/telemetry/span.ts
```diff
@@ -22,6 +22,19 @@ export interface Span {
 		properties?: CallerProperties,
 		measurements?: CallerMeasurements,
 	): Promise<T>;
+	/** Emit a point-in-time log event correlated with this span. */
```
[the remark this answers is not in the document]
EhabY: Updated to match the `phase` doc style and call out the absent fields:

```ts
/** Emit a point-in-time log event under this span. No framework-set `result` or `durationMs`. */
```

test/unit/telemetry/service.test.ts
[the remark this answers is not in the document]
EhabY: Update: ended up with a middle ground. Behavior tests still check only the visible contract (no event leak), but added one dedicated `"warns once per post-emit method call"` test that exercises all seven post-emit methods (`setProperty`, `setMeasurement`, `markAborted`, `markFailure`, `log`, `logError`, `phase`) and verifies a warn fires for each. Keeps the regression guard you flagged while keeping the behavior tests focused.

test/unit/telemetry/service.test.ts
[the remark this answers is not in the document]
EhabY: Same fix as the sibling thread: the phase post-emit warn is now covered by the dedicated `"warns once per post-emit method call"` test alongside the other six methods.

src/telemetry/service.ts
```diff
@@ -200,6 +219,29 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 					{ traceId, parentEventId: eventId, traceLevel },
 				);
 			},
+			log: (
+				logName: string,
+				logProperties: CallerProperties = {},
+				logMeasurements: CallerMeasurements = {},
+			): void => {
+				if (completed) {
+					warnPostEmit("log", logName);
```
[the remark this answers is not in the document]
EhabY: Fair point. Changed to `call ignored` so it reads accurately for all seven callers, including `phase` (which runs the fn but suppresses the span-level effect).

src/telemetry/service.ts
```diff
@@ -178,7 +178,22 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
 		let mark: "aborted" | "error" | undefined;
 		const warnPostEmit = (op: string, name: string): void => {
 			this.logger.warn(
-				`Telemetry span '${eventName}' ${op}('${name}') called after emit; mutation dropped`,
+				`Telemetry span '${eventName}' ${op}('${name}') called after emit; call ignored`,
```
mtojek, line 181: wasn't "mutation dropped" more precise?

test/unit/telemetry/service.test.ts
```diff
@@ -287,40 +370,71 @@ describe("TelemetryService", () => {
...
 
-			// Mutations dropped: emitted event is unchanged.
 			expect(h.sink.events[0].properties.late).toBeUndefined();
 			expect(h.sink.events[0].measurements.lateMs).toBeUndefined();
 			expect(h.sink.events[0].properties.result).toBe("success");
+		});
 
-			// Each post-emit mutation logs a warning.
-			expect(vi.mocked(h.logger.warn).mock.calls.length).toBe(warnBefore + 4);
-			expect(vi.mocked(h.logger.warn).mock.calls[warnBefore][0]).toContain(
```
mtojek, line 312: why did you lose these assertions?

src/telemetry/service.ts
```diff
@@ -178,7 +178,22 @@ export class TelemetryService implements vscode.Disposable, TelemetryReporter {
...
+			);
+		};
+		const emitSpanLog = (
+			logName: string,
+			logProperties: CallerProperties,
+			logMeasurements: CallerMeasurements,
+			error?: unknown,
+		): void => {
+			const safeName = this.#sanitizeChildName(logName, "log");
```
mtojek, line 190: should we check if trace level is off?

test/unit/telemetry/service.test.ts
```diff
@@ -165,6 +176,78 @@ describe("TelemetryService", () => {
...
+
+			const [log, logError, child, parent] = h.sink.events;
+			expect(log.eventName).toBe("op.phase.checkpoint");
+			expect(log.traceId).toBe(parent.traceId);
+			expect(log.parentEventId).toBe(child.eventId);
+			expect(logError.eventName).toBe("op.phase.oops");
+			expect(logError.traceId).toBe(parent.traceId);
+			expect(logError.parentEventId).toBe(child.eventId);
+			expect(logError.error).toMatchObject({ message: "boom" });
+			expect(child.parentEventId).toBe(parent.eventId);
```
mtojek, line 247: maybe include measurements?

mtojek (approved):
I don't see anything blocking; left a few comments 👍

src/telemetry/service.ts
EhabY: "mutation dropped" only some of the calls, `log` / `logError` / `phase` emit (or skip) child events, they don't mutate

test/unit/telemetry/service.test.ts
EhabY: This was moved into it's own test, I do not want these tests to test side-effects!

src/telemetry/service.ts
EhabY: Not needed. `traceLevel` is frozen at trace start which return `NOOP_SPAN` if it's off

test/unit/telemetry/service.test.ts
EhabY: Done!

mtojek (approved):
Thanks!

PR #976, opened 2026-05-22
fix: recover mTLS auth after settings races

### LINKED ISSUE #973: mTLS users signed out on remote connect when companion settings extension activates after Coder
## Symptoms

mTLS users whose `coder.tls*` and `coder.headerCommand` settings are written by a companion extension see, shortly after the remote workspace opens:

1. "Session expired. You have been signed out." toast.
2. `Header Command setting changed. Reload window to apply.`

The workspace itself opens successfully, then both messages appear.

## Root cause

The Coder extension activates on `onResolveRemoteAuthority:ssh-remote` (earliest possible), so it reads the settings that the companion extension persisted on a previous session. `setupCoderRemote` runs cleanly and `watchSettings` (`src/remote/remote.ts:459`) arms with those values as baseline. The damage happens **after** that, when the companion extension activates and rewrites the same settings.

VS Code's `WorkspaceConfiguration.update` is per-key and emits a separate `onDidChangeConfiguration` event for each call. There is no atomic bulk write. So a companion that does:

```ts
await config.update('coder.tlsCertFile', newCert, Global);
await config.update('coder.tlsKeyFile',  newKey,  Global);
await config.update('coder.tlsCaFile',   newCa,   Global);
await config.update('coder.headerCommand', newCmd, Global);
```

leaves the merged config in a partially-updated state between each `await`, visible to all other extensions. Two consequences cascade:

- **Watcher fires legitimately.** When `coder.headerCommand` changes value, `configWatcher.ts:21-40` detects the drift and surfaces the reload prompt. The SSH `ProxyCommand` already on disk has `--header-command "OLD_VALUE"` baked in (`remote.ts:359`), so the prompt is technically correct, but for a companion that is just refreshing settings to the same effective value, it is spurious noise.
- **In-flight requests 401 on the transient state.** The request interceptor (`src/api/coderApi.ts:505-526`) reads TLS settings and `coder.headerCommand` at request time. A request from `Inbox`, `WorkspaceMonitor`, `agentMetadata`, or workspace polling that lands between two of the companion's `update` calls sees a mismatched cert/key pair or a cleared `coder.headerCommand`, the server returns 401, and `AuthInterceptor` invokes `handleAuthFailure` (`src/extension.ts:104-124`), which suspends the session and shows the toast. There is no retry.

`DeploymentManager.registerAuthListener` (`src/deployment/deploymentManager.ts:198-233`) only listens to `onDidChangeSessionAuth` (token changes), not to settings, so once suspended the session stays suspended until the user clicks "Log In", even though the settings stabilise milliseconds later.

## Proposed fix

| # | Change | File |
|---|---|---|
| 1 | Single silent retry in `AuthInterceptor.handleError` when a 401 arrives shortly after a `coder.tls*` / `coder.headerCommand` change event. The failing request used pre-change settings; a retry picks up post-change settings. Gated by the existing `_retryAttempted` flag. | `src/api/authInterceptor.ts` |
| 2 | In `DeploymentManager`, watch `coder.tls*` / `coder.headerCommand`. When `!isAuthenticated()` and a deployment is still set (suspended state), re-attempt `setDeploymentIfValid` on change, debounced. Restores the session automatically without user action. | `src/deployment/deploymentManager.ts` |
| 3 | Coalesce `watchSettings` events within a short window (e.g. 250 ms). A companion rewriting four keys in sequence should not produce four prompts, and a value that flips empty then back to the same value should be a no-op. | `src/configWatcher.ts` / `src/remote/remote.ts:979-1004` |
| 4 | Drop the `needToken(...)` half of the pre-check; keep only `!context.baseUrl`. Covers the separate first-connect-from-fresh case, where a companion that has never persisted settings has not run yet and the current check incorrectly demands a token from an mTLS user. | `src/remote/remote.ts:174-183` |

#1 is the load-bearing fix for the toast. #2 is the backstop that restores the session if #1 ever misses. #3 quiets the reload prompt when the companion settles back to the same effective value. #4 is independent of the customer's current symptom but is a real bug on first-connect-from-fresh.

## Repro

1. Coder deployment requiring mTLS.
2. Companion extension that writes `coder.tlsCertFile`, `coder.tlsKeyFile`, `coder.tlsCaFile`, `coder.headerCommand` to user scope via sequential `update` calls, with an activation event later than `onResolveRemoteAuthority:ssh-remote`.
3. Open a remote Coder workspace.

Expected: connection establishes silently.
Actual: workspace opens, then "Session expired" toast and "Header Command changed" reload prompt appear.

EhabY:
Closes #973

A session suspended by an mTLS or `coder.headerCommand` failure stayed suspended even after the user fixed the setting. Two races caused it: a request could 401 from a mid-flight settings change with no way to recover, and nothing re-tried once the fix landed.

## What changes

- **Suspended sessions auto-recover** once auth settings become valid again, with no logout/login round-trip.
- **Settings-race 401s retry silently** under the new settings instead of escalating to an interactive prompt.
- **Retried requests don't carry stale headers** from a previous header-command run.
- **Verify isn't overwritten by stale work.** Logout, deployment switch, or dispose during an in-flight auth verify wins; the verify quietly bails.
- **Cross-window login is robust.** A window observing an invalid token from another window keeps listening for a follow-up valid write.
- **Config-change side-effects coalesce.** Reload prompts, recovery, and WebSocket reconnects fire once edits settle, not on every keystroke.
- **No stale deployment in storage** when logout races a deployment write.

Tests cover silent retry, retry escalation, debounce semantics, suspended-state recovery (including the racing-`clearDeployment` case), and the header-cleanup-vs-session-token behaviour. `pnpm typecheck`, `pnpm lint`, and `pnpm test:extension` clean.

### DIFF
--- a/src/api/authInterceptor.ts
+++ b/src/api/authInterceptor.ts
@@ -9,7 +9,6 @@ import type * as vscode from "vscode";
 import type { ServiceContainer } from "../core/container";
 import type { SecretsManager } from "../core/secretsManager";
 import type { Logger } from "../logging/logger";
-import type { RequestConfigWithMeta } from "../logging/types";
 import type { OAuthSessionManager } from "../oauth/sessionManager";
 
 import type { CoderApi } from "./coderApi";
@@ -62,17 +61,6 @@ export class AuthInterceptor implements vscode.Disposable {
 			throw error;
 		}
 
-		if (this.shouldRetryAfterAuthConfigChange(error)) {
-			return this.retryAfterAuthConfigChange(error);
-		}
-
-		if (error.config) {
-			const config = error.config as { _retryAttempted?: boolean };
-			if (config._retryAttempted) {
-				throw error;
-			}
-		}
-
 		const baseUrl = this.client.getHost();
 		if (!baseUrl) {
 			throw error;
@@ -82,43 +70,30 @@ export class AuthInterceptor implements vscode.Disposable {
 		return this.recoverFromUnauthorized(error, hostname);
 	}
 
-	private shouldRetryAfterAuthConfigChange(error: AxiosError): boolean {
-		if (!error.config) {
-			return false;
-		}
-
-		const config = error.config as RequestConfigWithMeta & {
-			_retryAttempted?: boolean;
-			_authConfigRetryAttempted?: boolean;
-			authConfigVersion?: number;
-		};
-		// Skip if any retry already happened: caps total attempts at 2.
-		if (config._retryAttempted || config._authConfigRetryAttempted) {
-			return false;
-		}
-
-		return this.client.hasAuthConfigChangedSince(config.authConfigVersion);
-	}
-
-	private retryAfterAuthConfigChange(error: AxiosError): Promise<unknown> {
-		if (!error.config) {
-			throw error;
-		}
-
-		const config = error.config as RequestConfigWithMeta & {
-			_authConfigRetryAttempted?: boolean;
-		};
-		config._authConfigRetryAttempted = true;
-		this.logger.debug(
-			"Authentication settings changed during request, retrying once",
-		);
-		return this.client.getAxiosInstance().request(config);
-	}
-
 	private recoverFromUnauthorized(
 		error: AxiosError,
 		hostname: string,
 	): Promise<unknown> {
+		const config = error.config;
+
+		// Checked before _retryAttempted so an OAuth-retry 401 caused by a
+		// fresh settings change still gets one silent attempt.
+		if (
+			config &&
+			!config._authConfigRetryAttempted &&
+			this.client.hasAuthConfigChangedSince(config.authConfigVersion)
+		) {
+			config._authConfigRetryAttempted = true;
+			this.logger.debug(
+				"Authentication settings changed during request, retrying once",
+			);
+			return this.client.getAxiosInstance().request(config);
+		}
+
+		if (config?._retryAttempted) {
+			throw error;
+		}
+
 		this.logger.debug("Received 401 response, attempting recovery");
 		return this.authTelemetry.traceAuthRecovery(async (recorder) => {
 			recorder.logReceived();
@@ -203,12 +178,9 @@ export class AuthInterceptor implements vscode.Disposable {
 			throw error;
 		}
 
-		const config = error.config as RequestConfigWithMeta & {
-			_retryAttempted?: boolean;
-		};
-		config._retryAttempted = true;
-		config.headers[coderSessionTokenHeader] = token;
-		return this.client.getAxiosInstance().request(config);
+		error.config._retryAttempted = true;
+		error.config.headers[coderSessionTokenHeader] = token;
+		return this.client.getAxiosInstance().request(error.config);
 	}
 
 	public dispose(): void {
--- a/src/api/authInterceptor.ts
+++ b/src/api/authInterceptor.ts
@@ -58,15 +58,19 @@ export class AuthInterceptor implements vscode.Disposable {
 			throw error;
 		}
 
+		if (error.response?.status !== 401) {
+			throw error;
+		}
+
+		if (this.shouldRetryAfterAuthConfigChange(error)) {
+			return this.retryAfterAuthConfigChange(error);
+		}
+
 		if (error.config) {
 			const config = error.config as { _retryAttempted?: boolean };
 			if (config._retryAttempted) {
 				throw error;
 			}
-		}
-
-		if (error.response?.status !== 401) {
-			throw error;
 		}
 
 		const baseUrl = this.client.getHost();
@@ -76,6 +80,39 @@ export class AuthInterceptor implements vscode.Disposable {
 		const hostname = toSafeHost(baseUrl);
 
 		return this.recoverFromUnauthorized(error, hostname);
+	}
+
+	private shouldRetryAfterAuthConfigChange(error: AxiosError): boolean {
+		if (!error.config) {
+			return false;
+		}
+
+		const config = error.config as RequestConfigWithMeta & {
+			_retryAttempted?: boolean;
+			_authConfigRetryAttempted?: boolean;
+			authConfigVersion?: number;
+		};
+		// Skip if any retry already happened: caps total attempts at 2.
+		if (config._retryAttempted || config._authConfigRetryAttempted) {
+			return false;
+		}
+
+		return this.client.hasAuthConfigChangedSince(config.authConfigVersion);
+	}
+
+	private retryAfterAuthConfigChange(error: AxiosError): Promise<unknown> {
+		if (!error.config) {
+			throw error;
+		}
+
+		const config = error.config as RequestConfigWithMeta & {
+			_authConfigRetryAttempted?: boolean;
+		};
+		config._authConfigRetryAttempted = true;
+		this.logger.debug(
+			"Authentication settings changed during request, retrying once",
+		);
+		return this.client.getAxiosInstance().request(config);
 	}
 
 	private recoverFromUnauthorized(
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -8,7 +8,10 @@ import {
 import { Api } from "coder/site/src/api/api";
 import * as vscode from "vscode";
 
-import { watchConfigurationChanges } from "../configWatcher";
+import {
+	CONFIG_CHANGE_DEBOUNCE_MS,
+	watchConfigurationChanges,
+} from "../configWatcher";
 import { ClientCertificateError } from "../error/clientCertificateError";
 import { toError } from "../error/errorUtils";
 import { ServerCertificateError } from "../error/serverCertificateError";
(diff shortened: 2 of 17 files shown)


src/api/authInterceptor.ts
```diff
@@ -58,17 +58,21 @@ export class AuthInterceptor implements vscode.Disposable {
 			throw error;
 		}
 
+		if (error.response?.status !== 401) {
+			throw error;
+		}
+
+		if (this.shouldRetryAfterAuthConfigChange(error)) {
+			return this.retryAfterAuthConfigChange(error);
+		}
```
code-asher: These feel like they would be natural to be located within `recoverFromUnauthorized`, so we could see all the possible 401 recovery paths in one place.  

That would let us avoid duplicating the `_retryAttempted` check as well.

src/api/authInterceptor.ts
```diff
@@ -78,6 +82,39 @@ export class AuthInterceptor implements vscode.Disposable {
...
+			_retryAttempted?: boolean;
+			_authConfigRetryAttempted?: boolean;
+			authConfigVersion?: number;
+		};
+		// Skip if any retry already happened: caps total attempts at 2.
+		if (config._retryAttempted || config._authConfigRetryAttempted) {
+			return false;
+		}
+
+		return this.client.hasAuthConfigChangedSince(config.authConfigVersion);
```
code-asher: This makes sense but I wonder if it would be possible to compare the relevant settings instead?  If any have changed, then we re-run the request.

Or I suppose the tracker could hash the values and use that as the version.

Hmm but although that might technically cover more cases, this is probably close enough, worst case is we run it once more with values that just failed (like if a user edits and then reverts the edit).

src/api/authInterceptor.ts
```diff
@@ -78,6 +82,39 @@ export class AuthInterceptor implements vscode.Disposable {
 		return this.recoverFromUnauthorized(error, hostname);
 	}
 
+	private shouldRetryAfterAuthConfigChange(error: AxiosError): boolean {
+		if (!error.config) {
+			return false;
+		}
+
+		const config = error.config as RequestConfigWithMeta & {
```
code-asher: Not a blocker, but there must be a way we can convince Typescript these properties exist.  I know in code-server we do something similar with Express to pass around things on the request

```typescript
declare global {
  // eslint-disable-next-line @typescript-eslint/no-namespace
  namespace Express {
    export interface Request {
      args: DefaultedArgs
      heart: Heart
      settings: SettingsProvider<CoderSettings>
      updater: UpdateProvider
      cookieSessionName: string
    }
  }
}
```

src/deployment/deploymentManager.ts
```diff
@@ -232,6 +242,50 @@ export class DeploymentManager implements vscode.Disposable {
...
+			do {
+				this.#recoveryPending = false;
+				const snapshot = this.#deployment;
+				if (this.#disposed || !snapshot || this.isAuthenticated()) {
+					return;
+				}
+				this.logger.debug(
+					"Authentication settings changed after session suspended, recovering",
+				);
+				await this.setDeploymentIfValid(
```
code-asher: I wonder if we need some renaming because just looking at this function it kinda looks like we get the current deployment, then set the deployment to the current deployment, which looks like it should be a no-op.

I know as a side effect this might un-suspend the deployment, but just hard to tell from here.

src/api/coderApi.ts
```diff
@@ -503,25 +529,37 @@ function setupInterceptors(
 	);
 
 	client.getAxiosInstance().interceptors.request.use(async (config) => {
+		const configWithAuthMeta = config as RequestConfigWithAuthMeta;
+
+		for (const key of configWithAuthMeta.headerCommandKeys ?? []) {
+			config.headers.delete(key);
+		}
```
code-asher: To make sure I understand what is happening, since this could be a retry, we are deleting headers from the previous run of the header command?  Since the next run may not output the same headers.

I think this property could use a bit of documentation.

src/login/loginCoordinator.ts
```diff
@@ -214,7 +214,8 @@ export class LoginCoordinator implements vscode.Disposable {
 							resolve({ success: true, token: auth.token, user });
 						} catch {
 							// Token from other window was invalid, ignore and keep waiting
-							// (or user can click Login/Cancel in the dialog)
+						} finally {
+							client.dispose();
```
code-asher, line 220: Ooo good catch there are probably at least a few clients that we were not disposing.

I wish we had Go's `defer`

src/configWatcher.ts
```diff
@@ -1,43 +1,74 @@
 import { isDeepStrictEqual } from "node:util";
 import * as vscode from "vscode";
 
+/**
+ * Debounce window for config-change reactions that fan out (recovery,
+ * reconnect, reload prompts). Keeps rapid edits in settings.json from
+ * flushing each side-effect per keystroke.
+ */
+export const CONFIG_CHANGE_DEBOUNCE_MS = 250;
```
code-asher: kinda wild to me that VS Code is not already debouncing setting updates 💀

src/configWatcher.ts
```diff
@@ -1,43 +1,74 @@
...
+
 /**
  * Watch for configuration changes and invoke a callback when values change.
- * The callback receives a map of changed settings to their new values, so
- * consumers can act on the new value without re-reading the configuration.
- * Only fires when actual values change, not just when settings are touched.
+ * Fires only when actual values change. With `debounceMs`, the first event
+ * opens a fixed collection window; subsequent events during the window are
+ * coalesced. This bounds latency even when events arrive faster than the
+ * window length (a reset-style debounce would starve).
```
code-asher: As I understand it, a debounce means every time the function is called, it resets the timer.

So this is not actually a debounce?  A throttle then?  It fires at most once per `debounceMs` (the window)?

What does it mean to bind latency?  Is this just saying we use a throttle to avoid the case where a user keeps editing settings within 250ms over and over, keeping the function from running?  Is this really a problem?  If the user keeps editing, we really probably do want to avoid using the now-stale values.

src/configWatcher.ts
```diff
@@ -1,43 +1,74 @@
...
+			return;
+		}
+		if (!options.debounceMs) {
+			detectAndFire();
+			return;
+		}
+		if (windowTimer) {
+			return; // already collecting in the open window
+		}
+		windowTimer = setTimeout(() => {
```
code-asher: yeah this looks like a throttle

I do feel like a debounce would make more sense though

code-asher (approved):

src/configWatcher.ts
EhabY: I had it at denounce first but I worried about denouncing for a long time since this can be triggered by editing the file OR the setting through the UI. 

I can bring back the denounce but should we limit this to some max time or am I being overly cautious? Do keep in mind that this is used for the CoderApi/Remote settings watch/deployment watch... Etc

Perhaps denounce but reduce it from 250ms?

src/configWatcher.ts
EhabY: They do but it's too short, also Netflix manually modifies the settings one by one so this coalesces them into one batch

src/api/authInterceptor.ts
EhabY: Done. `recoverFromUnauthorized` owns all three paths

src/api/authInterceptor.ts
EhabY: Done via module augmentation

src/api/authInterceptor.ts
EhabY: Honestly I feel like we can keep this because some of the settings are non-deterministic (like `coder.headerCommand`)

src/api/coderApi.ts
EhabY: Yes. Documented on the augmented field in [`src/api/axios.d.ts`](src/api/axios.d.ts)

src/deployment/deploymentManager.ts
EhabY: Renamed `setDeploymentIfValid` to `verifyAndApplyDeployment`!

src/login/loginCoordinator.ts
EhabY: True..

src/configWatcher.ts
EhabY: Restored debounce instead of throttle since this is for config watching

src/configWatcher.ts
EhabY: Restored to 200ms

code-asher: I think we would need to know more about the use case that triggers that scenario before we can confidently say whether a debounce or throttle would be more appropriate for it.

Maybe a user has some thing that rapidly updates settings to new values constantly and they expect the plugin to react periodically.

Or maybe a user has something that rapidly updates settings over a long period of time but it does eventually finish and they expect the plugin to wait until they are done.

Both seem kinda crazy to me though lol

code-asher: My main thought is, if the idea is that we need the debounce/throttle because it is bad to try with the settings only partially updated, then the debounce is the only thing that could somewhat reasonably guarantee that.  A throttle forces you to update everything within 250ms which sounds tough if the settings are being manually updated, depending on the number of settings (even a 200ms debounce could be tight though)

PR #991, opened 2026-06-04
feat: add shared workspaces view

### LINKED ISSUE #992: Add a Shared Workspaces view
Add a **Shared Workspaces** view to the Coder sidebar so users can see workspaces that other people have shared with them, alongside their own.

Tracked by #991.

## What it does
- New **Shared Workspaces** tree view in the Coder activity bar, shown only when authenticated (`when: coder.authenticated`), sitting beneath **My Workspaces**.
- Lists workspaces returned by the `shared:true` query, with the owner shown (`owner / workspace`) since these belong to other users.
- Filters out workspaces the current user owns and shared outward, so the view shows only what was shared *with* them.
- Has its own search/filter command (`coder.searchSharedWorkspaces`) and refreshes on the same interval and sign-in/out events as the other views.

## Implementation notes
- `WorkspaceProvider` is now query-driven via a per-view config (`showOwner` / `showMetadata` / `excludeOwn`), so My / All / Shared share one implementation. `WorkspaceQuery.Shared = "shared:true"` was added.
- Workspace refresh keys off a shared session state (`WorkspaceSessionState`) exposed by `DeploymentManager`, with a dedicated `SessionStore` owning sign-in/out transitions and revision tracking. This drives clear-and-refetch when the user signs in or out, and prevents stale results from races during deployment switches.
- View registered as `sharedWorkspaces` in `package.json` and wired in `extension.ts`.

## Verification
`pnpm typecheck`, `pnpm lint`, and the full unit suite (1854 passed / 1 skipped) all green.

EhabY:
## Summary

- Add a `Shared Workspaces` tree view with search and manual refresh actions.
- Query `shared:true`, then filter out workspaces owned by the signed-in user inside `WorkspaceProvider` using the session's user id.
- Own the session in `DeploymentManager`/`SessionStore`; every sign-in/sign-out replaces the immutable `SessionData` object, so providers detect concurrent session changes by comparing identity and drop stale fetches.

## Testing

- `pnpm format:check`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm test:extension`

<details>
<summary>Implementation plan and decision log</summary>

- Cherry-picked PR #957 onto current `origin/main` and resolved the `DeploymentManager` conflict while preserving current main's recovery and concurrency guards.
- Verified `coder/coder` shared-workspace semantics: `shared:true` matches any workspace with non-empty user/group ACLs, including current-user-owned workspaces that were shared out, so the extension must filter out `workspace.owner_id === currentUserId` client-side.
- Avoided `shared_with_user:me` because server-side `me` support is not implemented there and direct-user filtering misses group shares.
- User chose manual refresh only for Shared Workspaces, matching `All Workspaces`, instead of polling like `My Workspaces`.
- Kept `WorkspaceProvider` free of current-user knowledge by driving per-view behavior from a `WORKSPACE_QUERY_CONFIG` table and reading the session through the read-only `SessionState` view exposed as `deploymentManager.session`.
- `SessionStore` owns the session and replaces the `SessionData` object on every sign-in/out; the provider keeps the object it read before each fetch and compares identity afterward to drop stale responses and to filter `shared:true` results by `owner_id`.
- A fetch that goes stale or arrives while another is in flight queues a re-fetch that runs once the current request settles, so requests never run in parallel and every session change ends with fresh data.

</details>

Generated with Coder Agents on behalf of @EhabY.

Fixes DEVEX-130


src/deployment/deploymentManager.ts
```diff
@@ -219,30 +235,37 @@ export class DeploymentManager implements vscode.Disposable {
...
 			async (auth) => {
-				if (this.#deployment?.safeHostname !== safeHostname) {
+				if (
+					this.#sessionStore.current.deployment?.safeHostname !== safeHostname
+				) {
 					return;
 				}
 
 				if (auth) {
 					if (this.isAuthenticated()) {
```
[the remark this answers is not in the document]
EhabY: Fixed. `verifyAndUpdateSession` now only calls `setDeployment` when the verified user actually changed. A plain token rotation just refreshes the client credentials in place, so the revision no longer bumps and the three trees are not torn down. Added a test asserting a same-user rotation updates the token without changing the revision.

src/deployment/deploymentManager.ts
```diff
@@ -255,6 +278,24 @@ export class DeploymentManager implements vscode.Disposable {
 		);
 	}
 
+	private async verifyAndUpdateSession(
```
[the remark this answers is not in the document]
EhabY: We kept verify-before-apply here on purpose. We do not want to push an unverified token onto the shared client, and the short verify delay is acceptable, so rather than applying eagerly we addressed the two concerns another way.

For the race where a concurrent 401 suspends the session mid-verify: instead of bailing whenever the state changed, we now recover. Once the new token verifies, if the session was merely suspended for this same deployment we re-sign-in with it, so the OAuth loop is not left dead.

For the failure path: we keep the old token instead of applying an unverified one. The next rotation or recovery retries. We still bail on logout or a switch to another deployment so a stale verify cannot clobber a newer session. Added tests for both the suspension recovery and the verify-failure case.

src/workspace/workspacesProvider.ts
```diff
@@ -51,89 +88,86 @@ export class WorkspaceProvider
...
-			throw new Error("not logged in");
-		} else if (url !== url2) {
-			// In this case we need to fetch from the new deployment instead.
-			// TODO: It would be better to cancel this fetch when that happens,
-			// because this means we have to wait for the old fetch to finish before
-			// finally getting workspaces for the new one.
+		// If the session changed while the request was in flight, this result is
+		// stale. Drop it and fetch again for the current session.
+		const latest = this.sessionState.getSnapshot();
+		if (latest.kind !== "signedIn" || latest.revision !== session.revision) {
```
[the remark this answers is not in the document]
EhabY: Fixed. `fetch()` is now a bounded loop (`MAX_FETCH_ATTEMPTS = 3`) with a `disposed` check at the top of each attempt, so a session that keeps changing cannot drive unbounded recursion or keep fetching after disposal.

test/unit/workspace/workspacesProvider.test.ts
```diff
@@ -0,0 +1,466 @@
...
+		const metadataItems = await provider.getChildren(metadataSection);
+		expect(metadataSection?.label).toBe("Agent Metadata");
+		expect(metadataItems[0]?.label).toBe("CPU: 42");
+
+		stream.pushError(new Error("boom"));
+		const [errorSection] = await provider.getChildren(agentItem);
+		expect(errorSection?.label).toBe("Failed to query metadata: boom");
+	});
+
+	it("empties the tree when cleared", async () => {
```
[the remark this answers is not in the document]
EhabY: Added. The test shows a provider, disposes it, fires a session change, then asserts no further fetch happens and `getChildren()` comes back empty.

test/unit/deployment/deploymentManager.test.ts
```diff
@@ -357,19 +408,24 @@ describe("DeploymentManager", () => {
...
 			// Wait for async handler
-			await new Promise((resolve) => setImmediate(resolve));
+			await flush();
 
 			expect(mockClient.host).toBe(TEST_URL);
 			expect(mockClient.token).toBe("");
 		});
 	});
 
 	describe("auth listener", () => {
```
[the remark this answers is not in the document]
EhabY: Added. The test signs in, makes the rotated-token verification reject, and asserts the client keeps the old credentials and stays signed in. That matches the verify-first choice in CRF-2.

test/mocks/testHelpers.ts
```diff
@@ -1184,3 +1208,88 @@ export class MockTerminalOutputChannel {
...
+
+/** Default user id reported by MockWorkspaceSessionState's signed-in snapshot. */
+export const TEST_CURRENT_USER_ID = "current-user";
+
+/** In-memory WorkspaceSessionState double with sign-in/out controls. */
+export class MockWorkspaceSessionState implements WorkspaceSessionState {
+	private revision = 0;
+	private readonly emitter =
+		new vscode.EventEmitter<WorkspaceSessionSnapshot>();
+	private snapshot: WorkspaceSessionSnapshot = {
```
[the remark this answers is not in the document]
EhabY: Added a test that starts signed out, signs in, and asserts the first fetch fires, which covers the real startup sequence. Left the mock default as signed-in since most tests rely on it.

src/workspace/workspacesProvider.ts
```diff
@@ -51,89 +88,86 @@ export class WorkspaceProvider
...
 		private readonly client: CoderApi,
 		private readonly logger: Logger,
-		private readonly isAuthenticated: () => boolean,
-		private readonly timerSeconds?: number,
+		private readonly sessionState: WorkspaceSessionState,
+		private readonly options: WorkspaceProviderOptions = {},
 	) {
-		// No initialization.
+		this.config = WORKSPACE_QUERY_CONFIG[getWorkspacesQuery];
+		this.sessionChangeDisposable = this.sessionState.onDidChange(() => {
```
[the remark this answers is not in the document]
EhabY: Documented with a test rather than adding a retry. The end state, where a failed poll stops polling until a manual refresh or session change, already matches the provider's contract. The test captures that behavior and confirms a manual refresh recovers.

src/deployment/sessionStore.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { User } from "coder/site/src/api/typesGenerated";
+
+import type {
+	WorkspaceSessionSnapshot,
+	WorkspaceSessionState,
+} from "../workspace/session";
+
+import type { Deployment } from "./types";
+
+/**
```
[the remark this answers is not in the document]
EhabY: Trimmed. Removed the comments that just restate the code and tightened the rest. Kept the why-comments, and kept short docs only where they add something the signature does not, like the full-versus-lean distinction on the two `SessionStore` accessors.

src/deployment/deploymentManager.ts
```diff
@@ -191,26 +211,22 @@ export class DeploymentManager implements vscode.Disposable {
 	 * Auth listener remains active so recovery can happen automatically if tokens update.
 	 */
 	public suspendSession(): void {
+		this.#sessionStore.signOut(this.#sessionStore.current.deployment);
+		this.clearSideEffects();
+	}
+
+	private clearSideEffects(): void {
```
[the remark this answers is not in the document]
EhabY: Renamed to `clearAuthState()`. It clears OAuth state, client credentials, and the auth contexts, so the name now lines up with the body.

src/workspace/session.ts
```diff
@@ -0,0 +1,14 @@
+import type * as vscode from "vscode";
+
+export type WorkspaceSessionSnapshot =
```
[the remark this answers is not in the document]
EhabY: Added a doc comment on `WorkspaceSessionSnapshot` explaining that `revision` increments on every sign-in or sign-out and how consumers snapshot it to detect a session change mid-flight.

test/mocks/testHelpers.ts
```diff
@@ -1184,3 +1208,88 @@ export class MockTerminalOutputChannel {
...
+			Promise.resolve({ workspaces: [], count: 0 }),
+	);
+
+	watchAgentMetadata(agentId: string) {
+		const stream = new MockEventStream<{ data: AgentMetadataEvent[] }>();
+		this.metadataStreams.set(agentId, stream);
+		return Promise.resolve(stream);
+	}
+
+	getAxiosInstance() {
```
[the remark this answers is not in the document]
EhabY: Removed `getAxiosInstance()` and the `baseURL` field it read from `MockWorkspacesClient`. Nothing uses them now that the provider reads `sessionState.getSnapshot()`.

src/extension.ts
```diff
@@ -100,7 +101,6 @@ async function doActivate(
 		? await secretsManager.getSessionAuth(deployment.safeHostname)
```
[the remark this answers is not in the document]
EhabY: Restored the blank line.

src/workspace/workspacesProvider.ts
```diff
@@ -160,17 +194,28 @@ export class WorkspaceProvider
...
 
-		return workspaceTreeItems;
+	private filterWorkspaces(
+		workspaces: readonly Workspace[],
+		session: Extract<WorkspaceSessionSnapshot, { kind: "signedIn" }>,
+	): readonly Workspace[] {
+		if (!this.config.excludeOwn) {
+			return workspaces;
+		}
+		// `shared:true` also returns workspaces we own and shared out; drop them
```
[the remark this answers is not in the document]
EhabY: Will add bodies to those commits explaining why `shared:true` includes current-user workspaces and why we filter by `owner_id` client-side, when we finalize the branch.

src/deployment/sessionStore.ts
```diff
@@ -0,0 +1,76 @@
+import * as vscode from "vscode";
```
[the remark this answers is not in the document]
EhabY: Updating the implementation-plan section to match the shipped code (`SessionStore`, `getSnapshot()`, `WORKSPACE_QUERY_CONFIG`) and dropping the stale `#authedUser`, `getCurrentUserId()`, `getAuthStateVersion()`, and `filterWorkspaces` references.

src/workspace/workspacesProvider.ts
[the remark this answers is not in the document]
EhabY: Done. Reworded the feature commit to explain the `shared:true` and `owner_id` filtering rationale, and gave the session-state refactor commit a body too, so every commit on the branch now carries one.

src/deployment/sessionStore.ts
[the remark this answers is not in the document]
EhabY: Done. The implementation-plan section now matches the shipped code (`SessionStore`, `getSnapshot()`, `WORKSPACE_QUERY_CONFIG`) and the removed `#authedUser`/`getCurrentUserId()`/`getAuthStateVersion()`/`filterWorkspaces` references are gone.

src/workspace/workspacesProvider.ts
```diff
@@ -51,126 +91,143 @@ export class WorkspaceProvider
...
+						continue;
+					}
 					const watcher = await createAgentMetadataWatcher(
 						agent.id,
 						this.client,
 					);
-					watcher.onChange(() => this.refresh());
+					// dispose() may have cleared the map mid-create; don't leak
+					// this watcher.
+					if (this.disposed) {
```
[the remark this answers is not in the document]
EhabY: Fixed. The post-await guard now also re-checks the session revision (via a `sessionChangedSince` helper shared with the mid-request check), so a sign-out or user switch during watcher creation drops the watcher and bails instead of overwriting the cleared tree with a stale session's workspaces.

test/unit/workspace/workspacesProvider.test.ts
```diff
@@ -0,0 +1,519 @@
...
```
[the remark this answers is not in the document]
EhabY: Added. The test makes every `getWorkspaces` response arrive stale by bumping the revision on each call, then asserts the provider stops after `MAX_FETCH_ATTEMPTS` calls and returns an empty tree instead of looping.

src/workspace/workspacesProvider.ts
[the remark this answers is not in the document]
EhabY: Added. The test makes `watchAgentMetadata` return a pending promise, disposes the provider while it is pending, then resolves it, and asserts the watcher's socket was closed and the tree is empty.

src/deployment/sessionStore.ts
```diff
@@ -0,0 +1,76 @@
...
+	readonly #onDidChange = new vscode.EventEmitter<WorkspaceSessionSnapshot>();
+
+	public readonly onDidChange = this.#onDidChange.event;
+
+	/** Full session state, including deployment and user. */
+	public get current(): SessionData {
+		return this.#data;
+	}
+
+	/** Lean projection for consumers that only track auth status and revision. */
```
code-asher: Why do we make this lean copy instead of just passing back the session data?  Seems like more memory for no benefit?

The PR description says this is to not "leak stale users" but I am not sure I understand what this means.  If a user signs in/out, and workspace tree has stored this snapshot, it will now be stale.  Even if the tree stores `#data` directly it would still be stale, since we set a new object rather than modifying it, it looks like?

And if the workspace tree is not storing the snapshot but rather calls `getSnapshot()` each time, this is still equivalent to just calling `current()` in terms of freshness.

src/deployment/sessionStore.ts
```diff
@@ -0,0 +1,76 @@
...
+			readonly deployment: Deployment;
+			readonly user: User;
+	  };
+
+/**
+ * Owns the deployment session. State changes only through signIn()/signOut(),
+ * each of which bumps the revision and notifies listeners.
+ *
+ * Consumers that only need auth status (like the workspace tree) take the lean
+ * WorkspaceSessionState projection instead of the full session.
```
code-asher: Nit: I think it is a bit of an anti-pattern to describe what should call this code because external changes may make the comment out of date ("find references" is better for this anyway).

Ideally it explains what functions should be call in what scenario, like `Prefer getSnapshot() over current() if you only need the auth status because (reason here)`.

src/deployment/deploymentManager.ts
```diff
@@ -90,53 +94,70 @@ export class DeploymentManager implements vscode.Disposable {
...
-	 * fresh logins and for un-suspending a session after auth settings or
-	 * a token become valid again. Bails if state moved during the verify
-	 * (logout, another login, dispose), so callers don't need a race guard.
+	 * Verify credentials and apply the deployment on success, signing in. Used
+	 * for fresh logins and for un-suspending a session once auth settings or a
+	 * token become valid again. Bails if state moved during the verify (logout,
+	 * another login, dispose), so callers don't need a race guard.
 	 */
-	public async verifyAndApplyDeployment(
+	public async verifyAndApplySession(
```
code-asher, line 114: I like the name change.  I even wonder if we should just have `SessionManager` instead of `DeploymentManager`.

src/deployment/deploymentManager.ts
```diff
@@ -36,18 +41,20 @@ import type * as vscode from "vscode";
...
+{
 	private readonly secretsManager: SecretsManager;
 	private readonly mementoManager: MementoManager;
 	private readonly contextManager: ContextManager;
 	private readonly logger: Logger;
 	private readonly telemetryService: TelemetryService;
 	private readonly deploymentTelemetry: DeploymentTelemetry;
 
-	#deployment: Deployment | null = null;
+	readonly #sessionStore = new SessionStore();
```
code-asher, line 53: It looks like all we do is expose the session store's methods directly on the deployment manager anyway?  Why not just put the methods directly on it?  Is there an advantage of a session store abstraction separate from the manager if it is never used in an abstract way?

src/deployment/deploymentManager.ts
```diff
@@ -90,53 +94,70 @@ export class DeploymentManager implements vscode.Disposable {
 	 * Get the current deployment state.
 	 */
 	public getCurrentDeployment(): Deployment | null {
-		return this.#deployment;
+		return this.#sessionStore.current.deployment;
+	}
+
+	public getSnapshot(): WorkspaceSessionSnapshot {
```
code-asher: `getCurrentAuthSession()` maybe?  `deploymentManager.getSnapshot()` feels non-obvious to me what it would return.

src/deployment/deploymentManager.ts
```diff
@@ -269,6 +296,48 @@ export class DeploymentManager implements vscode.Disposable {
 		);
 	}
 
+	/**
+	 * Handle a token change while already signed in: verify the new token before
+	 * applying it, so an unverified token never reaches the live client.
+	 */
+	private async verifyAndUpdateSession(
```
code-asher, line 300: Yeah seeing this really makes me feel like the deployment manager should be called a session manager, where a session means a deployment url and credentials for it.

src/workspace/session.ts
```diff
@@ -0,0 +1,21 @@
+import type * as vscode from "vscode";
+
+/**
+ * A point-in-time view of the deployment session for workspace providers.
+ *
+ * `revision` increments on every sign-in or sign-out. Consumers snapshot the
+ * revision before an async call and compare afterward to detect that the
+ * session changed while the call was in flight.
+ */
+export type WorkspaceSessionSnapshot =
```
code-asher: I see that `Workspace` here is in reference to the workspace providers, but I feel like this should really be called something like `SessionSnapshot`; this data has nothing actually to do with workspaces on their own, even if they will be used in the workspace providers.

It feels odd to me that `SessionStore` for example implements `WorkspaceSessionState`; why should the current session be dependent on the workspace views?  In practice the dependency is the other way around.

src/workspace/session.ts
```diff
@@ -0,0 +1,21 @@
+import type * as vscode from "vscode";
+
+/**
+ * A point-in-time view of the deployment session for workspace providers.
+ *
+ * `revision` increments on every sign-in or sign-out. Consumers snapshot the
+ * revision before an async call and compare afterward to detect that the
+ * session changed while the call was in flight.
```
code-asher: If we just grabbed the session data we could compare the object and have no need of an extra revision property I believe.

Although, I feel like, if possible, a better path than checking the session after a call is to immediately abort any calls whenever the session changes (via `onDidChange`).  But idk how easy this is, probably would have to support it in the api we import from coder I imagine.

src/workspace/workspacesProvider.ts
```diff
@@ -20,18 +20,56 @@ import {
...
+		excludeOwn: true,
+	},
+} as const satisfies Record<WorkspaceQuery, WorkspaceQueryConfig>;
+
+export interface WorkspaceProviderOptions {
+	readonly refreshIntervalMs?: number;
+}
+
+// Bounds fetch() retries when the session keeps changing mid-request.
+export const MAX_FETCH_ATTEMPTS = 3;
```
code-asher: Is there a downside to unbounding this?  If the session changes, we really should start a new request, even if it is the fourth time.

src/workspace/workspacesProvider.ts
```diff
@@ -51,126 +91,150 @@ export class WorkspaceProvider
...
-		private readonly isAuthenticated: () => boolean,
-		private readonly timerSeconds?: number,
+		private readonly sessionState: WorkspaceSessionState,
+		private readonly options: WorkspaceProviderOptions = {},
 	) {
-		// No initialization.
+		this.config = WORKSPACE_QUERY_CONFIG[getWorkspacesQuery];
+		this.sessionChangeDisposable = this.sessionState.onDidChange(() => {
+			this.clear();
+			void this.fetchAndRefresh();
```
code-asher, line 97: Oh wait we do fetch every time it changes.  This means the retry is not necessary right?  Would we not potentially end up with two requests in parallel?  This new one here, and potentially a retry if one was already in flight.

src/workspace/workspacesProvider.ts
```diff
@@ -51,126 +91,150 @@ export class WorkspaceProvider
...
-		if (!url2) {
-			throw new Error("not logged in");
-		} else if (url !== url2) {
-			// In this case we need to fetch from the new deployment instead.
-			// TODO: It would be better to cancel this fetch when that happens,
-			// because this means we have to wait for the old fetch to finish before
-			// finally getting workspaces for the new one.
-			return this.fetch();
-		}
+			// Session changed mid-request; this result is stale, so retry.
```
code-asher: Yeah seems like here we should just return early or abort the request, since a new request will be tried anyway.

src/workspace/workspacesProvider.ts
```diff
@@ -51,126 +91,150 @@ export class WorkspaceProvider
...
-		return workspaceTreeItems;
+	private filterWorkspaces(
+		workspaces: readonly Workspace[],
+		session: Extract<WorkspaceSessionSnapshot, { kind: "signedIn" }>,
+	): readonly Workspace[] {
+		if (!this.config.excludeOwn) {
+			return workspaces;
+		}
+		// `shared:true` also returns workspaces we own and shared out; drop them
+		// to leave only those shared with us.
```
code-asher: Nit: I think this comment would make more sense on `excludeOwn: true` above where we set it for the shared workspace filter.

code-asher (approved):
Gave it a run and it looks good to me!  Mostly just architectural comments below, and I think the retry is not necessary?

src/deployment/sessionStore.ts
EhabY: Dropped the snapshot entirely in 947a12b5. Since every transition replaces the SessionData object, comparing identity detects mid-flight changes just as well, so the revision and the lean projection are gone. The provider now compares `session.current` before and after the await.

src/deployment/sessionStore.ts
EhabY: Reworded. The docs now describe the contract (compare the object across an await to detect a change) instead of naming callers.

src/deployment/deploymentManager.ts
EhabY: Kept DeploymentManager since there is already a SessionManager in `src/oauth`, and this class also owns persistence, recovery, and cross-window sync. The session itself is now exposed as `deploymentManager.session`.

src/deployment/deploymentManager.ts
EhabY: The manager no longer mirrors the store, it just exposes it as a read-only `session` view. Kept the store as its own class because it guards the one invariant everything relies on: every change replaces the object and fires the event.

src/deployment/deploymentManager.ts
EhabY: Gone, it is `deploymentManager.session.current` now.

src/deployment/deploymentManager.ts
EhabY: Same as the thread above: kept the name to avoid clashing with the OAuth SessionManager, but the session data now lives behind `deploymentManager.session`.

src/workspace/session.ts
EhabY: Agreed, the dependency was backwards. Deleted `src/workspace/session.ts`; the contract is now `SessionState` in `sessionStore.ts` and the provider imports it from the deployment layer.

src/workspace/session.ts
EhabY: Done, providers now compare the SessionData object by identity and the revision is gone. On aborting in-flight calls: the SDK getWorkspaces does not take an abort signal, so that needs an upstream change. The identity check plus the queued re-fetch covers it for now.

src/workspace/workspacesProvider.ts
EhabY: No cap anymore. A change that lands mid-flight queues a re-fetch that runs once the current request settles, so every change gets a fresh request.

src/workspace/workspacesProvider.ts
EhabY: Close, but the `fetching` guard meant the onDidChange fetch was dropped while one was in flight, so the retry was the only thing picking up the change (and also why two requests could never run in parallel). Reworked it to match your reading: a mid-flight call now just sets `refetchPending` and fetchAndRefresh loops once the request settles. Still never parallel, and a fetch that fails after a session change now retries instead of leaving the tree empty.

src/workspace/workspacesProvider.ts
EhabY: It does now: a stale result returns null and is dropped, and the queued re-fetch issues the new request.

src/workspace/workspacesProvider.ts
EhabY: Moved.

src/workspace/workspacesProvider.ts
code-asher: Oh right duh, good point.

code-asher: Ideally we would immediately send the new request and ignore the old request once it comes back, otherwise when the session changes you could be waiting for the current request and then the next request before you finally get the correct data.  Probably not a big deal in practice though.

PR #993, opened 2026-06-08
feat: add telemetry for CLI and remote setup

### LINKED ISSUE #985: Telemetry: instrument CLI resolution and remote setup phases
Follow-up from the local telemetry audit of coder/vscode-coder#953 / coder/vscode-coder#903.

## Scope

Main already has `remote.setup`, `cli.download`, HTTP rollups, WebSocket connection telemetry, and SSH process/network telemetry. Keep this issue focused on missing phases and privacy/cardinality fixes rather than adding duplicate lifecycle events.

Add or refine:

* a parent `cli.resolve` trace covering cache lookup, version check, lock wait, download decision, download-disabled fallback, and fallback-to-existing-binary outcome
* reuse existing `cli.download` for actual downloads and its existing `verify` child phase; do not create a second download event
* one compact `cli.configure` trace for CLI credential/config writes with config mode, credential source category, result, bounded failure category, and duration
* a few additional `remote.setup` child phases where they answer “where did setup spend time?”, especially CLI resolve/configure, compatibility check, workspace monitor setup, SSH monitor setup, and VS Code handoff
* WebSocket telemetry cleanup: strip query strings, bucket/normalize routes, and add bounded backoff/attempt bucket data only if it does not create per-attempt noise

## Avoid duplicating

* `cli.download` already captures download result/duration, reason, downloaded bytes, and signature verification outcome.
* `remote.setup` already captures setup duration/result and child phases for workspace lookup, workspace readiness, agent resolution, and SSH config write.
* `connection.state_transitioned`, `connection.opened`, `connection.dropped`, and `connection.reconnect_resolved` already cover WebSocket lifecycle/reconnects.
* `ssh.process.discovered/lost/recovered/replaced/disposed` and `ssh.network.sampled` already cover SSH process and network lifecycle.

## Out of scope

* Do not add `ssh.process.spawned`; VS Code Remote SSH owns spawning and we only discover the process.
* Do not add a vague `ssh.connection.stable` unless a concrete stability threshold and support use case are defined.
* Do not add per-attempt WebSocket events if existing aggregate reconnect telemetry can answer the question.

## Privacy

Do not log raw query strings, private key paths, raw command lines, tokens, or unnormalized user-controlled paths. Use bounded route buckets and failure categories.

## Acceptance criteria

* CLI cache hit, version mismatch, disabled-download fallback, lock wait, actual download, verification, and fallback outcomes are distinguishable without duplicate download events.
* Remote setup has enough child phases to identify the slow/failing setup segment.
* WebSocket route telemetry strips query strings and avoids high-cardinality paths.
* SSH telemetry remains limited to existing lifecycle/network signals unless a precise new gap is identified.
* Tests cover representative success/failure/fallback paths.

---

Generated by Coder Agent from the telemetry audit of coder/vscode-coder#953. Updated after reviewing existing telemetry on main.

EhabY:
## Summary

Adds structured, bounded telemetry traces for the CLI and remote-setup flows:

- **`cli.resolve`** parent trace with bounded child phases — `cache_lookup`, `version_check`, `download_decision`, `lock_wait`/`lock_wait_recheck`, and `fallback_to_existing_binary` — surfaced on the parent as a single `outcome`, `cache_source`, `version_check`, `download_reason`, and a typed `failure_category`.
- **`cli.download`** with `reason` and a `downloaded_bytes` measurement, plus a `cli.download.verify` child phase for signature verification.
- **`cli.configure`** with `silent`, `credential_source`, a typed `failure_category`, result, and duration.
- **`remote.setup`** child phases for CLI resolve/configure, compatibility check, workspace lookup/monitor/ready, agent resolve, SSH config/monitor, and connection handoff, with a typed `outcome` for non-throwing early exits.
- **WebSocket reconnect** aggregate telemetry extended with `attempts`, `max_backoff_ms`, and `total_duration_ms` measurements.

Attribute names follow the #996 conventions: snake_case keys and subject-first phase names.

## Implementation notes

- Telemetry is threaded explicitly via a `trace`/`span` callback arg. Public methods (`fetchBinary`, `configure`) are thin telemetry wrappers over private implementations, keeping instrumentation out of the business logic.
- `cli.configure` failures are categorized from a typed `CredentialFileError` thrown by the storage layer rather than from error-shape heuristics.
- Multi-arg internal methods group their value params into an `options` object, with the cross-cutting `trace`/`span` as the trailing argument.
- The `cliManager` unit and telemetry tests share a `setupCliManager()` harness — per-test setup returning local mocks plus scenario builders — so the two files no longer duplicate setup.

## Verification

- `pnpm typecheck`
- `pnpm lint`
- `pnpm test` (1928 passed, 1 skipped)
- `pnpm format:check`

Closes #985

### DIFF
--- a/src/instrumentation/cli.ts
+++ b/src/instrumentation/cli.ts
@@ -1,0 +1,212 @@
+import { isAbortError } from "../error/errorUtils";
+
+import type { CallerPropertyValue } from "../telemetry/event";
+import type { TelemetryService } from "../telemetry/service";
+import type { Span } from "../telemetry/span";
+
+export type CliCacheSource = "file-path" | "directory" | "not-found";
+export type CliDownloadReason = "missing" | "version_mismatch" | "unreadable";
+export type CliResolveOutcome =
+	| "cache_hit"
+	| "downloaded"
+	| "lock_wait_cache_hit"
+	| "download_disabled_fallback"
+	| "fallback_to_existing_binary";
+export type CliVersionCheckOutcome =
+	| "missing"
+	| "match"
+	| "mismatch"
+	| "unreadable";
+export type CliConfigureFailureCategory =
+	| "cancelled"
+	| "filesystem"
+	| "credential_store"
+	| "unknown";
+export type CliResolveFailureCategory =
+	| "downloads_disabled"
+	| "download"
+	| "fallback_declined";
+
+export class CliTelemetry {
+	public constructor(private readonly telemetry: TelemetryService) {}
+
+	public resolve<T>(fn: (trace: CliResolveTrace) => Promise<T>): Promise<T> {
+		return this.telemetry.trace("cli.resolve", (span) =>
+			fn(new CliResolveTrace(span)),
+		);
+	}
+
+	public download<T>(
+		reason: CliDownloadReason,
+		fn: (span: Span) => Promise<T>,
+	): Promise<T> {
+		return this.telemetry.trace("cli.download", fn, { reason });
+	}
+
+	public configure<T>(
+		options: { silent: boolean; hasToken: boolean },
+		fn: (trace: CliConfigureTrace) => Promise<T>,
+	): Promise<T> {
+		return this.telemetry.trace("cli.configure", (span) => {
+			span.setProperty("silent", options.silent);
+			span.setProperty(
+				"credentialSource",
+				options.hasToken ? "session_token" : "empty_token",
+			);
+			return fn(new CliConfigureTrace(span));
+		});
+	}
+}
+
+export class CliResolveTrace {
+	public constructor(private readonly span: Span) {}
+
+	public setDownloadsEnabled(enabled: boolean): void {
+		this.span.setProperty("downloadsEnabled", enabled);
+	}
+
+	public setOutcome(outcome: CliResolveOutcome): void {
+		this.span.setProperty("outcome", outcome);
+	}
+
+	public setFailure(category: CliResolveFailureCategory | "unknown"): void {
+		this.span.setProperty("failureCategory", category);
+	}
+
+	public async cacheLookup<T extends { readonly source: CliCacheSource }>(
+		fn: () => Promise<T>,
+	): Promise<T> {
+		const result = await tracedPhase(
+			this.span,
+			"cache_lookup",
+			"source",
+			(r) => r.source,
+			fn,
+		);
+		this.span.setProperty("cacheSource", result.source);
+		return result;
+	}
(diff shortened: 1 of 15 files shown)


test/unit/core/cliManager.telemetry.test.ts
```diff
@@ -0,0 +1,361 @@
...
+			async (reason) => {
+				const t = setupCliManager();
+				prepare[reason](t);
+				t.withSuccessfulDownload();
+
+				await t.manager.fetchBinary(t.mockApi);
+
+				t.expectProps("cli.download", { reason, result: "success" });
+				expect(t.event("cli.download").measurements).toMatchObject({
+					durationMs: expect.any(Number),
```
ethanndickson, line 130: should `durationMs` be written as camel case where this is created?

src/instrumentation/cli.ts
```diff
@@ -0,0 +1,224 @@
...
+
+	public downloadDecision(
+		reason: CliDownloadReason,
+		action: CliDownloadAction,
+	): Promise<void> {
+		this.span.setProperty("download_reason", reason);
+		return this.span.phase("download_decision", () => Promise.resolve(), {
+			reason,
+			outcome: action,
+		});
```
ethanndickson: Why is this a no-op phase as opposed to a log?

src/instrumentation/cli.ts
```diff
@@ -0,0 +1,224 @@
...
+					const result = await fn();
+					span.setProperty("failure_category", categorizeResolveFailure(error));
+					return result;
+				} catch (fallbackError) {
+					span.setProperty(
+						"failure_category",
+						categorizeResolveFailure(fallbackError),
+					);
+					throw fallbackError;
+				}
```
ethanndickson, line 153: Should this call `setFailure` too?

ethanndickson (approved):

test/unit/core/cliManager.telemetry.test.ts
EhabY: This is an internally managed measurement, so we use JS/TS styles (camel case), when this gets translated to whatever then it is converted to snake case (JSON)

src/instrumentation/cli.ts
EhabY: Refactored this to follow the new conventions of `error.type` and the usage of `recordError`

src/instrumentation/cli.ts
EhabY: Nice catch, changed to log!

Issue #999, opened 2026-06-11
telemetry: schema-first event registry (typed catalog + generated docs)


EhabY:
## Problem

Telemetry events, properties, and value unions are declared per-domain in `src/instrumentation/*` and documented by hand in `src/instrumentation/EVENTS.md`. Nothing structurally prevents:

- emitting an event name that isn't cataloged anywhere,
- adding/renaming a property without updating the docs,
- drifting from the conventions in `src/instrumentation/CONVENTIONS.md` (a recent audit found camelCase keys, kebab-case enum values, and caller-set `result` that review missed).

## Proposal

Adopt a schema-first registry as the single source of truth for the telemetry surface, the way OTel itself defines semantic conventions in a registry and generates code + docs from it (Weaver).

TS-native sketch:

- A central `TelemetryEventMap` interface declaring every event with its properties and measurements (value unions spelled out).
- `TelemetryService.trace`/`log`/`logError` constrained to `keyof TelemetryEventMap`; `Span` becomes generic (`Span<E>`) so `setProperty`/`setMeasurement` keys and values are type-checked per event. An undeclared event or attribute key becomes a compile error.
- `EVENTS.md` generated from the registry (script walking the interface + JSDoc), with a CI `--check` mode so the doc can never go stale.

## Known costs / open questions

- Every instrumentation class and its tests are touched by the `Span<E>` generic.
- Phase names compose at runtime (`parent.child`), so child phases need registry entries or looser typing.
- The registry partially duplicates per-domain unions the instrumentation classes already export; decide whether the registry imports those types or replaces them.

## Cheaper interim guards (could land independently)

- Type the event-name parameter of `TelemetryService.trace`/`log` as a union of registered names (names are already top-level string literals; near-zero churn).
- ESLint restriction banning raw `telemetry.trace(`/`.log(` outside `src/instrumentation` (the layering rule from CONVENTIONS.md).
- Golden event catalog aggregated from the test suite's sinks, checked like the existing OTLP goldens.

PR #1000, opened 2026-06-11
docs(telemetry): add event inventory

### LINKED ISSUE #987: Telemetry: document local telemetry behavior and privacy guidance
Follow-up from the local telemetry audit of #953 / #903.

## Problem

Before we encourage users to export/share local telemetry, we should document what is collected, where it lives, and the basic privacy/cardinality rules future instrumentation should follow.

This issue is intentionally documentation/guidance only. Implementation work for export hardening and new instrumentation is tracked separately in #983, #984, #985, and #986.

### User-facing documentation

Document:

- what `coder.telemetry.level = local` means
- where local telemetry files are stored
- retention/rotation behavior
- what kinds of data may appear in exports
- how to disable local telemetry
- how to export/review telemetry before sharing
- how telemetry relates to support bundles once #952 lands

### Lightweight contributor guidance

Add a short telemetry guidance section for future instrumentation:

- use bounded event names and properties
- avoid raw query strings, file paths, tokens, task prompts/messages, and arbitrary user-provided content
- prefer enums/buckets over free-form error strings or high-cardinality dimensions
- be deliberate about deployment URLs, workspace names, agent names, machine IDs, and session IDs
- keep metrics low-cardinality

## Out of scope

- local telemetry management commands such as clear/open/show disk usage
- Chrome Trace/Perfetto timeline export
- a full schema registry or static validation framework
- support-bundle telemetry inclusion, which is already tracked in #952
- export hardening work, which is tracked in #983
- additional instrumentation, which is tracked in #984, #985, and #986

## Acceptance criteria

- Users can understand what local telemetry is, where it is stored, how long it is retained, how to disable it, and what exports may contain.
- Contributors have concise privacy/cardinality guidance for adding telemetry events.
- The docs link to the related implementation issues where appropriate.

---

Updated by Coder Agent from the telemetry audit of #953.

EhabY:
## What

Adds `src/instrumentation/EVENTS.md`: a catalog of every telemetry event the extension emits, grouped by signal kind (spans / logs / metrics) with per-kind index tables, each event's properties and measurements, value unions spelled out verbatim, conditionality notes, the framework envelope (per-event fields, session context, OTLP resource/metric export mapping), and a short section on consuming JSON/OTLP exports and support bundles. Sibling to `CONVENTIONS.md` (#996), which covers how to add new events; this covers what exists.

Also rounds out the contributor privacy guidance in `CONVENTIONS.md` from #987: no file paths or user-provided content in properties, low-cardinality metric attributes, and deliberate identifying dimensions.

## Why

Until now the only way to answer "what do we track and with which keys?" was reading every instrumentation module. A single reference makes dashboard work, privacy review, and convention audits cheap. Hand-written for now — generating and CI-checking it from a typed registry is tracked in #999.

The user-facing half of #987 (telemetry level, storage, retention, the export flow, support bundles) lands on the website in coder/coder#26215; this PR covers the contributor-guidance half.

Closes #987

### DIFF
--- a/src/instrumentation/EVENTS.md
+++ b/src/instrumentation/EVENTS.md
@@ -88,9 +88,9 @@ date range in one of two formats:
   above, for direct inspection or ad-hoc processing.
 - **OTLP**: a zip of standard OTLP/JSON envelopes (spans in `traces.json`,
   logs in `logs.json`, metric events as data points in `metrics.json`) plus
-  a `manifest.json` describing the export. Feed these to anything that
-  ingests OTLP/JSON, such as an OpenTelemetry Collector pipeline or your
-  observability backend's import tooling.
+  a `manifest.json` describing the export. Feed these to any OTel-compatible
+  tool that ingests OTLP/JSON, such as an OpenTelemetry Collector pipeline or
+  your observability backend's import tooling.
 
 The **Coder: Create Support Bundle** command also packs the raw JSONL
 telemetry files into its bundle (under `vscode-logs/telemetry`), so a support
--- a/src/instrumentation/EVENTS.md
+++ b/src/instrumentation/EVENTS.md
@@ -1,0 +1,466 @@
+# Telemetry events
+
+Every event, property, and measurement the extension emits. Conventions for
+adding new ones live in `CONVENTIONS.md`.
+
+## Event index
+
+| Category                                        | Events                                                                                                                                                                                                                                                                  |
+| ----------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| [Activation](#activation)                       | `activation`, `activation.deployment_init`                                                                                                                                                                                                                              |
+| [Auth](#auth)                                   | `auth.login`, `auth.logout`, `auth.login_prompted`, `auth.token_refresh.completed`, `auth.token_refresh.deduped`, `auth.unauthorized_intercepted`, `auth.session_lookup`, `auth.credential.store`, `auth.credential.clear`                                              |
+| [CLI](#cli)                                     | `cli.resolve`, `cli.download`, `cli.configure`                                                                                                                                                                                                                          |
+| [Commands](#commands)                           | `command.invoked`, `command.diagnostic.completed`                                                                                                                                                                                                                       |
+| [Deployment](#deployment)                       | `deployment.suspended`, `deployment.recovered`, `deployment.cross_window.detected`, `deployment.auth_config.recovery_failed`                                                                                                                                            |
+| [Remote setup](#remote-setup)                   | `remote.setup`                                                                                                                                                                                                                                                          |
+| [SSH](#ssh)                                     | `ssh.process.discovered`, `ssh.process.lost`, `ssh.process.recovered`, `ssh.process.replaced`, `ssh.process.disposed`, `ssh.network.sampled`                                                                                                                            |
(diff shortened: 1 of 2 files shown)


src/instrumentation/EVENTS.md
```diff
@@ -0,0 +1,466 @@
...
+
+Framework-managed envelope fields (wire keys, JSONL):
+
+| Field             | Meaning                                                      |
+| ----------------- | ------------------------------------------------------------ |
+| `event_id`        | Unique per event (OTel span id, 16 hex)                      |
+| `event_name`      | The event names below                                        |
+| `timestamp`       | ISO 8601 emission time                                       |
+| `event_sequence`  | Monotonic per-session counter                                |
+| `schema_version`  | Wire format version (currently 1)                            |
```
matifali: Is it 1 or 1.0?
Do we plan to follow semantic versioning on this?

src/instrumentation/EVENTS.md
```diff
@@ -0,0 +1,466 @@
...
+## SSH
+
+Emitted by `SshTelemetry`.
+
+### `ssh.process.discovered` (span)
+
+| Attribute      | Values                                    |
+| -------------- | ----------------------------------------- |
+| `found`        | `true`, `false` (whether a PID was found) |
+| `attempts` (M) | discovery attempts                        |
```
matifali: What does `(M)` mean here?

src/instrumentation/EVENTS.md
```diff
@@ -0,0 +1,466 @@
...
+heartbeat.
+
+| Attribute            | Values                                                             |
+| -------------------- | ------------------------------------------------------------------ |
+| `p2p`                | `true`, `false`                                                    |
+| `preferred_derp`     | DERP region name                                                   |
+| `latency_ms` (M)     | exports as gauge `ssh.network.sampled.latency` with unit `ms`      |
+| `download_mbits` (M) | exports as gauge `ssh.network.sampled.download` with unit `Mbit/s` |
+| `upload_mbits` (M)   | exports as gauge `ssh.network.sampled.upload` with unit `Mbit/s`   |
+
```
matifali: Can we separte `span` and `log` type to two sections?

matifali: also `metrics`

matifali: So the current headings should change from sections to their types.

`## HTTP` --> `## Metrics`

matifali (commented):
Can we add a section on how users are expected to use these OTEL exports? OR we assume the reader is already aware? 
Something like you can import them or use them in your existing ... system, etc., would help too with not going into details or specific examples.

src/instrumentation/EVENTS.md
EhabY: It's just a simple integer (so 1), we bump it if there's a breaking change

src/instrumentation/EVENTS.md
EhabY: It means measurement (as opposed to a property), expanded `M` into `measurement`

src/instrumentation/EVENTS.md
EhabY: Sure, I split them but I feel like this scattered related events since it puts logs/spans/metrics for similar events and scope in different places

matifali: I am fine either way. This was my opinion on grouping them by type.
How do some other tools with OTLP do that?

EhabY: Apparently it's mixed, with leanings more toward domain > signal kind so I'll do that

matifali (approved):

PR #1011, opened 2026-06-17
feat: support configurable CLI global config

### LINKED ISSUE #185: Add support for sharing login/auth between Coder CLI and VS Code extension
A customer has requested that it be possible to share login/auth between Coder CLI and the VS Code extension.

This is useful e.g. under strict token policies where tokens expire after 24 hours and daily login is necessary. This feature would reduce the need to login from 2 to 1 time a day.

The main problem with this approach: **We can't assume that it's OK to use or overwrite the config in e.g. `~/.config/coderv2`**, they may be configured to talk to different instances, etc.

I can think of one option, and one improvement on top of it:
1. Add a VS Code config option where you can manually specify config directory (e.g. if the user sets it to `~/.config/coderv2`, then that is the source of truth for config (read/write)
2. (Improvement) On first use, detect existing Coder config in the default paths and, if present, ask the user if they want to use it

If no config directory is specified, the behavior would be the current behavior.

EhabY:
## Summary

- Adds `coder.globalConfig` as a machine-scoped directory setting for the CLI `--global-config` path, with `CODER_CONFIG_DIR` fallback and existing per-deployment storage as the default.
- Applies the resolved config directory consistently through `PathResolver`, CLI auth flags, credential file reads/writes/deletes, active remote reload prompts, and support bundle settings.
- Preserves keyring precedence while allowing file-backed CLI credentials to be reused when keyring auth is not active.

Closes #185.

## Testing

- `pnpm test:extension ./test/unit/cliConfig.test.ts ./test/unit/core/pathResolver.test.ts ./test/unit/core/cliCredentialManager.test.ts ./test/unit/login/loginCoordinator.test.ts ./test/unit/supportBundle/settings.test.ts`
- `pnpm test`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format:check`
- `pnpm build`
- `git diff --check`

Generated by Coder Agents.

<details>
<summary>Implementation plan</summary>

- Add a dedicated `coder.globalConfig` setting for a directory path and keep `coder.globalFlags` from overriding managed `--global-config` / `--use-keyring` flags.
- Resolve global config path in `PathResolver`: `coder.globalConfig` first, then `CODER_CONFIG_DIR`, then the existing `<globalStorage>/<safeHostname>` default.
- Preserve keyring precedence in CLI auth resolution; use `--url` for active supported keyring auth, otherwise `--global-config <resolvedDir>`.
- Read file-backed CLI credentials from the resolved config directory when keyring auth is not active so CLI login can be shared with the extension.
- Add reload prompt/support bundle handling for the setting.
- Cover behavior with targeted unit tests for path resolution, CLI flags, credential manager, login fallback, and support bundle settings.

</details>


src/core/cliCredentialManager.ts
```diff
@@ -122,37 +126,38 @@ export class CliCredentialManager {
 	}
 
 	/**
-	 * Read a token via `coder login token --url`. Returns trimmed stdout,
-	 * or undefined on any failure (resolver, CLI, empty output).
-	 * Throws AbortError when the signal is aborted.
+	 * Read a token from CLI-managed credentials. Uses `coder login token --url`
+	 * when keyring auth is active, otherwise reads the file credentials under
+	 * --global-config. Returns undefined on any failure (resolver, CLI, empty
```
code-asher: Why not just use `coder login token` for both paths?  It works for file-based storage as well.  We will have call it with the global flags of course so it reads the right files.

~This would also have the advantage of being able to get the token from the default config location, since right now we require the user to fill out a separate variable.~ nvm idk why I said that, we do still read credentials from the default location with the current setup.

src/core/cliCredentialManager.ts
```diff
@@ -317,3 +375,16 @@ export class CliCredentialManager {
 		);
 	}
 }
+
+function sameCredentialUrl(storedUrl: string, expectedUrl: string): boolean {
+	return cleanCredentialUrl(storedUrl) === cleanCredentialUrl(expectedUrl);
+}
+
+function cleanCredentialUrl(url: string): string {
```
code-asher: imo `normalize` would be a clearer name than `clean`

src/core/cliCredentialManager.ts
```diff
@@ -317,3 +375,16 @@ export class CliCredentialManager {
 		);
 	}
 }
+
+function sameCredentialUrl(storedUrl: string, expectedUrl: string): boolean {
```
code-asher: Maybe `sameNormalizedUrl` or something to make it clear it is performing comparison on normalized versions

src/core/pathResolver.ts
```diff
@@ -10,13 +10,15 @@ export class PathResolver {
...
-	 * Return the directory for the deployment with the provided hostname to
-	 * where the global Coder configs are stored.
+	 * Return the directory where the global Coder configs are stored.
 	 *
 	 * The caller must ensure this directory exists before use.
 	 */
 	public getGlobalConfigDir(safeHostname: string): string {
-		return path.join(this.basePath, safeHostname);
+		return (
+			PathResolver.resolveOverride("coder.globalConfig", "CODER_CONFIG_DIR") ||
```
code-asher: Instead of separate variables could we just have users add the flag to `coder.globalFlags`?  Then we change the order so theirs takes precedence.

If we combine this with always running `coder login token` then we never need to actually know where the global dir is ourselves.

src/core/cliCredentialManager.ts
EhabY: I had made a refactor and forgot to push :sob: 

Pls review the last commit now!

src/core/pathResolver.ts
EhabY: In that case all deployments would share the same location by default which is why we introduced this per-deployment storage and the internal use of the `global-config`

src/core/cliCredentialManager.ts
EhabY: `coder login token` was only added in 2.31.0 (January of this year), we still need handle the older deployments and thus we need to read the file system

EhabY: We could do something like this:

- Is binary resolvable and >= 2.31?
  - Use `coder login token` for both paths (delegates format knowledge to the CLI).
  - Otherwise keep the direct file read.

src/core/cliCredentialManager.ts
EhabY: Done

src/core/cliCredentialManager.ts
code-asher: I think it is reasonable to scope the feature to 2.31.0 personally, unless product has asked us to make older versions work?

src/core/pathResolver.ts
code-asher: I am not sure I follow, we should keep our current default.  The simplest solution would be to just prepend our default so the user's takes precedence when the CLI parses the flags (at least, I am pretty sure when the flag is duplicated the last takes precedence).

We could also check for the flag and avoid adding our default if we wanted the flags to be a bit "cleaner" without repeats.

That does make me think though, I wonder if we should provide a variable so users can configure their own multi-deployment setup with different paths to avoid collisions.  So they could do something like  `--global-config /path/to/my/global/configs/${env:CODER_URL}` or something like that.

src/core/cliCredentialManager.ts
EhabY: We're keeping the direct file read to support deployments older than 2.31.0, where `coder login token` doesn't exist. Even scoped to 2.31+, shelling out to read the url/session files the extension wrote is more work (binary resolution + process spawn).

I am confused by the "scope to 2.31.0" since this would affect all reads and not just the when users have a custom `global-config`

src/core/pathResolver.ts
EhabY: In file mode the extension reads/writes the url+session files itself (and defaults the binary cache under this dir), so it needs the path directly, a flag in `coder.globalFlags` only reaches CLI invocations, not our own `fs` calls, which is exactly why `--global-config` is stripped from user flags today.

We could potentially extract the global-config from the `globalFlags` but this means we need to parse arguments and deal with that headache. Also, we do actually support `${env:XXX}` format here anyway!

src/core/cliCredentialManager.ts
code-asher: > shelling out to read the url/session files the extension wrote is more work

We should really be using `coder login` to write these as well IMO (this is how the Toolbox extension does it).   As far as work, I suppose it is more overhead, but to me it feels like an anti-pattern for one tool to read/write the config of another tool, there is no public API or guarantee so we rely on internal knowledge of how the configs happen to work.

> We're keeping the direct file read to support deployments older than 2.31.

If we have to do this, then we gotta do it.  But if not, we could have a single code path and only support this token-sharing feature for 2.31.0 and above.

> I am confused by the "scope to 2.31.0" since this would affect all reads and not just the when users have a custom global-config

Oh yeah this was unclear, I just meant scoping the token read feature, not the whole thing.  So for versions lower than 2.31.0, `readToken` would always return `undefined`.  In other words, if a user wants the plugin to read credentials created by the CLI, they would need to update to 2.31.0 or higher.

Actually, I guess we have two features mixed up in this PR, the global config, and also reading the token from the CLI when using file storage.

src/core/pathResolver.ts
code-asher: Good point, I forgot we do that and thought we stored them with `coder login` like the Toolbox plugin (we should switch though).

> and defaults the binary cache under this dir

Oh wait this is a really good point, IMO the user setting the `--global-config` flag should *not* affect the binary path, it should keep going into the plugin's data directory (or the binary dir if they configured that of course).  `--global-config` should be for the CLI only, so it makes no sense to store the plugin's data there too (currently they are the same by coincidence basically).

I just feel like we are going down a path of adding more and more global settings when we already have the one, but not gonna block over it.

> Also, we do actually support ${env:XXX} format here anyway!

True true, I meant to make sure we are setting `CODER_URL` since not sure we are doing that right now.

EhabY: I can make it something like this (remove the setting, use the CLI whenever possible):

## Global config / credential changes

Custom CLI config dir: was the `coder.globalConfig` setting; now `--global-config` in `coder.globalFlags` (2.31+ only, flag-only, never touches the extension's own files).

### Behavioral diff (only what changed)

| Case | Old | New |
|---|---|---|
| File write, **2.31+** | extension writes the files | `coder login --global-config` |
| File read, **2.31+** | extension reads the files | `coder login token --global-config` |
| File read, **<2.31** | extension reads the files | returns nothing (no CLI read path) |
| Keyring read, **<2.29** | extension reads the files | returns nothing |
| Custom config dir | redirected our `fs` + flag + binary cache | flag only, file mode, 2.31+ |
| Binary cache dir | followed the override | always plugin dir (`coder.binaryDestination` still applies) |

Unchanged: keyring write/read on supported versions, all writes/reads on `<2.31` default dir, and delete (`fs.rm` + `coder logout`).

**Residual:** with an override, `deleteToken` still cleans only the default dir.

code-asher: I think that broadly makes sense (`old` here means this PR right?  Not current extension behavior).

One note though on `File write, 2.31+`, this has been supported since 0.25.0 (well technically longer but we also need the `--use-token-as-session` flag added in 0.25.0 otherwise it uses the token to generate a different one).  So we can always use this, not just on 2.31+.

For the `Custom config dir` row, it says `file mode` but we want to set for keyring mode as well right?  So it knows which URL it is currently configured with.

Good call on `deleteToken`, makes sense to only clear it if the directory is the one controlled by the plugin.  We could use `coder logout` instead for this as well.

code-asher: Actually we always delete the keyring token when logging out regardless so maybe it makes sense to do the same for file mode.  Not really sure what users expect here.

src/core/cliCredentialManager.ts
EhabY: Done. Write now uses `coder login --use-token-as-session`, read uses `coder login token` (2.31+, returns nothing below that), and delete uses `coder logout`. We no longer read or write the credential files ourselves except a pre-0.25 write fallback. Read and write floors differ (0.25 write, 2.31 read), which is fine.

But actually, can we drop support to pre-0.25? That feels old enough IMO

src/core/pathResolver.ts
EhabY: Dropped the setting. A custom dir is now just `--global-config` in `coder.globalFlags`, honored on 2.31+ (file mode - since keyring would filter out the global config flag anyway). Binary cache no longer follows it, it stays in the plugin dir.

EhabY:
With https://github.com/coder/vscode-coder/pull/1011/changes/7e5d31510b4db2b8bdb2b4e11bb34a0076252895 we have the following. @code-asher I think this simplified things a bit and we can further cut down if we drop support for <0.25 (very old tbh - 3 years old).

# Global config / credential changes (vs `main`)

`<dir>` = the per-deployment config dir, default `<globalStorage>/<safeHost>`, or a user `--global-config` in `coder.globalFlags` (honored on 2.31.0+, file mode only).

## WRITE

| Mode / version | `main` | new |
|---|---|---|
| keyring (mac/win, 2.29+) | `coder login --use-token-as-session <url>` | same |
| file, 0.25+ | extension writes `<dir>/session` + `<dir>/url` | `coder login --use-token-as-session <url> --global-config <dir>` |
| file, <0.25 | extension writes the files | extension writes the files (fallback) |

## READ

| Mode / version | `main` | new |
|---|---|---|
| keyring (2.31+) | `coder login token --url <url>` | same |
| file, 2.31+ | nothing (only keyring was read) | `coder login token --url <url> --global-config <dir>` |
| <2.31 | nothing | nothing |

## DELETE

`rm <default>` files always runs: it is the only deleter when `coder logout` is unavailable (`<0.25` or binary unresolvable), and it sweeps the default dir, which `coder logout --global-config <override>` does not touch.

| Mode / version | `main` | new |
|---|---|---|
| keyring (2.29+) | `coder logout --url <url> --yes` + rm `<default>` files | same |
| file, 0.25+ | rm `<default>` files only | `coder logout --url <url> --global-config <dir> --yes` + rm `<default>` files |
| file, <0.25 / no binary | rm `<default>` files | rm `<default>` files (no `coder logout`) |

## Config dir / `--global-config`

| | `main` | new |
|---|---|---|
| Custom dir | not configurable | `--global-config` in `coder.globalFlags` (2.31+, file mode) |
| Binary cache | under `<default>` | always plugin dir (`coder.binaryDestination` still applies) |

## What it means for the user (vs `main`)

- New: point the extension at a shared CLI dir (`--global-config=~/.config/coderv2` in `coder.globalFlags`) to share login with the terminal `coder`, on 2.31+.
- New: on 2.31+ the extension picks up file based CLI credentials (`main` only read the keyring).
- File writes/deletes now go through `coder login` / `coder logout` instead of the extension touching the files directly.
- No change for keyring users.

src/core/cliCredentialManager.ts
code-asher: 100% agree we can drop support for <0.25, the Toolbox plugin does not support it either.

code-asher: ~In fact this was way back when we still had to use `vscodessh` which we straight up do not support anymore anyway, so it is impossible to use <0.25 already.~

Wait no I misread the check lol, we do still support `vscodessh` :facepalm:  ignore me.

But!  Maybe we can just update that minimum check for 0.14 to 0.25.

src/core/container.ts
```diff
@@ -83,12 +82,9 @@ export class ServiceContainer implements vscode.Disposable {
...
 				}
-				try {
-					return await this.cliManager.locateBinary(url);
-				} catch {
-					const client = CoderApi.create(url, "", this.logger);
-					return this.cliManager.fetchBinary(client);
-				}
+				// Locate-only: never download from credential ops; the connect and
+				// CLI-command flows fetch the binary first.
+				return this.cliManager.locateBinary(url);
```
EhabY, line 87: locate-only means a keyring user with no binary yet won't get their terminal token picked up at first login until something fetches the binary (main would download for them). Keep it, or revert to download-on-demand?

I'm hesitant about downloading on login/logout because that doesn't seem very UX friendly

src/core/cliCredentialManager.ts
EhabY: Made the minimum 0.25 and dropped the file write/read :eyes:

src/core/container.ts
code-asher: Yeah I think this is good, if the binary is missing then there is no credential share/sync (yet) and that seems fair to me.

src/settings/cli.ts
```diff
@@ -51,27 +51,39 @@ function buildGlobalFlags(
...
-			? ["--url", escAuth(auth.url)]
-			: ["--global-config", escAuth(auth.configDir)];
+	const userFlags = getExpandedUserGlobalFlags(configs);
+
+	// Honor a user `--global-config` only when allowOverride (file mode, 2.31+);
+	// otherwise strip it and emit our own so it matches where we read/write.
+	const honorOverride =
+		auth.mode === "global-config" &&
+		auth.allowOverride &&
+		userFlags.some((flag) => isFlag(flag, "--global-config"));
```
code-asher: Why are we not able to allow the override with the keyring?  

Currently we happen to know the CLI only stores a url and token there, but this could change in the future.  No harm in letting it be set, I think.

And to take it a bit further, IMO a config dir is better than `--url` because the user can run the CLI directly as well (like for example say we decide to expose it on PATH in the terminal or something so users can run commands directly).

src/util/uri.ts
```diff
@@ -0,0 +1,47 @@
...
+export function normalizeUrl(value: string): string {
+	return removeTrailingSlashes(value.trim());
+}
+
+/**
+ * Return the URL for opening Coder pages in the browser.  Uses the
+ * `coder.alternativeWebUrl` setting when configured, otherwise returns
+ * the connection URL unchanged.
+ */
+export function resolveUiUrl(connectionUrl: string): string {
```
code-asher: I know this already existed, but I think maybe `resolveCoderDashboardUrl` or something might be a better name, `Ui` is pretty generic

src/util/uri.ts
```diff
@@ -0,0 +1,47 @@
...
+	const alt = normalizeUrl(
+		vscode.workspace
+			.getConfiguration("coder")
+			.get<string>("alternativeWebUrl") ?? "",
+	);
+	return alt || connectionUrl;
+}
+
+/**
+ * Open a path on the Coder deployment in the user's browser, applying
```
code-asher: A nit, but the comment says it opens on the Coder deployment but really `connectionUrl` could be anything from the perspective of this function.

> Open a path resolved against coder.alternativeWebUrl if not empty, otherwise resolve against connectionUrl

or something like that

src/core/cliCredentialManager.ts
```diff
@@ -175,33 +182,109 @@ export class CliCredentialManager {
...
+		});
+		if (!cli) {
+			return { kind: "none" };
+		}
+		if (isKeyringEnabled(configs) && cli.featureSet.keyringAuth) {
+			return cli.featureSet.keyringTokenRead
+				? { kind: "keyring", binPath: cli.binPath }
+				: { kind: "none" };
+		}
+		if (cli.featureSet.keyringTokenRead) {
```
code-asher: It feels a little weird to me that we are checking a keyring capability in the keyring disabled branch.  

The actual feature is the `token` subcommand which works for both types so maybe we should rename this to `tokenRead` or something.

code-asher (approved):

src/core/container.ts
code-asher: Hrm I tested and I guess it could be a little weird for first time login because if you set a global config to share with a cli you already have downloaded and configured, it does not work because VS Code has not downloaded a binary yet.

code-asher: Which I think means the read feature will basically never be useful, because pretty much the one time you need to read the login is when you have never logged in before.  :thinking:   (I think??)

EhabY: Yeaaah it's less useful now but consistent (CLI always does the action). It could be useful still if they login every day for example so they will have the CLI but the token will be taken from the CLI login

src/settings/cli.ts
EhabY: Because we have a setting `coder.useKeyring`, if we allow `--global-config` to pass through then this will result in hiding the user setting above. I am confused by the `--url` point though, like if we do not pass `--global-config` then the CLI defaults to keyring which is why we need to pass `--url`

src/util/uri.ts
EhabY: Done

src/util/uri.ts
EhabY: Fixed!

src/core/cliCredentialManager.ts
EhabY: Ah true actually, resolved!

src/settings/cli.ts
code-asher: Is `--global-config` somehow tied to `--use-keyring`?

...ah and looking at code, apparently it is.  Why???  That makes no sense...it would be basically no extra work to support `--global-config` with the keyring, and then the CLI could write/read the current URL to the file, which would make it unnecessary to pass `--url` and make it possible for the user to invoke the CLI directly.  And what if the CLI adds extra config files in the future??  Using keyring storage means the CLI is just not allowed to ever store any kind of configuration ever?

Aaaaanyway I have more thoughts about that hahahah but I guess it is too late for that conversation now.  I see why you had to do it this way.  :stuck_out_tongue:

code-asher: Actually the CLI does seem to still read and write the URL from the global config when using the keyring?  We just are not allowed to change the global config now so it always writes to the default config location.  Which makes the whole thing even weirder.  But I guess technically that means `--url` is not required.

Which makes sense since otherwise it would be unusable on the command line, users would have to constantly add `--url` to every command.

src/core/container.ts
code-asher: Fair point, if the token has a low lifetime or something and expires, then they only have to log into the CLI again rather than log into both the CLI and VS Code, as long as they authenticate in that order (with keyring I suppose they can do it in either order, and only have to log in once).

No idea how common that is but yeah that is still useful.

We can always change this later.

src/settings/cli.ts
code-asher: Not saying we have to remove `--url`, just an observation.

In fact we *do* need `--url` since with keyring all write the URL to the default location, multiple deployments would step on each other's toes.  And I guess this means if you log into a new deployment with VS Code (or Toolbox) any CLI the user is managing themselves will also get logged into that deployment.

src/core/container.ts
EhabY: Yepp and it would work for both file (above 2.29) and keyring actually but yes, we can change it layer if need be

code-asher: File would only work from CLI to VS Code right, not VS Code to CLI?  Since we set `--global-config`, it would not be shared with the user's "main" CLI, wherever they downloaded it.

src/settings/cli.ts
EhabY: Hahaha agreed, I vaguely remember the decisions here after the fact (Nov of last year I think???) but yeah it is a bit confusing and limiting

src/core/container.ts
code-asher: Ohhh wait yeah we store it without setting `--global-config`.  Gotcha.

src/settings/cli.ts
EhabY: https://github.com/coder/coder/pull/20943

src/core/container.ts
EhabY: Oh we do set the `global-config` always so yes it would only share with the CLI for the file version IF they pass `--global-config` which is fair tbh

code-asher: Ooops I was looking at the keyring spawn :facepalm: 

Yeah I agree that seems fair.

PR #1012, opened 2026-06-17
fix: propagate proxy settings to SSH environment

### LINKED ISSUE #1010: ProxyCommand for the SSH tunnel does not receive http.proxy
## Problem

When conne ting to a workspace through the VS Code extension from behind a corporate HTTP proxy, the connection hangs at "Setting up SSH Host…" and eventually times out. The root cause is that the ProxyCommand generated by the extension runs coder ssh --stdio without any proxy configuration: no proxy flag is passed, and no proxy environment variable is injected into the subprocess. The coder CLI therefore reaches the deployment only if HTTP_PROXY/HTTPS_PROXY happen to already be present in the environment VS Code inherited.

The extension does honor http.proxy for its own API client but that setting is never propagated to the SSH tunnel subprocess, so it has no effect on the actual connection.

## Expected behavior

The extension uses the configured proxy (http.proxy, falling back to HTTP_PROXY/HTTPS_PROXY, and honoring coder.proxyBypass/http.noProxy) for the SSH tunnel, so the connection succeeds.

## Environment

- OS: Windows
- VS Code with the Remote - SSH extension (ms-vscode-remote.remote-ssh)
- Coder extension version: 1.15.0
- Network: behind a corporate HTTP proxy; the Coder deployment is reachable only via that proxy
- http.proxy configured in settings.json

EhabY:
## Summary

Propagate VS Code's proxy settings to the spawned `coder ssh` ProxyCommand so connections work behind a proxy. The `coder` CLI has no proxy flag; it reads `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` from its environment like any Go HTTP client, so the job is getting those variables into that process.

* `HTTP_PROXY`/`HTTPS_PROXY` from `http.proxy` (trimmed; blank ignored).
* `NO_PROXY` from `coder.proxyBypass`, falling back to `http.noProxy`. Bypass matching is left to the CLI's Go `NO_PROXY` parser, so it applies per-connection to whatever host the CLI dials.
* Watch these settings to prompt for a reload when the SSH proxy environment changes.

A single `applySshEnvironment` call applies the env via two mechanisms, covering both Remote-SSH modes:

* `process.env`: inherited by ssh spawned as a child process (`useLocalServer=true`).
* `EnvironmentVariableCollection`: inherited by ssh spawned in a terminal (`useLocalServer=false`), which can't see `process.env`.

Both run unconditionally, since the spawn path isn't reliably predictable from the setting and neither mechanism covers the other's mode. The collection only touches local terminals, so workspace terminals are unaffected. We mutate the environment instead of writing the SSH config, keeping credentialed URLs off disk and windows independent.

## Testing

* `pnpm lint`
* `pnpm typecheck`
* `pnpm test`

## Manual testing matrix

Verified the proxy variables reach the spawned `coder ssh` ProxyCommand by injecting a marker through each mechanism and reading the live ssh process's environment (an `ssh` wrapper via `remote.SSH.path` dumping `env`).

| Editor | Remote-SSH mode | Reaches ssh? | Via |
| -- | -- | -- | -- |
| VS Code (MS) | `useLocalServer=true` | ✅ | `process.env` |
| VS Code (MS) | `useLocalServer=false` | ✅ | `EnvironmentVariableCollection` |
| Cursor / Antigravity / Devin / VSCodium | default | ✅ | `process.env` (local-server path) |

Notes:

* `useLocalServer=false` runs the connection ssh as the shell of a hidden terminal, so it inherits the terminal environment (the collection), not the extension host's `process.env`. Modifying `process.env` does not reach terminals; the collection is the documented mechanism ([microsoft/vscode#88718](<https://github.com/microsoft/vscode/issues/88718>)).
* Remote-SSH also runs an `ssh -V` capability probe spawned directly by the extension host, which always shows `process.env` regardless of mode. Check the real connection process (the `-D ... bash` one), not the probe.

Closes coder/vscode-coder#1010


src/core/dismissibleNotifier.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const DISMISS = "Don't Show Again";
+
+/** globalState keys under which "Don't Show Again" dismissals are stored. */
+export const DISMISSIBLE_NOTIFICATION_KEYS = [
+	"coder.proxyUseLocalServerWarningDismissed",
+] as const;
+
+export type DismissibleNotificationKey =
+	(typeof DISMISSIBLE_NOTIFICATION_KEYS)[number];
```
code-asher: Why not `enum`?  This looks basically like an enum.

Or, if the idea is that we want to pass string literals then we could just do `type DISMISSIBLE_NOTIFICATION_KEYS = "key" | "key2"` since it does not look like we use the const for anything right now.

But if we will need to iterate over it or something like that in the future, then ignore me.

src/remote/environment.ts
```diff
@@ -0,0 +1,124 @@
...
+ * the spawned `coder ssh` ProxyCommand inherits them. For now that is just the
+ * proxy configuration (HTTP_PROXY/HTTPS_PROXY/NO_PROXY): the coder CLI reads them
+ * like any Go HTTP client and has no proxy flag. We mutate process.env rather
+ * than baking values into the SSH config so no credentialed proxy URL is written
+ * to disk and multiple windows onto the same workspace stay independent.
+ *
+ * Best-effort: only processes spawned afterwards inherit the change. MS VS Code
+ * with `remote.SSH.useLocalServer=false` spawns ssh off a path that does not
+ * inherit, so propagation there needs `useLocalServer=true`. Returns a disposable
+ * that restores the previous values.
```
code-asher: How would the location of the SSH binary change whether environment variables are inherited?  Or by path do you mean the code path and not the binary path?

If I understand correctly, `true` spawns a single SSH process, and then VS Code multiplexes all windows connected to that remote using that single SSH connection.

And then `false` spawns SSH in a hidden terminal per window.  And I guess when they do it this way, they are probably explicitly setting `env` which prevents inheriting as a side effect?

src/remote/environment.ts
```diff
@@ -0,0 +1,124 @@
...
+ * a proxy actually applies to a deployment via `.HTTP_PROXY`.
+ */
+export function getSshProxyEnvironment(
+	baseUrl: string,
+	cfg: Pick<WorkspaceConfiguration, "get">,
+): SshEnvironment {
+	const httpProxy = getSetting(cfg, "http.proxy");
+	const noProxy = getSetting(cfg, "coder.proxyBypass") ?? getHttpNoProxy(cfg);
+	const proxy = httpProxy
+		? getProxyForUrl(baseUrl, httpProxy, noProxy, undefined)
```
code-asher: Why do we call this differently here than in the plugin?  We could just pass the fallback as the fourth option to get the same behavior, right?

I think maybe it is so we can use `noProxy` below but this seems a bit wrong because `getProxyForUrl` also checks other variables that were not included here (like `npm_config_no_proxy`).  Maybe `getProxyForUrl` could return both `proxy` and `noProxy`?

Actually, there is no need to even set `noProxy` right?  Because `getProxyForUrl` will return a blank string if any `noProxy` is in effect.

Or, maybe better to skip `getProxyForUrl` and just pass everything in and let the CLI figure it out.  For example maybe the CLI tries to use a different region rather than the main deployment domain or something.

src/remote/environment.ts
```diff
@@ -0,0 +1,124 @@
...
+					delete env[key];
+				}
+			}
+		},
+	};
+}
+
+function getEnvKeys(env: Environment, key: string): string[] {
+	const keys = Object.keys(env).filter(
+		(envKey) => envKey.toLowerCase() === key.toLowerCase(),
```
code-asher: Environment variables are case-sensitive (at least, in Linux, Windows might be different) so I think we should not compare with lowercase or we technically cause collisions.

src/remote/environment.ts
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	let disposed = false;
+	return {
+		dispose: () => {
+			if (disposed) {
+				return;
+			}
+			disposed = true;
+			for (let i = previous.length - 1; i >= 0; i--) {
```
code-asher: Is this not using a `for of` like above because it needs to iterate backwards?  The keys are unique though so I am not sure why the order matters.

src/remote/environment.ts
```diff
@@ -0,0 +1,124 @@
...
+	cfg: Pick<WorkspaceConfiguration, "get">,
+): SshEnvironment {
+	const httpProxy = getSetting(cfg, "http.proxy");
+	const noProxy = getSetting(cfg, "coder.proxyBypass") ?? getHttpNoProxy(cfg);
+	const proxy = httpProxy
+		? getProxyForUrl(baseUrl, httpProxy, noProxy, undefined)
+		: "";
+
+	return {
+		...(proxy ? { HTTP_PROXY: proxy, HTTPS_PROXY: proxy } : {}),
```
code-asher: Is it right to set both `http_proxy` and `https_proxy`?  If the user set `http_proxy` I imagine we should not set it on `https_proxy`, and maybe vice versa.  idk why someone would have Coder on an http URL though, other than for maybe testing.

src/remote/environment.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * The proxy portion of the SSH environment. Exposed so callers can check whether
+ * a proxy actually applies to a deployment via `.HTTP_PROXY`.
+ */
+export function getSshProxyEnvironment(
+	baseUrl: string,
+	cfg: Pick<WorkspaceConfiguration, "get">,
+): SshEnvironment {
+	const httpProxy = getSetting(cfg, "http.proxy");
```
code-asher: nbd but could just do `cfg.get` right?  A blank or even untrimmed string with spaces would not break anything here (and we are not doing it elsewhere when we get the proxy URL, so if it does break we should do it there as well).

src/remote/remote.ts
```diff
@@ -818,6 +837,54 @@ export class Remote {
...
+				return;
+			}
+			if (cfg.get<boolean>("remote.SSH.useLocalServer") !== false) {
+				return;
+			}
+
+			const ENABLE = "Enable Local Server";
+			const choice = await this.dismissibleNotifier.showDismissible(
+				"coder.proxyUseLocalServerWarningDismissed",
+				"Your proxy settings may not reach the SSH connection because `remote.SSH.useLocalServer` is disabled. Enable it so Coder can apply the proxy to the connection.",
```
code-asher: Have we validated this?  I wonder if VS Code might actually explicitly include `http_proxy` when spawning, it does seem reasonable.  But maybe not.

src/remote/remote.ts
```diff
@@ -818,6 +837,54 @@ export class Remote {
...
+			// Use the jsonc writer, not cfg.update which can hang during remote setup.
+			// No reload needed: ssh hasn't spawned yet, so it picks this up on connect.
+			const ok = await applySettingOverrides(
+				this.pathResolver.getUserSettingsPath(),
+				[{ key: "remote.SSH.useLocalServer", value: true }],
+				this.logger,
+			);
+			if (!ok) {
+				this.logger.warn("Failed to enable remote.SSH.useLocalServer");
+			}
```
code-asher: Can we change the setting through the API or does it only let us change our own settings?  I think I remember trying it through the API before but it was not possible.

EhabY: We could but I'm keeping the JSONC settings writer because VS Code config updates can hang during remote setup, and this needs to land before ssh spawns

src/core/dismissibleNotifier.ts
EhabY: Applied for now since it's simpler!

src/remote/environment.ts
EhabY: Rephrased to mean the code path, and yes it's something internal but it probably overwrites the `env`

src/remote/environment.ts
EhabY: Changed this to pass proxy + `NO_PROXY` through and let the CLI/Go HTTP client apply bypass rules per request

src/remote/environment.ts
EhabY: Changed for Linux and macOS but for Windows `process.env` is case-insensitive so this still updates the existing entry

src/remote/environment.ts
EhabY: True, changed to `for of`

src/remote/environment.ts
EhabY: Simplified to `cfg.get` for proxy settings and kept the existing `http.noProxy` join behavior!

src/remote/environment.ts
EhabY: There's only one source here: VS Code's core `http.proxy`, which it applies to all traffic regardless of scheme. No separate http/https values to split apart, so we just fan it out to both

src/remote/remote.ts
EhabY: Validated against the Remote-SSH bundle (v0.125):

- `useLocalServer: true` -> ssh is spawned inheriting `process.env`, so our vars reach it.
- `useLocalServer: false` -> ssh runs in a `createTerminal()` whose env comes from VS Code's launch snapshot, not our mutated `process.env`. That's the case this warns about.

The only proxy vars VS Code sets itself go into the remote bootstrap script (the `remote.SSH.httpProxy`/`httpsProxy` settings), never the local ssh spawn.

This actually raises a point, we can this work everywhere but only if we attach the proxy to the terminal env, this means that any terminal created in this session would have those env variables so this could create a security (a credentialed `user:pass@host` URL for example)

EhabY: I've applied this in the commit https://github.com/coder/vscode-coder/pull/1012/changes/ad86f9a24d0666dd612c2b15bdc9200feaf5ffc7

Now the proxy command gets the variable all of the time regardless of the settings!

src/remote/environment.ts
code-asher: `http.proxy` is only populated if the user fills it out right?  The plugin also supports the environment variables `http_proxy` and `https_proxy` (and others) but not sure if VS Code will populate the `http.proxy` setting if these are set.

Looks like we are only passing `http.proxy` and `coder.proxyBypass || http.noProxy`.  Instead we should copy everything used in `proxy.ts` (or break it out to share in some way).

That includes the setting, and the environment variables`npm_config_http_proxy`, `npm_config_https_proxy`, `http_proxy`, `https_proxy`, `npm_config_proxy`, and `all_proxy` (both lowercase and uppercase), then for no proxy we have the settings plus `npm_config_no_proxy` and `no_proxy` environment variables (both lowercase and uppercase).

code-asher: Oh right but for the local server case those would already be inherited anyway  :thinking:   Would it make sense to add them for the non-local server case?  Unless VS Code already preserves them.

src/remote/environment.ts
```diff
@@ -0,0 +1,117 @@
...
+		...(httpProxy ? { HTTP_PROXY: httpProxy, HTTPS_PROXY: httpProxy } : {}),
+		...(noProxy ? { NO_PROXY: noProxy } : {}),
+	};
+}
+
+function applyEnvironment(
+	values: SshEnvironment,
+	env: Environment,
+): { dispose(): void } {
+	const previous: PreviousValue[] = [];
```
code-asher: nbd at all but to me it feels more natural to store the vars in the same structure as `process.env` (so a record rather than an array) where an `undefined` value means to delete it when cleaning up.

src/remote/environment.ts
```diff
@@ -0,0 +1,117 @@
...
+	const httpProxy = trimmed(cfg.get<string | null>("http.proxy"));
+	const noProxy =
+		trimmed(cfg.get<string | null>("coder.proxyBypass")) ??
+		joinNoProxy(cfg.get<string[]>("http.noProxy"));
+
+	return {
+		...(httpProxy ? { HTTP_PROXY: httpProxy, HTTPS_PROXY: httpProxy } : {}),
+		...(noProxy ? { NO_PROXY: noProxy } : {}),
+	};
+}
```
code-asher, line 78: also nbd but you could just do

```
return {
  HTTP_PROXY: httpProxy,
  HTTPS_PROXY: httpProxy,
  NO_PROXY: noProxy
}
```

since you skip setting undefined values

code-asher (approved):
Nice find on the terminal env!

src/remote/remote.ts
code-asher: Yikes that is a pretty bad bug

src/remote/environment.ts
EhabY: I just tested actually and in both cases all env variables are inherited still, it's just that `useLocalServer=false` does not inherit from the current `process.env`. In that case, we do not need to handle env variables, only the setting -> env variable

src/remote/environment.ts
EhabY: True, applied your suggestion!

src/remote/environment.ts
EhabY: Yep this is much cleaner actually

PR #1018, opened 2026-06-30
feat: show announcement banners

### LINKED ISSUE #101: Show service banners
Per a user request, we should find a way to surface the service banner in VS Code when it's set on the dashboard.

For many (if not most) deployments, developers rarely visit the dashboard and could miss important PSAs such as scheduled downtime.

However we implement this, the banner must be dismissible (maybe reappearing if it's changed or on a timer) to avoid corrupting the temple that is the IDE.

EhabY:
Closes #101

## Summary

- Surface active Coder deployment announcements in a compact status bar item: `$(megaphone) Coder` or `$(megaphone) Coder N`.
- Show popup notifications only for banners the user has not seen on that deployment, respecting `coder.disableNotifications`.
- Add a `Coder: View Announcements` quick pick to read full messages.
- Centralize Coder status bar item ids, names, and priorities in `util/statusBar` so their relative order lives in one place.

## Testing

- `pnpm typecheck`
- `pnpm lint`
- `pnpm format:check`
- `pnpm test:extension`
- `pnpm build`

<details>
<summary>Implementation notes</summary>

- Modern deployments mirror `announcement_banners[0]` into the deprecated `service_banner`, so the latter is only read when `announcement_banners` is absent (older deployments); nothing is double-counted.
- Banner fingerprints hash the message alone (the API exposes no stable banner ids): editing the text re-notifies, reordering or color-only changes do not.
- Seen fingerprints are stored per deployment and unioned on write, and banners are marked seen only when the user could actually see them (popup shown, or the quick pick opening); announcements suppressed by `coder.disableNotifications` stay eligible for a popup after notifications are re-enabled.
- Overlapping fetches are generation-guarded, so a delayed response never overwrites newer banners or the seen store.
- Non-modal popups are never awaited (they may auto-dismiss without settling); the popup's View action opens the picker from the just-fetched banners without a second request.
- Fetch cadence: session change, a 30-minute authenticated poll, and a manual refresh before the quick pick opens. Failures keep the last banners, log a warning, and only surface an error popup on the manual path; the "no announcements" message appears only after a successful refresh.
- The status bar registry preserves pre-existing placement: `agentMetadata` keeps its id and effective priority 0, and items that never had an explicit id stay id-less so users' show/hide preferences survive.
- Follow-ups: #1023 (move memento-backed state out of `SecretsManager`), #1024 (default REST timeout; like the other pollers, a hung request currently stalls the announcement poll).

</details>

Generated by Coder Agents on behalf of @EhabY.

## Visuals

image

image


src/announcements/banners.ts
```diff
@@ -0,0 +1,66 @@
+import { createHash } from "node:crypto";
+
+import type {
+	AppearanceConfig,
+	BannerConfig,
+} from "coder/site/src/api/typesGenerated";
+
+const POPUP_MESSAGE_MAX_LENGTH = 120;
```
mtojek: out of curiosity, why such limit? should we add a comment if somebody wants to increase it?

src/announcements/banners.ts
```diff
@@ -0,0 +1,66 @@
...
+	const banners = appearance.announcement_banners ?? [
+		appearance.service_banner,
+	];
+	return banners
+		.map((banner) => toAnnouncement(banner))
+		.filter((banner): banner is Announcement => banner !== undefined);
+}
+
+export function statusText(count: number): string {
+	return `$(megaphone) Coder${count === 1 ? "" : ` ${count}`}`;
```
mtojek, line 34: could you post a screenshot? I'm trying to visualize this.

src/announcements/banners.ts
```diff
@@ -0,0 +1,66 @@
...
+function bannerKey(message: string): string {
+	return createHash("sha256").update(message).digest("hex").slice(0, 16);
+}
+
+function truncate(message: string): string {
+	// Slice code points so emoji are never cut in half.
+	const chars = [...message];
+	return chars.length <= POPUP_MESSAGE_MAX_LENGTH
+		? message
+		: `${chars.slice(0, POPUP_MESSAGE_MAX_LENGTH - 1).join("")}…`;
```
mtojek: nit: should we apply smart truncation, I mean not midword ;)?

src/announcements/banners.ts
```diff
@@ -0,0 +1,66 @@
...
+): Announcement | undefined {
+	const message = banner?.message?.trim();
+	if (!banner?.enabled || !message) {
+		return undefined;
+	}
+	return { message, key: bannerKey(message) };
+}
+
+function bannerKey(message: string): string {
+	return createHash("sha256").update(message).digest("hex").slice(0, 16);
```
mtojek, line 121: thinking loud: in theory we can present two different banners with the same content, right?

src/announcements/manager.ts
```diff
@@ -0,0 +1,208 @@
...
+		}
+		const notifiable = !areNotificationsDisabled(
+			vscode.workspace.getConfiguration(),
+		);
+		if (options.notify && notifiable) {
+			this.showPopup(unseen);
+		}
+		// Mark seen only what the user could see; suppressed banners notify later.
+		if (!options.notify || notifiable) {
+			await this.secretsManager.setSeenBanners(
```
mtojek: Hm... has user actually seen the banner or not yet?

src/announcements/manager.ts
```diff
@@ -0,0 +1,208 @@
...
+	statusText,
+	statusTooltip,
+} from "./banners";
+
+import type { CoderApi } from "../api/coderApi";
+import type { SecretsManager } from "../core/secretsManager";
+import type { SessionState } from "../deployment/sessionStore";
+import type { Logger } from "../logging/logger";
+
+export const REFRESH_INTERVAL_MS = 30 * 60 * 1000;
```
mtojek, line 24: hm... should it be configurable?

src/announcements/banners.ts
```diff
@@ -0,0 +1,66 @@
...
+	readonly message: string;
+	/** Stable fingerprint used to track which banners were already seen. */
+	readonly key: string;
+}
+
+export function normalizeBanners(
+	appearance: AppearanceConfig,
+): readonly Announcement[] {
+	// Modern servers mirror announcements[0] into the deprecated service_banner.
+	const banners = appearance.announcement_banners ?? [
```
mtojek, line 23: is there possibility that `appearance.announcement_banners` is `[]`?

test/unit/announcements/banners.test.ts
```diff
@@ -0,0 +1,123 @@
...
+		expect(popupMessage(announcements("Maintenance tonight"))).toBe(
+			"Coder announcement: Maintenance tonight",
+		);
+		expect(popupMessage(announcements("First", "Second"))).toBe(
+			"Coder has 2 new deployment announcements.",
+		);
+	});
+
+	it("truncates long popup messages without splitting emoji", () => {
+		expect(popupMessage(announcements("a".repeat(121)))).toBe(
```
mtojek: nit: you can add one more case with 120 ;)

test/unit/announcements/manager.test.ts
```diff
@@ -0,0 +1,317 @@
...
+
+		nextAppearance(client, []);
+		await manager.refresh({ notify: true });
+		nextAppearance(client, ["Maintenance tonight"]);
+		await manager.refresh({ notify: true });
+
+		expect(vscode.window.showInformationMessage).not.toHaveBeenCalled();
+	});
+
+	it("shows newly seen banners on a different deployment", async () => {
```
mtojek, line 205: I'm thinking about scenario when the banner is displayed, user wants to click it, but in the meantime all banners are cleared, worth it?

src/announcements/banners.ts
EhabY: Yes but there's no other indication sadly, we do not have a "created_at" or something to indicate ID/order

src/announcements/manager.ts
EhabY: We should the banner once at least then never surface it as a notification (it can still be accessed from the status bar item or command execution)

src/announcements/banners.ts
mtojek: Ah crap! ok let's leave it 👍

src/announcements/banners.ts
EhabY: image

Hovering would show the summary (like all of the enabled banners but a single line each maximum)

test/unit/announcements/banners.test.ts
EhabY: Done!

mtojek (approved):
LGTM!

test/unit/announcements/manager.test.ts
EhabY: Added such a test case!

src/announcements/banners.ts
EhabY: Done!

EhabY: Applied!

src/announcements/banners.ts
EhabY: I'll remove this actually

src/announcements/manager.ts
EhabY: I've kept it fixed since manual clicking or login already refreshes, this is purely for bg refreshes

src/announcements/banners.ts
EhabY: It could be, but then we ignore the deprecated `service` field

mtojek (approved):
I love the final effect!

PR #1019, opened 2026-06-30
fix: honor disabled proxy support for proxy settings

### LINKED ISSUE #1017: Version 1.15.1 is setting HTTP_PROXY/HTTPS_PROXY without honoring remote workspace vscode config
It appears that the new coder extension is detecting and setting HTTP_PROXY/HTTPS_PROXY that gets pulled from the vscode user settings. Downgrading to 1.15.0 resolved that issue for me.

local user settings.json:
```
    "http.proxy": "http://127.0.0.1:3636",
    "http.proxySupport": "off",
    "http.useLocalProxyConfiguration": false,
    "workbench.browser.enableRemoteProxy": false
```

remote settings.json
```
	"git.useIntegratedAskPass": false,
	"github.gitAuthentication": false
```

After launching my coder instance HTTP_PROXY/HTTPS_PROXY env vars are consistently being set to "http://127.0.0.1:3636" and there isn't a way to disable that behavior. Version 1.15.0 does not do this!

EhabY:
## Summary

* Stop deriving SSH proxy env vars from VS Code proxy settings when `http.proxySupport` is `off`.
* Apply the same `http.proxySupport: off` behavior to Coder API HTTP, WebSocket, and SSE paths by ignoring VS Code-derived proxy settings and proxy authorization.
* Preserve inherited proxy environment variables instead of scrubbing them.
* Watch `http.proxySupport` so disconnected WebSocket streams reconnect with updated proxy behavior, and SSH sessions prompt for reload when it changes.

Fixes coder/vscode-coder#1017

## Tests

* `pnpm test:extension ./test/unit/api/utils.test.ts ./test/unit/api/coderApi.test.ts ./test/unit/remote/environment.test.ts`
* `pnpm test:extension`
* `pnpm format:check CHANGELOG.md src/remote/environment.ts test/unit/remote/environment.test.ts src/api/utils.ts src/api/coderApi.ts test/unit/api/utils.test.ts test/unit/api/coderApi.test.ts`
* `pnpm typecheck`
* `pnpm lint`

<details><summary>Investigation notes</summary>

Issue coder/vscode-coder#1017 is a regression from the v1.15.1 fix for coder/vscode-coder#1010. The extension started translating VS Code proxy settings into `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` for the `coder ssh` ProxyCommand, but only read `http.proxy`, `http.noProxy`, and `coder.proxyBypass`.

VS Code's `http.proxySupport` setting includes `off`, which means extension proxy support is disabled. With this patch, `http.proxySupport: "off"` disables only VS Code-derived proxy env/config injection. Existing process env proxy variables remain untouched, matching the selected remediation path.

For Coder API traffic, HTTP requests use `createHttpAgent` via Axios interceptors; WebSockets pass the same agent to `ws`; SSE fallback uses the same Axios instance through the streaming fetch adapter. Therefore the shared `createHttpAgent` behavior covers HTTP, WebSocket, and SSE connections.

`workbench.browser.enableRemoteProxy` was checked but is specific to VS Code browser/tunnel proxy behavior, not the SSH ProxyCommand or Coder API client.

Generated by Coder Agents.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 
 ### Fixed
 
-- Honor `http.proxySupport: off` when deriving the SSH proxy environment from
-  VS Code settings.
+- Honor `http.proxySupport: off` when deriving proxy settings for SSH and API
+  connections from VS Code settings.
 
 ## [v1.15.1](https://github.com/coder/vscode-coder/releases/tag/v1.15.1) 2026-06-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@
      from published versions since it shows up in the VS Code extension changelog
      tab and is confusing to users. Add it back between releases if needed. -->
 
+## Unreleased
+
+### Fixed
+
+- Honor `http.proxySupport: off` when deriving the SSH proxy environment from
+  VS Code settings.
+
 ## [v1.15.1](https://github.com/coder/vscode-coder/releases/tag/v1.15.1) 2026-06-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,20 @@
 
 ## Unreleased
 
+> **Breaking:** API requests now respect `http.proxySupport: off`. Previously
+> the extension applied VS Code's proxy settings (`http.proxy`, `http.noProxy`,
+> `coder.proxyBypass`) to API requests even when `http.proxySupport` was `off`.
+> Now `off` ignores those settings and only proxy environment variables are
+> used. If you set those settings and relied on them while proxy support was
+> `off`, either set `http.proxySupport` to `on` to keep using them, or move the
+> values into the `HTTP_PROXY`/`HTTPS_PROXY` (from `http.proxy`) and `NO_PROXY`
+> (from `http.noProxy`/`coder.proxyBypass`) environment variables.
+
 ### Fixed
 
 - Honor `http.proxySupport: off` when deriving proxy settings for SSH and API
-  connections from VS Code settings.
+  connections, so VS Code's proxy settings are ignored while inherited proxy
+  environment variables still apply.
 
 ## [v1.15.1](https://github.com/coder/vscode-coder/releases/tag/v1.15.1) 2026-06-26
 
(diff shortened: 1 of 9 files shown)


src/api/utils.ts
```diff
@@ -28,8 +28,16 @@ export async function createHttpAgent(
...
-	const httpNoProxy = cfg.get<string[]>("http.noProxy");
+	// "off" ignores VS Code proxy config; inherited env proxies still apply.
+	const proxyEnabled = cfg.get<string>("http.proxySupport") !== "off";
+	const proxySetting = <T>(key: string) =>
+		proxyEnabled ? cfg.get<T>(key) : undefined;
+	const proxyAuthorization = proxySetting<string | null>(
+		"http.proxyAuthorization",
+	);
+	const httpProxy = proxySetting<string | null>("http.proxy");
+	const coderProxyBypass = proxySetting<string | null>("coder.proxyBypass");
```
code-asher, line 39: We should update the description for `coder.proxyBypass` because it suggests it has an effect when `http.proxySupport` is `on` or `off` but with `off` it will no longer have any effect.

While we are in there maybe should try deprecating it saying to prefer `http.noProxy`.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
      from published versions since it shows up in the VS Code extension changelog
      tab and is confusing to users. Add it back between releases if needed. -->
 
+## Unreleased
+
+### Fixed
+
+- Honor `http.proxySupport: off` when deriving proxy settings for SSH and API
+  connections from VS Code settings.
```
code-asher: IMO we should say this specifically disables reading plugin and VS Code proxy settings but environment variables will still be respected.

Also I suppose technically this is a breaking change, right?  Because before `coder.proxyBypass` would still apply if you set the proxy via env vars and had proxy support set to `off`.

idk why anyone would actually do that, but might make sense to call out the breaking change and say to set `no_proxy` in the env instead if anyone was doing this.

code-asher (approved):

CHANGELOG.md
code-asher: Actually I guess more than that is a breaking change, if someone had `http.proxy` and support off, previously the API would use it, so maybe we can call that out and lay out the recommended fix if someone was relying on that behavior (turn on proxy support I imagine is the fix, or use env vars instead).

src/api/utils.ts
EhabY: Oh yes, added a deprecation message and changed the description

CHANGELOG.md
EhabY: Added more info about this in the changelog!

PR #1026, opened 2026-07-07
feat: use shared_with_user filter for the shared workspaces view

EhabY:
## Summary

- The Shared Workspaces view now queries `shared_with_user:<current user id>` instead of fetching `shared:true` and filtering out own workspaces client-side. The server-side filter returns exactly the workspaces shared *with* the user, so `excludeOwn`/`filterWorkspaces()` are removed.
- When the server rejects the workspaces query with HTTP 400 (deployments older than 2.27.0), the view is hidden via a new `coder.sharedWorkspacesSupported` context key instead of showing an empty tree. The flag resets on any session change (login/logout/deployment switch) and on `coder.refreshWorkspaces`, so the view re-probes without restarting VS Code.
- Added `initialSize` weights to the sidebar views: My Workspaces `2`, Shared `1`, All `1` (50/25/25 when all are visible). Only affects fresh installs / reset layouts.

## Compatibility

`shared_with_user:` shipped in the same server release as `shared:true` (both merged Sept 2025, first released in **v2.27.0** — coder/coder#19807 and coder/coder#19875), so there is no deployment where the old query works and the new one doesn't. `shared_with_user:me` was intentionally avoided since "me" support only landed in coder/coder#26494 (v2.35+); the user UUID works on every supported version and is immune to renames.

On pre-2.27 deployments the previous behavior was a logged fetch error and a permanently empty Shared view; it is now hidden entirely.

<details>
<summary>Implementation plan / decision log</summary>

1. `workspacesProvider.ts`
   - `WorkspaceQuery` enum values become plain identifiers (`mine`/`all`/`shared`); the actual search query moves into `WORKSPACE_QUERY_CONFIG` as `getQuery(session)`, built at fetch time because the Shared query needs the signed-in user's ID.
   - UUID over username: `parseUser` on the server accepts both, but the UUID avoids rename/quoting issues.
   - New `onQueryRejected` option, invoked only for axios HTTP 400 responses. Network errors/5xx keep the existing behavior (warning log + empty tree) so transient failures never hide the view.
2. `extension.ts` / `contextManager.ts`
   - New context key `coder.sharedWorkspacesSupported`, default `true`.
   - Shared provider sets it to `false` on `onQueryRejected`.
   - Reset to `true` on `deploymentManager.session.onDidChange` and in the `coder.refreshWorkspaces` command (covers in-place server upgrades: re-showing the view triggers its fetch, which hides it again if still unsupported). Hiding the view also stops its polling, since visibility drives the fetch loop.
3. `package.json`
   - `sharedWorkspaces` view gated on `coder.authenticated && coder.sharedWorkspacesSupported` (same pattern as `coder.isOwner` on `allWorkspaces`).
   - `initialSize` is a stable views-contribution property honored when the extension owns the view container (it does).
4. Tests: query expectations updated; the client-side ownership-filtering tests are replaced with `onQueryRejected` coverage (400 fires it once, non-400 does not).

</details>

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm format:check`
- `pnpm test` — 143 files, 2149 passed / 1 skipped

---

> Generated by Coder Agents on behalf of @EhabY.

### DIFF
--- a/package.json
+++ b/package.json
@@ -316,20 +316,23 @@
 					"id": "myWorkspaces",
 					"name": "My Workspaces",
 					"visibility": "visible",
-					"icon": "media/logo-white.svg"
+					"icon": "media/logo-white.svg",
+					"initialSize": 2
 				},
 				{
 					"id": "sharedWorkspaces",
 					"name": "Shared Workspaces",
 					"visibility": "visible",
 					"icon": "media/logo-white.svg",
-					"when": "coder.authenticated"
+					"initialSize": 1,
+					"when": "coder.authenticated && coder.sharedWorkspacesSupported"
 				},
 				{
 					"id": "allWorkspaces",
 					"name": "All Workspaces",
 					"visibility": "visible",
 					"icon": "media/logo-white.svg",
+					"initialSize": 1,
 					"when": "coder.authenticated && coder.isOwner"
 				}
 			],
--- a/package.json
+++ b/package.json
@@ -325,7 +325,7 @@
 					"visibility": "visible",
 					"icon": "media/logo-white.svg",
 					"initialSize": 1,
-					"when": "coder.authenticated && coder.sharedWorkspacesSupported"
+					"when": "coder.authenticated"
 				},
 				{
 					"id": "allWorkspaces",
@@ -371,6 +371,16 @@
 				"view": "coder.tasksLogin",
 				"contents": "Sign in to view and manage Coder tasks.\n[Login](command:coder.login)",
 				"when": "!coder.authenticated && coder.loaded"
+			},
+			{
+				"view": "sharedWorkspaces",
+				"contents": "Shared workspaces require Coder 2.27.0 or newer.",
+				"when": "coder.authenticated && !coder.sharedWorkspacesSupported"
+			},
+			{
+				"view": "sharedWorkspaces",
+				"contents": "No workspaces have been shared with you.",
+				"when": "coder.authenticated && coder.sharedWorkspacesSupported"
 			}
 		],
 		"commands": [
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -194,17 +194,16 @@ async function doActivate(
 		output,
 		deploymentManager.session,
 		{
-			// Deployments older than 2.27.0 reject the shared workspaces query;
-			// hide the view for them (see `when` in package.json).
+			// Older deployments reject this query; show a message instead of an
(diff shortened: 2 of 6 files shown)


src/extension.ts
```diff
@@ -184,9 +184,24 @@ async function doActivate(
 		client,
 		output,
 		deploymentManager.session,
+		{
+			// Deployments older than 2.27.0 reject the shared workspaces query;
```
code-asher: Instead of hiding are we able to show a message?  "Your deployment does not support shared workspaces" or something like that.

Partially because it feels a bit wonky when it appears for a second then disappears.

But I have a suspicion the tree view is not that flexible 😢   I think we already are not showing errors except in the logs.  So maybe this is not something we can do unless/until we switch to web views.

package.json
```diff
@@ -316,20 +316,23 @@
 					"id": "myWorkspaces",
 					"name": "My Workspaces",
 					"visibility": "visible",
-					"icon": "media/logo-white.svg"
+					"icon": "media/logo-white.svg",
+					"initialSize": 2
```
code-asher, line 320: oo nice

src/extension.ts
```diff
@@ -319,6 +334,9 @@ async function doActivate(
 		commands.navigateToWorkspaceSettings.bind(commands),
 	);
 	commandManager.register("coder.refreshWorkspaces", () => {
+		// Re-probe shared workspaces support in case the deployment was upgraded;
+		// showing the view triggers its fetch, which hides it again on rejection.
+		contextManager.set("coder.sharedWorkspacesSupported", true);
```
code-asher: Something seems to be a bit odd with the context here, I simulated by changing the query to something invalid and when I refresh the shared workspaces panel appears and does not disappear, then I refresh again and it disappears, refresh again and it reappears.

Based on the logs it appears the odd-numbered refreshes are not actually triggering requests.

src/extension.ts
EhabY: Turns out yes — this repo already uses `viewsWelcome` for the login prompts. Replaced the `when` clause that hid the view with a `viewsWelcome` message instead (plus one for the supported-but-empty case). Bonus: this also removes the visibility toggle causing the race in your other comment.

src/extension.ts
EhabY: Fixed by keeping the view visible at all times instead of hiding it based on context

code-asher (approved):
Nice!!

PR #1029, opened 2026-07-09
refactor: move memento-backed state out of SecretsManager

### LINKED ISSUE #1023: Split non-secret memento state out of SecretsManager (or rename it)
`SecretsManager` has grown beyond VS Code `SecretStorage`: it now also persists non-secret state in the workspace `Memento` — deployment access timestamps (`coder.access.*`) and seen announcement banners (`coder.seenBanners.*`, added in the announcement banners PR).

The name implies everything in it is secret, which is confusing when reading or reviewing storage code.

Options:

- Split the memento-backed state into a separate manager (e.g. `StateManager`/`MementoStore`) and keep `SecretsManager` secrets-only.
- Or rename `SecretsManager` to something storage-neutral (e.g. `StorageManager`) that owns both backends.

Either way, per-deployment cleanup (`clearAllAuthData`) needs to keep clearing both stores together.

## Recommended approach

Keep `SecretsManager` secrets-only and move the memento-backed state into the existing `MementoManager` (`src/core/mementoManager.ts`), which is already the home for non-secret persisted state:

1. Move `getSeenBanners`/`setSeenBanners`/`clearSeenBanners` and the `coder.access.*` timestamp reads/writes to `MementoManager` (same keys and formats, so no data migration).
2. Inject `MementoManager` into `SecretsManager`. Secret writes keep stamping deployment access and `getKnownSafeHostnames` keeps ordering by it; `SecretsManager` orchestrates but no longer stores non-secret data.
3. `clearAllAuthData` stays on `SecretsManager` as the single per-deployment cleanup entry point, delegating the memento clears to `MementoManager` so both stores stay in sync.
4. `AnnouncementManager` switches its seen-banner dependency from `SecretsManager` to `MementoManager`.

Renaming `SecretsManager` (e.g. to `StorageManager`) was considered and rejected: the name gets vaguer while the class would still mix two backends.

EhabY:
Closes #1023

`SecretsManager` had grown to also persist non-secret state in the workspace `Memento`: deployment access timestamps (`coder.access.*`), surfaced announcement banners (`coder.surfacedBanners.*`), and the legacy `url` key. This moves all of it into `MementoManager`, keeping `SecretsManager` secrets-only.

- `SecretsManager` now takes a `MementoManager` instead of a raw `Memento`. Secret writes still stamp deployment access, `getKnownSafeHostnames` still orders by it, and `clearAllAuthData` remains the single per-deployment cleanup entry point, delegating the memento clears so both stores stay in sync.
- `AnnouncementManager` depends on `MementoManager` directly. The surfaced-banner read-modify-write moved into `MementoManager.addSurfacedBanners`, so any future atomic memento update can be adopted in one place.
- Keys and formats are unchanged, so no data migration.

Follow up to https://github.com/coder/vscode-coder/pull/1018

### DIFF
--- a/src/announcements/manager.ts
+++ b/src/announcements/manager.ts
@@ -16,7 +16,7 @@ import {
 import { AnnouncementsPreview } from "./preview";
 
 import type { CoderApi } from "../api/coderApi";
-import type { SecretsManager } from "../core/secretsManager";
+import type { MementoManager } from "../core/mementoManager";
 import type { SessionState } from "../deployment/sessionStore";
 import type { Logger } from "../logging/logger";
 
@@ -43,7 +43,7 @@ export class AnnouncementManager implements vscode.Disposable {
 	public constructor(
 		private readonly client: Pick<CoderApi, "getAppearance">,
 		private readonly sessionState: SessionState,
-		private readonly secretsManager: SecretsManager,
+		private readonly mementoManager: MementoManager,
 		private readonly logger: Logger,
 	) {
 		this.statusBarItem.command = "coder.viewAnnouncements";
@@ -149,7 +149,7 @@ export class AnnouncementManager implements vscode.Disposable {
 		this.banners = banners;
 
 		const surfacedKeys = new Set(
-			this.secretsManager.getSurfacedBanners(session.deployment.safeHostname),
+			this.mementoManager.getSurfacedBanners(session.deployment.safeHostname),
 		);
 		const newBanners = banners.filter(
 			(banner) => !surfacedKeys.has(banner.key),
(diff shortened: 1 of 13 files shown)


src/core/mementoManager.ts
```diff
@@ -72,6 +79,68 @@ export class MementoManager {
...
+		);
+		const result = SurfacedBannersSchema.safeParse(raw);
+		return result.success ? result.data : [];
+	}
+
+	/**
+	 * Merge banner keys into the surfaced set. The read-modify-write lives
+	 * here so a future atomic Memento update can be adopted in one place.
+	 */
+	public async addSurfacedBanners(
```
mtojek, line 117: shouldn't this function be thread safe?

src/announcements/manager.ts
```diff
@@ -168,36 +168,20 @@ export class AnnouncementManager implements vscode.Disposable {
 			return;
 		}
 		this.showPopup(newBanners);
-		await this.markSurfaced(banners, surfacedKeys);
+		await this.markSurfaced(banners);
```
mtojek, line 171: re removing `surfacedKeys`: is it intended? code simplification?

src/core/mementoManager.ts
```diff
@@ -72,6 +79,68 @@ export class MementoManager {
...
+			await this.memento.update(`${SURFACED_BANNERS_PREFIX}${safeHostname}`, [
+				...merged,
+			]);
+		}
+	}
+
+	/** Clear all per-deployment state (access timestamp, surfaced banners). */
+	public async clearDeploymentData(safeHostname: string): Promise<void> {
+		await Promise.all([
+			this.memento.update(
```
mtojek, line 133: shouldn't we clear deployment url too?

src/core/container.ts
```diff
@@ -50,7 +50,7 @@ export class ServiceContainer implements vscode.Disposable {
 		this.mementoManager = new MementoManager(context.globalState);
 		this.secretsManager = new SecretsManager(
 			context.secrets,
-			context.globalState,
+			this.mementoManager,
```
mtojek, line 53: nit: do we need an assertion to check if `mementoManager` has been initialized?

src/core/mementoManager.ts
EhabY: It could be yes but we could also have multiple windows connected to the same deployment which we cannot prevent. I suppose we can prevent it within the same window only though

src/announcements/manager.ts
EhabY: Yep we simplified by just reading internally (which is sync anyway)

src/core/container.ts
EhabY: I'm pretty sure that TS would not compile if we try to pass a local variable BEFORE initializing

mtojek: Ok, no need fo guard then 👍 

EDIT:
btw as initialized, I didn't mean TS initialization, but actually preparing the internal state of the manager.

mtojek (approved):
Nice!

src/core/mementoManager.ts
EhabY: This is out of scope of the memento refactor IMO, it's a different code path

src/core/container.ts
EhabY: Just confirmed and it would not compile!

EhabY: Oh it would be initialized if the object is created (essentially it's mostly a delegation of the VS code types)

PR #1041, opened 2026-07-18
ci: replace Chromatic with Pixel

EhabY:
Replaces Chromatic with [Pixel](https://pixel.coder.com) for visual regression testing, mirroring the `coder/coder` setup. Resolves DEVEX-634.

- New `pixel` CI job: installs Chromium, runs `pixel-storybook`, which builds the Storybook and snapshots every story in light and dark. Main builds auto-approve via `PIXEL_AUTO_REVIEW`, like Chromatic's `autoAcceptChanges` did, so squash merges don't block CI.
- `pixel.jsonc` holds the project ID and snapshot matrix. Chrome-only at the default viewport, since the webviews only ever render inside VS Code.
- Deps: `chromatic` out, `@coder/pixel-storybook` and `playwright-core` in, plus a build-script allowance for `odiff-bin` (image diffing).
- New `.prettierrc.json` stops Prettier from adding trailing commas in `.jsonc` files, which VS Code's JSONC mode warns about.

`PIXEL_KEY` is already set on the repo. The job skips forks and Dependabot since they can't read it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,49 +81,26 @@ jobs:
         run: ${{ runner.os == 'Linux' && 'xvfb-run -a' || '' }} pnpm test:integration --label "VS Code ${{ matrix.vscode-version }}"
         shell: bash
 
-  chromatic:
-    name: Chromatic
+  pixel:
+    name: Pixel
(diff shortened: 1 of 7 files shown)


johnstcn (approved):
Stamp 👍

PR #1042, opened 2026-07-20
feat: collect remote editor logs in support bundles

EhabY:
## Summary

- Collect local Remote-SSH, Coder proxy, and remote editor-server logs in support bundles.
- Pass remote files through Coder's repeatable `--workspace-file` support, gated to Coder 2.36.0+.
- Preserve the selected workspace agent and fail closed when a safe remote path cannot be determined.

## Remote log coverage

For each resolved server data directory, collect these narrow known layouts:

- `data/logs/**/*.log` — editor runtime, extension-host, and related logs
- `.*.log` — legacy server startup logs
- `cli/servers/*/log.txt` — exec-server startup logs

Both startup layouts are included rather than gated by the current `remote.SSH.useExecServer` value: the setting can change after connection, Open Remote SSH uses the legacy layout independently, and supported IDE forks do not expose one consistent authoritative mode signal.

## Log path resolution

Resolution is deterministic from local state; no remote connection is needed:

1. Interpret the selected Remote-SSH extension's `remote.SSH.serverInstallPath` for the workspace's SSH host. Relative paths are anchored to `~/` since SSH resolves them against the home directory.
2. Fall back to `~/{product-server-folder}` from the editor's `product.json`, the server's own default.
3. Omit remote logs rather than scan broad paths or pass unresolved environment variables.

The SSH host comes from the live authority when connected and is otherwise reconstructed from the workspace and agent (first agent for workspace-level items, matching the CLI), so sidebar, picker, and palette flows resolve identically.

An earlier revision also read `VSCODE_AGENT_FOLDER` from the active exec server. Dropped after verifying each implementation's actual code (local extension bundles for Microsoft, Cursor, Windsurf, and Antigravity; source for Open Remote SSH):

- Microsoft and Cursor derive that variable from `serverInstallPath` themselves and export it at server launch, overriding any user-set value, so settings resolve the identical path.
- Open Remote SSH never uses the variable; it passes `--server-data-dir` instead.
- Windsurf and Antigravity have no install path setting under any config prefix and hardcode `$HOME/{product-server-folder}`.
- `getRemoteExecServer` is not a passive read: when not connected it spawns a real SSH connection and installs the VS Code CLI on the remote, so it could never safely serve the sidebar flow.

The resolver uses the same first supported Remote-SSH extension selected during activation.

| Remote-SSH extension | Host matching | `serverInstallPath` behavior |
| --- | --- | --- |
| Microsoft Remote-SSH | Exact host | Parent directory; append the product server folder |
| Cursor Remote SSH | Exact host | Parent directory; append the product server folder without duplicating an existing suffix (verified in its installer script) |
| Open Remote SSH | Exact host, then most-specific wildcard, then `*` | Final server data directory; use directly |
| Windsurf Remote SSH | Not applicable | No setting exists; always `~/.devin-server` (formerly `~/.windsurf-server`) |
| Antigravity Remote SSH | Not applicable | No setting exists; always `~/.antigravity-server` |

## Testing

- `pnpm test` — 2,255 passed, 1 skipped
- `pnpm lint`
- `pnpm typecheck`
- `pnpm format:check`
- `git diff --check`

Generated by Coder Agents.


src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,193 @@
...
+		remotePath.join(value, "cli", "servers", "*", "log.txt"),
+	];
+}
+
+async function getActiveServerDataPath(
+	remoteAuthority: string,
+): Promise<RemoteServerDataPath | undefined> {
+	try {
+		if (!vscodeProposed.workspace.getRemoteExecServer) {
+			return undefined;
```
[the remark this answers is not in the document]
EhabY: Fixed in faffb1c5. The resolver and workspaceFiles now take a Logger. Thrown errors and rejected unsafe values warn; an unavailable exec server or absent configuration stays quiet.

src/supportBundle/workspaceFiles.ts
```diff
@@ -0,0 +1,73 @@
...
+		? await getRemoteServerDataPath({
+				remoteAuthority,
+				serverDataFolderName,
+			})
+		: undefined;
+
+	const serverDataRoot: RemoteServerDataPath | undefined =
+		serverDataPath ??
+		(serverDataFolderName
+			? { value: `~/${serverDataFolderName}`, style: "posix" }
```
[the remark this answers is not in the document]
EhabY: Checked the agent side (agent/agentfiles/bundlefiles.go): matching runs doublestar over filepath.ToSlash(abs), and ~/ expands via filepath.Join, so forward-slash globs also match on Windows agents. The posix fallback is portable; a comment now documents that contract. The two-definitions gap is closed by the CRF-7 fix, which applies the same unsafe character rejection to the folder name.

src/commands.ts
```diff
@@ -1133,15 +1146,22 @@ export class Commands {
...
 			return {
 				status: "selected",
+				agentName: item instanceof AgentTreeItem ? item.agent.name : undefined,
 				client: this.extensionClient,
 				workspaceId: createWorkspaceIdentifier(item.workspace),
 			};
 		}
 		if (this.workspace && this.remoteWorkspaceClient) {
+			const remoteAuthority = vscodeProposed.env.remoteAuthority;
+			const authorityParts = remoteAuthority
```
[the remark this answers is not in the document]
EhabY: Fixed in faffb1c5. The parse is wrapped in try/catch and a malformed authority degrades agentName to undefined with a warning; covered by a command-layer test.

src/commands.ts
```diff
@@ -1133,15 +1146,22 @@ export class Commands {
 		if (item) {
```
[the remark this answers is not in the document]
EhabY: Added commands.supportBundle.test.ts in faffb1c5: agent from an AgentTreeItem selection, agent and authority from the active connection, malformed authority degradation, and the 2.36 feature gate.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,193 @@
...
+	for (const [pattern, installPath] of Object.entries(pathMap)) {
+		const score = hostnamePatternScore(hostname, pattern);
+		if (score > 0 && (!bestMatch || score > bestMatch.score)) {
+			bestMatch = { path: installPath, score };
+		}
+	}
+	return bestMatch?.path;
+}
+
+function hostnamePatternScore(hostname: string, pattern: string): number {
```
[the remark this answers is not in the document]
EhabY: Keeping as is. The scorer mirrors Open Remote SSH's matchHostnamePattern, not OpenSSH: the cited source escapes ? and matches it literally (serverSetup.ts L28-47), exactly like this copy. Reusing computeSshProperties would treat ? as a single-character wildcard and diverge from the implementation whose config this interprets. The literal tiers keep the copy line-comparable with upstream. The new wildcard precedence test from CRF-9 locks the behavior.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,193 @@
...
+
+/** Resolve the active remote server's data directory when possible. */
+export async function getRemoteServerDataPath({
+	remoteAuthority,
+	serverDataFolderName,
+}: RemoteServerDataPathOptions): Promise<RemoteServerDataPath | undefined> {
+	const configuredServerDataPath = serverDataFolderName
+		? getConfiguredServerDataPath(remoteAuthority, serverDataFolderName)
+		: undefined;
+	return configuredServerDataPath ?? getActiveServerDataPath(remoteAuthority);
```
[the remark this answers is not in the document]
EhabY: Keeping the configured-first order, with the rationale now documented in faffb1c5. server.main.ts L39 resolves the data folder as args['server-data-dir'] || VSCODE_AGENT_FOLDER || join(homedir, serverDataFolderName), and supported implementations pass serverInstallPath as that flag, so the resolver mirrors the server's own precedence tier for tier. Inverting would lose whenever both signals exist and disagree, since the flag outranks the env var.

src/supportBundle/workspaceFiles.ts
```diff
@@ -0,0 +1,73 @@
...
+		return isSafeServerDataFolderName(product.serverDataFolderName)
+			? product.serverDataFolderName
+			: undefined;
+	} catch {
+		return undefined;
+	}
+}
+
+/** Return whether the value is a single portable path segment. */
+function isSafeServerDataFolderName(value: unknown): value is string {
```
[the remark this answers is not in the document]
EhabY: Fixed in faffb1c5. isSafeServerDataFolderName now applies the shared hasUnsafePathChars rejection, with glob-case tests.

test/unit/core/cliExec.test.ts
```diff
@@ -271,7 +271,14 @@ describe("cliExec", () => {
 				bin,
 			);
 			configs.set("coder.headerCommand", "my-header-cmd");
-			await cliExec.supportBundle(env, "owner/workspace", outputPath);
+			await cliExec.supportBundle(env, "owner/workspace", {
```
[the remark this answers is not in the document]
EhabY: Added in faffb1c5: a default-shape test asserts the exact argv with no positional agent and no --workspace-file flags.

test/unit/supportBundle/remoteServerDataPath.test.ts
```diff
@@ -0,0 +1,240 @@
...
+
+		await expect(
+			getRemoteServerDataPath({ remoteAuthority, serverDataFolderName }),
+		).resolves.toEqual({
+			value: "C:\\Users\\coder\\editor\\.vscode-server",
+			style: "win32",
+		});
+	});
+
+	it("uses Open Remote SSH's most specific matching path as the final folder", async () => {
```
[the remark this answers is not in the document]
EhabY: Added in faffb1c5: a row without the exact host asserts coder-vscode.* beats *.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/** Reject variables and glob syntax that could broaden workspace collection. */
+function isSafeAbsolutePath(
+	value: unknown,
+	style: RemoteServerDataPath["style"],
+): value is string {
+	return (
+		typeof value === "string" &&
+		path[style].isAbsolute(value) &&
```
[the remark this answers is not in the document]
EhabY: Fixed in faffb1c5. isSafeAbsolutePath now rejects .. segments, matching the folder name guard, with a test row.

src/commands.ts
```diff
@@ -440,8 +444,17 @@ export class Commands {
 					throw new SupportBundleUnsupportedCliError();
 				}
 
+				const workspaceFiles = env.featureSet.supportBundleWorkspaceFiles
```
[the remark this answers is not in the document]
EhabY: Fixed in faffb1c5. Resolution is raced against the progress signal via a new raceWithAbort helper, so cancel rejects promptly; the underlying call is abandoned since the proposed API takes no signal.

src/supportBundle/workspaceFiles.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+interface RemoteEditorLogOptions {
+	readonly appRoot?: string;
+	/** The active authority, only when it targets the support bundle workspace. */
+	readonly remoteAuthority?: string;
+}
+
+/** Return known remote server log globs for the target workspace. */
+export async function getRemoteEditorLogFiles(
```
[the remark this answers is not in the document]
EhabY: Renamed to getRemoteEditorLogGlobs in faffb1c5; it also now requires appRoot instead of reading vscode.env directly.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,230 @@
...
+		: undefined;
+	return configured ?? getActiveServerDataPath(remoteAuthority, logger);
+}
+
+/** Append known editor log locations using the remote operating system's path format. */
+export function toRemoteLogGlobs({
+	value,
+	style,
+}: RemoteServerDataPath): readonly string[] {
+	const remotePath = path[style];
```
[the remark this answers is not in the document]
EhabY: Fixed in 193f5ccd. toRemoteLogGlobs now emits forward-slash globs for all styles (normalizing a win32 base's backslashes), keeping style for validation only, and the comment notes the agent normalizes to forward slashes before matching. The win32 test now asserts the forward-slash output.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,230 @@
...
+
+function hostnamePatternScore(hostname: string, pattern: string): number {
+	if (hostname === pattern) {
+		return 1000;
+	}
+	if (pattern === "*") {
+		return 1;
+	}
+	const expression = pattern
+		.replace(/[.+?^${}()|[\]\\]/g, "\\$&")
```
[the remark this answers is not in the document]
EhabY: Added in 193f5ccd: a test where coder-vscode.?xample--owner--workspace.agent must not match the real host, falling back to the * entry. Under OpenSSH ?-as-wildcard semantics that pattern would win, so the test fails if anyone aligns this with computeSshProperties.

test/unit/commands.supportBundle.test.ts
```diff
@@ -0,0 +1,179 @@
...
+			expect.anything(),
+			"owner/ws",
+			expect.objectContaining({ agentName: "main" }),
+		);
+		expect(getRemoteEditorLogGlobs).toHaveBeenCalledWith(
+			expect.objectContaining({ remoteAuthority }),
+		);
+	});
+
+	it("degrades the agent to undefined for a malformed Coder authority", async () => {
```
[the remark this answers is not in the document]
EhabY: Fixed in 193f5ccd. The setup harness now returns its logger, and the malformed-authority test asserts the "Could not resolve the connected agent" warning alongside the degraded return.

src/supportBundle/remoteServerDataPath.ts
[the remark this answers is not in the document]
EhabY: Fixed

test/unit/commands.supportBundle.test.ts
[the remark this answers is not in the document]
EhabY: Update: a later cleanup pass moved this assertion into a dedicated logging describe with loose matching (expect.any(String) + expect.any(Error)) per this repo's convention of not asserting exact log messages. The warn breadcrumb is still locked. The branch was also squashed to two commits, so earlier SHAs in this thread are orphaned.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,234 @@
...
+		if (!execServer) {
+			return undefined;
+		}
+		const { env, osPlatform } = await execServer.env();
+		const value = env.VSCODE_AGENT_FOLDER;
+		if (value === undefined) {
+			return undefined;
+		}
+		const style = pathStyleForPlatform(osPlatform);
+		if (!isSafeAbsolutePath(value, style)) {
```
code-asher: I feel like a relative path should be OK, and any other characters should be treated as literal (escaped) rather than rejected.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,234 @@
...
+
+		const remotePlatforms = config.get<Record<string, string>>(
+			"remotePlatform",
+			{},
+		);
+		const style = configuredPathStyle(
+			installPath,
+			remotePlatforms[parts.sshHost],
+		);
+		if (!isSafeAbsolutePath(installPath, style)) {
```
code-asher: Similarly here

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,234 @@
...
+ */
+export async function getRemoteServerDataPath({
+	remoteAuthority,
+	serverDataFolderName,
+	logger,
+}: RemoteServerDataPathOptions): Promise<RemoteServerDataPath | undefined> {
+	const configured = serverDataFolderName
+		? getConfiguredServerDataPath(remoteAuthority, serverDataFolderName, logger)
+		: undefined;
+	return configured ?? getActiveServerDataPath(remoteAuthority, logger);
```
code-asher: I see VS Code also has a fallback to `home + serverDataFolderName || .vscode-remote`; we should add that as well right?

**Edit**: Ah wait I think we do that elsewhere, but I feel like it would make sense to have it here.

src/supportBundle/workspaceFiles.ts
```diff
@@ -0,0 +1,86 @@
...
+	getRemoteServerDataPath,
+	hasUnsafePathChars,
+	toRemoteLogGlobs,
+} from "./remoteServerDataPath";
+
+import type { Logger } from "../logging/logger";
+
+interface ProductConfiguration {
+	/**
+	 * Default remote server data directory beneath the user's home directory.
```
code-asher: Probably just a me issue lol but I was reading this as `serverDataFolderName` being `home + some path` but I think it is just the `some path` part right?  Maybe we say something like `the data directory relative to the user's home directory` (but again this is probably just me).

src/supportBundle/workspaceFiles.ts
```diff
@@ -0,0 +1,86 @@
...
+			})
+		: undefined;
+	if (serverDataPath) {
+		return toRemoteLogGlobs(serverDataPath);
+	}
+	// The agent expands `~/` against the remote home directory and normalizes
+	// separators before glob matching, so the posix style is portable here.
+	if (serverDataFolderName) {
+		return toRemoteLogGlobs({
+			value: `~/${serverDataFolderName}`,
```
code-asher: It should default to `.vscode-remote` if not included it looks like?

src/supportBundle/workspaceFiles.ts
```diff
@@ -0,0 +1,86 @@
...
+			? product.serverDataFolderName
+			: undefined;
+	} catch (error) {
+		logger.warn("Could not read the editor's product metadata", error);
+		return undefined;
+	}
+}
+
+/** Return whether the value is a single portable path segment. */
+function isSafeServerDataFolderName(value: unknown): value is string {
```
code-asher: Do we really need to worry about this?  It resolves to whatever it resolves to.  If it has glob syntax we should escape those though.

src/commands.ts
```diff
@@ -1133,15 +1154,29 @@ export class Commands {
...
+				agentName: item instanceof AgentTreeItem ? item.agent.name : undefined,
 				client: this.extensionClient,
 				workspaceId: createWorkspaceIdentifier(item.workspace),
 			};
 		}
 		if (this.workspace && this.remoteWorkspaceClient) {
+			const remoteAuthority = vscodeProposed.env.remoteAuthority;
+			// Agent resolution is best-effort; a malformed authority must not
+			// block diagnostics for the whole workspace.
+			let agentName: string | undefined;
```
code-asher: Seems chill but makes me think we should set a `this.agent` alongside `this.workspace` or something to keep track of the current agent.

code-asher (commented):
Nice!   This is a partial review and I have not tried it out yet but here is what I have so far.

src/supportBundle/remoteServerDataPath.ts
EhabY: Done

src/supportBundle/remoteServerDataPath.ts
EhabY: I dropped this check and passed through verbatim.

The agent still requires absolute or `~/` paths and expands env vars, but a bad pattern just becomes a note in the bundle manifest now

src/supportBundle/remoteServerDataPath.ts
EhabY: Moved here. It now resolves flag > env > `~/<folder || .vscode-remote>` and always returns a path.

src/supportBundle/workspaceFiles.ts
EhabY: Reworded :pray:

src/supportBundle/workspaceFiles.ts
EhabY: Yes, defaults to `.vscode-remote` now

src/supportBundle/workspaceFiles.ts
EhabY: Cut down to a string check, and escaped glob patterns

src/commands.ts
EhabY: Done!

src/supportBundle/remoteServerDataPath.ts
code-asher: We could do something like prepend `~/` to relative paths since that is where they would resolve I think, since SSH's cwd is the home directory.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,233 @@
...
+ * implementations pass `serverInstallPath` as that flag, so its
+ * interpretation outranks the environment variable.
+ * @see https://github.com/microsoft/vscode/blob/085b6a1465387d070516ba8a640ccfed66417796/src/vs/server/node/server.main.ts#L39
+ */
+export async function getRemoteServerDataPath({
+	remoteAuthority,
+	serverDataFolderName,
+	logger,
+}: RemoteServerDataPathOptions): Promise<RemoteServerDataPath> {
+	const configured =
```
code-asher: The ternaries here with undefined branches feel a bit difficult to parse to me.  I think it is something like this:

```javascript
let dataPath
if (remoteAuthority && serverDataFolderName)
  dataPath = getConfigured(...)
}
if (!dataPath && remoteAuthority) {
  dataPath = await getActive(...)
}
return dataPath || { ... }
```

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,233 @@
...
+	const active =
+		configured ??
+		(remoteAuthority
+			? await getActiveServerDataPath(remoteAuthority, logger)
+			: undefined);
+	// The agent expands `~/` against the remote home directory, matching the
+	// server's default when neither override is present.
+	return (
+		active ?? {
+			value: `~/${serverDataFolderName ?? ".vscode-remote"}`,
```
code-asher: Is a blank folder name valid?  Should this be `||`?  I think we want `||` since that is what VS Code is using.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,233 @@
...
+	const configured =
+		remoteAuthority && serverDataFolderName
+			? getConfiguredServerDataPath(
+					remoteAuthority,
+					serverDataFolderName,
+					logger,
+				)
+			: undefined;
+	const active =
+		configured ??
```
code-asher: Similarly I think this should be `||`.

**Edit**: wait nope not this, this is an object not a string, I just got lost in the ternaries lol

src/supportBundle/workspaceFiles.ts
```diff
@@ -0,0 +1,62 @@
...
+			productJson,
+		) as ProductConfiguration;
+		return typeof serverDataFolderName === "string" && serverDataFolderName
+			? serverDataFolderName
+			: undefined;
+	} catch (error) {
+		logger.warn("Could not read the editor's product metadata", error);
+		return undefined;
+	}
+}
```
code-asher: Should this go in `remoteServerDataPath.ts` and be called by `getRemoteServerDataPath`?  It feels more directly related to the server data path than to workspace files, IMO.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,233 @@
...
+		style === "win32" ? value.replaceAll("\\", "/") : value,
+	);
+	return [
+		path.posix.join(base, "data", "logs", "**", "*.log"),
+		path.posix.join(base, ".*.log"),
+		path.posix.join(base, "cli", "servers", "*", "log.txt"),
+	];
+}
+
+async function getActiveServerDataPath(
```
code-asher: I think the difference between `getActive` and `getConfigured` is non-obvious, what do you think about names like `getServerAgentFolderEnv` and `getServerInstallPathSetting` maybe?

Probably not the best names either but just looking at the names from high-level it is not clear to me why `configured` takes precedence over `active`, just based on the names it makes me feel it should be the other way around, like the "active" path should take precedence over the "configured" path.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,233 @@
...
+	getRemoteSshExtension,
+	type RemoteSshExtensionId,
+} from "../remote/sshExtension";
+import { parseRemoteAuthority } from "../util/authority";
+import { vscodeProposed } from "../vscodeProposed";
+
+import type { Logger } from "../logging/logger";
+
+interface RemoteServerDataPathOptions {
+	readonly remoteAuthority?: string;
```
code-asher, line 17: Thinking out loud...if support bundles are generated for agents not currently connected, then we might end up getting the completely wrong path right?

Could we perhaps reconstruct the authority for these?  Like from a sidebar item I think we could call `toRemoteAuthority()`.  Currently if I try to generate a support bundle from the sidebar, even for the workspace I am currently connected to, I get the wrong paths.  Works great from the command palette though!

Alternatively, we could try passing in the workspace and agent here and then construct the remote authority directly here as needed.

Also, for the support bundle command on a workspace directly (rather than an agent) we could use the first agent since I think this is what the support command does, although maybe we could check if one is connected and use that, if not then fall back to the first.

But maybe the env resolver only works when actively connected.  In that case...we could resolve the env when connecting and then store that for later use?  Just wanna avoid the case where someone generates a bundle and it goes to the wrong place depending on whether they happen to be connected or not.

src/supportBundle/remoteServerDataPath.ts
```diff
@@ -0,0 +1,233 @@
...
+		// Its installer strips this suffix before consistently re-appending it.
+		const parentPath =
+			extensionId === "anysphere.remote-ssh" &&
+			remotePath.basename(installPath) === serverDataFolderName
+				? remotePath.dirname(installPath)
+				: installPath;
+		return {
+			value: remotePath.join(parentPath, serverDataFolderName),
+			style,
+		};
```
code-asher, line 149: So just to make sure I understand, if we have `serverDataFolderName` as `foo` and `serverInstallPath` as `/home/coder/foo` then:
 
MS: end path is `/home/coder/foo/foo`
Cursor: end path is `/home/coder/foo`

Also if `serverDataFolderName` is blank, do they default to `.vscode-server` at this point?  Or nah?

Oh...just tested VS Code and it actually created a `.vscode-server-oss` directory when the folder is blank or missing.  Although shortly after I get a connection error.

Now, why would someone edit the product.json and remove the path?  No idea lol but I guess it could happen.

code-asher (commented):
I have one main concern about generating the support bundle from the sidebar but otherwise looking good!

src/supportBundle/remoteServerDataPath.ts
code-asher: I did some more testing and it looks like, at least with VS Code, it always overrides `VSCODE_AGENT_FOLDER`.  So there seems no way to set it anyway?  I tried setting it in my profile and bashrc.  So it might be we have no need to check it, seems like it only existed as a way for the remote SSH plugin to tell the server the install path rather than a way for users to configure the path.

src/supportBundle/remoteServerDataPath.ts
EhabY: Done. Relative install paths get anchored to `~/` after style detection; `~`, `$`, and `%` prefixed values pass through verbatim so the agent can still expand them.

src/supportBundle/remoteServerDataPath.ts
EhabY: Restructured. With the exec-server check gone (see the sidebar thread) it collapsed to a single if plus a fallback return.

src/supportBundle/remoteServerDataPath.ts
EhabY: Switched to `||`; a blank name now falls back to `.vscode-remote`, same as server.main.ts.

src/supportBundle/workspaceFiles.ts
EhabY: Moved. getRemoteServerDataPath now takes appRoot and reads product.json itself.

src/supportBundle/remoteServerDataPath.ts
EhabY: Moot now: `getActiveServerDataPath` is gone entirely (rationale in the sidebar thread), so only the settings-based resolver is left.

src/supportBundle/remoteServerDataPath.ts
EhabY: Dropped the env check entirely and went settings-only. I checked the installed bundles (MS, Cursor, Windsurf, Antigravity) plus Open Remote SSH's source to be sure: MS and Cursor compute the dir from `serverInstallPath` and export `VSCODE_AGENT_FOLDER` themselves at launch, which is why your profile/bashrc values were ignored. Open Remote SSH never uses the var and passes `--server-data-dir` instead. Windsurf and Antigravity have no install path setting under any prefix and hardcode `$HOME/<serverDataFolderName>`, which is exactly our fallback. So the env read could only ever repeat what local settings already say.

Sidebar and picker flows now reconstruct the authority with toRemoteAuthority, so connected and disconnected runs resolve the same paths!

src/supportBundle/remoteServerDataPath.ts
EhabY: Yes exactly, MS ends up with `/home/coder/foo/foo` and Cursor with `/home/coder/foo` (their installer strips the trailing `/<serverDataFolderName>` before re-appending).

The current sequence of checks now should cover all cases :eyes:

src/supportBundle/remoteServerDataPath.ts
code-asher: beautiful

code-asher (approved):
Lovely!

PR #1054, opened 2026-07-30
Improve credential lifecycle and log hygiene

EhabY:
> Review priority: 7/10. Independent hygiene fixes, one logical change per commit.
> Diff: production +364/-162, tests +427/-33

Small, independent hardening changes around credential handling and what ends up in logs.

- **Header command**: parse errors reference the offending line by number instead of echoing command output, and command execution no longer logs the command string or its stdout/stderr. Failures log the exit code only.
- **HTTP logging**: header redaction is case-insensitive and covers the standard credential-bearing headers (`Authorization`, `Cookie`, `Set-Cookie`, `X-Api-Key`, `Coder-Session-Token`, `Proxy-Authorization`) plus every header applied by the configured header command. Known credential fields (`access_token`, `refresh_token`, `client_secret`, `code`, `code_verifier`, `id_token`, `password`, `token`) are redacted from logged bodies on all paths, including JSON and form-encoded strings. Redaction is copy-on-write, and log serialization is bounded in depth and length.
- **Logout**: OAuth refresh and access tokens are revoked at the server (best-effort, one request pipeline for both), and logout reports incomplete credential cleanup with a retry hint instead of unconditional success. Revocation is wired to logout only; deployment switch deliberately keeps credentials for switching back.
- **Support bundle**: a modal dialog lists the collected data categories before the bundle is created. Previously the CLI's own warning was suppressed with `--yes` and nothing was shown in its place.
- **Migration cleanup**: legacy file-based `url`/`session` credentials are deleted after migration into secret storage instead of being left on disk. The migration logic now lives in `src/remote/migration.ts`.

Each change ships with unit tests. No settings or command surface changes beyond the new dialog.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -534,10 +534,10 @@ export class Commands {
 			"A support bundle may contain sensitive information. It collects:",
 			"",
 			"\u2022 Deployment and workspace diagnostics",
-			"\u2022 Coder extension logs and workspace connection logs from recent VS Code windows",
+			"\u2022 Coder extension and connection logs from recent VS Code windows",
 			"\u2022 Remote SSH extension logs",
-			"\u2022 Locally recorded telemetry entries",
-			"\u2022 A snapshot of the Coder extension settings",
+			"\u2022 Locally recorded telemetry",
+			"\u2022 Coder extension settings",
 			"",
 			"Review the bundle before sharing it.",
 		].join("\n");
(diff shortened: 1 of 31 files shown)


src/commands.ts
```diff
@@ -523,6 +528,27 @@ export class Commands {
 		});
 	}
 
+	/** Modal disclosure of what a support bundle collects; the CLI's own prompt is suppressed. */
+	private async confirmSupportBundleCollection(): Promise<boolean> {
```
mtojek: nit: I have a feeling that this could go in a separate changeset.

mtojek (approved):
LGTM

src/commands.ts
EhabY: I've kept this here but added a new changelog entry to it's clearer :pray:

mtojek: Works for me!

PR #1058, opened 2026-08-02
fix(cli): ignore SemVer build metadata in cache checks

linonetwo:
## Summary

- compare cached CLI and server versions with SemVer equality instead of raw string equality
- ignore SemVer build metadata while still requiring the same major, minor, patch, and prerelease identifiers
- use the same comparison in initial cache lookup, telemetry outcome, and post-lock/download race checks
- add regression coverage for a release CLI matching a server version that includes source-revision metadata

## Problem

Coder server builds can report a release version with source metadata, for example:

```text
server: v2.34.0+3006da5
CLI:    v2.34.0
```

These versions have the same SemVer precedence and represent the same release, but `CliManager` currently compares their raw strings in three places. The extension therefore treats a valid cached CLI as stale and downloads it again on every connection.

We encountered this while using a reproducibly built and signed CLI from the same Coder v2.34.0 release source as the server. The unnecessary download was approximately 56 MB each time and materially delayed VS Code Remote startup. This can also affect any release packaging where one artifact retains VCS build metadata and another does not.

## Behavior

`versionsMatch` validates both strings and delegates equality to `semver.eq`, whose equality semantics ignore build metadata:

- `v2.34.0` matches `v2.34.0+3006da5`
- `v2.34.0-rc.1` matches `v2.34.0-rc.1+3006da5`
- `v2.34.0` does not match `v2.34.1`
- invalid versions do not match

The existing server-version validation remains unchanged. Signature verification and download behavior are also unchanged; this only prevents a download when an executable cached CLI is already the same SemVer release.

## Tests

- `node node_modules/vitest/vitest.mjs --run test/unit/core/cliManager.test.ts test/unit/core/cliManager.telemetry.test.ts test/unit/core/cliManager.concurrent.test.ts` — 88 passed
- `pnpm exec prettier --check src/core/cliManager.ts test/unit/core/cliManager.test.ts`
- `pnpm exec eslint src/core/cliManager.ts test/unit/core/cliManager.test.ts`

### DIFF
--- a/src/core/cliManager.ts
+++ b/src/core/cliManager.ts
@@ -56,6 +56,16 @@ type SingleVerifyResult =
 	| { kind: "bypassed" }
 	| { kind: "sig_unavailable"; status: number };
 
+function versionsMatch(actual: string, expected: string): boolean {
+	const actualVersion = semver.valid(actual);
+	const expectedVersion = semver.valid(expected);
+	return (
+		actualVersion !== null &&
+		expectedVersion !== null &&
+		semver.eq(actualVersion, expectedVersion)
+	);
+}
+
 export class CliManager {
 	private readonly binaryLock: BinaryLock;
 	private readonly cliTelemetry: CliTelemetry;
@@ -166,7 +176,10 @@ export class CliManager {
 				this.checkResolvedBinary(restClient, resolved),
 			);
 
-		if (existingVersion === buildInfo.version) {
+		if (
+			existingVersion !== null &&
+			versionsMatch(existingVersion, buildInfo.version)
+		) {
 			this.output.debug("Existing binary matches server version");
 			trace.setOutcome("cache_hit");
 			return resolved.binPath;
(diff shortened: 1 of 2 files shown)


EhabY:
Thanks for digging into this.

I don't think we can ignore build metadata unconditionally though. On our dogfood deployment every build off `main` reports the last tag plus a commit SHA, so consecutive server versions differ *only* in build metadata:

v2.35.3-devel+79724ab0ba  →  v2.35.3-devel+aabbccddee

`semver.valid()` strips that and `semver.eq` ignores it, so `versionsMatch` returns true for both. The cache path is per-hostname rather than per-version, so version equality is our only staleness signal, and the extension would keep an arbitrarily old CLI until the next tag bump.

Your case looks like one-sided metadata (server has it, CLI doesn't), which is narrower. Which server build reports `v2.34.0+3006da5`? Our release builds emit the bare tag with no metadata, so I'd like to understand where it's coming from before we settle on a rule.

linonetwo:
Sorry, I was just digging too deep into "Coder over tailnet direct connect" https://github.com/coder/tailscale/pull/128#issuecomment-5169353336 and have to release my fork of vscode extension.

But now, I think it's overcomplicated. Let me close this.

EhabY:
Sure, let me know if you need help on the VS Code extension side!

Issue #1057, opened 2026-08-02
Cursor: extension mangles deployment hostname (and drops --header-command) when hostname starts with "coder."


Daniel-OD:
## Summary

When connecting to a Coder deployment from **Cursor** via the "Open in Cursor" workspace button, if the deployment's hostname's first label is literally `coder` (e.g. `coder.example.com`), the extension incorrectly substitutes that label with the editor's identifier (`cursor`) when constructing the per-editor global-config directory — and then **uses that mangled string as the real deployment URL**, not just as a local cache-directory name. This breaks DNS resolution and login entirely.

This does **not** happen in VS Code — there the config directory correctly stays `coder.example.com`. It's specific to running the extension inside Cursor (and presumably other VS Code forks with a different `env.appName`/product identifier).

Additionally, once that's worked around, a second, independent bug appears: the SSH `ProxyCommand` block the extension generates for Cursor never includes `--header-command`, even when `coder.headerCommand` is correctly set in Cursor's `settings.json` — even though the equivalent VS Code-generated block for the *same* deployment *does* include it correctly. This breaks any deployment sitting behind an auth proxy that requires custom headers (e.g. Cloudflare Access service tokens).

## Steps to reproduce

1. Self-host a Coder deployment whose hostname's first label is `coder` (e.g. `coder.example.com`), ideally behind a reverse proxy that requires a custom auth header (Cloudflare Access, etc.) so `coder.headerCommand` is needed.
2. Install `coder.coder-remote` in Cursor (in addition to / instead of VS Code).
3. Set `coder.headerCommand` in Cursor's `settings.json` to a script that outputs the required header(s).
4. From the Coder dashboard, click "Open in Cursor" for a workspace.

## Observed behavior

- The "Coder" output channel logs:
  ```
  [info] Handling Coder URI {"route":"open","safeHostname":"coder.example.com",...}
  [warning] Failed to handle URI {"path":"/open",...,"error":"Cannot read properties of undefined (reading 'some')"}
  [warning] Failed to store token at login: No CLI binary found at .../coder.coder-remote/cursor.example.com/bin/coder-darwin-arm64
  ```
  Note the CLI binary path uses `cursor.example.com`, not `coder.example.com`.

[...]

PR #1059, opened 2026-08-03
fix: enforce session auth hostname invariant

EhabY:
## Summary

- enforce that session auth URLs match their secret-storage hostname on reads and writes
- treat malformed or mismatched stored auth as a missing session with sanitized warnings
- catch invalid file-based auth migration at its recovery boundary and continue to login

Addresses #1057.

## Validation

- `pnpm test` — 2,264 passed, 1 skipped
- `pnpm typecheck`
- `NODE_OPTIONS=--max-old-space-size=4096 pnpm lint`
- `pnpm format:check`

<details>
<summary>Implementation plan</summary>

1. Validate the hostname invariant centrally in `SecretsManager`.
2. Let writes throw; degrade invalid reads to a missing session.
3. Catch migration failures at the file-migration recovery boundary.
4. Cover read/write enforcement, OAuth exact-URL validation, and migration fallback.

</details>

This pull request was generated by Coder Agents.

### DIFF
--- a/src/core/secretsManager.ts
+++ b/src/core/secretsManager.ts
@@ -63,15 +63,17 @@ export class SecretsManager {
 	}
 
 	private assertSessionAuthHostname(safeHostname: string, url: string): void {
-		let authHostname: string | undefined;
+		let authHostname: string;
 		try {
 			authHostname = toSafeHost(url);
 		} catch {
-			authHostname = undefined;
+			throw new Error(
+				`Session auth hostname mismatch: expected "${safeHostname}", got an invalid URL`,
+			);
 		}
 		if (authHostname !== safeHostname) {
 			throw new Error(
-				"Session auth URL does not match its deployment hostname",
+				`Session auth hostname mismatch: expected "${safeHostname}", got "${authHostname}"`,
 			);
 		}
 	}
@@ -218,6 +220,12 @@ export class SecretsManager {
 		return result.data;
 	}
 
+	/**
+	 * Store session auth for a deployment.
+	 *
+	 * @throws If the auth URL is invalid or its hostname does not match the
+	 * deployment.
+	 */
 	public async setSessionAuth(
 		safeHostname: string,
 		auth: SessionAuth,
--- a/src/core/secretsManager.ts
+++ b/src/core/secretsManager.ts
@@ -68,7 +68,7 @@ export class SecretsManager {
 			authHostname = toSafeHost(url);
 		} catch {
 			throw new Error(
-				`Session auth hostname mismatch: expected "${safeHostname}", got an invalid URL`,
+				`Session auth hostname mismatch: expected "${safeHostname}", got an invalid URL "${url}"`,
 			);
 		}
 		if (authHostname !== safeHostname) {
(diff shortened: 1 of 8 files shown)


src/core/secretsManager.ts
```diff
@@ -62,6 +62,22 @@ export class SecretsManager {
 		return `${prefix}${safeHostname}`;
 	}
 
+	private assertSessionAuthHostname(safeHostname: string, url: string): void {
+		let authHostname: string;
+		try {
+			authHostname = toSafeHost(url);
+		} catch {
+			throw new Error(
+				`Session auth hostname mismatch: expected "${safeHostname}", got an invalid URL`,
```
code-asher: We should include `url` in the error here I think so callers can more easily identify why the URL is invalid

src/remote/remote.ts
```diff
@@ -823,10 +823,16 @@ export class Remote {
...
-				url: url.value.trim(),
-				token: token.value.trim(),
-			});
+			try {
+				await this.secretsManager.setSessionAuth(safeHostname, {
+					url: url.value.trim(),
+					token: token.value.trim(),
+				});
+			} catch {
+				this.logger.warn("Failed to migrate session auth from files", {
```
code-asher: This should also log the error imo otherwise it is a bit cryptic.  I see there is a test that explicitly tests that it was sanitized but the error is kinda useless without the details.

src/core/secretsManager.ts
```diff
@@ -189,15 +205,34 @@ export class SecretsManager {
...
 		const result = SessionAuthSchema.safeParse(data);
-		return result.success ? result.data : undefined;
+		if (!result.success) {
+			return undefined;
+		}
+		try {
+			this.assertSessionAuthHostname(safeHostname, result.data.url);
+		} catch {
+			this.logger.warn(
+				"Ignoring session auth with invalid deployment hostname",
```
code-asher: This should log the error too imo rather than overriding it with a less detailed message

code-asher (approved):
Looks good but how does this fix #1057?  Supposedly we are replacing `coder.example.com` with `cursor.example.com` but I see nothing in this diff that should affect that, just new assertions.

Unless it is related to the file migration somehow?  But it seems like that would imply we somehow mangled the domain to begin with so the root bug must still be around somewhere.

EhabY:
The theory here is that this was because of a failed migration but this does not really fix it because there is nothing in our codebase that could do this, I have a suspicion that this is related to global settings not being applied to remote connection which is already fixed on main.

Also this would catch the users error much earlier and show more info, so hopefully we can understand how that happened.

src/core/secretsManager.ts
EhabY: True, added `url` here!

src/remote/remote.ts
EhabY: Now we log the error as well, this should be safe and not include sensitive details

src/core/secretsManager.ts
EhabY: Done

PR #1060, opened 2026-08-03
fix: read SSH settings from the active extension's section

EhabY:
## Problem

`getSshConfigPath()` read `remote.SSH.configFile` unconditionally. Two of the five Remote-SSH implementations we support forked `open-remote-ssh` and renamed the entire settings section, so that key does not exist for them:

| Extension | Section it reads |
|---|---|
| `ms-vscode-remote.remote-ssh` | `remote.SSH` |
| `anysphere.remote-ssh` (Cursor) | `remote.SSH` |
| `jeanp413.open-remote-ssh` | `remote.SSH` |
| `codeium.windsurf-remote-openssh` (Windsurf/Devin) | `remote.devinSSH`, legacy `remote.windsurfSSH` |
| `google.antigravity-remote-openssh` | `remote.antigravitySSH` |

Verified by reading the installed bundles: Antigravity resolves `getConfiguration("remote.antigravitySSH").get("configFile")` and contributes only `remote.antigravitySSH.*`; Windsurf uses a dual-read helper over `remote.devinSSH` and `remote.windsurfSSH`, preferring the former. Neither references `remote.SSH` anywhere.

So on Antigravity or Windsurf with a custom SSH config file configured, we wrote the workspace host block into `~/.ssh/config` while the extension launched `ssh -F <their file>` — host not found, connection fails.

## Change

`SETTING_SECTIONS` maps each known extension to the sections it consults, and `getRemoteSshSetting()` returns the first non-empty value across them. Unknown/absent extension falls back to `remote.SSH`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@
 
 ### Fixed
 
+- Write the SSH config to the file the active Remote-SSH extension actually
+  reads. Windsurf/Devin and Antigravity renamed the whole `remote.SSH` settings
+  section, so a custom config file set as `remote.devinSSH.configFile`,
+  `remote.windsurfSSH.configFile`, or `remote.antigravitySSH.configFile` was
+  ignored and the workspace host was written to `~/.ssh/config` instead, where
+  those editors never looked for it.
 - Apply a 60-second default timeout to REST requests, so requests hung on a
   half-open TCP connection don't stall pollers forever.
 - Change `coder.binarySource`, `coder.binaryDestination`, `coder.headerCommand`,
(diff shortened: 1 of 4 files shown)


matifali:
@EhabY should we also start respecting ting `devin://` in our Frontend code on coder/coder and https://registry.coder.com/modules/coder/windsurf

It is not part of this, but just came to mind while reading it.

EhabY:
Oh yes we probably should, the only difference now is that the app name has been changed from Windsurf to Devin, so `windsurf` -> `devin` everywhere

code-asher (approved):
Makes sense!

PR #1061, opened 2026-08-03
fix: write workspace SSH hosts to a shared glob include that ssh reads first

### LINKED ISSUE #1055: Cannot connect when ~/.ssh/config has a catch-all "Host *" ProxyCommand declared before the Coder block
## Summary

If `~/.ssh/config` contains a catch-all `Host *` block with a `ProxyCommand` **declared before** the extension's own block, the extension refuses to connect with a modal `Unexpected SSH Config Option` and there is no way to proceed. This is the normal state of affairs for anyone whose SSH config is generated by a config manager — [`assh`](<https://github.com/moul/assh>) is the case I hit, and it is not fixable from the assh side.

## Reproduce

1. Have `~/.ssh/config` contain, in this order:

   ```
   Host *
     ProxyCommand /opt/homebrew/bin/assh connect --port=%p %h
   
   # --- START CODER VSCODE coder.example.com ---
   Host coder-vscode.coder.example.com--*
     ProxyCommand "…/coder-darwin-arm64" … ssh --stdio … %h
   # --- END CODER VSCODE coder.example.com ---
   ```
2. Open a workspace in VS Code Desktop.
3. Modal appears:

   > Your SSH config is overriding the "ProxyCommand" property to "/opt/homebrew/bin/assh connect --port=%p %h" when it expected "…coder-darwin-arm64 … ssh --stdio … %h" for the "[coder-vscode.coder.example.com--user--workspace.dev](<http://coder-vscode.coder.example.com--user--workspace.dev>)" host. Please fix this and try again!

   The only button is **Reload Window**, after which `closeRemote()` runs and the connection is aborted.

Confirmed with OpenSSH that the extension's reading is correct:

```console
$ /usr/bin/ssh -G coder-vscode.coder.example.com--user--workspace.dev | grep -i proxycommand
proxycommand /opt/homebrew/bin/assh connect --port=%p %h
```

## Root cause

* `sshConfig.ts` `appendBlock()` puts the Coder block at the **end** of the file when no existing marker block is found.
* OpenSSH uses the **first** value obtained for each keyword, so an earlier `Host *` block wins.
* `remote.ts` (`updateSSHConfig`) then compares expected against effective for `ProxyCommand`, `UserKnownHostsFile` and `StrictHostKeyChecking`, and on any mismatch shows the modal, calls `closeRemote()` and throws.

So the extension writes its block into a position where it is guaranteed to lose, then treats losing as fatal.

## Why users of config managers cannot fix this

With `assh` specifically, the global `ProxyCommand` is not user configuration that can be removed — it is how assh works, and it is emitted unconditionally for the defaults block ([`pkg/config/host.go`](<https://github.com/moul/assh/blob/master/pkg/config/host.go>)):

```go
if h.isDefault {
    if h.noAutomaticRewrite {
        fmt.Fprintf(w, "  ProxyCommand %s connect --no-rewrite --port=%%p %%h\n", asshBinaryPath)
    } else {
        fmt.Fprintf(w, "  ProxyCommand %s connect --port=%%p %%h\n", asshBinaryPath)
    }
}
```

Adding a per-host entry for `coder-vscode.*` to `assh.yml` does not help either, because for non-default hosts assh writes `ProxyCommand` as a **comment** and resolves it internally at connect time:

```go
} else {
    if h.ProxyCommand != "" {
        fmt.Fprintf(w, "  # ProxyCommand %s\n", h.ProxyCommand)
    }
}
```

assh also regenerates `~/.ssh/config` from `assh.yml`, so manually reordering the blocks does not survive.

## Current workaround

Setting `remote.SSH.configFile` to a dedicated file works, since `getSshConfigPath()` honours it:

```ts
const configured = vscode.workspace.getConfiguration().get<string>("remote.SSH.configFile");
return expandPath(configured || path.join("~", ".ssh", "config"));
```

The cost is that Remote-SSH then only lists hosts from that file, so every other SSH host disappears from the picker. This workaround is not mentioned in the error message or, as far as I can find, in the docs.

## Suggested fixes

Roughly in order of preference:

1. **Write the Coder block above any catch-all block** rather than appending to the end, so the extension's values win by SSH's own first-match-wins rule. This overlaps with the placement idea in coder/vscode-coder#695.
2. **Set the values on the command line instead of relying on config precedence** — Remote-SSH's `ProxyCommand` can be supplied per-host in a file the extension fully controls, or options can be passed as `-o` flags, which beat config file entries.
3. **Make the check non-fatal**, or add an escape hatch. There is currently no setting to bypass it; I enumerated every entry in `contributes.configuration`.
4. **At minimum, name the workaround in the error message.** The modal says "Please fix this and try again!" without saying that the conflicting block may be machine-generated, or that `remote.SSH.configFile` exists.

## Environment

|  |  |
| -- | -- |
| Extension | `coder.coder-remote` v1.15.2 |
| Coder | v2.34.7 (AGPL) |
| OpenSSH | OpenSSH_10.2p1, LibreSSL 3.3.6 |
| OS | macOS (darwin arm64) |
| SSH config manager | assh |

Related: coder/vscode-coder#695.

EhabY:
Fixes SSH config precedence without letting VS Code-based editors overwrite one another's generated Coder hosts.

### Shared SSH config directory

Each editor writes one generated file per deployment into a directory shared by all editors under the platform data dir (`$XDG_DATA_HOME`, `~/Library/Application Support`, `%APPDATA%`):

```text
~/.local/share/coder.coder-remote/ssh/vscode--dev.coder.com.conf
~/.local/share/coder.coder-remote/ssh/cursor--dev.coder.com.conf
```

The user's SSH config gets a single editor-agnostic block, placed first because OpenSSH uses the first value it finds:

```ssh-config
# --- START CODER ---
# Moves back to the top on connect; override options via coder.sshConfig.
Include "~/.local/share/coder.coder-remote/ssh/*.conf"
# --- END CODER ---
```

On connect, the extension:

1. fully regenerates the current (editor, deployment) file, so its mtime marks the last connect;
2. adds or moves the shared Include block to the top of the effective user config;
3. removes the released direct block for that deployment from the effective user config;

Correctness without locking: each generated file has a single writer, and concurrent writers of the same file produce equivalent content, so last-writer-wins is correct. The Include block is byte-identical from every editor, so once placed the user's config stops changing and concurrent editors converge instead of racing. Host patterns are namespaced per editor and deployment, so glob ordering between the files is irrelevant.

Other deployments' direct blocks and the deployment-unaware historical block are preserved; hosts matching the historical `coder-vscode--*` pattern parse as foreign, so their windows keep connecting through the preserved block instead of failing setup.

Since first-line placement now guarantees the managed options apply, the post-write check that recomputed them and aborted the connection with "Unexpected SSH Config Option" on a mismatch (typically from a catch-all `Host *`) is removed. Validation of deployment-sent options (the deny list) is unchanged.

Generated files carry a do-not-edit header, since they are rewritten on every connection. A new **Coder: Open Generated SSH Configuration File** command opens them read-only: a connected window opens its own deployment's file, a local window offers a deployment picker.

Include paths inside `$HOME` are emitted as `~/...`, so quirks in the home path (spaces, `%`, glob characters) never reach the argument ssh reads. Paths outside `$HOME` are quoted with Windows separators normalized and glob characters escaped; CR, LF, NUL, `"`, and `%` are rejected because ssh cannot read them back (OpenSSH 9.9+ percent-expands Include arguments and exits fatally on unknown tokens).

### Which file gains the include

Microsoft's and Cursor's Remote-SSH pass `remote.SSH.configFile` to ssh with `-F`, and VSCodium's fork parses that file in-process, so the setting is honored there. Antigravity and Windsurf/Devin renamed the setting but spawn ssh without `-F`, so ssh always reads `~/.ssh/config`; on those two the setting (or a stale synced `remote.SSH.configFile`) is ignored rather than writing the include where the connection never looks.

### Editor-specific authorities

The SSH host prefix uses `vscode.env.uriScheme` unchanged — `coder-vscode`, `coder-cursor`, `coder-windsurf`, `coder-vscode-insiders` — and the generated wildcard host and CLI `--ssh-host-prefix` use the same prefix.

### Legacy authority migration

Only the actual historical `coder-vscode` authority is migrated in other editors; foreign `coder-*` authorities are ignored rather than retargeted.

> [!IMPORTANT]
> **Migration reopens the window automatically.** Connecting to a legacy authority in another editor retargets it and reopens the same folder or saved workspace once, preserving the full URI (scheme, path, query, and fragment). Empty windows reopen with the retargeted authority.

An untitled multi-root workspace cannot be reopened without dropping folders, so it keeps connecting through the old host and a modal explains how to finish the switch (save the workspace, then reload). Recent-folder entries with legacy authorities stay compatible when reopening a workspace. Plain, URI-wrapped, and nested `ssh-remote+` authorities retain their surrounding wrappers when retargeted.

### Concurrent config changes

All writes are atomic temp-file renames with retries for transient Windows errors, following the CLI `config-ssh` shape: read, merge, skip when unchanged, write. No locks or optimistic retries; every editor writes an identical Include block, so concurrent updates converge, and a lost update stays shadowed by the include (first value wins) until the next connect rewrites it.

## Compatibility

- New integration tests run the real `ssh -G` against generated configs: a quoted glob include in a path with spaces wins over a later `Host *`, escaped glob characters in the directory name resolve, two deployments and two editors are served through one glob include, and deleting the generated directory degrades gracefully instead of breaking ssh.
- Verified against the `jeanp413/open-remote-ssh` source (VSCodium's Remote-SSH): it parses the config in-process via `ssh-config`, follows top-level quoted Includes (`untildify` + glob), and implements first-value-wins, on both current and years-old releases.
- Verified against `PowerShell/openssh-portable` source and release history: every in-box Windows ssh.exe resolves Include arguments through `glob(3)` (wildcard or not) since the first bundled 7.7p1, 8.1p1+ (Windows 10 2004+) also honors quoted forward-slash absolute paths, and a missing include directory is a non-fatal "matched no files".

## Diff size

Relative to `main` (`15d4c8f2`):

| Area | Added | Removed | Net |
| --- | ---: | ---: | ---: |
| Production (`src/`) | 507 | 295 | +212 |
| Tests (`test/`) | 1,163 | 1,146 | +17 |
| Docs and manifest | 75 | 13 | +62 |

## Validation

- `pnpm test`: 2,374 passed, 1 skipped (includes the real-OpenSSH tests, skipped where ssh is absent)
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format:check`
- `pnpm build`
- `git diff --check`

<details>
<summary>Implementation plan and decisions</summary>

- Use `vscode.env.uriScheme` unchanged and require a non-empty editor ID where it is consumed.
- Generate one config per (editor, deployment) in a shared platform data dir; the filename carries the editor and deployment, and the user's config includes the directory with a glob.
- Use `START|END CODER <hostname>` markers in generated files and one parameterless `START|END CODER` include block in the user's config. Legacy `CODER VSCODE <hostname>` markers are still matched for user-config cleanup.
- Emit home-relative include paths with a tilde; reject characters ssh cannot read back rather than half-escape them.
- Migrate only `coder-vscode`, preserving wrappers and complete URI/workspace state; parse deployment-unaware `coder-vscode--*` hosts as foreign.
- Keep an unsaved untitled multi-root workspace on the old host with an explanatory modal rather than dropping folders or failing setup.
- Preserve other deployments' direct blocks for lazy migration; sweep generated files unused for a week (connects always rewrite, so mtime marks the last connect).
- Ignore the renamed config-file setting on editors whose ssh never reads it; keep honoring `remote.SSH.configFile` where ssh (or an in-process parser) does.
- Keep the Node `typeof`-based filesystem interface; tests mock `node:fs/promises` with memfs instead of hand-rolled fixtures.
- Open the generated file read-only from the new command, since the extension rewrites it on every connection.
- Match the CLI's read-merge-write on the user's config instead of a lock or optimistic retries; generated files are single-writer last-write-wins.

</details>

Generated by Coder Agents for @EhabY.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,6 +41,13 @@ Host coder-vscode.dev.coder.com--*
 	LogLevel ERROR
 ```
 
+Which file that entry goes in depends on the Remote - SSH extension. Microsoft's
+and Cursor's pass `remote.SSH.configFile` to ssh with `-F`, and VSCodium's parses
+the file itself instead of running ssh, so all three connect through it.
+Antigravity and Windsurf/Devin renamed the setting but spawn ssh without `-F`, so
+ssh reads `~/.ssh/config` regardless; we ignore it there rather than write the
+host where the connection never looks.
+
 If any step fails, we show an error message. Once the error message is closed
 we close the remote so the Remote - SSH connection does not continue to
 connection. Otherwise, we yield, which lets the Remote - SSH continue.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,39 +14,57 @@ The `ssh-remote` scheme is registered by Microsoft's Remote - SSH extension and
 indicates that it should connect to the provided host name using SSH.
 
 The host name takes the format
-`coder-vscode.<domain>--<username>--<workspace>`. This is parsed by the CLI
-(which is invoked via SSH's `ProxyCommand`) to route SSH to the right workspace.
+`coder-<editor>.<domain>--<username>--<workspace>`, where `<editor>` comes from
+that product's URI scheme, such as `vscode`, `cursor`, or `windsurf`. The CLI is
+invoked through SSH's `ProxyCommand` with this prefix so it can route SSH to the
+right workspace. A legacy `coder-vscode` authority opened in another editor is
+reopened once with that editor's prefix; legacy recent-folder entries remain
+compatible when opening the same workspace.
 
 The Coder Remote extension also registers for the
 `onResolveRemoteAuthority:ssh-remote` [extension activation
 event](https://code.visualstudio.com/api/references/activation-events) to hook
 into this process, running before the Remote - SSH extension actually connects.
 
-On activation of this event, we check if `vscode.workspace.workspaceFolders`
-contains the `coder-vscode` prefix, and if so we delay activation to:
+On activation of this event, we check whether the remote authority belongs to
+the current editor, and if so we delay activation to:
 
 1. Parse the host name to get the domain, username, and workspace.
 2. Ensure the workspace is running.
 3. Download the matching server binary to the client.
 4. Configure the binary with the URL and token, asking the user for them if they
    are missing. Each domain gets its own config directory.
-5. Add an entry to the user's SSH config for `coder-vscode.<domain>--*`.
+5. Write an entry for `coder-<editor>.<domain>--*` to `ssh-config` in the
+   extension's global storage directory.
+6. Add a per-editor `Include` block at the top of the user's SSH config.
 
 ```text
-Host coder-vscode.dev.coder.com--*
-	ProxyCommand "/tmp/coder" --global-config "/home/kyle/.config/Code/User/globalStorage/coder.coder-remote/dev.coder.com" ssh --stdio --network-info-dir "/home/kyle/.config/Code/User/globalStorage/coder.coder-remote/net" --ssh-host-prefix coder-vscode.dev.coder.com-- %h
-	ConnectTimeout 0
-	StrictHostKeyChecking no
-	UserKnownHostsFile /dev/null
-	LogLevel ERROR
+# --- START CODER cursor ---
+Include "~/.config/Cursor/User/globalStorage/coder.coder-remote/ssh-config"
+# --- END CODER cursor ---
+
+# --- START CODER vscode ---
+Include "~/.config/Code/User/globalStorage/coder.coder-remote/ssh-config"
+# --- END CODER vscode ---
 ```
 
-Which file that entry goes in depends on the Remote - SSH extension. Microsoft's
-and Cursor's pass `remote.SSH.configFile` to ssh with `-F`, and VSCodium's parses
-the file itself instead of running ssh, so all three connect through it.
-Antigravity and Windsurf/Devin renamed the setting but spawn ssh without `-F`, so
-ssh reads `~/.ssh/config` regardless; we ignore it there rather than write the
-host where the connection never looks.
+Each included file contains only that editor's host entries:
+
+```text
+Host coder-cursor.dev.coder.com--*
+  ProxyCommand "/tmp/coder" --global-config "/home/kyle/.config/Cursor/User/globalStorage/coder.coder-remote/dev.coder.com" ssh --stdio --network-info-dir "/home/kyle/.config/Cursor/User/globalStorage/coder.coder-remote/net" --ssh-host-prefix coder-cursor.dev.coder.com-- %h
+  ConnectTimeout 0
+  StrictHostKeyChecking no
+  UserKnownHostsFile /dev/null
+  LogLevel ERROR
+```
+
+Which main file gains the include depends on the Remote - SSH extension.
+Microsoft's and Cursor's pass `remote.SSH.configFile` to ssh with `-F`, and
+VSCodium's parses the file itself instead of running ssh, so all three connect
+through it. Antigravity and Windsurf/Devin renamed the setting but spawn ssh
+without `-F`, so ssh reads `~/.ssh/config` regardless; we ignore the renamed
+setting there rather than add the include where the connection never looks.
 
 If any step fails, we show an error message. Once the error message is closed
 we close the remote so the Remote - SSH connection does not continue to
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,21 +34,25 @@ the current editor, and if so we delay activation to:
 3. Download the matching server binary to the client.
 4. Configure the binary with the URL and token, asking the user for them if they
    are missing. Each domain gets its own config directory.
-5. Write an entry for `coder-<editor>.<domain>--*` to `ssh-config` in the
-   extension's global storage directory.
-6. Add a per-editor `Include` block at the top of the user's SSH config.
+5. Write an entry for `coder-<editor>.<domain>--*` to a per-editor,
+   per-deployment file in a data directory shared by every editor, such as
+   `~/.local/share/coder.coder-remote/ssh/cursor--dev.coder.com.conf`.
+6. Keep a shared `Include` block at the top of the user's SSH config that
+   globs the whole directory. Every editor writes the identical block, so
+   concurrent writers converge on the same content. The `CODER INCLUDE <id>`
+   marker convention lets other Coder integrations recognize the block, since
+   Coder-managed includes route disjoint hosts and are order-independent.
 
 ```text
-# --- START CODER cursor ---
-Include "~/.config/Cursor/User/globalStorage/coder.coder-remote/ssh-config"
-# --- END CODER cursor ---
-
-# --- START CODER vscode ---
-Include "~/.config/Code/User/globalStorage/coder.coder-remote/ssh-config"
-# --- END CODER vscode ---
+# --- START CODER INCLUDE CODER-REMOTE ---
+# Managed by each editor's Coder extension (coder.coder-remote).
+# Moves back to the top on connect; override options via coder.sshConfig.
+Include "~/.local/share/coder.coder-remote/ssh/*.conf"
+# --- END CODER INCLUDE CODER-REMOTE ---
 ```
 
-Each included file contains only that editor's host entries:
+Each generated file contains only its own editor's host entries for one
+deployment:
 
 ```text
 Host coder-cursor.dev.coder.com--*
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ the current editor, and if so we delay activation to:
 
 ```text
 # --- START CODER INCLUDE CODER-REMOTE ---
-# Managed by each editor's Coder extension (coder.coder-remote).
+# Managed by the Coder extension for VS Code and its forks.
 # Moves back to the top on connect; override options via coder.sshConfig.
 Include "~/.local/share/coder.coder-remote/ssh/*.conf"
 # --- END CODER INCLUDE CODER-REMOTE ---
--- a/package.json
+++ b/package.json
@@ -478,12 +478,20 @@
 			},
 			{
 				"command": "coder.viewLogs",
-				"title": "Coder: View Logs",
+				"title": "View Logs",
+				"category": "Coder",
 				"icon": "$(list-unordered)"
 			},
 			{
+				"command": "coder.openSshConfig",
+				"title": "Open Generated SSH Configuration File",
+				"category": "Coder",
+				"icon": "$(file-code)"
+			},
+			{
 				"command": "coder.exportTelemetry",
-				"title": "Coder: Export Telemetry",
+				"title": "Export Telemetry",
+				"category": "Coder",
 				"icon": "$(save)"
 			},
 			{
@@ -614,6 +622,10 @@
 				},
 				{
 					"command": "coder.viewLogs",
+					"when": "true"
+				},
+				{
+					"command": "coder.openSshConfig",
 					"when": "true"
 				},
 				{
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -62,7 +62,6 @@ import { applySettingOverrides, buildSshOverrides } from "./sshOverrides";
 import { SshProcessMonitor } from "./sshProcess";
 import {
 	computeSshProperties,
-	findSshPropertyProblems,
 	sshSupportsSetEnv,
 	type SshProperties,
 } from "./sshSupport";
@@ -83,6 +82,13 @@ import type { PathResolver } from "../core/pathResolver";
 import type { SecretsManager } from "../core/secretsManager";
 import type { Logger } from "../logging/logger";
 import type { LoginCoordinator } from "../login/loginCoordinator";
+
+/**
+ * Our own config, included from the user's. Keep the tilde: relative includes
+ * always resolve against ~/.ssh, and an absolute path would not survive a
+ * config synced between machines.
+ */
+const CODER_SSH_CONFIG_PATH = "~/.ssh/coder/config";
 
 export interface RemoteDetails extends vscode.Disposable {
 	safeHostname: string;
@@ -897,19 +903,23 @@ export class Remote {
 		featureSet: FeatureSet,
 		cliAuth: CliAuth,
 	): Promise<SshProperties> {
-		const sshConfigFile = this.getSshConfigPath();
-
-		const sshConfig = new SshConfig(sshConfigFile, this.logger);
+		// Our blocks live in our own file; the user's only gains the include.
+		const sshConfig = new SshConfig(this.getSshConfigPath(), this.logger);
 		await sshConfig.load();
+		const coderConfig = new SshConfig(
+			expandPath(CODER_SSH_CONFIG_PATH),
+			this.logger,
+		);
+		await coderConfig.load();
 
 		// Options the user set themselves win the merge below, so they are exempt
 		// from the deny list. Both sources are local and already trusted: whoever
-		// can write the SSH config could write any Host block directly, and the
-		// post-write check below still pins the critical options.
+		// can write the SSH config could write any Host block directly.
 		const userConfigSsh = vscode.workspace
 			.getConfiguration("coder")
 			.get<string[]>("sshConfig", []);
 		const userConfig = parseSshConfig(userConfigSsh);
+		// The CLI writes its block to the user's config, so read it from there.
 		const configSshOptions = parseCoderSshOptions(sshConfig.getRaw());
 
 		let deploymentSshConfig = {};
@@ -973,51 +983,15 @@ export class Remote {
 			sshValues.SetEnv = "CODER_SSH_SESSION_TYPE=vscode";
 		}
 
-		await sshConfig.update(safeHostname, sshValues, sshConfigOverrides);
+		// Write our file before including it, so the include never dangles.
+		await coderConfig.update(safeHostname, sshValues, sshConfigOverrides);
+		await sshConfig.updateInclude(CODER_SSH_CONFIG_PATH, safeHostname);
 
-		// A user can provide a "Host *" entry in their SSH config to add options
-		// to all hosts. We need to ensure that the options we set are not
-		// overridden by the user's config.
-		const computedProperties = computeSshProperties(
+		// Mirror SSH's parse order; RemoteCommand can come from the user's config.
+		return computeSshProperties(
 			hostName,
-			sshConfig.getRaw(),
+			`${coderConfig.getRaw()}\n${sshConfig.getRaw()}`,
 		);
-		const problems = findSshPropertyProblems(computedProperties, {
-			ProxyCommand: sshValues.ProxyCommand,
-			UserKnownHostsFile: sshValues.UserKnownHostsFile,
-			StrictHostKeyChecking: sshValues.StrictHostKeyChecking,
-		});
-		if (problems.length > 0) {
-			await this.failSshConfigCheck(hostName, problems);
-		}
-
-		return computedProperties;
-	}
-
-	/** Show every unexpected option at once, close the remote, and abort. */
-	private async failSshConfigCheck(
-		hostName: string,
-		problems: string[],
-	): Promise<never> {
-		const title = `Unexpected SSH Config Option${problems.length > 1 ? "s" : ""}`;
-		const detail =
-			`Your SSH config sets unexpected values for the "${hostName}" host. ` +
-			`Please fix the following and try again:\n\n` +
-			problems.map((problem) => `- ${problem}.`).join("\n");
-		const result = await vscodeProposed.window.showErrorMessage(
-			title,
-			{
-				useCustom: true,
-				modal: true,
-				detail,
-			},
-			"Reload Window",
-		);
-		if (result === "Reload Window") {
-			await this.reloadWindow();
-		}
-		await this.closeRemote();
-		throw new Error("SSH config mismatch, closing remote");
 	}
 
 	private watchSettings(
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -40,10 +40,9 @@ import { getHeaderCommand } from "../settings/headers";
 import { escapeCommandArg, expandPath } from "../util";
 import {
 	type AuthorityParts,
-	classifyRemoteAuthority,
+	classifySshHost,
 	parseRemoteAuthority,
 	retargetRemoteAuthority,
-	toCurrentAuthorityHostPrefix,
 } from "../util/authority";
 import { createStatusBarItem } from "../util/statusBar";
 import { vscodeProposed } from "../vscodeProposed";
@@ -54,7 +53,6 @@ import { migrateAuthToSecretsStorage } from "./migration";
 import {
 	SshConfig,
 	type SshValues,
-	cleanupStaleSshConfigs,
 	mergeSshConfigValues,
 	parseCoderSshOptions,
 	parseSshConfig,
@@ -164,9 +162,11 @@ export class Remote {
 
 		// parseRemoteAuthority returned null for foreign hosts, so this is
 		// either the current editor's authority or a migratable legacy one.
-		if (classifyRemoteAuthority(parts) === "legacy") {
-			await this.migrateLegacyAuthority(remoteAuthority, startupMode);
-			return;
+		if (classifySshHost(parts.sshHost) === "legacy") {
+			if (await this.migrateLegacyAuthority(remoteAuthority, startupMode)) {
+				return;
+			}
+			// Not reopened: keep going so the legacy host still connects.
 		}
 
 		this.logger.info("Setting up remote connection", {
@@ -676,8 +676,7 @@ export class Remote {
 			this.logger.info("Updating SSH config...");
 			return await this.updateSSHConfig(
 				workspaceClient,
-				context.parts.safeHostname,
-				context.parts.sshHost,
+				context.parts,
 				binaryPath,
 				logDir,
 				featureSet,
@@ -727,10 +726,15 @@ export class Remote {
 		return undefined;
 	}
 
+	/**
+	 * Reopen the window on this editor's own authority. Returns false when the
+	 * workspace cannot be reopened losslessly, so the caller connects over the
+	 * legacy host instead.
+	 */
 	private async migrateLegacyAuthority(
 		remoteAuthority: string,
 		startupMode: StartupMode,
-	): Promise<void> {
+	): Promise<boolean> {
 		const migratedAuthority = retargetRemoteAuthority(remoteAuthority);
 		const workspaceFile = vscode.workspace.workspaceFile;
 		const workspaceFolders = vscode.workspace.workspaceFolders ?? [];
@@ -738,28 +742,26 @@ export class Remote {
 			workspaceFile?.scheme === "untitled" ? undefined : workspaceFile;
 		if (!savedWorkspaceFile && workspaceFolders.length > 1) {
 			this.logger.warn(
-				"Cannot migrate an unsaved multi-root workspace",
+				"Cannot migrate an unsaved multi-root workspace; connecting over the legacy host",
 				remoteAuthority,
 			);
-			const choice = await vscodeProposed.window.showWarningMessage(
-				"Opening the remote over the old coder-vscode SSH host",
-				{
-					modal: true,
-					useCustom: true,
-					detail:
-						"This editor now uses its own SSH hosts, but switching an unsaved multi-root workspace would drop its folders. " +
-						"To switch, save the workspace, then reload the window.",
-				},
-				"Learn More",
-			);
-			if (choice === "Learn More") {
-				await vscode.env.openExternal(
-					vscode.Uri.parse(
-						"https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces",
-					),
-				);
-			}
-			return;
+			// Fire-and-forget: the connection proceeds either way.
+			void vscode.window
+				.showWarningMessage(
+					"This workspace still opens over the old coder-vscode SSH host. " +
+						"To switch it to this editor's own host, save the workspace, then reload the window.",
+					"Learn More",
+				)
+				.then(async (choice) => {
+					if (choice === "Learn More") {
+						await vscode.env.openExternal(
+							vscode.Uri.parse(
+								"https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces",
+							),
+						);
+					}
+				});
+			return false;
 		}
 
 		await this.serviceContainer
@@ -779,12 +781,13 @@ export class Remote {
 				}),
 				false,
 			);
-			return;
+			return true;
 		}
 		await vscode.commands.executeCommand("vscode.newWindow", {
 			remoteAuthority: migratedAuthority,
 			reuseWindow: true,
 		});
+		return true;
 	}
 
 	private async resolveRemoteBinary(workspaceClient: Api): Promise<string> {
@@ -959,13 +962,14 @@ export class Remote {
 	// all Coder entries.
 	private async updateSSHConfig(
 		restClient: Api,
-		safeHostname: string,
-		hostName: string,
+		parts: AuthorityParts,
 		binaryPath: string,
 		logDir: string,
 		featureSet: FeatureSet,
 		cliAuth: CliAuth,
 	): Promise<SshProperties> {
+		// Taken from the authority, so an unmigrated legacy host keeps working.
+		const { hostPrefix, safeHostname } = parts;
 		// One file per (editor, deployment); the user's config gains one shared include.
 		const sshConfig = new SshConfig(this.getMainSshConfigPath(), this.logger);
 		await sshConfig.load();
@@ -1017,8 +1021,6 @@ export class Remote {
 			userConfig,
 		);
 
-		const hostPrefix = toCurrentAuthorityHostPrefix(safeHostname);
-
 		const proxyCommand = await this.buildProxyCommand(
 			binaryPath,
 			safeHostname,
@@ -1046,15 +1048,14 @@ export class Remote {
 
 		// Write our file before including it, so the include never dangles.
 		await coderConfig.update(sshValues, sshConfigOverrides);
-		const sharedSshConfigDir = this.pathResolver.getSshConfigDir();
-		await sshConfig.updateInclude(sharedSshConfigDir, safeHostname);
-		// Our file was just written, so only other unused deployments are swept.
-		// Never throws, and the connection does not depend on it.
-		void cleanupStaleSshConfigs(sharedSshConfigDir, this.logger);
+		await sshConfig.updateInclude(
+			this.pathResolver.getSshConfigDir(),
+			safeHostname,
+		);
 
 		// Mirror SSH's parse order; RemoteCommand can come from the user's config.
 		return computeSshProperties(
-			hostName,
+			parts.sshHost,
 			`${coderConfig.getRaw()}\n${sshConfig.getRaw()}`,
 		);
 	}
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -952,7 +952,7 @@ export class Remote {
 		return ["--log-dir", escapeCommandArg(logDir), "-v"];
 	}
 
-	private getSshConfigPath(): string {
+	private getMainSshConfigPath(): string {
 		const configured = getRemoteSshConfigFile();
 		return expandPath(configured || path.join("~", ".ssh", "config"));
 	}
@@ -969,9 +969,9 @@ export class Remote {
 		cliAuth: CliAuth,
 	): Promise<SshProperties> {
 		// Our blocks live in our own file; the user's only gains the include.
-		const sshConfig = new SshConfig(this.getSshConfigPath(), this.logger);
+		const sshConfig = new SshConfig(this.getMainSshConfigPath(), this.logger);
 		await sshConfig.load();
-		const coderConfigPath = this.pathResolver.getSshConfigPath();
+		const coderConfigPath = this.pathResolver.getIncludedSshConfigPath();
 		const coderConfig = new SshConfig(coderConfigPath, this.logger);
 		await coderConfig.load();
 
--- a/src/remote/remote.ts
+++ b/src/remote/remote.ts
@@ -57,7 +57,7 @@ import {
 	parseSshConfig,
 	validateDeploymentSshOptions,
 } from "./sshConfig";
-import { getRemoteSshSetting } from "./sshExtension";
+import { getRemoteSshConfigFile } from "./sshExtension";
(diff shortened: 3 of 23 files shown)


src/remote/sshConfig.ts
```diff
@@ -93,6 +100,44 @@ export function parseSshConfig(lines: string[]): Record<string, string> {
...
+			managed = false;
+		} else if (!managed && BLOCK_START_REGEX.test(line)) {
+			index = lineIndex;
+			break;
+		}
+	}
+	if (index === -1) {
+		return -1;
+	}
+	while (index > 0 && isUserComment(lines[index - 1])) {
```
code-asher: I think we need to account for blank lines as well.

Also comments could be indentend, would be a kinda weird thing to do in general but I could see it for commented-out settings.

```
# the second best host
Host *
  Port 2222
  # ConnectTimeout 0

# the best host
Host *
```

src/remote/sshConfig.ts
```diff
@@ -53,6 +53,13 @@ const defaultFileSystem: FileSystem = {
 // Matches an SSH config key at the start of a line (e.g. "ConnectTimeout", "LogLevel").
 const sshKeyRegex = /^[a-zA-Z0-9-]+/;
 
+// Matches a line that opens a Host or Match block. The trailing separator
+// keeps it from matching keys that merely start the same way, like HostName.
+const BLOCK_START_REGEX = /^[ \t]*(Host|Match)[ \t=]/i;
```
code-asher: I think it would be better to go the include path, because for example I have my own `Include` at the top of my config files, and that has `Host` entries in it.  

Without prepending to the top we would not be able to override those included hosts.

Also, then we would have no need to parse `Host` or `Match` or comments, etc.

src/remote/sshConfig.ts
```diff
@@ -223,16 +272,47 @@ export class SshConfig {
...
 			this.logger.debug("Replacing SSH config block", safeHostname);
 			this.replaceBlock(block, newBlock);
 		} else {
-			this.logger.debug("Appending new SSH config block", safeHostname);
-			this.appendBlock(newBlock);
+			this.logger.debug("Inserting SSH config block", safeHostname);
+			if (block) {
+				this.removeBlock(block);
+			}
+			this.insertBlock(newBlock);
```
code-asher: Not a big deal at all but for simplicity we could just always remove and insert.

src/remote/sshConfig.ts
code-asher: Also this would fix the problem of global settings overriding our settings I think?

src/remote/sshConfig.ts
EhabY: Not relevant anymore

src/remote/sshConfig.ts
EhabY: Indeed, I went with the `Include` approach since it seems superior in all ways. Only limitation is OpenSSH has to be >= 7.3 which is 9-10 years old

src/remote/sshConfig.ts
EhabY: This is no longer an issue since we use `Include` (always at the top)

src/remote/remote.ts
```diff
@@ -77,6 +77,13 @@ import type { SecretsManager } from "../core/secretsManager";
 import type { Logger } from "../logging/logger";
 import type { LoginCoordinator } from "../login/loginCoordinator";
 
+/**
+ * Our own config, included from the user's. Keep the tilde: relative includes
+ * always resolve against ~/.ssh, and an absolute path would not survive a
+ * config synced between machines.
+ */
+const CODER_SSH_CONFIG_PATH = "~/.ssh/coder/config";
```
code-asher: Thoughts on using the same dir as the setting?  I like to move my config from `.ssh` to `~/.config/ssh` and I prevent anything from writing to my root home dir so for my setup this can fail.

So something like `path.dirname(getSshConfigPath())` (where supported).

Or actually if we put it in the extension's global storage dir or something I think that would be chill.  Less chance of collision that way too, if the user already has `~/.ssh/coder/config` for some reason.

src/remote/sshConfig.ts
```diff
@@ -233,6 +245,41 @@ export class SshConfig {
...
+	async updateInclude(includePath: string, safeHostname: string) {
+		const original = this.getRaw();
+		const block = this.getBlock(safeHostname);
+		if (block) {
+			this.logger.debug("Removing superseded SSH config block", safeHostname);
+			this.removeBlock(block);
+		}
+		const include = [
+			INCLUDE_START,
+			"# Your Coder workspaces, managed by the Coder VS Code extension. Keep first:",
```
code-asher: The way I read this is that I am being asked to keep it first, which implies to me I could move it, but actually if I move it we just move it back.

Maybe we say something like "this block will always be moved to the top of the config before connecting; to override settings use coder.sshConfig instead"

Or we could respect whatever the current placement is, if the user has decided to move it.  I think either option is fine.  I do personally prefer to override via the ssh config over VS Code options, but it is not a huge deal.

src/remote/sshExtension.ts
```diff
@@ -11,31 +11,25 @@ export const REMOTE_SSH_EXTENSION_IDS = [
 export type RemoteSshExtensionId = (typeof REMOTE_SSH_EXTENSION_IDS)[number];
 
 /**
- * Sections each extension reads, in order. The rebranded forks renamed the
- * whole `remote.SSH` section, so reading it directly misses them.
+ * Extensions that spawn ssh without `-F`, so it reads ~/.ssh/config whatever
+ * their renamed setting says. Honoring one would write the workspace host
```
code-asher: woah crazy they have a setting and then just not use it lol

code-asher (commented):
Nice, I like how minimal the edits are to my main config now.  One comment on the path for our config.

src/remote/remote.ts
EhabY: That was my first thought but then did some research (see the PR description):

> The tilde is deliberate. Relative includes resolve against ~/.ssh wherever the including file lives (ssh_config(5): "Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration file", verified from a config outside ~/.ssh), so deriving the location from a custom configFile would need an absolute path.

I'm not see sure about absolute paths tbh

src/remote/sshExtension.ts
EhabY: They did this because they forked the open source Remote SSH extension but did not implement everything when they switched from JS handling of SSH to OS-native :sob:

src/remote/sshConfig.ts
EhabY: Good catch, the wording implied a choice we do not honour. It always moves back to the top, so it now says so and points at the setting that does let you override us:

```
# --- START CODER VSCODE INCLUDE ---
# Your Coder workspaces, managed by the Coder VS Code extension.
# This block moves back to the top on every connect, since SSH uses the first
# value it finds. To override these options, use the coder.sshConfig setting.
Include ~/.ssh/coder/config
# --- END CODER VSCODE INCLUDE ---
```

I went with moving it rather than respecting placement, since honouring a moved include brings back the whole class of bug this PR is fixing (an `Include` or `Host *` above ours silently wins, and `ProxyCommand` losing is fatal). `coder.sshConfig` lands inside our included block, so it still beats anything in your own config.

EhabY: Reworded this :pray:

src/remote/remote.ts
EhabY: Turns out following your setting is more workable than I said: `dirname` of the unexpanded value keeps the tilde, so it doesn't have to be absolute but I'd still rather not. It makes the location a moving target, and since this PR ignores that setting on Antigravity/Windsurf, one machine could end up with our config in two places depending on which editor connected last. Sharing a single file between editors is fine either way, since our block markers and `Host` patterns are identical and they already overwrite each other today.

Your collision point was right, though. If `~/.ssh/coder` already exists as a file, say a key named `coder`, `mkdir` throws ENOTDIR and the connect just dies. I'll use `~/.ssh/coder.config` instead so there's no directory involved.

That leaves your setup working as long as `~/.ssh` exists, which it always should anyway.

EhabY: | | fixed `~/.ssh/coder.config` | `dirname(configFile)` |
|---|---|---|
| `~/.ssh` missing or write-blocked | fails | works, the one real win |
| Relocated config respected | no | yes |
| After `configFile` changes | old include still points at the live file | frozen copy left behind, stale ProxyCommands |
| `${userHome}` / `${env:}` in the setting | n/a | forces an absolute include, silently |
| Across editors | same path everywhere | differs, since 2 of 5 ignore the setting |
| Path handling | one constant string | separators and quoting per platform, fails silently |

Summary ^, honestly not too pressed on either :thinking:

code-asher: `~/` is absolute though, it does not resolve relatively.  Just a shorthand for `$HOME/` or `/home/$USER` typically.

> one machine could end up with our config in two places depending on which editor connected last

Why would this be a problem?  The plugin settings are already duplicated for each editor, and SSH config is just more settings.  Not to mention everything else that is duplicated like the binary, logs, etc etc

> I'll use ~/.ssh/coder.config instead so there's no directory involved.

I mean, it might be less prone to collisions, but it still fails to solve the collision problem.  Why not just put it in an extension-controlled directory?  Manipulating the user's config dir will never be reliable.

> That leaves your setup working as long as ~/.ssh exists, which it always should anyway

Not in my case, I `chattr +i ~` so nothing can create `.ssh`.

code-asher: > Why would this be a problem?

Thinking about it more, it could be weird because the main config would keep overwriting the include block and point to a different config every time each editor connected.  I guess not an actual bug or anything but it could look janky.  So that means instead of using global storage we could use a shared path like `$XDG_DATA_HOME/vscode-coder/config` or something like that.

code-asher: > Relocated config respected

We write every time we connect so I think this would not be an issue.  

> it could be weird because the main config would keep overwriting the include block and point to a different config every time each editor connected

I just realized we already have a maybe bigger problem with the proxy command since the paths would change depending on what editor last ran (the binary, the config, even the log paths).  If we have to worry about this, we might actually need per-editor configs anyway to make sure the log paths do not go to a different editor (say you connect with VS Code, then with Windsurf, then VS Code loses connection and reconnects, it would now be logging into Windsurf's paths and using the binary Windsurf downloaded, which might not even be the same version depending on the settings, and they might have different `coder.sshConfig` values).

`START CODER VSCODE` and `START CODER WINDSURF` etc with unique host name prefixes.

src/remote/sshConfig.ts
code-asher: > honouring a moved include brings back the whole class of bug this PR is fixing

Well before it was something that happened without user interaction, this time the user would have to explicitly choose to move it.  It gives the user more control, but yeah of course more control means they can shoot themselves in the foot lol

Personally I am not too bothered either way since I only use VS Code for testing ahaha, most people are probably fine editing `coder.sshConfig` I would imagine.

src/remote/remote.ts
EhabY: I've applied what we discussed in the DMs yesterday, do check out the PR description for the full details but I like this solution much more and it seems to work well (migration has a window reload but that's automatic and only once per legacy URL)

src/remote/remote.ts
```diff
@@ -910,15 +982,18 @@ export class Remote {
 			}
 		}
 
-		const sshConfigFile = this.getSshConfigPath();
-
-		const sshConfig = new SshConfig(sshConfigFile, this.logger);
+		// Our blocks live in our own file; the user's only gains the include.
+		const sshConfig = new SshConfig(this.getSshConfigPath(), this.logger);
 		await sshConfig.load();
+		const coderConfigPath = this.pathResolver.getSshConfigPath();
```
code-asher: With these functions named the same it might be a bit confusing, what if we did something like `getMainSShConfigPath()` and `getIncludedCoderSshConfigPath()` or something like that?

src/remote/sshConfig.ts
```diff
@@ -213,138 +248,191 @@ export class SshConfig {
...
 			this.logger.debug("Replacing SSH config block", safeHostname);
-			this.replaceBlock(block, newBlock);
+			merged = this.replaceRange(raw, existing, block);
 		} else {
 			this.logger.debug("Appending new SSH config block", safeHostname);
-			this.appendBlock(newBlock);
+			merged = raw ? `${raw.trimEnd()}\n\n${block}` : block;
 		}
-		await this.save();
+		return merged.startsWith(CODER_SSH_CONFIG_HEADER)
```
code-asher: I had a silly thought, if the user edits this file (despite the warning lol) and prepends something, we would add a duplicate comment.

Maybe we should check to see if the comment is elsewhere in the config and remove it so we can put it back to the top?

Or alternatively we read in all the blocks, and then regenerate the entire file so anything the user added outside the blocks gets wiped.

**Edit**: may not be an issue if we end up changing the file structure.

src/remote/sshConfig.ts
```diff
@@ -213,138 +248,191 @@ export class SshConfig {
...
+		const endCount = countSubstring(markers.end, raw);
+		if (startCount !== endCount) {
 			throw new SshConfigBadFormat(
-				`Malformed config: ${this.filePath} has an unterminated START CODER VSCODE ${safeHostname} block. Each START block must have an END block.`,
+				`Malformed config: ${this.filePath} has an unterminated "${markers.start}" block. Each START block must have an END block.`,
 			);
 		}
-
-		if (startBlockCount > 1 || endBlockCount > 1) {
+		if (startCount > 1) {
```
code-asher, line 375: Why the removal of `endCount > 1`?  We could have one start marker and two end markers.

I suppose we could assume the first end marker is the correct one, but ultimately the config is malformed and there is no way to be sure.

src/remote/sshConfig.ts
```diff
@@ -213,138 +248,191 @@ export class SshConfig {
...
-
-		lines.push(this.endBlockComment(safeHostname));
-		return {
-			raw: lines.join("\n"),
-		};
+	): string {
+		let rest = raw;
+		const editorBlock = this.findBlock(rest, blockMarkers(editorId));
+		if (editorBlock) {
+			rest = this.removeRange(rest, editorBlock);
```
code-asher: It occurs to me that if you use multiple editors they will keep moving their blocks on top of each other.

Not an actual issue or anything, but it could be noisy if the user has their SSH config source controlled or automatic backups or something like that and the file keeps changing unnecessarily.

Maybe not worth solving (up to you), but we could extract all the include blocks, order them and re-insert so they are stable.  Or leave them in place if they already exist (I know we talked about not doing that so the user has less control but idk if it would really be an issue).

**Edit**: may not be an issue if we end up changing the file structure.

src/remote/sshConfig.ts
```diff
@@ -213,138 +248,191 @@ export class SshConfig {
...
+
+	private mergeDeployment(
+		raw: string,
+		safeHostname: string,
+		block: string,
+	): string {
+		let merged: string;
+		const existing =
+			this.findBlock(raw, blockMarkers(safeHostname)) ??
+			this.findBlock(raw, legacyDeploymentMarkers(safeHostname));
```
code-asher: `mergeDeployment` always applies to our controlled per-editor config I think, right?

So I think there should never be any legacy markers in it since it is a new thing.

src/remote/sshConfig.ts
```diff
@@ -213,138 +248,191 @@ export class SshConfig {
...
+		const before = raw.slice(0, range.start).trimEnd();
+		const after = raw.slice(range.end).trimStart();
+		return [before, after].filter(Boolean).join("\n\n");
 	}
 
-	private appendBlock(block: Block) {
-		const raw = this.getRaw();
+	private async mutate(mutation: Mutation): Promise<void> {
+		let snapshot = this.getRaw();
+		for (let attempt = 0; attempt < UPDATE_ATTEMPTS; attempt++) {
```
code-asher: If there is some failure that is retryable, we should be explicit about that and check the error (permissions error for example will probably not go away on the next attempt).

This is in case multiple editors are writing to the file concurrently right?  Not sure what the code for that is, `EBUSY` or something?

src/remote/sshConfig.ts
```diff
@@ -213,138 +248,191 @@ export class SshConfig {
...
+			}
+			snapshot = latest;
 		}
-	}
 
-	private withIndentation(text: string) {
-		return `  ${text}`;
+		this.raw = snapshot;
+		throw new Error(
+			`Failed to update SSH config at ${this.filePath} because it kept changing. Please try again.`,
```
code-asher: Maybe we add something like "kept changing, likely by another IDE" or something, otherwise the error is a bit vague.

src/remote/sshConfig.ts
```diff
@@ -379,12 +467,26 @@ export class SshConfig {
...
+				if (latest !== expectedRaw) {
+					await this.fileSystem.unlink(tempPath).catch((unlinkErr: unknown) => {
+						this.logger.warn(
+							"Failed to clean up conflicted temp SSH config file",
+							tempPath,
+							unlinkErr,
+						);
+					});
+					return false;
+				}
```
code-asher: I see the notes about the race being acceptable but if we are anticipating this to be a problem enough to try to address it then I feel like we should solve it fully.

Which I think would mean a lock in `$XDG_DATA_HOME`.

Alternatively...we could put the configs in `$XDG_DATA_HOME`, then use a wildcard include for the main config, that way even if they step on each other's toes writing the main config the end result would always be correct.

Multiple instances of one editor on different deployments could still step on each other's toes for their own configs though.  We could do a config per deployment.  A setup where we have something like these files:

```
~/.local/share/coder.coder-remote/ssh/vscode--dev.coder.com.conf
~/.local/share/coder.coder-remote/ssh/vscode--cdrstable.dev.conf
~/.local/share/coder.coder-remote/ssh/cursor--dev.coder.com.conf
```

And this include:

```
Include ~/.local/share/coder.coder-remote/ssh/*.conf
```

src/remote/sshConfig.ts
```diff
@@ -402,11 +504,14 @@ export class SshConfig {
 		}
 	}
 
-	public getRaw() {
-		if (this.raw === undefined) {
-			throw new Error("SshConfig is not loaded. Try sshConfig.load()");
+	private async readForConflict(): Promise<string> {
```
code-asher: This should perhaps just be `read` since it is not actually doing anything about conflicts (just happens to be used that way).

src/util/authority.ts
```diff
@@ -8,44 +12,89 @@ export interface AuthorityParts {
...
+	}
+	if (
+		currentPrefix !== LegacyAuthorityPrefix &&
+		sshHost.startsWith(`${LegacyAuthorityPrefix}.`)
+	) {
+		return "legacy";
+	}
+	// Deployment-unaware hosts like coder-vscode--ws stay foreign; their
+	// preserved config block still routes them.
+	return sshHost.startsWith("coder-") ? "foreign" : undefined;
```
code-asher: Nit but in the code there is no difference between how `foreign` and `undefined` are handled so we could just return one or the other, which is what we are doing in practice anyway with the `?? "foreign"` in `classifyRemoteAuthority()`.

Or if we anticipate this distinction to be useful in the future we should remove the `??` and let the undefined bubble up.

code-asher (commented):
I think this is the right direction!  The main concern I have is if we should use a different file structure to account for concurrency, or just not worry about concurrency at all.

Also I have not been able to test it yet because every time I build I get a timeout for one of the dependencies.  Not sure why `pnpm` insists on downloading every time I build...

src/remote/sshConfig.ts
EhabY: While that is def cleaner, I think the fact that people use multiple VS Code editors is very unlikely and so we can even live with the race fully (what are the chances you try to open two editors at the same exact time AND this is the first time that we add it?).

One improvement we can do, is that we do not move the include to the top if the top has other "vscode" includes (since those are mutually exclusive).

My issue with `XDG_DATA_HOME` is that we have to handle each platform, and we have to ensure it's not deleted or cleaned up by mistake. While using the global storage seems more like a natural place IMO

EhabY: Actually the more that I think about it the more I like your idea since it means we have a single file for each (editor, deployment), though in all fairness we can duplicate this in the `globalStorage` as well.

I'll double check if we can find this path easily on all platforms and make sure globs work on all platforms equally, if so then yeah maybe that's the way to go.

The only downside is file cleanup, maybe we always write the full file, and then we run a cleanup on editor startup we clean up those files if they haven't been touched in a week or something. Potentially we can let ANY editor remove these files so even for other editors to fix the staleness :thinking:

src/remote/remote.ts
EhabY: Done

src/remote/sshConfig.ts
EhabY: This is moot in the new approach

src/remote/sshConfig.ts
EhabY: Restored with more meaningful messages

src/remote/sshConfig.ts
EhabY: Solved with the file structure change

src/remote/sshConfig.ts
EhabY: We currently only retry if the file content changes, any other error (like `EACCES`) fails loudly

src/remote/sshConfig.ts
EhabY: Moved, will test on Linux and Windows 🙏

src/remote/sshConfig.ts
EhabY: No longer an issue now

src/remote/sshConfig.ts
EhabY: Done

src/remote/sshConfig.ts
EhabY: Indeed, renamed

src/util/authority.ts
EhabY: Collapsed it and simplified :pray:

src/remote/sshConfig.ts
EhabY: Update: removed the retries entirely and matched `coder config-ssh` on this file: read, merge, skip if unchanged, atomic temp+rename. Every editor writes the same include bytes now, so there's nothing worth retrying.

src/remote/sshConfig.ts
EhabY: Update: this error is gone along with the retry loop (see the other thread).

src/remote/sshConfig.ts
EhabY: I've tested this on Windows and Linux and it works perfectly!

I do actually like this solution, please take another look :eyes:

src/remote/remote.ts
```diff
@@ -718,6 +727,66 @@ export class Remote {
...
+				remoteAuthority,
+			);
+			const choice = await vscodeProposed.window.showWarningMessage(
+				"Opening the remote over the old coder-vscode SSH host",
+				{
+					modal: true,
+					useCustom: true,
+					detail:
+						"This editor now uses its own SSH hosts, but switching an unsaved multi-root workspace would drop its folders. " +
+						"To switch, save the workspace, then reload the window.",
```
code-asher: I think we have to still do remote setup in this case rather than aborting.

Because currently in this case we skip the remote setup, which may mean the connection will fail (for example if the SSH config needs to be regenerated), and now there will be no way to save the workspace file since it requires being connected.

src/remote/remote.ts
```diff
@@ -973,51 +1044,19 @@ export class Remote {
...
 		}
 
-		await sshConfig.update(safeHostname, sshValues, sshConfigOverrides);
+		// Write our file before including it, so the include never dangles.
+		await coderConfig.update(sshValues, sshConfigOverrides);
+		const sharedSshConfigDir = this.pathResolver.getSshConfigDir();
+		await sshConfig.updateInclude(sharedSshConfigDir, safeHostname);
+		// Our file was just written, so only other unused deployments are swept.
+		// Never throws, and the connection does not depend on it.
+		void cleanupStaleSshConfigs(sharedSshConfigDir, this.logger);
```
code-asher: IMO we should not do this; IDE sessions can be long lived (leaving editors open for weeks, the only interruption being putting the machine to sleep).

If we need to clean up, we could have the plugin periodically touch the file or something like that.  But I feel like the number of deployment/IDE combinations will be so small it is probably not worth the effort.

We were not previously cleaning them up either, maybe we can revisit if we end up needing this?

src/remote/sshConfig.ts
```diff
@@ -62,6 +66,45 @@ const KEY_ONLY_REGEX = /^[a-zA-Z0-9-]+$/;
...
+		start: `# --- START CODER VSCODE ${safeHostname} ---`,
+		end: `# --- END CODER VSCODE ${safeHostname} ---`,
+	};
+}
+
+/** Shared include block; identical bytes from every editor, so writers converge. */
+const INCLUDE_MARKERS: BlockMarkers = {
+	start: "# --- START CODER ---",
+	end: "# --- END CODER ---",
+};
```
code-asher, line 89: I think we need a more unique prefix because JetBrains may end up doing something similar and having its own wildcard includes.

Maybe we can still use `VSCODE`?  Kinda unfortunate though because it implies just the main VS Code IDE.

`CODER CODE-OSS` maybe?  To indicate anything based off the Code OSS?  `CODER VSCODE FAMILY` perhaps?  `CODER.CODER-REMOTE`?

Also...if we do this with JetBrains (and maybe one day the cli) they will fight over who gets to be at the top which seems unfortunate.  Still wondering if we should leave it in place if it already exists in the file.

Or...not sure how sketchy this is but we could share a directory with all the plugins and the cli.  Put it in `$data/coderv2/ssh` for example.  Or sub-directories for each plugin, and include with `*/*.conf`.

src/remote/sshConfig.ts
```diff
@@ -514,11 +508,24 @@ export class SshConfig {
 		}
 	}
 
-	public getRaw() {
-		if (this.raw === undefined) {
-			throw new Error("SshConfig is not loaded. Try sshConfig.load()");
-		}
+	private async discardTemp(tempPath: string): Promise<void> {
+		await this.fileSystem.unlink(tempPath).catch((unlinkErr: unknown) => {
```
code-asher: nbd but with `await` I feel like `try`/`catch` would be more consistent with the rest of the codebase

src/remote/sshExtension.ts
```diff
@@ -11,31 +11,24 @@ export const REMOTE_SSH_EXTENSION_IDS = [
...
-	// Windsurf became Devin and reads both, preferring the new name.
-	"codeium.windsurf-remote-openssh": ["remote.devinSSH", "remote.windsurfSSH"],
-	"anysphere.remote-ssh": ["remote.SSH"],
-	"ms-vscode-remote.remote-ssh": ["remote.SSH"],
-	"google.antigravity-remote-openssh": ["remote.antigravitySSH"],
-};
+const IGNORED_CONFIG_FILE: readonly RemoteSshExtensionId[] = [
+	"google.antigravity-remote-openssh",
+	"codeium.windsurf-remote-openssh",
+];
```
code-asher, line 20: Just musing out loud, but I had a thought that if these plugins change to add support for that setting, then suddenly the Coder extension will stop working and we will have no idea until a bug report comes in.

But trying to test it automatically would be tricky so that might just be the way it has to be.

src/util/authority.ts
```diff
@@ -73,15 +127,51 @@ export function parseRemoteAuthority(authority: string): AuthorityParts | null {
 	};
 }
 
+export function classifyRemoteAuthority(
+	parts: AuthorityParts,
+): AuthorityClassification {
+	return classifySshHost(parts.sshHost);
+}
```
code-asher: very much a nit so please feel free to ignore, but I feel like the caller could just call `classifySshHost` itself without the function wrapper.

src/util/authority.ts
```diff
@@ -8,44 +12,94 @@ export interface AuthorityParts {
...
+function currentAuthorityPrefix(): string {
+	return `coder-${currentEditorId()}`;
+}
+
+function getSshHostStart(authority: string): number | undefined {
+	if (authority.startsWith(sshRemotePrefix)) {
+		return sshRemotePrefix.length;
+	}
+
+	for (const wrapper of [`@${sshRemotePrefix}`, `://${sshRemotePrefix}`]) {
```
code-asher, line 45: Oo nice I like the idea, seems better compared to splitting on `+`.

Just a thought but it could maybe be more stable to pass in the host by parsing the URL via `new URL(authority).host` or `URI.parse` or something so we avoid having to sort of jankily handle `@` and `://`.

CONTRIBUTING.md
```diff
@@ -14,33 +14,58 @@ The `ssh-remote` scheme is registered by Microsoft's Remote - SSH extension and
...
-	StrictHostKeyChecking no
-	UserKnownHostsFile /dev/null
-	LogLevel ERROR
+# --- START CODER cursor ---
+Include "~/.config/Cursor/User/globalStorage/coder.coder-remote/ssh-config"
+# --- END CODER cursor ---
+
+# --- START CODER vscode ---
+Include "~/.config/Code/User/globalStorage/coder.coder-remote/ssh-config"
+# --- END CODER vscode ---
```
code-asher: This is out of date now

package.json
```diff
@@ -478,12 +478,20 @@
 			},
 			{
 				"command": "coder.viewLogs",
-				"title": "Coder: View Logs",
+				"title": "View Logs",
+				"category": "Coder",
```
code-asher, line 482: ooo neat did not know about category

src/remote/remote.ts
EhabY: Oooh good catch, I made it continue using the legacy authority which means it might fight with the VS Code one (though unlikely and this is better than breaking). This is also self-healing so the next run without an untitled workspace should migrate it!

src/remote/remote.ts
EhabY: Fair enough, I've removed this. Though tbh having something that deletes every month for example should be safe. Like on the next reconnect it's recreated, so the only issue is if we do not reconnect BUT the SSH itself reconnects.

src/remote/sshConfig.ts
EhabY: I've went with `CODER.CODER-REMOTE` since this is the extension ID so it seems like a natural fit + I added a comment explaining this.

Regarding the include fight, how about we just ignore `CODER` include blocks at the top? Of course, every IDE has to follow through but that's always gonna be the case. In fact, I've just changed the block to `START CODER INCLUDE CODER-REMOTE`, Jetbrains can do `START CODER INCLUDE JETBRAINS`, etc...
I have not implemented the ignore logic here though since we do not have another client yet. wdyt?

src/remote/sshConfig.ts
EhabY: Changed!

src/remote/sshExtension.ts
EhabY: Yeahhh, if they suddenly fix it we'd have to make a small change here :/

src/util/authority.ts
EhabY: Done!

CONTRIBUTING.md
EhabY: Good catch, updated!

package.json
EhabY: Yepp, it adds the `Coder:` prefix ONLY for the command palette

src/util/authority.ts
EhabY: I gave it a shot, but two of the three shapes we receive are not URLs: the bare `ssh-remote+host` form has no scheme, and in nested remotes like `dev-container+abc@ssh-remote+host` the `@` is VS Code authority nesting, not userinfo. `new URL (and URI.parse)` throws on those, so we'd keep the same string handling anyway.

Where it does parse, `new URL` also lowercases the host, which would mangle case-sensitive usernames and workspace names (those currently are all lowercase but yeah).

src/remote/remote.ts
code-asher: When VS Code reconnects (whether after waking from sleep or whatever) our plugin does not run setup again though right?  People could leave their IDEs running indefinitely so it would only be a matter of time before it broke (unless we do the touch thing ofc).

src/remote/sshConfig.ts
code-asher: It does seem unfortunate it can only handle Coder blocks, like if some other tool put things at the top or something, but that is probably good enough.  People can always give VS Code a separate config if it becomes a problem for them anyway!

src/util/authority.ts
code-asher: Oh I thought the remote authority always had `vscode://` for the scheme?  But yeah if it does come bare in some cases I suppose we could handle that, whether as a catch or a pre-emptive check, or always prepending the scheme.

Nested authorities should parse just fine though for the purposes of getting the SSH host?

```
> new URL("vscode://attached-container+def@dev-container+abc@ssh-remote+host").host
ssh-remote+host
```

Not saying we have to do this though, was just a thought.  The only way the current method breaks anyway is if somehow `ssh-remote` shows up twice but I think that is impossible (unless one day it become possible to nest remotes like `vscode://ssh-remote+abc@ssh-remote+host` or something but that seems crazy).

src/remote/sshConfig.ts
code-asher: Also, this is something we can easily change later if we do end up needing to tweak something!

code-asher: I double-checked the Toolbox plugin ID just to be sure we will not overwrite, it is `coder-toolbox` so I think we should be OK.

code-asher: OK I had one more thought lol

If someone is just looking at the config, will it be obvious that `coder-remote` refers to VS Code or similar editor?  The wording makes it sound just as likely to be JetBrains or anything else (people are probably not familiar with the extension IDs). 

Am I overthinking it lol

code-asher (approved):
Nicely done!

src/remote/remote.ts
EhabY: Yep, reconnecting does not re-run the setup. Though worst case is that the window reloads but I've removed the clean up here anyway

src/remote/sshConfig.ts
EhabY: Hahahah that's actually fair, I changed the comment to say this now:

```
# --- START CODER INCLUDE CODER-REMOTE ---
# Managed by the Coder extension for VS Code and its forks.
# Moves back to the top on connect; override options via coder.sshConfig.
Include "~/.local/share/coder.coder-remote/ssh/*.conf"
# --- END CODER INCLUDE CODER-REMOTE ---
```

src/util/authority.ts
EhabY: Bare is the only shape we get (`env.remoteAuthority` and `Uri.authority` are scheme-less), and retargeting needs the offset, not the host, so URL means prepend a scheme, parse, then find the index anyway.

I've added more documentation for this :pray:

PR #1063, opened 2026-08-03
fix: ignore the SSH config file setting where ssh never reads it

EhabY:
Follow-up to #1060, which shipped an incorrect premise.

## Problem

#1060 assumed every Remote-SSH extension connects through the config file its `configFile` setting names. Two of them do not.

Antigravity and Windsurf/Devin spawn ssh as `-v -T [-o RemoteCommand=none] -D <port> <dest> bash -s` — **no `-F`** anywhere in either bundle (the only `-F` token is a `grep -F` in their remote install script). Their `getSSHConfigPath()` feeds the host tree, the file watcher and the "Open SSH Configuration File" command, and nothing else. ssh therefore reads its own default `~/.ssh/config` no matter what `remote.antigravitySSH.configFile` or `remote.devinSSH.configFile` says.

So after #1060, a user who set one of those got the workspace host written to a file the connection never reads, and the connection failed. Before #1060 that user had a cosmetic mismatch only. The same applies to a stale `remote.SSH.configFile` synced in from VS Code, which those editors also never pass to ssh.

## Change

Ignore the setting on those two extensions and read `remote.SSH.configFile` everywhere else, because the three extensions that genuinely connect through it all use that name:

| Extension | Connects through the setting? |
|---|---|
| `ms-vscode-remote.remote-ssh` | yes, passes it with `-F` |
| `anysphere.remote-ssh` | yes, passes it with `-F` |
| `jeanp413.open-remote-ssh` | yes, parses the file itself (bundles `ssh2`, never runs ssh) |
| `google.antigravity-remote-openssh` | no |
| `codeium.windsurf-remote-openssh` | no |

That makes #1060's per-extension section map inert, so it is removed: the renamed sections belong exclusively to extensions we now skip. The rule is documented in `CONTRIBUTING.md` next to the SSH config flow.

Known cosmetic trade-off: on those two editors, a user pointing the setting elsewhere will not see Coder workspaces in the extension's host tree. Their own hosts in that file are equally unreachable for the same upstream reason, so this restores the pre-#1060 behaviour rather than introducing a new gap.

## Testing

`getRemoteSshConfigFile` is covered for the three honouring extensions, for both ignoring ones (including with a stale `remote.SSH.configFile` present), for no extension installed, and for nothing configured. Full suite passes, plus lint and typecheck.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,12 +25,10 @@
 
 ### Fixed
 
-- Write the SSH config to the file the active Remote-SSH extension actually
-  reads. Windsurf/Devin and Antigravity renamed the whole `remote.SSH` settings
-  section, so a custom config file set as `remote.devinSSH.configFile`,
-  `remote.windsurfSSH.configFile`, or `remote.antigravitySSH.configFile` was
-  ignored and the workspace host was written to `~/.ssh/config` instead, where
-  those editors never looked for it.
+- Ignore the SSH config file setting on Antigravity and Windsurf/Devin. They
+  launch ssh without pointing it at a config file, so it always reads
+  `~/.ssh/config`, and honoring the setting wrote the workspace host where the
+  connection never looked.
 - Apply a 60-second default timeout to REST requests, so requests hung on a
   half-open TCP connection don't stall pollers forever.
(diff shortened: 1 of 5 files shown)


EhabY:
Folded into #1061.

PR #1070, opened 2026-08-06
feat: deprecate the Tasks panel and hide it on Coder 2.35 and above

EhabY:
Coder Tasks was deprecated in June 2026 (moving to ESR) and is going to be removed from deployments starting v2.37, replaced by Coder Agents. This deprecates the extension's Tasks panel accordingly:

- The activity bar item is now titled **Coder Tasks (Deprecated)**, which shows in its hover.
- The panel is only shown on deployments that predate the deprecation (**2.34 and below**). On newer deployments, and while signed out or before the version is known, the activity bar icon is hidden entirely.

### DIFF
--- a/package.json
+++ b/package.json
@@ -313,7 +313,7 @@
 				},
 				{
 					"id": "coderTasks",
-					"title": "Coder Tasks",
+					"title": "Coder Tasks (Deprecated)",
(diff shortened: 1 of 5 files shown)


matifali (approved):
LGTM.

PR #1071, opened 2026-08-09
fix(api): validate API responses at the boundary

### LINKED ISSUE #1050: Validate API responses at the boundary instead of trusting the generated types
## Problem

`CoderApi` extends the SDK's `Api`, which casts response bodies to the generated TypeScript types with no runtime check. Any endpoint that answers 2xx with a body that isn't the expected shape flows into the extension as if it were valid, and we crash later at the first unguarded property access — far from the cause, with a message that says nothing about which URL or endpoint was wrong.

Concrete case: if `/api/v2/users/me` returns 200 with something that isn't a `User` (a non-Coder service on the configured URL, an HTML error page from a proxy, a partial body), login *succeeds* and we fail here instead:

```ts
// src/deployment/deploymentManager.ts
const isOwner = user?.roles.some((r) => r.name === "owner") ?? false;
// TypeError: Cannot read properties of undefined (reading 'some')
```

The user sees `Cannot read properties of undefined (reading 'some')`. Via a deep link it surfaces as a `Failed to handle URI` modal with that same text. Note the `?.` guards `user` but not `roles`, so the error is doubly misleading — it can only happen when `user` *is* defined.

The same exposure exists at every other `getAuthenticatedUser()` call site (`loginCoordinator` ×4, `oauth/authorizer`, `DeploymentManager.#verifyCredentials`) and for every other response we read into without validating.

## Proposal

**1. Validate in one place.** zod is already in use (`DeploymentSchema`, `src/api/api-helper.ts`). Add a `UserSchema` and override the method on `CoderApi`, which covers all six call sites without touching any of them:

```ts
override async getAuthenticatedUser(): Promise<User> {
  return parseApiResponse(
    UserSchema,
    await super.getAuthenticatedUser(),
    "/api/v2/users/me",
  );
}
```

Keep schemas permissive — only the fields we actually read, unknown fields allowed — so a newer deployment adding fields never breaks login. The point is to catch "this isn't a user object", not to mirror the API.

**2. Throw a typed error with an actionable message**, not a raw `ZodError`:

```ts
export class InvalidApiResponseError extends Error {
  constructor(endpoint: string, url: string, options?: { cause?: unknown }) {
    super(
      `${url} did not return a valid Coder API response for ${endpoint}. ` +
        `Check that the URL points to a Coder deployment.`,
      options,
    );
  }
}
```

Login and URI-handling paths already funnel errors through `errToStr`, so this reaches the modal as-is. `src/command/diagnosticFlow.ts` already does the equivalent for CLI output (`… did not match the expected format`) — worth matching that wording and its "check `Output > Coder`" hint. Log the `ZodError` details rather than showing them.

**3. Guard the immediate crash**: `user?.roles?.some(...)`. One character, but it only relocates the failure — validation is the real fix, so this shouldn't land alone.

## Scope

Start with the endpoints on the login and connect paths (`users/me`, workspace, agents, deployment SSH config). No need to schema-ify the whole SDK surface; the value is at the boundary where a URL might not be a Coder deployment at all.

EhabY:
## Summary

Fixes coder/vscode-coder#1050

The SDK casts 2xx bodies to the generated TypeScript types with no runtime check, so a proxy error page, a non-Coder service, or a partial body flowed in as valid and crashed far from the cause — e.g. `user?.roles.some(...)` in `deploymentManager.ts` throwing `Cannot read properties of undefined`.

Responses are now validated at the HTTP call and fail with an error naming the URL and endpoint.

## What changed

* `src/api/responseValidation.ts` — `parseApiResponse` + `InvalidApiResponseError` (the Zod failure kept as `cause`), the schemas, and `VALIDATED_RESPONSES` mapping each SDK method to its schema. Schemas are `looseObject`s listing only fields the extension reads; a field is required only if every supported deployment sends it, so unknown fields pass through untouched and newer ones must be `.optional()`.
* `src/api/coderApi.ts` — `wrapWithValidation` iterates that map in the constructor. Also reimplements `waitForBuild`: the SDK polls inside a voided IIFE that swallows errors, so a validation failure would hang callers forever.
* `src/oauth/validation.ts` plus `metadataClient.ts`, `authorizer.ts`, `sessionManager.ts` — same treatment for the endpoints hit before a session exists (metadata, client registration, token exchange and refresh). `validateRequiredEndpoints` folded into the schema via `.min(1)`. Errors blame the endpoint's own origin, which may differ from the deployment.

## Behavior change

* A roles-less `/api/v2/users/me` now fails login with a clear error instead of silently treating the user as non-owner.
* A malformed build response rejects `waitForBuild` instead of hanging "Stopping workspace for update..." forever.

## Notes

* Wrapping is by reassignment: SDK `Api` methods are arrow-function instance properties, so `super.getAuthenticatedUser()` does not compile (ts(2855)) and `override` fields would depend on declaration order.
* No axios response interceptor — URL-to-schema matching is brittle with path params, and non-`AxiosError` rejections would confuse `isApiError` handling.
* Every required field was verified present in `codersdk` at v0.25.0, the floor `featureSet.ts` declares. The three collection fields we require (`resources`, `roles`, `ssh_config_options`) are built with `make(...)` upstream, so they serialize as `[]`/`{}` rather than Go's `null`.

## Testing

Table-driven: one case per schema pinning the minimal body an old deployment sends, one per validated method for pass and reject, and a guard that fails when a method joins `VALIDATED_RESPONSES` without a test case. `pnpm test` (2416 tests), `typecheck`, and `lint` all green.

---

*Generated by [Coder Agents](<https://coder.com/docs>) on behalf of* @EhabY

### DIFF
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -52,6 +52,7 @@ import { getRefreshCommand, refreshCertificates } from "./certificateRefresh";
 import {
 	parseApiResponse,
 	SSHConfigResponseSchema,
+	TemplateSchema,
 	UserSchema,
 	WorkspaceBuildSchema,
 	WorkspaceResourcesSchema,
@@ -61,16 +62,13 @@ import { createHttpAgent } from "./utils";
 
 import type {
 	GetInboxNotificationResponse,
+	ProvisionerJob,
 	ProvisionerJobLog,
 	ServerSentEvent,
-	SSHConfigResponse,
-	User,
 	Workspace,
 	WorkspaceAgent,
 	WorkspaceAgentLog,
 	WorkspaceBuild,
-	WorkspaceOptions,
-	WorkspaceResource,
 } from "coder/site/src/api/typesGenerated";
 import type { ClientOptions } from "ws";
 import type { z } from "zod";
@@ -128,6 +126,7 @@ export class CoderApi extends Api implements vscode.Disposable {
 		private readonly authConfigTracker: AuthConfigTracker,
 	) {
 		super();
+		wrapWithValidation(this);
 		this.configWatcher = this.watchConfigChanges();
 	}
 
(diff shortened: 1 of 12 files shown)


src/api/coderApi.ts
```diff
@@ -747,6 +775,24 @@ function wrapResponseTransform(
...
 }
 
+/**
+ * Validate the fields the extension reads on each response, since the SDK
+ * casts bodies to the generated types with no runtime check. The methods
+ * are instance arrow properties, so wrapping is by reassignment;
+ * `override` fields would depend on declaration order.
+ */
+function wrapWithValidation(api: CoderApi): void {
+	const methods = api as unknown as Record<
```
jeremyruppel: nit: `as unknown as` is a major code smell and loses type information. any way to avoid it here?

jeremyruppel (approved):
zod ftw! one type nit but otherwise this looks fantastic. nice one!

src/api/coderApi.ts
EhabY: Oh I removed this but then refactored and it was added again lol. I'll remove this :pray:

EhabY: Fixed by using TS correctly :)!

PR #1073, opened 2026-08-11
feat(logging): tag logs with a per-session ID

aqandrew:
## What

Introduce a per-session identifier and make the extension tag every log line with it, so all logs for a session can be correlated by searching a single ID.

This is **Phase 1 of 3** for [DEVEX-661](https://linear.app/codercom/issue/DEVEX-661) — "VS Code: add `session_id` to all requests, logs, existing telemetry, and CLI invocations".

## Changes

- Add `SessionLogger` (`src/logging/sessionLogger.ts`) — a `Logger` that wraps the Coder output channel and prefixes every message with `[<sessionId>]`.
- Generate the session ID once in `ServiceContainer` (reusing the existing `newSessionId()` that already backs the telemetry session, so logs, telemetry, requests, and the CLI all share one ID), and expose it via `getSessionId()` for the later phases.

Because every service already receives `Logger` by injection, session-tagged logging propagates with no call-site changes.

## Design notes

The code owner (Ehab) suggested composing the session ID and the VS Code logger inside `container.ts`'s `ServiceContainer`; this PR follows that approach. The telemetry `sessionId` from `newSessionId()` *is* the RFC `session_id` (confirmed with Ehab — reuse it rather than minting a second ID).

## Testing

- `pnpm typecheck`, targeted `pnpm lint`, and `pnpm test:extension` (full suite: 2108 passing).
- New unit tests for `SessionLogger` prefixing and argument forwarding.

<details>
<summary>Implementation plan (DEVEX-661)</summary>

Thread a single per-session identifier (16-byte / 32-char lowercase hex) so that
logs, API requests, existing telemetry, and the CLI `ssh` invocation the
extension drives can all be correlated by one `session_id`.

**In-scope RFC requirements**

| Req | Summary | Where |
| --- | ------- | ----- |
| 1 | Generate 16-byte / 32-hex session ID | Reuse existing `newSessionId()` |
| 2 | Session ID on every client log | Phase 1 (this PR) |
| 3 | `session_id` on every API request via baggage | Phase 2 |
| 4.2 | `session_id` on VS Code telemetry | Phase 2 (already covered by reusing the ID) |
| 5.2/5.3 | `CODER_TRACE_SESSION_ID` via `process.env` + terminal env collection | Phase 2 |
| 7 | Log workspace/agent/lifecycle state changes at `info` | Phase 3 |
| 16 | Default `--log-dir` + old-log cleanup | Already implemented — verify only |

**Out of scope:** coderd tracing middleware (req 6), agent/coordination-protocol
changes and the CLI `ssh` subcommand behavior (reqs 8–15) live in `coder/coder`.
Req 13 (in-memory log buffer flushed on connection failure) is deferred to a
follow-up.

**Key decisions**

- D1 (resolved): the telemetry `sessionId` *is* the RFC `session_id` — one ID,
  generated once in `ServiceContainer`.
- D2: session scope is per activation; Remote-SSH opens each workspace in a fresh
  activation, so container-scoped ≈ per-connection.
- D3 (resolved): prefix every log line with the full 32-hex ID as
  `[<sessionId>] <message>`.
- D4: `baggage: session_id=<hex>` header.

**Phases**

1. Session ID + `SessionLogger` composition in the container (this PR).
2. Thread the ID to API requests (baggage), telemetry (already covered), and the
   CLI via `CODER_TRACE_SESSION_ID` on `process.env` + the terminal env
   collection.
3. Log workspace / agent / lifecycle state transitions at `info`.

</details>

---

🤖 Generated by Coder Agents.


src/core/container.ts
```diff
@@ -139,6 +148,10 @@ export class ServiceContainer implements vscode.Disposable {
 		return this.logger;
 	}
 
+	getSessionId(): string {
+		return this.sessionId;
+	}
```
EhabY: Should we expose this here? I don't think it's used

src/logging/sessionLogger.ts
```diff
@@ -0,0 +1,42 @@
+import type { Logger } from "./logger";
```
EhabY: We can actually simplify this using functional programming:

```ts
import type { Logger } from "./logger";

/**
 * Wraps a {@link Logger} so every message is prefixed, letting all lines that
 * share a prefix (a session ID, a workspace name) be found with one search.
 * Extra arguments are forwarded untouched.
 */
export function prefixLogger(inner: Logger, prefix: string): Logger {
	const tag = (message: string) => `${prefix} ${message}`;
	return {
		trace: (message, ...args) => inner.trace(tag(message), ...args),
		debug: (message, ...args) => inner.debug(tag(message), ...args),
		info: (message, ...args) => inner.info(tag(message), ...args),
		warn: (message, ...args) => inner.warn(tag(message), ...args),
		error: (message, ...args) => inner.error(tag(message), ...args),
		show: () => inner.show(),
	};
}
```

Maybe even call this `prefixLogger.ts`

src/logging/sessionLogger.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export class SessionLogger implements Logger {
+	constructor(
+		private readonly inner: Logger,
+		private readonly sessionId: string,
+	) {}
+
+	private prefix(message: string): string {
+		return `[${this.sessionId}] ${message}`;
+	}
```
EhabY: Should we use `shortId` for the session ID? This is very verbose now and conflict is extremely unlikely to happen

Image

Also I'm not sure if this is clear that the first number is the session ID while the second number is the request ID

src/logging/sessionLogger.ts
aqandrew: Love that idea, thanks for the suggestion! Done in e5dec0e. Along with the rename so that we can use it for prefixes other than session ids if needed

src/logging/sessionLogger.ts
aqandrew: Yeah we definitely should use `shortId`. What do you think of wrapping both the session id and request id in square brackets, and labeling them? So the top line would become

```
2026-08-13 14:53:01.305 [trace] [session ece5f89f] ← [request 38b0a2b4] 200 GET /api/v2/workspaces?q=owner%3Ame (158 kB) 1.35s
```

EhabY: Yes I like this, this is clearer!

src/core/container.ts
```diff
@@ -42,7 +45,13 @@ export class ServiceContainer implements vscode.Disposable {
 	private readonly commandManager: CommandManager;
 
 	constructor(context: vscode.ExtensionContext) {
-		this.logger = vscode.window.createOutputChannel("Coder", { log: true });
+		this.outputChannel = vscode.window.createOutputChannel("Coder", {
+			log: true,
+		});
+		// One session ID per activation, shared by logs, API requests,
+		// telemetry, and the CLI so all data for a session correlates.
```
EhabY, line 53: Nit: this comment will go once we have a the file with a session ID defined at runtime

EhabY (approved):

src/logging/sessionLogger.ts
aqandrew: Done in 1e6b0d5. Now using `shortId` for both, wrapped and labeled, so a line reads:

```
2026-08-13 14:53:01.305 [trace] [session ece5f89f] ← [request 38b0a2b4] 200 GET /api/v2/workspaces?q=owner%3Ame (158 kB) 1.35s
```

The full session ID is unchanged on the wire (`client_session_id` baggage, `CODER_TRACE_SESSION_ID`); only the log representation is shortened.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

PR #1074, opened 2026-08-11
feat: attach client_session_id to requests and set the CLI session env var

aqandrew:
## What

Propagate the session ID from Phase 1 to outbound API requests and to the CLI, so server-side logs, telemetry, and the spawned `coder ssh` process can all be correlated with the extension's session.

This is **Phase 2 of 3** for [DEVEX-661](https://linear.app/codercom/issue/DEVEX-661). It builds on #1073.

## Changes

- **Requests (RFC req 3):** add a `sessionId` argument to `CoderApi.create` and attach the session ID to every request via the W3C `baggage` header using the `client_session_id` key (`baggage: client_session_id=<hex>`), on both the REST default headers and the WebSocket handshake headers. Threaded through all `CoderApi.create` call sites, including the pre-auth OAuth/login/deployment clients, so every request carries it.
- **Telemetry (RFC req 4.2):** already satisfied — the shared ID is the telemetry `sessionId`, which ships on every event's context.
- **CLI (RFC reqs 5.2/5.3):** extend `applySshEnvironment` to also set `CODER_TRACE_SESSION_ID` on both `process.env` and the terminal environment collection, so the spawned `coder ssh` ProxyCommand reuses the plugin's session ID instead of generating its own.

## Testing

- `pnpm typecheck`, targeted `pnpm lint`, full `pnpm test:extension` (2108 passing).
- New tests: `baggage: client_session_id=<hex>` present/absent on `CoderApi`; `CODER_TRACE_SESSION_ID` applied to `process.env` and the terminal collection and restored on dispose.

---

🤖 Generated by Coder Agents.

### DIFF
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -80,6 +80,9 @@ const coderSessionTokenHeader = "Coder-Session-Token";
 /** W3C baggage header used to propagate the session ID to the server. */
 const baggageHeader = "baggage";
 
+/** Baggage key that carries the client's session ID. */
+const sessionIdBaggageKey = "client_session_id";
+
 /**
  * Default timeout for REST requests, so requests hung on half-open TCP
  * connections (e.g. after system sleep) don't stall pollers forever.
@@ -157,7 +160,7 @@ export class CoderApi extends Api implements vscode.Disposable {
 		client.getAxiosInstance().defaults.timeout = DEFAULT_REQUEST_TIMEOUT_MS;
 		if (sessionId) {
 			client.getAxiosInstance().defaults.headers.common[baggageHeader] =
-				`session_id=${sessionId}`;
+				`${sessionIdBaggageKey}=${sessionId}`;
 		}
 		client.setCredentials(baseUrl, token);
 
@@ -392,7 +395,7 @@ export class CoderApi extends Api implements vscode.Disposable {
 		const headers = {
 			...(token ? { [coderSessionTokenHeader]: token } : {}),
 			...(this.sessionId
-				? { [baggageHeader]: `session_id=${this.sessionId}` }
+				? { [baggageHeader]: `${sessionIdBaggageKey}=${this.sessionId}` }
 				: {}),
 			...configs.options?.headers,
 			...headersFromCommand,
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -393,11 +393,11 @@ export class CoderApi extends Api implements vscode.Disposable {
 		 */
 		const headers = {
 			...(token ? { [coderSessionTokenHeader]: token } : {}),
+			...configs.options?.headers,
+			...headersFromCommand,
 			...(this.sessionId
 				? { [baggageHeader]: `${SESSION_ID_BAGGAGE_KEY}=${this.sessionId}` }
 				: {}),
-			...configs.options?.headers,
-			...headersFromCommand,
 		};
 
 		const baseUrl = new URL(baseUrlRaw);
--- a/src/api/coderApi.ts
+++ b/src/api/coderApi.ts
@@ -79,9 +79,9 @@ import type {
 const coderSessionTokenHeader = "Coder-Session-Token";
 
 /** W3C baggage header used to propagate the session ID to the server. */
-const baggageHeader = "baggage";
-
+const BAGGAGE_HEADER = "baggage";
 const SESSION_ID_BAGGAGE_KEY = "client_session_id";
+const SESSION_ID_BAGGAGE = `${SESSION_ID_BAGGAGE_KEY}=${sessionId}`;
 
 /**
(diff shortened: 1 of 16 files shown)


src/api/coderApi.ts
```diff
@@ -77,6 +77,11 @@ import type {
 
 const coderSessionTokenHeader = "Coder-Session-Token";
 
+/** W3C baggage header used to propagate the session ID to the server. */
+const baggageHeader = "baggage";
+
+const sessionIdBaggageKey = "client_session_id";
```
EhabY: Should this be similar to the session token header: `Client-Session-Id`? Where is the RFC actually?

Nit: but use UPPERCASE for file level consts

src/api/coderApi.ts
```diff
@@ -379,6 +393,9 @@ export class CoderApi extends Api implements vscode.Disposable {
 		 */
 		const headers = {
 			...(token ? { [coderSessionTokenHeader]: token } : {}),
+			...(this.sessionId
+				? { [baggageHeader]: `${sessionIdBaggageKey}=${this.sessionId}` }
+				: {}),
```
EhabY: Should this be the last option so it's never overriddable?

src/api/coderApi.ts
```diff
@@ -130,13 +136,16 @@ export class CoderApi extends Api implements vscode.Disposable {
...
+	 * NOOP_TELEMETRY_REPORTER for throwaway clients). When a session ID is
+	 * provided it is attached to every request via the `baggage` header so the
+	 * server can correlate requests with the session's logs and telemetry.
 	 */
 	static create(
 		baseUrl: string,
 		token: string | undefined,
 		output: Logger,
 		telemetry: TelemetryReporter = NOOP_TELEMETRY_REPORTER,
+		sessionId?: string,
```
EhabY: Do we ever not pass the session ID?

Ignore if you do the global session ID export

src/core/container.ts
```diff
@@ -114,6 +114,7 @@ export class ServiceContainer implements vscode.Disposable {
 			new AuthTelemetry(this.telemetryService),
 			this.oauthCallback,
 			context.extension.id,
+			this.sessionId,
```
EhabY: Hmmmm this is passed a lot and deep, how about we just add something like `core/sessionId.ts` that exports a constant? We can then remove `newSessionId()` from the `ids.ts` file 

This would remove most of the deltas here that are caused by passing this variable around

src/remote/environment.ts
EhabY: Not related but I got confused so can we rename this type to `SshProxyEnvironment`?

src/api/coderApi.ts
aqandrew: DMed you the RFC link! It specifically says the key should be lowercase snake_case

src/api/coderApi.ts
aqandrew: Good idea, I don't see much reason to override the baggage in `configs` / command args. Done in 1acc53b7af569fb29b1eb478a5bfe62e2451bd84

src/core/container.ts
aqandrew: 🔥 done in f625b5075892004af0d35ba952724b2ed884b7b4. The diff is way smaller now

src/api/coderApi.ts
aqandrew: no / went with the global exported session id!

src/remote/environment.ts
aqandrew: Yep! Done in fec1170b25d3b9cdfad5b0d1ebb6e692d141ea19

src/api/coderApi.ts
```diff
@@ -77,6 +78,11 @@ import type {
 
 const coderSessionTokenHeader = "Coder-Session-Token";
 
+/** W3C baggage header used to propagate the session ID to the server. */
+const baggageHeader = "baggage";
```
EhabY: nit: UPPERCASE the variables here, also we could just do instead of defining the key then redoing this everywhere:

```ts
SESSION_ID_BAGGAGE = `${SESSION_ID_BAGGAGE_KEY}=${sessionId}`
```

src/core/sessionId.ts
```diff
@@ -0,0 +1,11 @@
+import { randomBytes } from "node:crypto";
+
+/**
+ * One session ID per activation, shared by logs, API requests, telemetry, and
+ * the CLI so all data for a session can be correlated by a single ID.
```
EhabY: This should be reworded to something like:

> one per extension host process, shared across logs, API requests, telemetry, and the CLI

Because there is a rare chance that an extension is deactivated then activated within the same window so it would share the same ID (which previously didn't but I'd argue that's better for the usecase here). Note that reloading a window or closing and opening does create a new process!

EhabY (approved):

PR #1075, opened 2026-08-11
feat(workspace): log workspace and agent state changes

aqandrew:
## What

Add info-level logging of workspace, agent, and lifecycle status transitions so connection debugging has a record of state changes — correlated by the session ID from #1073.

This is **Phase 3 of 3** for [DEVEX-661](https://linear.app/codercom/issue/DEVEX-661) (RFC req 7). It builds on #1074.

## Changes

State handling is split into three layers, one job each:

- **detect** — `WorkspaceStateObserver` / `WorkspaceAgentObserver` (`src/instrumentation/workspace.ts`) each detect a transition once and report a transition object. Stateful but effect-free (no logger/telemetry imports). The agent observer tracks every agent by ID and reports agents that disappear.
- **record** — `recordWorkspaceState` / `recordAgentState` emit the `workspace.state_transitioned` / `workspace.agent.state_transitioned` telemetry events.
- **log** — inline in `WorkspaceMonitor.update()`, with flat scalar payloads matching the repo's other log sites. Logs `state observed` on first sight, `state changed` afterward, and `agent <name> removed` when an agent disappears.

`WorkspaceMonitor` is now the single agent-observation site, so `WorkspaceStateMachine` no longer tracks agent telemetry.

> [!IMPORTANT]
> **Telemetry data change.** `workspace.agent.state_transitioned` now covers **every agent for the whole monitored session** (deduped per agent), rather than only the connected agent during connection setup. In practice most workspaces have one agent, so volume is unchanged; multi-agent workspaces and flaky connections emit more. There is no sampling in `src/telemetry/`, so this lands as-is. `EVENTS.md` is updated; heads-up to whoever owns the funnel queries.

## Testing

- `pnpm typecheck`, targeted `pnpm lint`, full `pnpm test` (2456 passing).
- Unit tests cover observer detection (first observation, dedup, per-agent independence, removal, reset), the `record*` event shapes, and the monitor's inline logging (observed vs changed, agent transitions across all agents, agent removal).

<details>
<summary>Refactor design (reviewer's three-layer shape)</summary>

Per the review, the goal was to remove the duplication between the telemetry observers and the state loggers by making transition detection happen once:

- **detect**: pure-detection observers report a transition object; no logger or telemetry imports.
- **record**: `recordWorkspaceState` / `recordAgentState` in `src/instrumentation/`, telemetry only.
- **log**: inline in the monitor, same as every other logging call site.

This deletes both `src/workspace/*Logger.ts` files and leaves one tracker per concern.

**Observer API decision:** observers return a transition object (workspace → `WorkspaceStateTransition | undefined`; agents → `{ transitions, removed }`) and `WorkspaceMonitor.update()` logs inline and calls `record*`. This was chosen over a constructor callback because `observe()` has a single synchronous caller, so a callback would only hide the side effects and complicate tests.

</details>

---

🤖 Generated by Coder Agents.


src/workspace/workspaceStateLogger.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/**
+ * Logs workspace, agent, and lifecycle status transitions at `info` level so
+ * connection debugging has a record of state changes correlated by the session
+ * ID. Tracks state per agent (keyed by agent ID) because a workspace can have
+ * several. Construct one per workspace; `WorkspaceMonitor` is the sole call
+ * site.
+ */
+export class WorkspaceStateLogger {
```
EhabY: There are lots of similarities between this and `WorkspaceStateTelemetry`, I suppose one is tracking the workspace states and one is tracking the agent states?

src/workspace/workspaceStateLogger.ts
```diff
@@ -0,0 +1,87 @@
...
+			});
+		}
+	}
+
+	private observeState(key: string, next: ObservedState): void {
+		const previous = this.observed.get(key);
+		if (
+			previous?.workspaceStatus === next.workspaceStatus &&
+			previous?.agentStatus === next.agentStatus &&
+			previous?.lifecycleState === next.lifecycleState
```
EhabY: IMO I think we should log once for each workspace state change (basically what the workspace telemetry does) and independently track the agent state since they could be independent

EhabY:
Should we include an entry in the changelog for the past 2 PRs + this? We only mention user-visible changes and these are worthy of it since it means sharing the logs or browsing them would give them an easy way to correlate logs to sessions. (also more logs so they can easily debug more)

src/workspace/workspaceStateLogger.ts
aqandrew: Done in c7dce24. Split the combined logger into `WorkspaceStateLogger` (one log per workspace **status** change, regardless of agent count) and `WorkspaceAgentLogger` (per-agent `status`/`lifecycle_state`, tracked independently), both wired in `WorkspaceMonitor`.

One scoping note: agent *logging* here covers **all** agents for the whole connection lifetime (via the monitor's watch socket), whereas agent *telemetry* (`WorkspaceAgentTelemetry`) stays selected-agent/connect-time only in `WorkspaceStateMachine`. I kept telemetry's scope as-is — let me know if you'd also like telemetry parity, and I'll do that as a separate change.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/workspaceStateLogger.ts
aqandrew: Addressed in c7dce24 — extracted a shared `TransitionTracker` now used by both the telemetry observers and the new loggers, so the transition-dedup logic lives in one place. The taxonomy is now explicit in the naming: `WorkspaceState{Telemetry,Logger}` (workspace status) and `WorkspaceAgent{Telemetry,Logger}` (agent status + lifecycle). The existing telemetry tests pass unchanged, confirming the observer refactor is behavior-preserving.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/workspaceMonitor.ts
```diff
@@ -136,6 +143,8 @@ export class WorkspaceMonitor implements vscode.Disposable {
 
 	private update(workspace: Workspace) {
 		this.telemetry.observe(workspace);
+		this.stateLogger.observe(workspace);
+		this.agentLogger.observe(workspace);
```
EhabY: These lines are the duplication I was getting at. `telemetry.observe` and `stateLogger.observe` take the same workspace and detect the same transition with two separate trackers, and the agent pair does it again.

Could we merge each pair into one observer that detects once and reports a transition, then let the caller decide?

```ts
this.stateObserver = new WorkspaceStateObserver(this.name, (t) => {
  this.logger.info(`Workspace ${this.name} state changed`, {
    from: t.from, to: t.to, transition: t.buildTransition, reason: t.buildReason,
  });
  recordWorkspaceState(telemetry, t);
});
```

Deletes both `workspace/*Logger.ts` files and leaves one tracker per concern. Telemetry event schema stays in `src/instrumentation/`, logging stays inline here, which matches how the rest of the repo does both.

One decision it forces: the comparators disagree. Telemetry dedupes on status plus transition plus reason (`instrumentation/workspace.ts:52`), the logger on status alone (`workspaceStateLogger.ts:23`). One tracker can't do both. I'd take the telemetry one and add transition and reason to the log payload, which also handles my note on `workspaceStateLogger.ts:38`.

src/workspace/workspaceAgentLogger.ts
```diff
@@ -0,0 +1,65 @@
...
+	WorkspaceAgentStatus,
+} from "coder/site/src/api/typesGenerated";
+
+import type { Logger } from "../logging/logger";
+
+/** Sentinel for the "from" side before any state is observed. `"unknown"` is a
+ * real server-reported value, so avoid it. */
+const INITIAL_STATE = "none";
+
+interface AgentState {
```
EhabY: This is the rest of my duplication comment. `TransitionTracker` shared the mechanism but not the state, comparator, or sentinel:

- `AgentState` here is `ObservedAgentState` (`instrumentation/workspace.ts:39`) minus `observedAtMs`
- the comparator at line 31 is byte identical to `instrumentation/workspace.ts:121`
- `INITIAL_STATE` is defined three times: line 14, `workspaceStateLogger.ts:12`, `instrumentation/workspace.ts:18`

Merging the pairs removes all three for free.

src/util/transitionTracker.ts
```diff
@@ -0,0 +1,36 @@
...
+	private readonly previous = new Map<string, T>();
+
+	public constructor(private readonly equals: (a: T, b: T) => boolean) {}
+
+	/**
+	 * Record `next` for `key`. Returns `{ from }` when it differs from the last
+	 * recorded value (`from` is `undefined` on the first observation), or
+	 * `undefined` when unchanged.
+	 */
+	public observe(next: T, key = ""): { from: T | undefined } | undefined {
```
EhabY: This contract is a bit confusing. Two nested undefineds meaning different things, outer for unchanged and inner `from` for first observation, from a method whose job is to mutate. A callback makes it one undefined with one meaning:

```ts
public observe(next: T, key: string, onChange: (from: T | undefined) => void): void
```

Callers lose their `if (!change) return` branches too. Also the `reset(key)` overload on line 29 is unused, only the no-arg form gets called.

Bigger question: should this file just fold into the observers file? After merging the pairs it has exactly one consumer, and there's no second candidate, since `connection.state_transitioned` takes `from`/`to` as arguments (`instrumentation/websocket.ts:68`) so the websocket layer tracks its own state. Easy to pull back out if that changes.

src/workspace/workspaceStateLogger.ts
```diff
@@ -0,0 +1,42 @@
...
+	) {}
+
+	public observe(workspace: Workspace): void {
+		const status = workspace.latest_build.status;
+		const change = this.tracker.observe(status);
+		if (!change) {
+			return;
+		}
+
+		this.logger.info(`Workspace ${this.workspaceName} state changed`, {
```
EhabY: Two things here. It logs status only, but telemetry emits `build.transition` and `build.reason` for the same transition (`instrumentation/workspace.ts:99`). `reason` is what tells an autostop from a user stop, usually the first thing you want from a dropped connection log.

And the payload is nested, which is new for this repo. Every other structured log call passes flat scalars (`remote/remote.ts:656`, `commands.ts:1573`, `remote/workspaceStateMachine.ts:288`), and telemetry uses flat dotted keys. Worth flattening both loggers, `workspaceAgentLogger.ts:50` wants `{ statusFrom, statusTo, lifecycleFrom, lifecycleTo }`.

src/workspace/workspaceAgentLogger.ts
```diff
@@ -0,0 +1,65 @@
...
+	readonly status: WorkspaceAgentStatus;
+	readonly lifecycleState: WorkspaceAgentLifecycle;
+}
+
+/**
+ * Logs agent status and lifecycle transitions at `info` level so connection
+ * debugging has a record of state changes correlated by the session ID. Mirrors
+ * `WorkspaceAgentTelemetry` but tracks every agent (keyed by agent ID) for the
```
EhabY: On the scope note in your reply: yes please, let's move agent telemetry here so logging and telemetry stop sitting in two different layers. `WorkspaceStateMachine` can drop `agentTelemetry` entirely, and `resetAgent()` plus both `reset()` methods go dead with it.

Worth flagging in the PR body though, since it's a data change and not a refactor. `workspace.agent.state_transitioned` grows two ways: all agents instead of just the connected one (linear in agent count, most workspaces have one), and the whole monitored session instead of connection setup. Dedup means a stable agent adds nothing, so the growth is really multi-agent workspaces and flaky connections. No sampling in `src/telemetry/`, so it lands as is, and `observed_duration_ms` stops being connect-time only. Probably wants an EVENTS.md update and a heads up to whoever owns the funnel queries.

src/workspace/workspaceAgentLogger.ts
```diff
@@ -0,0 +1,65 @@
...
+				status: agent.status,
+				lifecycleState: agent.lifecycle_state,
+			};
+			const change = this.tracker.observe(next, agent.id);
+			if (!change) {
+				continue;
+			}
+
+			this.logger.info(
+				`Workspace ${this.workspaceName} agent ${agent.name} state changed`,
```
EhabY: Two small ones. `WorkspaceMonitor` calls `update()` from its constructor, so every connection opens with a `state changed` line where nothing changed. Maybe `state observed` when `from` is the sentinel, in both loggers.

And nothing logs when an agent disappears, so a rebuild gives you lines for the new agents and no trace of the old ones leaving. Map growth itself is bounded and fine.

EhabY (commented):
Nice cleanup, the split into two loggers resolves my second comment. My first one about duplication is only half done, and fixing it properly is the one structural change I'm after, left inline on `workspaceMonitor.ts:145`.

Shape I have in mind, three layers with one job each:

- detect: the observers, pure, no logger or telemetry imports, report a transition object
- record: `recordWorkspaceState` / `recordAgentState` in `src/instrumentation/`, telemetry only
- log: inline in the monitor, same as every other logging call site here

Sorry this got a bit out of hand, I saw an opportunity for some refactoring and unifying :)

src/workspace/workspaceMonitor.ts
aqandrew: Done in 97ef3df — the workspace state log payload now includes `transition` and `reason` (flat), taking the telemetry observer semantics. The `TransitionTracker` equality check considers all of `status`, `transition`, and `reason`, so a transition change while the status stays `running` still logs.

🤖 Posted by Coder Agents on behalf of @aqandrew.

src/util/transitionTracker.ts
aqandrew: Done in df7596b — folded `TransitionTracker` into `src/instrumentation/workspace.ts` and removed `src/util/transitionTracker.ts`. The loggers now import the class from the observers file, and the tracker tests moved into `test/unit/instrumentation/workspace.test.ts`.

🤖 Posted by Coder Agents on behalf of @aqandrew.

src/workspace/workspaceStateLogger.ts
aqandrew: Fully addressed in 1ecdb98. The parallel logger classes are gone — workspace and agent state are each detected once by a single observer (`WorkspaceStateObserver` / `WorkspaceAgentObserver`), which the monitor logs inline and records to telemetry via `recordWorkspaceState` / `recordAgentState`. One tracker per concern now.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/workspaceAgentLogger.ts
aqandrew: Addressed in 1ecdb98. The duplicated `AgentState` type, the byte-identical comparator, and the three `INITIAL_STATE` definitions are gone. There's now a single `WorkspaceAgentObserver` with one comparator, and one exported `INITIAL_STATE` in `instrumentation/workspace.ts`.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/workspaceStateLogger.ts
aqandrew: Addressed in 1ecdb98. The inline workspace log now emits flat `{ from, to, transition, reason }`, and the agent log emits flat `{ statusFrom, statusTo, lifecycleFrom, lifecycleTo }` — no nested objects. `transition` and `reason` ride along from the observer's transition object.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/workspaceAgentLogger.ts
aqandrew: Done in 1ecdb98 — `WorkspaceStateMachine` no longer constructs or observes agent telemetry (`agentTelemetry` and its reset are gone); `WorkspaceMonitor` is the single observation site. `EVENTS.md` is rescoped in 86aaae3, and I've added the data-change heads-up to the PR body (all agents, whole session, deduped, no sampling).

One note: I kept `resetAgent()` in the state machine because it still clears the selected agent (`this.agent`) on rebuild — only the telemetry `reset()` line was removed.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/workspaceAgentLogger.ts
aqandrew: Both done in 1ecdb98. The first observation now logs `state observed` (sentinel `from`) and later ones `state changed`, for both the workspace and each agent. And the agent observer reports agents that disappear since the previous observation, which the monitor logs as `agent <name> removed` — so a rebuild leaves a trace of the old agents leaving.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/workspaceMonitor.ts
```diff
@@ -6,7 +6,13 @@ import { formatDistanceToNowStrict } from "date-fns";
 import * as vscode from "vscode";
 
 import { createWorkspaceIdentifier, errToStr } from "../api/api-helper";
-import { WorkspaceStateTelemetry } from "../instrumentation/workspace";
+import {
+	INITIAL_STATE,
+	recordAgentState,
+	recordWorkspaceState,
+	WorkspaceAgentObserver,
+	WorkspaceStateObserver,
```
EhabY: Those should be moved into `src/workspace` IMO because they are just agent/state observers without relating to instrumentation (it's a pure event subscription)

src/instrumentation/workspace.ts
```diff
@@ -28,6 +28,43 @@ const PROVISIONING_STATUSES: ReadonlySet<WorkspaceStatus> = new Set([
...
 
+/**
+ * Tracks the last observed value per key and reports the previous value each
+ * time it changes. Shared by the workspace/agent observers and the corresponding
+ * loggers so transition detection lives in one place.
+ *
+ * A single tracked entity (e.g. a workspace) can omit the key; callers that
+ * track many entities (e.g. agents keyed by ID) pass a distinct key each time.
+ */
+export class TransitionTracker<T> {
```
EhabY: These are pure (no telemetry or logger imports), but that's exactly why I don't think they belong here. `CONVENTIONS.md` says `instrumentation/` is telemetry only, one class per domain. Can we move `TransitionTracker`, `WorkspaceStateObserver`, `WorkspaceAgentObserver` (lines 39-243), and the transition types over to `src/workspace/observers.ts`? They're always used together (`workspaceMonitor.ts` calls both every `update()`), so one file beats splitting into `stateObserver.ts` / `agentObserver.ts`. Tests should probably follow too (`test/unit/instrumentation/workspace.test.ts:214-527`).

src/instrumentation/workspace.ts
```diff
@@ -28,6 +28,43 @@ const PROVISIONING_STATUSES: ReadonlySet<WorkspaceStatus> = new Set([
...
+	private readonly previous = new Map<string, T>();
+
+	public constructor(private readonly equals: (a: T, b: T) => boolean) {}
+
+	/**
+	 * Record `next` for `key`. Returns `{ from }` when it differs from the last
+	 * recorded value (`from` is `undefined` on the first observation), or
+	 * `undefined` when unchanged.
+	 */
+	public observe(next: T, key = ""): { from: T | undefined } | undefined {
```
EhabY: Still doing a double undefined check here: `if (!change) return` for "nothing changed," then `change.from` for "was this the first observation," which gets re-checked again downstream. The fold-into-one-file pass (`df7596b9`) moved `TransitionTracker` next to the observers but didn't actually merge it in. Since it's only used by these two, what if we inline the tracking directly into each observer and drop the generic class? Each one could shape its own result instead of sharing this wrapper.

src/instrumentation/workspace.ts
```diff
@@ -41,121 +78,225 @@ interface ObservedAgentState {
...
 /**
- * Emits `workspace.agent.state_transitioned` as the agent's `status` and
- * `lifecycle_state` change. The agent has two state dimensions so the event
- * carries `status.*` and `lifecycle_state.*` properties. Construct one per
- * workspace.
+ * Detects agent status/lifecycle changes for every agent in a workspace,
+ * keyed by agent ID so each is tracked independently, and reports agents that
+ * have disappeared since the previous observation. Stateful but effect-free.
+ * Construct one per workspace.
  */
```
EhabY: A lot of these doc comments just restate the name, e.g. `:81`, `:94`, `:109`, `:114` (the four transition/observation interfaces), and the two observer class docs at `:120-125` and `:183-188` basically describe each other. Can we trim these? (`:16` and `:341-344` earn their keep though, those explain something non-obvious.)

src/workspace/workspaceMonitor.ts
```diff
@@ -135,12 +141,54 @@ export class WorkspaceMonitor implements vscode.Disposable {
...
+		this.logger.info(`Workspace ${this.name} ${verb}`, {
+			from: transition.from ?? INITIAL_STATE,
+			to: transition.to,
+			transition: transition.buildTransition,
+			reason: transition.buildReason,
+		});
+		recordWorkspaceState(this.telemetry, this.name, transition);
+	}
+
+	/** Detect agent status/lifecycle changes and removals, logging and recording each. */
```
EhabY: Same thing here, `:151` and `:168` just restate `observeState` / `observeAgents`. Worth cutting these along with the ones in `instrumentation/workspace.ts`.

src/instrumentation/EVENTS.md
```diff
@@ -550,6 +550,11 @@ Opening a workspace from any entry point.
 
 #### `workspace.agent.state_transitioned`
 
+Emitted for every agent in the workspace, deduped per agent, for the whole
+monitored session (not only the connected agent during connection setup).
+Transitions are sampled from the workspace event stream, so intermediate hops
+between samples may coalesce into a single transition.
```
EhabY: Both actually coalesce the same way today, same `TransitionTracker.observe()` check, tested symmetrically for workspace and agent. But this coalescing note only shows up here, `workspace.state_transitioned` above (`:540-549`) doesn't have it even though it's the same mechanism. Intentional, or just missing from the workspace doc?

src/instrumentation/workspace.ts
aqandrew: Yeah definitely, done in 1dbc216

src/workspace/workspaceMonitor.ts
aqandrew: Totally fair, removed in 83d6125e9090ba9a72984f17c707dd7c7c50034e

src/instrumentation/workspace.ts
aqandrew: Done in 8e6ca43. Moved `TransitionTracker`, `WorkspaceStateObserver`, `WorkspaceAgentObserver`, and the transition types to `src/workspace/observers.ts` (`PROVISIONING_STATUSES` and the `Observed*` interfaces went with them). `recordWorkspaceState` / `recordAgentState` stay in `instrumentation/` and import the transition types from the new file. The observer and `TransitionTracker` tests moved to `test/unit/workspace/observers.test.ts`; the `record*` tests stayed put.

I left `INITIAL_STATE` in `instrumentation/workspace.ts` — the observers return `from: undefined` and never touch it; it's the sentinel the telemetry/log `from` value falls back to, so it lives with `record*` and the monitor's inline logging. Happy to relocate it if you'd rather it sit elsewhere.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/instrumentation/EVENTS.md
aqandrew: Not intentional — fixed in 0c10123. Lifted the coalescing note up under `### Logs` so it covers both `workspace.state_transitioned` and `workspace.agent.state_transitioned`, and left the agent-only "every agent / whole session" note under the agent event.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/instrumentation/workspace.ts
aqandrew: Done in 7919d7c. Dropped the generic `TransitionTracker` and inlined the tracking into each observer: `WorkspaceStateObserver` holds a single previous-state field, `WorkspaceAgentObserver` a `Map` keyed by agent ID. Each compares inline and shapes its own result directly, so the `if (!change)` / `change.from` double-check is gone. Removed the tracker's unit tests; the observer tests cover the behavior.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/observers.ts
```diff
@@ -0,0 +1,181 @@
...
+			from: previous?.status,
+			to: status,
+			buildTransition,
+			buildReason,
+			durationMs: previous ? now - previous.observedAtMs : undefined,
+			buildDurationMs,
+		};
+	}
+
+	public reset(): void {
```
EhabY: Both `reset()` methods are dead, this one and `:177`. Nothing in `src/` calls either; the only callers are `test/unit/workspace/observers.test.ts:106` and `:211`. Delete both methods and their two tests, since `buildStartedAtMs` and the maps now live and die with the monitor.

src/workspace/observers.ts
```diff
@@ -0,0 +1,181 @@
...
+		this.buildStartedAtMs = undefined;
+	}
+}
+
+/**
+ * Construct one per workspace.
+ */
+export class WorkspaceAgentObserver {
+	/** Previous observed state per agent ID, tracked independently. */
+	private readonly previous = new Map<string, ObservedAgentState>();
```
EhabY, line 117: `previous` and `names` (`:132`) always hold the same keys: a first sight always reports a transition, so both get set together, and the removal loop deletes from both. Fold the name into the state and drop `names`:

```ts
interface ObservedAgentState {
	readonly name: string;
	readonly status: WorkspaceAgentStatus;
	readonly lifecycleState: WorkspaceAgentLifecycle;
	readonly observedAtMs: number;
}
```

The removal loop then iterates `previous`. One caveat: a rename with no state change keeps the old name, which can't happen without a rebuild handing out new IDs.

src/workspace/observers.ts
```diff
@@ -0,0 +1,181 @@
...
+	};
+	readonly lifecycleState: {
+		readonly from: WorkspaceAgentLifecycle | undefined;
+		readonly to: WorkspaceAgentLifecycle;
+	};
+	/** Time since the previous observation of this agent; `undefined` on the first. */
+	readonly durationMs: number | undefined;
+}
+
+export interface RemovedAgent {
```
EhabY: `RemovedAgent` wraps a single `name`. Make it `readonly removed: string[]`, so the monitor reads `for (const name of removed)` and the test asserts `["second"]`. Also no need to export if this is only used internally

src/workspace/observers.ts
```diff
@@ -0,0 +1,181 @@
...
+	readonly buildTransition: WorkspaceBuild["transition"];
+	readonly buildReason: WorkspaceBuild["reason"];
+	/** Time spent in the previous state; `undefined` on the first observation. */
+	readonly durationMs: number | undefined;
+	/** Set only on the observation where a provisioner run resolves. */
+	readonly buildDurationMs: number | undefined;
+}
+
+/** Reported by `WorkspaceAgentObserver`. */
+export interface AgentStateTransition {
```
EhabY, line 51: `AgentStateTransition` is the last nested payload left. Both consumers flatten it on arrival. Put `statusFrom`, `statusTo`, `lifecycleFrom`, and `lifecycleTo` on the transition itself.

src/workspace/observers.ts
```diff
@@ -0,0 +1,181 @@
...
+		if (
+			previous?.status === status &&
+			previous?.buildTransition === buildTransition &&
+			previous?.buildReason === buildReason
+		) {
+			return undefined;
+		}
+		this.previous = { status, buildTransition, buildReason, observedAtMs: now };
+
+		const wasProvisioning =
```
EhabY: `buildStartedAtMs` is set only while provisioning and cleared as soon as it isn't, so `wasProvisioning` is exactly `buildStartedAtMs !== undefined` and both locals can go:

```ts
let buildDurationMs: number | undefined;
if (PROVISIONING_STATUSES.has(status)) {
	this.buildStartedAtMs ??= now;
} else if (this.buildStartedAtMs !== undefined) {
	buildDurationMs = now - this.buildStartedAtMs;
	this.buildStartedAtMs = undefined;
}
```

src/instrumentation/workspace.ts
```diff
@@ -1,159 +1,73 @@
...
-
-/** Statuses where a provisioner job is actively running. */
-const PROVISIONING_STATUSES: ReadonlySet<WorkspaceStatus> = new Set([
-	"pending",
-	"starting",
-	"stopping",
-	"canceling",
-	"deleting",
-]);
+export const INITIAL_STATE = "none";
```
EhabY: `workspaceMonitor.ts:10` imports `INITIAL_STATE` from `instrumentation/` to build a **log** payload, which is the same layering argument that moved the observers out. Move the constant to `observers.ts`, next to the types whose docs say "`undefined` on the first observation", and let `instrumentation/` import it from there.

EhabY (commented):
Shape looks right now, the three layers each do one job and the earlier duplication is gone. Just some small clean up left :pray: 

**Changelog.** Still missing. The Unreleased section has no entry for the session ID or the state logs.

**`src/instrumentation/CONVENTIONS.md:9`.** The checklist says one instrumentation class per domain wrapping `TelemetryService`, and this PR replaces two such classes with the stateless `recordWorkspaceState` and `recordAgentState` functions. Either keep the methods  but update the conventions or convert them into a class if there's some shared state that is useful.

src/workspace/workspaceMonitor.ts
```diff
@@ -135,12 +141,52 @@ export class WorkspaceMonitor implements vscode.Disposable {
...
 		this.updateContext(workspace);
 		this.updateStatusBar(workspace);
 	}
 
+	private observeState(workspace: Workspace) {
+		const transition = this.stateObserver.observe(workspace);
+		if (!transition) {
+			return;
+		}
+		const verb =
```
EhabY: Same ternary at `:170`. One module-level helper covers both and reads better at the call site:

```ts
const stateVerb = (from: string | undefined) =>
	from === undefined ? "state observed" : "state changed";
```

test/unit/workspace/workspaceMonitor.test.ts
```diff
@@ -64,11 +69,19 @@ describe("WorkspaceMonitor", () => {
...
+			client,
+			stream,
+			config,
+			statusBar,
+			contextManager,
+			logger,
+		};
 	}
 
 	describe("telemetry", () => {
```
EhabY, line 87: This block now re-tests the observers. Dedup on reason (`:132`), build-duration resolution (`:173`), and the duration measurements (`:96`) are all observer behavior, and `observers.test.ts:56`, `:75`, and `:90` cover each of them directly. Now that detection lives in one place, keep the monitor tests on wiring only: the first update logs and records once, a later change logs and records again. That drops roughly 100 lines of triple coverage.

test/unit/workspace/observers.test.ts
```diff
@@ -0,0 +1,215 @@
...
+		expect(transitions).toHaveLength(1);
+		expect(transitions[0]).toMatchObject({
+			agentName: "main",
+			status: { from: undefined, to: "connecting" },
+			lifecycleState: { from: undefined, to: "created" },
+			durationMs: undefined,
+		});
+	});
+
+	it("dedupes an unchanged agent", () => {
```
EhabY, line 106: `WorkspaceStateObserver` has "reports a change when only transition or reason changes" (`:75`), but the agent side never changes lifecycle on its own. Add the symmetric case: same `status`, new `lifecycle_state`, one transition reported.

test/unit/workspace/workspaceMonitor.test.ts
```diff
@@ -193,6 +206,137 @@ describe("WorkspaceMonitor", () => {
...
+
+			expect(logger.info).toHaveBeenCalledWith(
+				expect.stringContaining("state changed"),
+				expect.objectContaining({ from: "running", to: "stopping" }),
+			);
+		});
+	});
+
+	describe("agent state", () => {
+		const agentWorkspace = (
```
EhabY: `agentWorkspace` here and `workspaceWith` at `observers.test.ts:21` are the same builder, and this one types its agents as `Array<Parameters<typeof createAgent>[0]>`. Move one into the shared test mocks and have it take `WorkspaceAgent[]`.

test/unit/workspace/workspaceMonitor.test.ts
```diff
@@ -193,6 +206,137 @@ describe("WorkspaceMonitor", () => {
...
+				new MockEventStream<ServerSentEvent>(),
+				createTestTelemetryService(sink),
+				agentWorkspace([
+					{ id: "a1", name: "first", status: "connected" },
+					{ id: "a2", name: "second", status: "connecting" },
+				]),
+			);
+
+			stream.pushMessage({
+				type: "data",
```
EhabY: Also at `:330`. These hand-roll `{ type: "data", data }` instead of using the file's own `workspaceEvent` helper (`:34`). Let that helper take a `Workspace` as well.

src/workspace/observers.ts
aqandrew: done in 4162366

src/workspace/observers.ts
aqandrew: done in 3a5ce54

src/instrumentation/workspace.ts
aqandrew: done in 8bf51f5b3a13f33ff38d233f8a248ecdcf285a18

src/workspace/workspaceMonitor.ts
aqandrew: done in 797069b1ebe309ce015670fb95f96dee1bc53e1c

src/workspace/observers.ts
aqandrew: done in 82ae2ddbc2202e62bf978134cd6864050a3656ad

src/workspace/observers.ts
aqandrew: Done in fb6ce76. Dropped the `RemovedAgent` wrapper — `AgentObservation.removed` is now `string[]`, the monitor reads `for (const name of removed)`, and the observer test asserts `["second"]`.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

src/workspace/observers.ts
aqandrew: Done in cffeadb. Dropped both locals — since `buildStartedAtMs` is only set while provisioning and cleared otherwise, the check folds into `else if (this.buildStartedAtMs !== undefined)`.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

test/unit/workspace/workspaceMonitor.test.ts
aqandrew: Done in 951628b. Dropped the dedup, build-duration, and duration tests here (covered directly in `observers.test.ts`). The monitor telemetry test now just checks wiring: the initial state records once and a later change records again.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

test/unit/workspace/observers.test.ts
aqandrew: Done in 83e1767. Added the symmetric case to `observers.test.ts`: same `status`, new `lifecycle_state` reports one transition.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

test/unit/workspace/workspaceMonitor.test.ts
aqandrew: Done in 399ba97. Moved the builder to `test/mocks/testHelpers.ts` as `workspaceWith(status, WorkspaceAgent[])` and used it from both `observers.test.ts` and `workspaceMonitor.test.ts`.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

test/unit/workspace/workspaceMonitor.test.ts
aqandrew: Done in 2fac677. Widened `workspaceEvent` to accept a full `Workspace` and used it at both `:302`/`:330` sites instead of hand-rolling `{ type: "data", data }`.

<sub>🤖 Posted by Coder Agents on behalf of @aqandrew.</sub>

EhabY (approved):
Looks great, thanks! :shipit:

PR #1079, opened 2026-08-14
feat(ui): add tree multi-select, type navigation, paging, and sticky scroll

EhabY:
## Summary

Second of three stacked PRs: #1065 → **this one** → #1078. Adds the interaction VS Code trees have beyond moving and selecting one row at a time.

- Multi-selection with `selectedItemIds`, anchored ranges, and modifier toggling
- Ctrl/Cmd+A scoped the way `list.selectAll` scopes it for a tree
- Buffered prefix and fuzzy type navigation
- PageUp and PageDown measured against the scroller's viewport
- Sticky scroll, with its own tab stop

Each of these reaches into the pure layers rather than sitting beside them, which is why they arrive together rather than as four more PRs: multi-selection puts a range anchor in the interaction state next to focus and selection, and sticky scroll needs the pointer policy to know which surface a click came from.

## Multi-selection

`multiSelect` switches the component to `selectedItemIds` and `onSelectedItemsChange` and sets `aria-multiselectable`. `multiSelectModifier` picks the toggle modifier, mirroring `workbench.list.multiSelectModifier`: `"ctrlCmd"` by default, or `"alt"`.

Shift-click and Shift+Arrow extend from the anchor, and the anchor survives a controlled selection arriving back in a different order, which is what a controlled list does on every echo. Modifier clicks take precedence over expansion, so Ctrl-clicking a branch never toggles it.

Ctrl/Cmd+A matches native rather than selecting everything: `list.selectAll` collects the visible descendants of the focused row's **parent**, and widens to include that parent once the scope is already fully selected. Verified against the command handler in the workbench bundle.

## Type navigation

Typing moves the active row by prefix, falling back to a fuzzy subsequence match, with a buffer that expires after 800ms and repeated-key presses cycling matches. Native list behavior, and the reason `treeTransition` carries a clock.

## Sticky scroll

Ancestors pin against the nearest scrolling ancestor with VS Code's semantics: the `stickyScrollMaxItemCount` cap, the 40% viewport cap, per-level indentation with twisties, and the deepest pinned row sliding out as its subtree ends. The pinned region is its own tab stop, where Arrow keys move among ancestors, Enter reveals and selects the real row, and a pinned twistie collapses the branch in place. The count and push-out math is pure, in `stickyState.ts`.

`stickyScroll` takes `true` for the native default of seven, or a number. Webviews receive no `workbench.tree.*` settings, so a consumer mirroring `enableStickyScroll` and `stickyScrollMaxItemCount` reads them in the extension host and passes them down.

## Diff composition

| Area | Added | Deleted |
| --- | ---: | ---: |
| Production UI source | 892 | 70 |
| Automated tests and shared helpers | 893 | 67 |
| Stories and visual fixtures | 99 | 7 |
| Documentation | 32 | 5 |
| **Total** | **1,916** | **149** |

The deletions are the single-selection shapes this PR generalizes: `selectRow` gains the multi paths, `pointerCommands` gains the gesture branch, and the selection props become a discriminated union. Commands carry every field rather than omitting defaults, so the pure-module tests and `StickyScroll` spell out `toggle`, `range`, `preserveHidden`, `page`, `extend`, and `recursive`.

Pinned sticky rows follow the base's delegated events: the sticky widget resolves which pinned row a click hit instead of handing every row its own handler, which keeps `TreeRow` memoized.

## Validation

- `pnpm test:webview --reporter=dot`: 45 files, 414 tests passed, up from 41 and 372 on #1065
- Focused Tree suite: 10 files, 96 tests passed
- Sticky pinning, its keyboard tab stop, and the shadow verified in Chrome against a running Storybook
- Ctrl/Cmd+A scoping checked against `list.selectAll` in the installed workbench bundle

### DIFF
--- a/packages/ui/README.md
+++ b/packages/ui/README.md
@@ -54,7 +54,7 @@ that override live.
 ## Tree
 
 `Tree` is controlled: `nodes` describe the hierarchy, `expandedIds` controls
-branches, and `selectedItemId` controls selection. Each
+branches, and the single- or multi-selection props control selection. Each
 visible node renders as a flat `treeitem`, while normal keyboard navigation
 keeps DOM focus on the `tree` container and identifies the active row with
 `aria-activedescendant`. Focus and selection are independent.
@@ -87,7 +87,8 @@ whose children are still loading. `icon`, `action`, and `className` customize
 the row. Actions stay live on plain hover, as in the native list, and are
 isolated from row selection and expansion.
 
-Arrow Up/Down, Home, and End move the active row through visible rows. Arrow Right
+Arrow Up/Down, Home, End, PageUp/PageDown, and buffered prefix/fuzzy typing
+move the active row through visible rows. Arrow Right
 expands a branch or enters it; Arrow Left collapses it or moves to its parent.
 
 `expandMode="singleClick"` is the default: clicking a branch selects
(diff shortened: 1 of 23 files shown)


jakehwll (approved):
Confirmed working in Storybook 🚂

PR #1083, opened 2026-08-18
docs: add instructions for creating/installing extension from .vsix file

aqandrew:
Added brief guidance on how to run local changes within one's own VS Code. This is helpful to me as a first-time VS Code extension developer 😄

Note: I tried running "Run Extension" per the existing instructions and got the below error. The instructions I added are a workaround for it

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -188,6 +188,13 @@ pnpm test:integration
 > this extension from a coder workspace. We currently recommend cloning the
 > repo locally
 
+To view your local changes to the extension within VS Code:
+
+1. Run `pnpm package`. This creates a file named coder-remote-VERSION.vsix in your repo directory.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -188,6 +188,13 @@ pnpm test:integration
 > this extension from a coder workspace. We currently recommend cloning the
 > repo locally
 
+To view your local changes to the extension within VS Code:
+
+1. Run `pnpm package`. This creates a file named coder-remote-VERSION.vsix in your repo directory.
+2. In the "Extensions" tab of VS Code, open the kebab menu and select "Install from VSIX...". You can also run "Extensions: Install from VSIX..." from the command palette. Select the VSIX file created by `pnpm package`.
```
EhabY, line 194: I would mention that this works on all VS Code forks as is, so this is the best way to test on multiple versions/forks

EhabY (approved):
Appreciated!

PR #1092, opened 2026-08-25
fix: keep legacy Coder authorities instead of reopening onto per-editor hosts

### LINKED ISSUE #1094: Workspace state lost after v1.16.1 moved workspaces onto per-editor SSH hosts
## Symptom

After upgrading to v1.16.1, Cursor users find their agent chats gone from workspaces they had been using. Window layout and other per-workspace extension state go with them. Reported by a customer.

## Cause

#1061 gave each editor its own SSH host prefix, so Cursor's hosts changed from `coder-vscode.` to `coder-cursor.`. That host name is part of the remote authority, the authority is part of every folder URI, and editors key `workspaceStorage` off a hash of that URI. A changed prefix therefore reads as a workspace the editor has never seen, and Cursor's chats are workspace-scoped.

`setup()` reopened the window on the new authority automatically (`remote.ts:165`), so it happened silently to every workspace opened before the upgrade, and picking the old entry from File > Open Recent bounced straight back onto the new one.

Nothing is deleted. The state is still on disk under the old address.

## Scope

Forks only: Cursor, Windsurf, VSCodium, and others. VS Code proper is unaffected, since its prefix is `coder-vscode` either way.

## Fix

#1092, targeting v1.16.2:

- Drops the automatic migration, so a `coder-vscode` authority stays supported and keeps the identity it already had.
- Reuses the host a workspace was last opened on when opening from the Coder panel, so only workspaces with no history get this editor's prefix.
- Serves the legacy prefix from a single generated SSH config file. Two files declaring the same host pattern left glob order deciding which editor's CLI and credentials a connection used, which is a live bug in v1.16.1.

## Recovery

Once v1.16.2 ships, a workspace that already moved comes back by reopening it from **File > Open Recent** and picking the entry from before v1.16.1. No data needs restoring.

## Known limitation

Two editors connecting to the same deployment over the legacy host share one generated file, last writer wins, and `setup()` runs on the `onResolveRemoteAuthority` activation event, so a reconnect does not rewrite it. This is the pre-1.16.1 behaviour and is tracked in #1092.

EhabY:
Fixes #1094: agent chats disappearing in Cursor after v1.16.1.

## Problem

The SSH host prefix became per-editor in #1061 (`coder-vscode.` to `coder-cursor.`). That host name is part of the remote authority, and editors key `workspaceStorage` off a hash of the folder URI, so a changed prefix reads as a workspace the editor has never seen. Cursor's chats are workspace-scoped, so they went with it, along with layout and every extension's `workspaceState`.

`setup()` reopened the window on the new authority automatically (`remote.ts:165`), so it happened silently to every workspace opened before the upgrade, and Open Recent bounced straight back.

## Why the prefix stays

It is load-bearing for the shared glob include: every editor's file is visible to every ssh invocation, and OpenSSH takes the first value it finds, so without namespaced patterns the alphabetically first filename decides whose `ProxyCommand` runs. Making the block editor-neutral is not safe either, since `--global-config` carries the session token and `coder.globalFlags`, `coder.sshFlags`, `coder.binaryDestination`, and `coder.sshConfig` are per-editor.

So the namespacing stays. What goes is everything that renames an existing workspace.

## Commit 1: keep legacy authorities

- Drop `migrateLegacyAuthority`. `updateSSHConfig` already takes `hostPrefix` from the authority, so a `coder-vscode` window still regenerates a matching block and connects. Users already moved onto `coder-cursor.` recover through Open Recent.
- Name a generated config file after the editor in the host prefix, not the one writing it, so the legacy prefix is served by one file.
- Derive that editor with `hostEditorId()` instead of storing what `hostPrefix` already encodes.
- Tests pick an editor through the vscode mock (`useEditor`) instead of overwriting `env.uriScheme`.

This also fixes a live bug: v1.16.1 took the prefix from the authority but always wrote to `<thisEditor>--<host>.conf`, so a legacy window in Cursor and one in VS Code declared the same `Host coder-vscode.dev--*` in two files. `cursor--` sorts first, so VS Code used Cursor's block and credentials on every connect.

## Commit 2: reopen on the host already in use

`openWorkspace` matched recent folders with `isRemoteAuthorityCompatible`, which accepts a legacy authority, then kept only the path and opened it under a freshly minted per-editor authority. It now reuses the matched folder's authority, so only workspaces with no history get this editor's prefix. Commit 3 replaces that path match with a host comparison.

## Commit 3: decide the host per workspace

Keying reuse on an exact folder path left holes. A multi-root workspace is recorded by its `.code-workspace` file, so it has no `folderUri` to match. A devcontainer authority carries a container payload, so only the identical container matched. A folder the agent supplies through `expanded_directory` has no entry at all until it is opened once. Each fell through to a per-editor authority while the rest of the workspace sat on the legacy host, and since an editor keys window state by the whole URI, the workspace looked like two.

A compatible entry can only be named with this editor's host or the legacy one, so matching paths was never choosing between more than those two. Comparing SSH hosts instead closes the holes and removes code.

- `recentlyOpened()` reads folder entries and `.code-workspace` entries, and tolerates an editor without the private command instead of failing the open.
- `reusableAuthority()` returns whichever of the two hosts a recent entry last connected over. Only the host carries over, never the entry's own authority, which can name another devcontainer of the same workspace.
- `openWorkspace`, `runOpenDevContainer`, and `toWorkspaceAuthority` all use it, so the panel, a dashboard link, a devcontainer link, and Open Recent agree, and a support bundle for a workspace you are not connected to looks for remote logs under the host that workspace opens on.

Per @code-asher's review, the tie goes to the most recently used host rather than to legacy. A workspace already moved onto a per-editor host comes back after one trip through Open Recent, and stays there.

Also in this commit:

- `Coder: Open Generated SSH Configuration File` lists the legacy file. It walked only this editor's prefix, so a local window could never open it.
- Open Recent marks the shared host `(legacy)`, since a workspace listed on both hosts was two identical lines, and the folder picker lists each folder once rather than once per host.
- `retargetRemoteAuthority` becomes `toLegacyAuthority`, a single direction, since nothing migrates any more.

## Known limitation

Two editors on the same deployment over the legacy host share one file, last writer wins, and `setup()` runs on the `onResolveRemoteAuthority` activation event, so a reconnect does not rewrite it. If Cursor connects, VS Code connects after it, and Cursor drops, Cursor's reconnect uses VS Code's block.

That is the pre-1.16.1 behavior and better than today, where alphabetical order always wins and never self-corrects. Closing it fully needs the per-editor prefix, which is what costs users their workspace identity.

## Testing

`pnpm test` (2465 passed, 6 skipped), `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm build`.

Migration tests are replaced with coverage that a folder, a saved multi-root workspace, an untitled multi-root workspace, and an empty window all connect over the legacy host without reopening. New coverage for host reuse from the panel and from a link, a multi-root workspace, a devcontainer, the generated config picker, the Open Recent suffix, and `toLegacyAuthority`.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -51,9 +51,13 @@ import {
 } from "./supportBundle/remoteServerDataPath";
 import { runExportTelemetryCommand } from "./telemetry/export/command";
 import {
+	LegacyEditorId,
+	currentEditorId,
 	hostEditorId,
 	isRemoteAuthorityCompatible,
 	parseRemoteAuthority,
+	sshHostOf,
+	toLegacyAuthority,
 	toRemoteAuthority,
 } from "./util/authority";
 import { openInBrowser, toSafeHost } from "./util/uri";
@@ -87,6 +91,12 @@ import type {
 	DuplicateWorkspaceIpc,
 	PongMessage,
 } from "./workspace/duplicateWorkspaceIpc";
+
+/** One entry from the private `_workbench.getRecentlyOpened` command. */
+interface RecentlyOpened {
+	folderUri?: vscode.Uri;
+	workspace?: { configPath?: vscode.Uri };
+}
 
 const NO_SSH_CONFIG_MESSAGE =
 	"No SSH config has been generated yet. It is written when you connect to a workspace.";
@@ -573,15 +583,12 @@ export class Commands {
 		);
 	}
 
-	/** Open this editor's generated SSH config, picking a deployment when several exist. */
+	/** Open the generated SSH config, picking a file when several exist. */
 	public async openSshConfig(): Promise<void> {
-		let configPath = this.connectedSshConfigPath();
+		const configPath =
+			this.connectedSshConfigPath() ?? (await this.pickSshConfigPath());
 		if (!configPath) {
-			const hostname = await this.pickSshHostname();
-			if (!hostname) {
-				return;
-			}
-			configPath = this.pathResolver.getSshConfigPath(hostname);
+			return;
 		}
 		try {
 			await openFile(configPath);
@@ -613,24 +620,33 @@ export class Commands {
 		}
 	}
 
-	/** Ask which deployment, when this editor has generated more than one config. */
-	private async pickSshHostname(): Promise<string | undefined> {
-		const hostnames = (
-			await readdirOrEmpty(this.pathResolver.getSshConfigDir())
-		)
-			.map((file) => this.pathResolver.parseSshConfigFile(file))
-			.filter((name) => name !== undefined);
-		if (hostnames.length === 0) {
+	/** Ask which config to open, of this editor's prefixes and the legacy one. */
+	private async pickSshConfigPath(): Promise<string | undefined> {
+		const files = await readdirOrEmpty(this.pathResolver.getSshConfigDir());
+		const items = [...new Set([currentEditorId(), LegacyEditorId])].flatMap(
+			(editorId) =>
+				files
+					.map((file) => this.pathResolver.parseSshConfigFile(file, editorId))
+					.filter((safeHostname) => safeHostname !== undefined)
+					.map((safeHostname) => ({
+						label: safeHostname,
+						// Which hosts it serves; two files can share a deployment.
+						description: `coder-${editorId}.${safeHostname}--*`,
+						path: this.pathResolver.getSshConfigPath(safeHostname, editorId),
+					})),
+		);
+		if (items.length === 0) {
 			vscode.window.showInformationMessage(NO_SSH_CONFIG_MESSAGE);
 			return undefined;
 		}
-		if (hostnames.length === 1) {
-			return hostnames[0];
+		if (items.length === 1) {
+			return items[0].path;
 		}
-		return vscode.window.showQuickPick(hostnames, {
+		const picked = await vscode.window.showQuickPick(items, {
 			title: "Open generated SSH configuration",
 			placeHolder: "Select a deployment",
 		});
+		return picked?.path;
 	}
 
 	/**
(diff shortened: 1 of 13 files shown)


EhabY:
Tracking issue: #1094.

src/commands.ts
```diff
@@ -1552,6 +1563,11 @@ export class Commands {
 					return { status: "cancelled", stage: "recent_folder_picker" };
 				}
 			}
+			// A compatible folder can still be on the legacy coder-vscode host.
+			// Reopen it there, since the editor keys window state by the whole URI.
+			remoteAuthority =
+				opened.find((f) => f.folderUri.path === folderPath)?.folderUri
+					.authority ?? remoteAuthority;
```
code-asher: Should we check for the legacy authority in addition to the path?

In the recents, users that already hit the bug should (in theory) have one from `cursor` and one from `vscode`, so it is possible it would pick the first and never reopen in the legacy one.

code-asher (approved):
Confirmed that it keeps using the legacy host when I launch the IDE again, and also when I disconnect then reconnect via the sidebar or URL when I have a legacy recent entry.  👍 

One potential quirk though.  When I first connect to the workspace, and then click "open folder" in the sidebar (or use a URL with `folder` set), apparently no recent entry is created??  I only get a recent if I specifically use the "open folder" link on the welcome screen, and even that only works if I open a *new* folder.  Does that happen to you as well?

So in that case connecting from the sidebar or URL gave me the new host even though I previously connected to the old one, since there is no recent entry to correlate.

Assuming it is not just a me issue, is there some other way we might have to check what workspaces have previously been open?

code-asher:
Huh, tested again and the recent is created so maybe just a weird issue with my setup?  I was wiping my editor `state.vscdb` file to get a fresh start so this time I tried deleting the entire `User` dir and it works.  So maybe deleting `state.vscdb` only partially deletes and leaves my IDE in a weird state or something.  In any case, it looks good to me :+1:

code-asher:
OK idk something seems weird with my VS Code.  I connect, I have a recent, if I close the remote and exit then relaunch, the entry is gone.

src/commands.ts
EhabY: Yes that's true but that's intended. Like imagine if a user connects to legacy one time, then the new `cursor` one then it would keep going back to the old entry even if it's months or years old

EhabY:
> apparently no recent entry is created?

I'm not entirely sure what the conditions are for creating a recents entry, it is not really consistent and could be that it only happens after connecting a few times? Or maybe connecting for a bit? In any case, this PR should maintain that legacy entry if it exists 🤔

src/commands.ts
code-asher: That is what users hitting this bug would want right?  Since their old one is where their chat history is.  We have to maintain this behavior forever.

code-asher: The only time a user would want non-legacy is if they are connecting for the first time, in which case it would never fall back anyway since there is no non-legacy recent.

EhabY: The idea was that those users can just do `File -> Open Recent -> ...` but yes you are right I think. Let me prefer the `coder-vscode` over other authorities

code-asher: We talked it over and decided to keep the behavior, it will use the latest recent, so to "migrate" the sidebar/url handler back to using the legacy prefix the user has to open once manually with the old prefix via the recents menu.

PR #1107, opened 2026-09-10
feat: default coder.useKeyring to true and harden shared credential handling

### LINKED ISSUE #1106: Default `coder.useKeyring` to `true`
## Background

`coder.useKeyring` is currently `false` by default. With it off, the session token is written to a plaintext `session` file in the extension's per-deployment CLI config directory. Storing long-lived credentials in cleartext on disk is a recurring finding for security-sensitive deployments, and the OS keyring path already exists and works.

The setting was made opt-in in #836 / #835 after the rollback in #828. The reason was behavioral, not technical: keyring mode makes the extension and the `coder` CLI share a single credential, so logging out of the extension also logs out the CLI. Keyring support itself is delegated entirely to the CLI (#816) and requires Coder CLI 2.29+ on macOS or Windows.

## Proposed change

Flip the default of `coder.useKeyring` to `true`. On macOS and Windows with a supported CLI, the session token is stored in Keychain / Credential Manager instead of a plaintext file. Linux and older CLIs are unchanged. Users can opt out with `coder.useKeyring: false`.

## Hardening to include

1. Always pass `--use-keyring=true` explicitly, so a user-set `CODER_USE_KEYRING=false` or `CODER_CONFIG_DIR` cannot cause the extension to write to or delete the CLI's own plaintext files.
2. Logout only revokes the keyring token if the extension stored it. Otherwise clear only the extension's own session state.
3. If the keyring token belongs to a different user, reuse the existing "replacing your session for X" confirmation rather than switching accounts silently.
4. If the keyring is unavailable, surface a clear error pointing at `coder.useKeyring: false`. No silent fallback to file storage.
5. Decide how a user-set `--global-config` in `coder.globalFlags` (#1011) interacts with keyring mode. Today the CLI silently forces file storage when that flag is set; the extension should either keep file mode explicitly or reject the combination rather than appear to use the keyring when it does not.

## Known trade-offs

- Logging out of the extension also logs out the `coder` CLI when they share the token. This is the behavior that caused the earlier rollback. It needs to be documented and communicated in the logout flow.
- Each login or connect updates the CLI's current deployment, and logout removes it. Users running one deployment in the CLI and another in the extension will need to re-run `coder login`.
- Connecting fails when the keychain is locked (over SSH/RDP, or after auto-lock). Intentional; the workaround is opting out.

Fully decoupling the extension's credential from the CLI's requires a CLI-side change, tracked in coder/coder#29105.

Created on behalf of @EhabY

EhabY:
Closes #1106.

## What changes

- `coder.useKeyring` defaults to `true`. On macOS and Windows with Coder CLI 2.29+, the session token also goes to the OS keyring through the CLI, so the terminal `coder` shares the session. Linux and older CLIs keep using a file. The extension's own secret storage is unchanged.
- Every CLI 2.29+ call passes `--use-keyring=<bool>`, so the setting wins over `CODER_USE_KEYRING`. `CODER_CONFIG_DIR` is honored like `--global-config`.
- The CLI store is either the **CLI's own** (`cli`: its config directory or the keyring, or a user directory from `--global-config` or `CODER_CONFIG_DIR`) or the **extension's** (`extension`: a file in the extension's per-deployment directory). `CliAuth` carries the choice and the flags follow from it.
- Logout asks "Sign out of the Coder CLI too?" when the CLI holds the same token as the extension (checked with `coder login token` on 2.32+; older CLIs cannot be read, so the CLI's own store always asks). A CLI signed in with a different token is left alone. The extension's own store is always logged out, since `coder logout` revokes the token. OAuth sessions skip the prompt because logout revokes their token anyway. Manage Stored Credentials asks the same for one deployment and signs the CLI out on Remove All where the token matches.
- Adopting the CLI's session for a different user than before asks first, reusing the link sign-in modal.
- Without a downloaded binary, credential operations are skipped with one info line instead of failing; connect downloads the binary and stores the token.
- `--allow-redirects` is passed to CLI 2.38+, whose `coder login` and `coder logout` otherwise fail for a deployment URL that redirects (coder/coder#29104). The extension already follows redirects.
- `coder version` is cached per binary until the file changes, and CLI reads or token checks are skipped when settings rule out the CLI's own store.
- CLI token reads require 2.32, where file mode checks the stored URL against `--url`. The keyring path keys entries by host only and cannot be checked from the extension; coder/coder#29290 tracks the CLI fix.
- A failed credential store shows the CLI's error with **Open Settings**. A logout that cannot remove every credential shows **Show Output**.

## Flags per situation

| Keyring | User directory | CLI | Store | Flags the extension adds |
| --- | --- | --- | --- | --- |
| any | any | < 2.29 | extension | `--global-config <ext dir> --url <url>` |
| on | any | ≥ 2.29 | cli | `--url <url> --use-keyring=true` |
| off | no | ≥ 2.29 | extension | `--global-config <ext dir> --url <url> --use-keyring=false` |
| off | yes | 2.29 to 2.31 | extension | `--global-config <ext dir> --url <url> --use-keyring=false` |
| off | yes | ≥ 2.32 | cli | `--url <url> --use-keyring=false` |

"Off" includes Linux. The extension's store drops the user's own `--global-config` item; the CLI's store keeps it. CLI 2.38+ also gets `--allow-redirects`.

## Trade-offs

- Signing in updates the CLI's session for that deployment, and signing out asks whether to sign the CLI out too. This is the behavior behind the earlier rollback (#828), now documented in the setting and left to the user at logout.
- A locked keyring fails the connection with an error instead of falling back to a file. The opt-out is `coder.useKeyring: false`.
- Telemetry: the `keyring_token` login method folds into `cli_token`; credential spans carry `store` (`cli`, `extension`) and `outcome` (`stored`, `logged_out`, `kept`, `no_binary`); `auth.logout` gains `user_dismissed`.

## Testing

`pnpm test`, `pnpm typecheck`, `pnpm lint`, and `pnpm format` pass. Developed on Linux; not yet verified on a keyring-capable machine: the shared token showing up in `coder login token`, `CODER_CONFIG_DIR` pass-through, and the locked-keychain error.


src/core/cliCredentialManager.ts
```diff
@@ -71,95 +54,63 @@ export class CliCredentialManager {
...
 		options?: { signal?: AbortSignal },
 	): Promise<void> {
 		return this.credentialTelemetry.traceStore(configs, async (span) => {
-			const transport = await this.resolveWriteTransport(url, configs);
-			span.setProperty(
-				"category",
-				transport.kind === "keyring" ? "keyring" : "file",
-			);
-			await this.cliLogin(transport, url, token, configs, options);
+			const cli = await this.resolveCli(url, configs);
```
code-asher, line 65: When I sign in for the first time I get an error popup "failed to store credentials: no cli binary found".  Maybe we should ignore `ENOENT` errors?  Same if I log out without having connected once.  Nothing is actually wrong at these points.

Arguably if the cli is deleted after a connect, then I could see considering it an error if it disappears, but at the same time we will download and configure it again, so maybe not a big deal.

I also get two log entries about it, one a warning (probably when trying to read the token), and then another after logging in (probably the write).  I feel like these should be infos at most and be a bit more descriptive (something like "tried to read token, but the cli has not been downloaded or was deleted") at least if we have not connected once before.

src/core/cliCredentialManager.ts
```diff
@@ -71,95 +54,63 @@ export class CliCredentialManager {
...
-			...this.credentialGlobalFlags(transport, url, configs),
-			"login",
-			"token",
-			"--url",
-			url,
-		];
-		const token = await this.runTokenRead(transport.binPath, args, options);
-		if (!token) {
+		// Keyring entries drop the scheme, so an http lookup returns the https token.
+		if (cli.auth.useKeyring && !url.startsWith("https:")) {
```
code-asher, line 100: I wonder if this check should be in the cli instead?  If it spits out an incorrect token given some input, that feels like a cli bug.

Then again, people probably update the VS Code extension faster so if this is a pressing issue then it might make sense to have it in both places (and I suppose we may need it in Toolbox too).

src/core/cliCredentialManager.ts
```diff
@@ -170,155 +121,116 @@ export class CliCredentialManager {
...
-			...this.credentialGlobalFlags(transport, url, configs),
-			"logout",
-			"--url",
-			url,
-			"--yes",
-		];
+		span.setProperty("store", cli.auth.store);
+		if (!(await this.ownsCliSession(cli, session, signal))) {
+			this.logger.info("Kept the CLI session for", url);
+			return true;
```
code-asher, line 164: Should be `return false` I think?  Because the doc comment on `deleteToken` says it returns true if every store was cleared, which is not the case here.

src/login/loginCoordinator.ts
```diff
@@ -432,29 +436,36 @@ export class LoginCoordinator implements vscode.Disposable {
...
 		}
-		return withLoginMethod(
-			cliCredential.source === "keyring" ? "keyring_token" : "cli_token",
-			result,
-		);
+		if (result.success && auth && auth.username !== result.user.username) {
+			const confirmed = await this.confirmSignIn(
+				deployment.url,
+				{
+					title: "Sign in with the Coder CLI session?",
```
code-asher, line 458: nice nice 👍

src/core/cliCredentialManager.ts
```diff
@@ -71,95 +54,63 @@ export class CliCredentialManager {
...
+				}
+				throw new CredentialCliError(error);
+			}
 		});
 	}
 
-	private async cliLogin(
-		transport: CliTransport,
+	/** Reads the CLI's token via `coder login token` (CLI 2.32+). Undefined on any failure. */
+	public async readToken(
```
code-asher, line 84: Similar to my other comment, we get a warning in the log if the cli has no session token, wonder if we could suppress this.

Then again, maybe not worth parsing the error output of `login token`, that does seem kinda janky.

src/core/cliCredentialManager.ts
```diff
@@ -170,155 +121,116 @@ export class CliCredentialManager {
...
-	): Promise<ReadTransport> {
-		// Reading is best-effort: a missing binary means no CLI credentials.
-		const cli = await this.resolveCli(url).catch((error) => {
-			this.logger.warn("Could not resolve CLI binary:", error);
-			return undefined;
-		});
-		if (!cli) {
-			return { kind: "none" };
+	/** A shared store is ours only if the CLI still holds the token this extension created. */
+	private async ownsCliSession(
```
code-asher: Are we actually tracking if the cli is the source?  When I tested it always clears the token from the cli, so this `ownsCliSession` seems to have no effect.

My test:

1. `login` on the command line
2. Launch VS Code, login
3. It picks up the token from the cli
4. Logout from VS Code
5. It clears the token from the cli

It also immediately seems to store the token via the cli even though it just got it from the cli, which does seem redundant, but probably harmless.

code-asher, line 182: This is neat but I feel like it is also unpredictable.  When a user goes to log out, sometimes it will log out the cli, sometimes not, depending on what they did potentially a long time ago which they may have long forgotten.

Or, it seems possible that a user may log in with VS Code, then rely on that shared token for JetBrains, then one day log out of VS Code and be surprised JetBrains also lost its token, for example.

Maybe we can prompt the user about whether to also clear the token from the shared storage or keyring, warning that anything else relying on the token would also be logged out?

src/settings/cli.ts
```diff
@@ -100,36 +101,47 @@ function isFlag(item: string, name: string): boolean {
...
+}
+
+/** True when `coder.useKeyring` is on and the platform supports it. */
 export function isKeyringEnabled(
 	configs: Pick<WorkspaceConfiguration, "get">,
 ): boolean {
-	return (
-		isKeyringSupported() && configs.get<boolean>("coder.useKeyring", false)
-	);
+	return isKeyringSupported() && configs.get<boolean>("coder.useKeyring", true);
```
code-asher, line 114: Reasoning out loud about the effects of this change: for users currently logged in, they still get their token from the secret store, and when they connect we store it in the keyring and configure the cli to use the keyring, so they should not be suddenly logged out or anything like that.  Does that sound right?

Or in other words, this only affects how we share tokens with the cli, it does not currently impact the plugin's own internal storage/copy of tokens.

The only potential breakage is if the keyring is misconfigured or something, then connecting could start failing despite no changes from the user.  But that seems reasonable, and they can fix by opting out, we even have a notification for it.

CHANGELOG.md
```diff
@@ -5,6 +5,34 @@
      from published versions since it shows up in the VS Code extension changelog
      tab and is confusing to users. Add it back between releases if needed. -->
 
+## Unreleased
+
+### Changed
+
+- Store session tokens in the OS keyring by default on macOS and Windows. The
```
code-asher, line 12: The wording to me makes it sound like the keyring is the only place they are stored now but we do still store them in VS Code's secret storage from what I can tell (looking at `persistSessionAuth`).

Could just be a me issue though lol.  But maybe "Additionally store session tokens in the OS keyring by default for use by the cli...does not change how the plugin stores its own tokens" or something

code-asher (approved):
Looks good to me.  I think maybe we should change whether we also log out the cli, but not blocking imo.

tbh I have not actually tested the keyring, being on Linux and all, but might find some time to run a VM later.

src/core/cliCredentialManager.ts
EhabY: Fixed. `locateBinary` returns `undefined` now instead of throwing, and the credential manager treats that as nothing to do: login skips the store and the read, and logout counts the CLI as cleared. Same for a deleted binary, since the next connect re-downloads and re-stores anyway.

The two warnings became one info line, `Skipped CLI credentials: the CLI binary is not downloaded`. A binary that exists but can't report its version still errors

src/core/cliCredentialManager.ts
EhabY: This is gone now, logout asks whether to sign out a shared CLI session

src/core/cliCredentialManager.ts
EhabY: Yeah, it's a CLI bug, so I dropped the check here entirely. We can't verify the scheme from the extension anyway: `coder login token` only prints the token and the keyring entry is keyed by host. File mode already checks `--url` against the stored URL since 2.32, the keyring path should do the same. Opened coder/coder#29290 for that.

src/core/cliCredentialManager.ts
EhabY: Now an info: `Could not read the CLI session (it may not be signed in):` plus the CLI error. Didn't parse the output, agreed that's gotta be janky

src/core/cliCredentialManager.ts
EhabY: Went with the prompt. If the store is shared, Logout asks "Sign out of the Coder CLI too?" with `Keep CLI Session` as the default and `Sign Out CLI` next to it. `Coder: Manage Stored Credentials` asks the same thing for a single deployment and signs the CLI out on `Remove All` (the confirm says so).

This is actually simpler now and easier to reason about!

src/settings/cli.ts
EhabY: Yep. Stored tokens are tried before the CLI and never touch the keyring, so existing sessions keep working. The next connect runs `coder login --use-keyring=true`, which writes the entry and points the CLI's `url` file at the deployment. The old private `url` and `session` files sit there until logout removes them

CHANGELOG.md
EhabY: Reworded this to make it clear now, but the VS Code storage itself is encrypted and never accessible from outside which is why it is very much an implementation detail :pray:

PR #1110, opened 2026-09-14
fix(ssh): repair managed Windows SSH config permissions

### LINKED ISSUE #1108: Generated Coder SSH config contains invalid/unknown user permissions
Remote SSH connections failing on Windows because the SSH config file generated by the Coder extension has invalid file permissions.

### Error

With latest vscode plugin Remote SSH fails with the following message:

```
Bad permissions. Try removing permissions for user:
UNKNOWN\UNKNOWN (S-1-5-21-354618XXXX-405751XXXX-XXXXXXX-XXXXXXXXXX)
on file C:/Users/<username>/AppData/Roaming/coder.coder-remote/ssh/vscode--xxxxx.de.conf.

Bad owner or permissions on
C:\Users\<username>\AppData\Roaming\coder.coder-remote\ssh\vscode--coder.zugang
```

### Workaround

Downgrading the Coder VS Code extension to version 1.15.2 resolves the issue for me.

With version 1.15.2, I can use the Coder-generated SSH configuration directly from my normal SSH config file, and VS Code Remote SSH connects successfully.

EhabY:
## Summary

Fixes #1108. Windows files inherit their permissions from the directory they live in, so a config the extension generates under `%APPDATA%\coder.coder-remote\ssh` can end up readable by other accounts. OpenSSH rejects such a file with `Bad owner or permissions` and skips the whole `Include`, so one bad file blocks every Coder host, not just the one it came from.

Before each managed write, the extension locks that directory down with built-in commands and lets its files inherit from it. No Windows Script Host, ADSI, native dependency, or elevation:

| Step                                                     | Command                                              |
| -------------------------------------------------------- | ---------------------------------------------------- |
| Read the current user's SID                              | `whoami.exe /user /fo csv /nh`                       |
| Clear the directory's own grants                         | `icacls.exe <dir> /reset`                            |
| Grant that user, SYSTEM, and Administrators full control | `icacls.exe <dir> /inheritance:r /grant:r <trustee>` |
| Clear each `*.conf` file so it inherits the directory    | `icacls.exe <file> /reset`                           |

- Resetting every direct `*.conf` match, not only the file being written, also repairs other deployments and editors on the first connection after an upgrade. The temporary file is reset before its rename.
- Check command exit codes. Never read ACLs back or parse SDDL.
- Reject links and non-files before the repair, but tolerate an entry that disappears mid-scan: two connecting windows share the directory, and one can rename its temporary config away between `readdir` and `lstat`.
- Leave your own SSH config, and anything outside the managed directory, alone.

`SshConfig` takes an optional `ManagedPermissions`. Its absence is the marker for "not Coder-managed", which is why it has no no-op default: a default would make the extension enumerate `~/.ssh` on every save on every platform. `createManagedPermissions()` returns `undefined` off Windows, so the whole repair path stays Windows-only.

## What each file changes

- `src/remote/windowsAcl.ts`: new. The `icacls` repair itself, exporting `WINDOWS_ACL`, `createManagedPermissions`, and `system32`.
- `src/remote/sshConfig.ts`: adds `ManagedPermissions`, splits `save()` into named steps, and makes the temp write exclusive (`wx`).
- `src/remote/remote.ts`: passes `createManagedPermissions()` to the per-deployment `SshConfig`.
- Tests and docs: mocked and native Windows suites, `CONTRIBUTING.md`, and `CHANGELOG.md`.
- CI: adds a `windows-11-arm` unit job, because the repair shells out to `icacls.exe` and `whoami.exe` and Windows on ARM64 is a platform users connect from. `test/env-check.ts` fails that job if it ever runs under x64 emulation, which would silently make it a duplicate.
- `.github/actions/setup/action.yml`: moves pnpm store caching from `setup-node` to `pnpm/action-setup`. `setup-node` restores on an exact lockfile hash, so a dependency bump refetched the whole store on every runner; `pnpm/action-setup` keys on the same hash but falls back to the last store for the platform.
- `pnpm-workspace.yaml`: turns off the `bufferutil` and `utf-8-validate` source builds. Neither ships a `win32-arm64` prebuild, so node-gyp compiled both on every cold ARM64 install. Both are dev-only and fall back to JavaScript, and platforms with a prebuild keep theirs.
- `.gitattributes` and `.vscodeignore`: a generated-file marker and a trailing newline, unrelated to the fix.

## Validation

`pnpm test` (2,751 passed, 6 skipped), `pnpm typecheck`, `pnpm lint`, and `pnpm format:check` pass. Windows CI covers x64 and ARM64 on a cold pnpm store. Against OpenSSH 9.6p1, a `*.conf` directory aborts the whole `Include` on Windows, where glibc instead reads zero lines and ignores it, so the extension reports that entry rather than trying to fix it with an ACL change.

## Limits

A directory-preparation failure stops the write. A single file-reset failure logs a warning and still attempts the connection, so OpenSSH may reject a fragment the repair missed. The two directory commands are not atomic: a failure after `/reset` can leave the directory with its parent's grants. The path and link checks stop mistakes, not an attacker racing the check.

Generated by Coder Agents on behalf of @EhabY.


EhabY:
## Windows environment check and remaining draft blocker

Run `34857781412` at `44a19ca` completed. Both x64 and ARM64 passed the production-path environmental regression under Node 22 and Electron 37/current:

- Ordinary temporary-directory inheritance works before repair.
- Adding an inheritable Everyone-write ACL makes the existing writer's included config fail Windows OpenSSH.
- A protected bad sibling remains a blocker even after directory/current-file repair.
- Production reconciliation fixes the included siblings; repeated atomic rewrite works; the external main config descriptor stays unchanged.

This supports an ACL-dependent problem, consistent with a working Windows 11 VM. Issue #1108 names an unknown SID on the generated AppData config but contains no owner/DACL dump, so its exact inheritance source is not established.

**CI remains red, and packaging is gated:** the added directory/hard-link regression fails on both architectures. `SetSecurityInfo` modifies the external hard-linked file's descriptor while securing its parent directory. The observed fixture difference is ACE ordering; it is evidence of unintended descriptor mutation, not proof of expanded effective access in that fixture. Five other native policy tests pass.

Microsoft documents automatic propagation and suppression when the supplied handle was opened with `MAXIMUM_ALLOWED`: https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-setsecurityinfo . Proposed next step is to test that mechanism while retaining inheritance for new children and explicit validated sibling repair, rather than introducing recursive traversal or removing the failing test. Awaiting user discussion of this additional native API decision.

Local production build/typecheck/lint/format and Electron extension suite passed (2253 tests). Actual universal production VSIX validation remains pending the native gate. No merge/release readiness is claimed.

Generated by Coder Agents on behalf of @EhabY.

vitest.config.mts
```diff
@@ -1,6 +1,12 @@
+import assert from "node:assert/strict";
 import path from "node:path";
 import { defineConfig } from "vitest/config";
 
+if (process.env.EXPECTED_ARCH) {
+	process.stdout.write(`Test runtime architecture: ${process.arch}\n`);
+	assert.equal(process.arch, process.env.EXPECTED_ARCH.toLowerCase());
+}
```
untra: putting execution logic and assertions directly into a configuration file is generally an anti-pattern.

maybe try instead moving this to a `env-check.ts` and:
```
export default defineConfig({
test: {
    ...
    globalSetup: './test/env-check.ts',
  },
```

test/unit/remote/windows/acl.native.test.ts
```diff
@@ -0,0 +1,300 @@
...
+		return await callback(backup);
+	} finally {
+		await fs.rm(backup, { force: true });
+	}
+}
+
+async function savedAcl(pathname: string): Promise<string> {
+	return withAclBackup(async (backup) => {
+		await execFile(system32("icacls.exe"), [pathname, "/save", backup]);
+		return await fs.readFile(backup, "utf16le");
```
untra: > Why does this have to be utf-16 ?

`icacls /save` Defaults to UTF-16LE
When the extension runs `icacls.exe [target] /save [backup]`, it asks Windows to dump the Access Control List (ACL) of the target SSH config file into a text file. By default, Windows writes this backup file using UTF-16LE (Little Endian) encoding.

EhabY: This was written in the prod file, but I've added a small comment here as well :pray:

vitest.config.mts
EhabY: Ohh nice idea, done!

test/unit/remote/windows/acl.native.test.ts
untra: just gotta make sure utf-16 files and strings don't leak out to much, are used minimally. I do see it in the prod code; I'm not a fan of alternative content encoding, as they can cause unintended data pollution.

> `icacls.exe`

I think another windows expert should give this PR a better review

phorcys420:
is there any reason you didn't use the `icacls` utility or PowerShell's `Set-Acl` all the way to set the permissions instead of depending on JScript (`assets/wsh/acl.js`)?

Scripts ran by the Windows Script Host are super sketchy patterns that AVs love to flag for no reason. Spawning PowerShell is similar but less so.

test/unit/remote/windows/acl.native.test.ts
EhabY: The utf-16 read is gone (except only in a test), I'll let a windows expert review the rest :pray:

EhabY:
Simplified this a lot more by removing the pre/post ACL checks that ran to make sure everything was applied correctly. We now apply this on all runs on Windows (which takes the same amount of time as the check anyway). The only windows specific logic is in `windowsAcl.ts`!

src/remote/windowsAcl.ts
```diff
@@ -0,0 +1,122 @@
...
+async function prepareDirectory(target: string): Promise<void> {
+	const directory = path.win32.normalize(target);
+	try {
+		checkPath(directory);
+		if (!(await lstat(directory)).isDirectory()) {
+			throw new Error("Expected a Coder-managed directory without links");
+		}
+		// Inheritable grants reach children even without /T, so vet them first.
+		for (const entry of await readdir(directory)) {
+			await checkRegularFile(path.win32.join(directory, entry));
```
fioan89: Could we tolerate ENOENT when checking these entries? Two connecting editor windows share this directory, and readdir can pick up another window’s temporary config just before it gets renamed (I think this happens in ssh config.save()) . The subsequent lstat then throws, aborting directory preparation and the connection.

fioan89 (approved):

src/remote/windowsAcl.ts
EhabY: Ohhh nice catch actually, also users could delete a file in the meantime and it shouldn't break!

aslilac (approved):
providing an authoritative stamp on @fioan89's behalf

matifali (approved):

PR #1115, opened 2026-09-16
fix: queue workspace updates as one build

### LINKED ISSUE #1095: Updating workspace via VS Code fails
Here are the logs from the Coder VS Code output panel when I try to update my workspace via the "Update workspace" button in VS Code:

```
2026-08-25 13:49:18.543 [info] [session 3c5e933a] Updating kevin/workstation {"status":"running"}
2026-08-25 13:49:20.446 [info] [session 3c5e933a] Workspace kevin/workstation state changed {"from":"running","to":"stopping","transition":"stop","reason":"initiator"}
2026-08-25 13:49:20.446 [info] [session 3c5e933a] Workspace kevin/workstation agent <ELIDED> removed
2026-08-25 13:49:35.695 [info] [session 3c5e933a] Workspace kevin/workstation state changed {"from":"stopping","to":"stopped","transition":"stop","reason":"initiator"}
2026-08-25 13:49:41.967 [info] [session 3c5e933a] Workspace kevin/workstation state changed {"from":"stopped","to":"starting","transition":"start","reason":"ssh_connection"}
2026-08-25 13:49:49.480 [warning] [session 3c5e933a] Update failed for kevin/workstation: "--global-config /Users/kevin/Library/Application Support/Code/User/globalStorage/coder.coder-remote/coder.stackav.io update kevin/workstation" exited with code 1: Encountered an error running "coder update", see "coder update --help" for more information
error: Trace=[start workspace: create workspace build: ]
A workspace build is already active.
A workspace build is already active.
```

The workspace does restart but ends up still running the same (non-updated) template version. Updating the workspace on the Coder website works fine as a workaround.

This is on macOS using the Coder VS Code extension version 1.16.1 in VS Code version  1.134.0 (Universal). We're running Coder v2.35.3 on the server.

I found https://github.com/coder/vscode-coder/issues/972 but I'm not sure if it's the same issue. The workspace template I am using does use dynamic parameters with one of the mentioned conditions (`A parameter only appears for certain combinations of other inputs (the static rich-parameters list does not include it).`).

EhabY:
## Problem

`coder update` runs a stop build and then a start build. Between the two, the extension's own `coder ssh` reconnect autostarts the workspace on the old template version, so the update's start build is rejected:

```
stopped → starting   reason: ssh_connection
"A workspace build is already active"
```

The workspace restarts on the version it already had, and the extension logs a warning and connects anyway.

## Changes

| Deployment     | Update path                                                                                    |
| -------------- | ---------------------------------------------------------------------------------------------- |
| 2.36 and later | One stop build that carries the start in `on_success`, so nothing can take the slot in between |
| Before 2.36    | Unchanged: `coder update`, or the REST API on CLIs before 2.24                                 |

The server owns the slot from the moment the stop build is posted, so a reconnect can no longer jump ahead of the update. `template_version_id` stays unset on the queued start, because pinning a follow-up build requires template update permission and the update targets the active version anyway. A workspace that is already stopped takes a lone start build, which can name the version.

While the queued build runs, the state machine follows it instead of starting the workspace itself, and it ignores monitor events from older builds.

On a failed update, a modal offers **Connect Anyway** instead of silently falling back to the old version. Declining aborts the connection.

### Feature sets

`FeatureSet` becomes `CliFeatureSet`, and a new `ServerFeatureSet` keys the 2.36 check to the deployment version instead of the CLI's. The Tasks check moves there too, bounded to 2.29 through 2.34, the releases that serve `/api/v2/tasks` ([coder/coder#20921](https://github.com/coder/coder/pull/20921) promoted it out of `/api/experimental`). Below 2.29 the panel used to show and then 404 on every poll.

## Validation

`pnpm lint`, `pnpm typecheck`, `pnpm format:check`, and `pnpm test` (2,729 passed, 1 skipped, 185 files). No live SSH-race reproduction or Windows smoke test.

Fixes #1095.


test/unit/remote/remote.test.ts
```diff
@@ -226,6 +235,56 @@ describe("Remote", () => {
...
+					"ssh --wait=yes --stdio --usage-app=vscode --network-info-dir <dir> --ssh-host-prefix coder-vscode.coder.example.com-- %h",
+			},
+		])("$name", async ({ flags, expected }) => {
+			const { remote, config } = createRemote();
+			if (flags) {
+				config.set("coder.sshFlags", flags);
+			}
+
+			const proxyCommand = await (
+				remote as unknown as RemoteInternals
```
jeremyruppel: `as unknown as` spotted! if this method is to be used in tests, it may need to be public?

EhabY: True actually, I've exposed those interfaces which gives us a specific contract and makes testing straightforward!

jeremyruppel (approved):

PR #1122, opened 2026-09-21
fix(cli): await trailing progress-log writes before settling download

### LINKED ISSUE #1120: Flaky test: concurrent CLI redownload times out waiting for the progress log to go away
`test/unit/core/cliManager.concurrent.test.ts` fails intermittently on CI. It passed on attempt 1 of [run 35406584668](https://github.com/coder/vscode-coder/actions/runs/35406584668) and failed on attempt 2 of the same commit, in [Unit Test (Linux, Electron 37)](https://github.com/coder/vscode-coder/actions/runs/35406584668/job/105798888468).

```
FAIL test/unit/core/cliManager.concurrent.test.ts > CliManager Concurrent Downloads > redownloads when version mismatch is detected concurrently
AssertionError: promise resolved "undefined" instead of rejecting
- Expected: Error { "message": "rejected promise" }
+ Received: undefined
 ❯ test/unit/core/cliManager.concurrent.test.ts:102:56
```

Line 102 is the progress-log half of `expectLockFilesRemoved`:

```ts
await vi.waitFor(async () => {
  await expect(fs.access(binaryPath + ".lock")).rejects.toThrow();
  await expect(fs.access(binaryPath + ".progress.log")).rejects.toThrow();
});
```

`fs.access` resolved, so `<binary>.progress.log` was still on disk when the poll gave up. The test starts three `fetchBinary` calls, two of which detect the version mismatch and redownload, and `Promise.all` resolves before every peer has finished releasing its progress log. `vi.waitFor` defaults to a 1s timeout at a 50ms interval, which a loaded runner can outlast.

#1043 added that polling for this same flake, so the window helps but is not always enough.

Worth considering:

- Give `vi.waitFor` an explicit timeout, which is the cheap fix but still time-based.
- Make the cleanup observable and await it, rather than polling the filesystem.
- Decide whether the progress log belongs in the assertion at all after a concurrent redownload, since the lock is the part the contract depends on.

EhabY:
Fixes #1120

## Root cause

`CliManager.download()` settled its promise on a `progressWrite` captured by value at the moment the read stream closed. A chunk write still in flight at that point was never awaited, so `clearProgress()` could delete `binary.progress.log` first and the trailing write would recreate it, leaving the file behind permanently. That is why the concurrent test failed only on the progress-log line and no `vi.waitFor` timeout could fix it.

## Fix

Two changes in `download()`:

- Progress-log writes are chained onto the previous one, so awaiting the final `progressWrite` awaits every write, not just the last.
- All three settle paths (success, read error, write error) settle only after the write stream emits `close`, which fs guarantees fires after every pending write callback.

Cleanup can no longer race a trailing write, on any scheduler.

## Tests

- New regression test holds fs write callbacks until `close()`, like a real stream on a loaded runner. Fails on unfixed code, passes with the fix.
- `withStreamError("write")` now emits `close` after `error`, matching real fs with `autoClose: true`.
- The concurrent test's `vi.waitFor` polling is removed; cleanup is complete when the download promises settle.

---

🤖 Generated by Coder Agents


test/unit/core/cliManagerHarness.ts
```diff
@@ -112,6 +112,33 @@ export function setupCliManager(basePath: string = BASE_PATH) {
...
 
+	/**
+	 * Serve a download whose fs write callbacks are held until close() is
+	 * called, like a real stream on a loaded runner.
+	 */
+	const withTrailingWriteFlush = () => {
+		vi.spyOn(fs, "createWriteStream").mockImplementation((writePath) => {
+			const pendingWrites: Array<() => void> = [];
+			const stream = new EventEmitter();
+			const writeStream = stream as unknown as fs.WriteStream;
```
aqandrew:
```suggestion
			const stream = new EventEmitter() as fs.WriteStream;
```

nit: We can avoid `as unknown as` by asserting that `stream` is a `fs.WriteStream`, then changing the `writeStream` references below to `stream` instead

aqandrew (approved):
Appreciate you DRYing up the settle paths in cliManager.ts too 🙂

test/unit/core/cliManagerHarness.ts
EhabY: Oh trueee, I've removed the `unknown` cast :pray:

